PR #5, opened 2021-12-16
Parameterize the revision

sbose78:
Without this, we wouldn't be able to have these pipeline definitions consumed for webhooks.


jduimovich (approved):
This is good, will approve to allow trigger testing. We will also need to update the other build pipeline definitions to have the exact same parameter definitions including revision.
TODO - add creds workspaces for authenticated registries

PR #14, opened 2022-01-14
Bump version to 0.1.3

Michkov:


jduimovich (commented):
These will be correct when we fix the build-definitions build to also create quay.io/redhat-appstudio/build-templates-bundle:v0.1.3

jduimovich (approved):

PR #16, opened 2022-01-21
Task bundles and release

Michkov:
Bundle each task into oci image pushed under `appstudio-tasks:${TASK_NAME}-${VERSION}`. Where VERSION is used by pipelines to point to proper task. VERSION is also used for appstudio-utils image. Pipeline in bundle is now independent on custom clusterTasks.

Release and dev deployment is done by building task bundles, pipelines bundle and  appstudio-utils with the same tag.

Production release requires to tag git repository first with the release.


jduimovich (approved):
Lets use this 1 task per bundle model for now. The pipelines are all in a single bundle but can also evolve. 
This will be tagged as v0.1.4 when merged into main.

PR #17, opened 2022-02-12
cleanup build directory fix

jduimovich:
This PR is intended to fix the cleanup task deleting subPaths: of a still running pipelinerun.

The issues found  
1) a single marker was used so if two of these runs execute concurrently one instance will delete the markers of another and those directories will be removed even if live. Using unique markers will prevent this. 
2) new PVs since the beginning of the cleanup will not contain a marker and would be deleted. This is handled by only processing the list of pvs you already marked and any one ones will have to wait until a future purge.

The script now 
- uses unique marker per run
- avoid race condition when new pvs arrive in middle of run by only using the same pv paths in both passes
also, the references to old versions (like v0.1.3) have been replaced in doc as vX.Y.Z so we don't need to update every time 
- fix reference in doc to hard coded versions as they are stale
The version reference is also in the tests, we can use a "test" tag  (./dev.sh test) which will update the images in your repo and can be tested with a fixed name 
- updated tests to use single "test" tag for now

@Michkov @AndrienkoAleksandr - FYI -  fix for subPaths were deleted while Pruns are active

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@ This repository contains components that are installed or managed by the managed
 
 This includes default Pipelines and Tasks. You need to have bootstrapped a working appstudio configuration from (see `https://github.com/redhat-appstudio/infra-deployments`) for the dev of pipelines or new tasks.
 
-Pipelines are delivered into App Studio via `quay.io/redhat-appstudio/build-templates-bundle:v0.1.3` (the tag will be updated every change)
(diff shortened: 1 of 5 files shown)


Michkov:
To point 2) in description (new PVs since the beginning of the cleanup will not contain a marker and would be deleted.) -> that's not true, directories with marker are deleted, so if the marker is missing then the folder is ignored.

I do not think that this is fixing the issue, but it's improvement.

Michkov (approved):

PR #19, opened 2022-02-16
Change structure and bundle more tasks together

Michkov:
- Create `hack` folder with scripts
- Have only pipelines in `pipelines` folder
- Have only tasks in `tasks` folder
- Use `hack/build-and-push.sh` for building all images


Michkov:
Tested with https://github.com/redhat-appstudio/infra-deployments/pull/56 to be able to do automatic build after merge.

jduimovich (approved):
This looks good.
My only concern is not being able to know which bundle contains a specific task.
The utilities which use task runs -- cleanup-pvc (quay.io/redhat-appstudio/appstudio-tasks:cleanup-build-directories-v0.1.5) can be computed as its a pattern, (same for show-pvc) but now will be in a number "part" bundle. 
So we either need a map (for tasks we want "exported" -- ie findable to run

ie something like this (and same for pipelines)
you use the name as the key and then if not present, you use "default"
This could be built manually by the scripts or computed when you add a bundle to the system ... 
```
apiVersion: v1
data:
  default_task_bundle: quay.io/redhat-appstudio/appstudio-tasks:SHA-PARTX  
  cleanup_build_directories: quay.io/redhat-appstudio/appstudio-tasks:SHA-PARTY    
   random_task: quay.io/redhat-appstudio/appstudio-tasks:SHA-PARTY  
kind: ConfigMap
name: 'task-to-bundle-map'

apiVersion: v1
kind: ConfigMap
metadata:
  name: build-pipelines-defaults 
data: 
  default_build_bundle:   quay.io/redhat-appstudio/build-templates-bundle:v0.1.4
  specific_pipeline:   quay.io/redhat-appstudio/build-templates-bundle:v0.1.5
```

Alternatively we just create multiple directory names in the task dir, 10 per and the names are up to us and could be more meaningful. 
task/utils/(up 10 tasks)
task/utils-2/(up to 10 tasks)
task/security/(up to 10 tasks)
task/security-2/(up to 10 tasks)

and at least you know from the path where the task will land but you need to know the directory structure which is fragile when we change things

And while the scripts are not part of the platform API, we would use this in testing to find and test specific tasks and/or pipelines. 
I think I prefer #1, since when the bundle limit changes, we get less bundles and if we have a mechism to map pipeline-name to bundle and task-name to bundle.  We can do the map in a follow-on PR

PR #22, opened 2022-02-21
remove clustertasks to work on kcp

jduimovich:
This PR contains the changes which remove references to clustertasks and replace with local task definitions. This is intended to enable pipelines on KCP

- added buildah, gitclone, s2i-java and s2i-nodejs tasks 
- changed references to all cluster tasks to local tasks

### DIFF
--- a/pipelines/noop.yaml
+++ b/pipelines/noop.yaml
@@ -29,8 +29,8 @@ spec:
   tasks:
     - name: start 
       taskRef:
-        kind: ClusterTask
-        name: openshift-client
+        bundle: ${CONTAINER_REPO}/appstudio-tasks:utils-task-${TAG}
+        name: utils-task
       params:
         - name: SCRIPT 
           value: |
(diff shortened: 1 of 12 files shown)


pipelines/noop.yaml
```diff
@@ -29,8 +29,8 @@ spec:
   tasks:
     - name: start 
       taskRef:
-        kind: ClusterTask
-        name: openshift-client
+        bundle: ${CONTAINER_REPO}/appstudio-tasks:utils-task-${TAG}
```
Michkov, line 32: there should not be this `bundle`, it will be automatically added when the name matches with the task name.

Michkov:
Is this final decision that we will not use ClusterTasks? Another option would be to upload them into kcp.

jduimovich:
>Transition of ClusterTasks to bundles and in-cluster Tekton Hub

The OpenShift pipelines q1/q2 plan has a stated goal to "Transition of ClusterTasks to bundles and in-cluster Tekton Hub"  so moving away from ClusterTasks is likely anyhow. Using clustertasks is problematic in multi-tenant environments since its a global config shared in all.

PR #24, opened 2022-03-01
Make the Buildah Task "tekton chains" friendly

Michkov:

PR #25, opened 2022-03-04
Make s2i-java and s2i-nodejs chains friendly too

simonbaird:
Tekton Chains expects to see an IMAGE_URL taskrun result containing
the image repo url so let's provide it.

Avoids a "Did not find anything to attest" error from the chains
controller.

Similar to #24 which was merged recently.

Also:
- Make the result descriptions and task step scripts consistent
  across each of the three tasks definitions
- Reword the IMAGE_URL description slightly

### DIFF
--- a/tasks/buildah.yaml
+++ b/tasks/buildah.yaml
@@ -46,10 +46,10 @@ spec:
     name: PUSH_EXTRA_ARGS
     type: string
   results:
-  - description: Digest of the image just built.
+  - description: Digest of the image just built
     name: IMAGE_DIGEST
-  - name: IMAGE_URL
-    description: Image repository where the built image would be pushed to
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
   steps:
   - image: $(params.BUILDER_IMAGE)
     name: build
--- a/tasks/s2i-java.yaml
+++ b/tasks/s2i-java.yaml
@@ -41,8 +41,10 @@ spec:
     name: BUILDER_IMAGE
     type: string
   results:
-  - description: Digest of the image just built.
+  - description: Digest of the image just built
     name: IMAGE_DIGEST
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
   steps:
   - args:
     - |-
@@ -132,7 +134,9 @@ spec:
   - image: $(params.BUILDER_IMAGE)
     name: digest-to-results
     resources: {}
-    script: cat $(workspaces.source.path)/image-digest | tee /tekton/results/IMAGE_DIGEST
+    script: |
+      cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
+      echo "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
   volumes:
   - emptyDir: {}
     name: varlibcontainers
--- a/tasks/s2i-nodejs.yaml
+++ b/tasks/s2i-nodejs.yaml
@@ -29,8 +29,10 @@ spec:
     name: BUILDER_IMAGE
     type: string
   results:
-  - description: Digest of the image just built.
+  - description: Digest of the image just built
     name: IMAGE_DIGEST
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
   steps:
   - command:
     - s2i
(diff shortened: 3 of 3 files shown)


simonbaird:
FYI @Michkov @sbose78 @jduimovich .

simonbaird:
If you prefer less tidying/tweaking of the script quote chars and result descriptions let me know.

simonbaird:
Do we need an upstream change in https://github.com/tektoncd/catalog/ for this too?

Michkov:
Looks good, I am for creating the same change in tektoncd/catalog. I do not like tasks duplicity here, so maybe I will rewrite the script to be able to include task to bundle directly from the catalog.

tasks/buildah.yaml
```diff
@@ -79,8 +79,8 @@ spec:
     name: digest-to-results
     resources: {}
     script: |
-      cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
-      echo "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
+      cat $(workspaces.source.path)/image-digest | tee $(results.IMAGE_DIGEST.path)
+      echo $(params.IMAGE) | tee $(results.IMAGE_URL.path)
```
zregvart, line 83:
```suggestion
      cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
      echo "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
```
I think we should keep the quotes, it's safer with weird paths/parameters. In a complementary PR #24 the quotes were added, I think we should be consistent with that here.

tasks/s2i-java.yaml
```diff
@@ -132,7 +134,9 @@ spec:
   - image: $(params.BUILDER_IMAGE)
     name: digest-to-results
     resources: {}
-    script: cat $(workspaces.source.path)/image-digest | tee /tekton/results/IMAGE_DIGEST
+    script: |
+      cat $(workspaces.source.path)/image-digest | tee $(results.IMAGE_DIGEST.path)
+      echo $(params.IMAGE) | tee $(results.IMAGE_URL.path)
```
zregvart, line 139:
```suggestion
      cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
      echo "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
```

We should quote these as well.

tasks/s2i-nodejs.yaml
```diff
@@ -87,7 +89,9 @@ spec:
   - image: $(params.BUILDER_IMAGE)
     name: digest-to-results
     resources: {}
-    script: cat $(workspaces.source.path)/image-digest | tee /tekton/results/IMAGE_DIGEST
+    script: |
+      cat $(workspaces.source.path)/image-digest | tee $(results.IMAGE_DIGEST.path)
+      echo $(params.IMAGE) | tee $(results.IMAGE_URL.path)
```
zregvart, line 94:
```suggestion
      cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
      echo "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
```
We should quote these as well.

zregvart (commented):
Suggested adding quotes for paths and parameters

tasks/buildah.yaml
simonbaird: Any tie-breaker votes on the quotes? :)

simonbaird:
Quotes are back in 2251e9c .

tasks/buildah.yaml
zregvart: I'm okay either way, I was thinking that we'd like to keep them as #24 introduced them there and it is a bit safer (say someone passes `IMAGE=*`). Looking at it again, it is not quoted for `buildah`, but I'm guessing buildah doesn't expand globs on it's own, so indeed the point about `buildah` failing first stands.

simonbaird: Yeah works for me.

simonbaird:
That was just to fix commit message.

simonbaird:
Looking in https://github.com/tektoncd/catalog/ I can see there are a handful of similar `digest-to-results` tasks where a similar change would probably be useful, but at this stage I'm not planning to do anything about it.

simonbaird:
@Michkov Is it good enough to merge? Do we need another approval or some CI to pass?

jduimovich (approved):
LGTM

PR #28, opened 2022-03-10
utility scripts and binaries to test tekton-chains

joejstuart:
* Adds the rekor-cli to the appstudio-utils image
*  Moving chains utility scripts from infra-deployments to the appstudio-utils image


Michkov (dismissed):

simonbaird:
FYI there is some duplication here, since these same bash scripts are being used in the infra-deployment under hack/chains . This ought to be fixed later - we don't want to maintain the same code in two places - but for now the motivation is to demonstrate how we can run rekor and cosign in a taskrun step defined in build-definitions, and reusing those same scripts is useful step towards that goal.

Michkov (approved):

PR #30, opened 2022-03-14
Set git-auth workspace only on git-clone task

Michkov:
Fix of private repo cloning. Using feature of git-clone task directly.
No interaction with other steps.


jduimovich (approved):
LGTM
- tested this with a private repo and git secrets mounted from namespace. 
Note, build.sh scripts peek repos using curl but do not use the secret so to test you need to specify the build pipeline you want to run.

PR #31, opened 2022-03-14
Adding HACBS basic sanity tests standalone task

jsztuka:
Initial script for HACBS test team basic sanity tests check.

### DIFF
--- a/tasks/sanity-label-check.yaml
+++ b/tasks/sanity-label-check.yaml
@@ -1,0 +1,34 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sanity-label-check
+spec:
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+  steps:
+    - name: inspect-image
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      script: "skopeo inspect docker://$(inputs.params.sourceImage) > /tekton/home/image_inspect.json"
+    - name: basic-sanity-checks-required-labels
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      command:
+        - /usr/bin/conftest
+      args:
+        - "test"
+        - "/tekton/home/image_inspect.json"
+        - "--policy"
+        - "/project/image/policy/required-labels.rego"
+        - "--output=json"
+    - name: basic-sanity-checks-deprecated-labels
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      command:
+        - /usr/bin/conftest
+      args:
+        - "test"
+        - "/tekton/home/image_inspect.json"
+        - "--policy"
+        - "/project/image/policy/deprecated-labels.rego"
+        - "--output=json"


jsztuka:
cc @dirgim

dirgim:
> Maybe dropping the `task` from name of the file is good idea?
Agreed, it seems to align with the existing naming convention.

jsztuka:
Also would like to create HACBS-test resource repo in app-studio org so we are getting policies from there.

tasks/basic-sanity-check-task.yaml
```diff
@@ -0,0 +1,52 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: basic-sanity-check
+spec:
+  stepTemplate:
+    volumeMounts:
+      - name: project
+        mountPath: /workspace/source/image
```
Michkov, line 10: unused

tasks/basic-sanity-check-task.yaml
```diff
@@ -0,0 +1,52 @@
...
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+  steps:
+    - name: clone-git-repo
+      image: alpine/git
```
Michkov, line 17: using dockerhub could be problematic do to pull limitations -> not possible to run in the office.

tasks/basic-sanity-check-task.yaml
```diff
@@ -0,0 +1,52 @@
...
+      script: "git clone https://github.com/dirgim/hacbs-conftest.git /project"
```
Michkov, line 18: Personally I would go with the docker image build from dirgim/hacbs-conftest.git, you would have an image with complete environment (skopeo, conftest and policies) -> easy to do overall test. You are managing content of the image and can easily revert to previous image.

tasks/basic-sanity-check-task.yaml
jsztuka: Removed

tasks/basic-sanity-check-task.yaml
jsztuka: Got rid of it, no use of cloning at all.

tasks/basic-sanity-check-task.yaml
jsztuka: Image updated so we have all in one place.

tasks/basic-sanity-check.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: basic-sanity-check
```
Dannyb48, line 5: Should we rename the task and the file to sanity-label-check? Since we plan on adding separate tasks for sanity-deprecated-image check and sanity-signed-rpm-check, etc. ?

Dannyb48, line 5: Should we rename the task and the file to sanity-label-check? Since we plan on adding separate tasks for sanity-deprecated-image check and sanity-signed-rpm-check, etc. ?

Michkov: Or the task could be just `conftest` and as parameter it could take the policy. And then you can have pipelines which will be composed of multiple tasks.

BTW there will be test pipelines or it will be part of other pipelines (eg. build, release)?

jsztuka: > Should we rename the task and the file.
That would make sense, since we are checking for labels.

> Or the task could be just conftest and as parameter it could take the policy. And then you can have pipelines which will be composed of multiple tasks.
This aproach is also possible, I would say we can keep this as a task for now and include it in a pipeline if it turns out that we want to have such approach.

> BTW there will be test pipelines or it will be part of other pipelines (eg. build, release)?
This should be part of other pipeline(s), currently I don't know about having standalone test pipeline.

Michkov (approved):

Dannyb48 (approved):
LGTM

PR #32, opened 2022-03-14
Allow to use custom registry secret for s2i builds

Michkov:
Use the same way for s2i build as for buildah. Allow to define push
secret for each pipelinerun.


jduimovich (approved):
LGTM
- tested with secrets in ns and push to quay.io

PR #33, opened 2022-03-16
consume the Dockerfile path

sbose78:
This is breaking staging.


sbose78:
Self approving and taking it in.

PR #34, opened 2022-03-17
Proposal: Hacbs split

Michkov:
- Duplicate the pipelines into hacbs folder
- Add sanity-label-check to hacbs pipelines

Proposal how to split hacbs pipelines. Downside is that there is a lot of duplicity. WDYT?


Dannyb48 (commented):
This is probably going to to be silly but could we just template-ize these pipeline yaml files (jsonnet, carvel-ytt, kapitan, or Jinja2) and then process the template using said template processor tool as part of the build-and-push.sh script? That way it generates the two different pipelines during build while still keeping it one file in repo?

jduimovich (approved):
Agree, the duplication is an issue, especially if you update the defaults.
The original idea around the prototype-compliance pipeline was that other pipelines would be created by "plugging in the tasks". So well defined stages in that pipeline would have defaults,  or be "noop out" if absent (optional stages for example) and that it could be customized for end users, including app studio.  We currently have no solution for this, so as a temporary model this is ok, but we need to work on this.

jduimovich:
@Dannyb48 ...
> This is probably going to to be silly but could we just template-ize these pipeline yaml files (jsonnet, carvel-ytt, kapitan, or Jinja2) and then process the template using said template processor tool as part of the build-and-push.sh script? That way it generates the two different pipelines during build while still keeping it one file in repo?

Not silly at all. That is intent of having a well defined default "pipelines as api" -- standard inputs, and standard customization (within reason)  The type of customization would include the ability to override tasks (hacbs) or "opt out" of these if not needed 2) have partners provide and plugin tasks 3) ability to select defaults (to generate variations of pipelines) quick-build, release-build which mix and match steps 4) dev mode and other customization.  In the long term this would be best to evolve into features of Tekton itself, so want to keep that in mind.  The one thing I'd like to avoid is any random new DSL with custom tools (which users have to learn and we have to maintain). I think the use of existing template systems could work. I had considered trying with kustomize initially (have not tried) and again want to do this while thinking about features that we could evolve into Tekton - dynamic lookup of tasks, conditional tasks(hidden), improved, version support and the associate complexity (ie a pipeline.lock type file to know what we generated). 

So, if you have a proposal on something that would make the mechanism of packaging easier I'd be interested. Note, some of the complexity in the packaging is due to bundle limits of 10 in tekton today. Hopefully if that is relaxed, some simplifcation is possible. 
good topic for the cabal.

jsztuka (approved):
Thanks for adding the task in pipelines.
LGTM

Michkov:
Closing in favor of #43

PR #35, opened 2022-03-18
Avoid performance issues when inspecting repo

lcarva:
By default, `skopeo inspect` will fetch a list of all the tags in the
repository. This can be an expensive operation if the repository has a
large amount of tags. A registry may paginate the results causing skopeo
to trigger an unbounded amount of queries to the registry.

Use the `--no-tags` parameter to avoid fetching the tags altogether
since they're not used.


lcarva:
@jsztuka, @jduimovich ptal

Michkov (approved):

PR #36, opened 2022-03-21
Cleanup AppStudio related code

Michkov:
- Remove of devfile-build pipeline -> handled by AppStudio
- Do not set parameters when they are the same as default one
- Move print summary into finally
- Remove dependency on appstudio-utils when it's not needed.
- Add rebuild parameter

Added also script for testing.


jduimovich (commented):
Looks good.
I tried to test the local test-build and the tasks and pipelines directories are missing kustomize.yaml files --- so they don't work. We could just directly apply the two directories instead for local tests

Michkov:
@jduimovich forgot to include the kustomization.yaml file in the commit... I am playing with idea to use kustomization to add organization specific tasks, that's why it's done via kustomization.

jduimovich:
Merge the auth workspace mount MR to your repo and this can be merged.

PR #37, opened 2022-03-21
cosign verify and verify-attestation Tekton tasks

zregvart:
This adds two tasks that take an `IMAGE` and a `PUBLIC_KEY` parameters
and verify the signature and attestation using the `cosign` binary.

I wanted the minimum of a shell script wrapper around `cosign`, so I
ended up not reusing the scripts in `appstudio-utils/util-scripts`.

The only clever bit in this is that the `PUBLIC_KEY` can be specified as
a PEM-encoded (BASE64 with header/footer) or as a valid path/URI
supported by `cosign`. This is to facilitate providing the `PUBLIC_KEY`
by means not directly supported by `cosign`, for example via an
environment variable.

In addition to that unit tests using shellspec[1] are added. To run,
install shellspec locally and run `$ shellspec` from the repository
root.

[1] https://shellspec.info/


simonbaird:
Looks good so far!

appstudio-utils/util-scripts/cosign-verify.sh
```diff
@@ -0,0 +1,37 @@
...
+  local output_file="$3"
+
+  cosign verify \
+    --key "${public_key}" \
+    --output-file "${output_file}" \
+    "${image_reference}"
+}
+
+# When included from shellspec we don't want to invoke cosign-verify
+${__SOURCED__:+return}
```
simonbaird, line 35: I'm confused here, because of the `echo "cosign $*"` don't we expect shellspec does run cosign, or at least its redefined version of it?

tasks/kustomization.yaml
```diff
@@ -11,4 +11,6 @@ resources:
 - buildah.yaml
 - git-clone.yaml
 - s2i-nodejs.yaml
-- s2i-java.yaml
\ No newline at end of file
+- s2i-java.yaml
+- cosign-verify.yaml
```
simonbaird, line 15: I was going to suggest calling it cosign-verify-image-sig for understandably, but actually it's better to stick with the convention established by cosign's own commands.

simonbaird (approved):

joejstuart:
This looks good. Only questions is, do we need the scripts under utils-scripts/tekton-chains anymore? If not, maybe we remove that in this PR?

zregvart:
> This looks good. Only questions is, do we need the scripts under utils-scripts/tekton-chains anymore? If not, maybe we remove that in this PR?

Removed those in [`328ff85` (#37)](https://github.com/redhat-appstudio/build-definitions/pull/37/commits/328ff853462a427f8cc6467f08bc1e27a50a1849), thanks for the suggestion!

lcarva (commented):
Excellent work! Just left a few minor comments/questions.

appstudio-utils/util-scripts/cosign-verify-attestation.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
lcarva, line 1: I always try to use the unofficial [Bash Strict Mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) in my bash scripts. Maybe it could be beneficial for us to start using it here?

tasks/cosign-verify-attestation.yaml
```diff
@@ -0,0 +1,37 @@
...
+      /appstudio-utils/util-scripts/$(context.task.name).sh "$(params.IMAGE)" "$(params.PUBLIC_KEY)" "$(results.VERIFY_ATTESTATION_JSON.path)"
+    name: verify
+    image: appstudio-utils
+    volumeMounts:
+    - name: ca-certificates
+      mountPath: /etc/ssl/certs/
+  volumes:
+  - name: ca-certificates
+    secret:
+      secretName: chains-ca-cert
```
lcarva, line 37: Should this be customizable? Not sure if "chains" is the right name within this context.

zregvart: Yeah, would make sense to use an optional workspace here. That actually gives me an idea to refactor this to use `$SSL_CERT_DIR` provided by Tekton which includes the service CA certificate and the CA bundle.

appstudio-utils/util-scripts/cosign-verify-attestation.sh
simonbaird: Yeah, I don't see any major downsides.

appstudio-utils/util-scripts/cosign-verify-attestation.sh
```diff
@@ -0,0 +1,37 @@
...
+#   <output file>     where to store the result in JSON format
+
+cosign-verify-attestation() {
+  local image_reference="$1"
+
+  local public_key="$2"
+  if [[ "${public_key}" == -----BEGIN* ]]; then
+    local tmp_public_key
+    tmp_public_key="$(mktemp)"
+    trap 'rm '"${tmp_public_key}" EXIT
```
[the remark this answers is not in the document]
simonbaird: Ack, ok.

spec/cosign-verify-attestation_spec.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+eval "$(shellspec -)"
```
Michkov, line 3: not installed

spec/cosign-verify-attestation_spec.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
Michkov, line 1: why it's in spec folder? could you move it to hack or test folder and put into README. When I see spec then I expect rpm spec.

zregvart:
@Michkov thanks for the comments, I've moved the specifications to the `test` directory and added documentation on how to execute those to the README.

simonbaird:
LGTM, nice PR

simonbaird (approved):

simonbaird:
I checked it out and ran the tests, worked good.

```
Running: /usr/bin/bash [bash 5.1.8(1)-release]
....

Finished in 0.11 seconds (user 0.11 seconds, sys 0.03 seconds)
4 examples, 0 failures
```

lcarva (approved):
Beautiful!

PR #38, opened 2022-03-25
Add a task for deprecated image check

staticf0x:


staticf0x:
@dirgim @jsztuka @Dannyb48 @hongweiliu17 PTAL

jsztuka (approved):
Just one nit
Maybe adding `-check` to metadata and file name, to distinguish what is the purpose of the task.
Otherwise LGTM

tasks/deprecated-image.yaml
```diff
@@ -0,0 +1,69 @@
...
+    # project
+    # └── policy
+    # data
+    # └── image_data.json
+    - name: query-pyxis
+      image: registry.access.redhat.com/ubi8/ubi
+      script: |
+        #!/usr/bin/env bash
+        echo "Querying Pyxis for $(inputs.params.source-image)..."
+        http_code=$(curl -s -o /data/image_data.json -w '%{http_code}' "https://catalog.redhat.com/api/containers/v1/repositories/registry/registry.access.redhat.com/repository/$(inputs.params.source-image)")
```
dirgim, line 34: Maybe have the `registry.access.redhat.com` as a parameter to allow better flexibility in future (e.g. the ISV images have a diferent value here).

hongweiliu17: +1, we may have different registry

tasks/deprecated-image.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - name: run-conftest
+      image: quay.io/redhat-appstudio/hacbs-test
+      script: |
+        #!/usr/bin/env sh
+        http_code=$(cat $(results.pyxis-http-code.path))
+
+        if [ "$http_code" == "200" ];
+        then
+          echo "Running conftest..."
+          conftest test /data/image_data.json -p $(inputs.params.policy-dir) -o json > /data/test_result.json && echo "All tests successful" || echo "Some tests failed"
```
hongweiliu17, line 55: Should we emit the task result then later task can consume it?

hongweiliu17:
Left comment in code. Others LGTM.

tasks/deprecated-image.yaml
dirgim: The conftest output can be too large for the Task result size limit. We'll have to work around that in a separate story

tasks/deprecated-image.yaml
staticf0x: Done!

staticf0x:
@dirgim @jsztuka @Dannyb48 @hongweiliu17 How about now?

jsztuka:
Thanks for changes, +1

dirgim (approved):
LGTM

tasks/deprecated-image-check.yaml
```diff
@@ -0,0 +1,72 @@
...
+    #
+    # project
+    # └── policy
+    # data
+    # ├── image_data.json
+    # └── test_result.json
+    - name: run-conftest
+      image: quay.io/redhat-appstudio/hacbs-test
+      script: |
+        #!/usr/bin/env sh
```
Dannyb48, line 52: We should move this If/else logic out of this step and push it up into the first step where we do the query of pyxis. Right now it makes the most sense to the keep the pyxis logic together. That way this step emulates like the label step and just performs the conftest command. Thoughts?

tasks/deprecated-image-check.yaml
```diff
@@ -0,0 +1,72 @@
...
+    # Download Pyxis metadata about the image
+    #
+    # project
+    # └── policy
+    # data
+    # └── image_data.json
+    - name: query-pyxis
+      image: registry.access.redhat.com/ubi8/ubi
+      script: |
+        #!/usr/bin/env bash
```
Dannyb48, line 35: I know it's completely legal to embed script logic within a tekton task but I wonder if we should encapsulate all this pyxis logic into a lightweight script in our hacbs-test repo. That way we can make the changes against the script rather than against the task. Keeps the tasks cleaner and less change since it's always interfacing with the same entry point.   Thoughts?

Dannyb48 (changes requested):
I think we can leave my comment about embed script code for another PR. I don't want to hold this up. But at a minimum I would move the HTTP Code checking in the second step into the first step so it stays with the pyxis logic.

staticf0x:
> I think we can leave my comment about embed script code for another PR. I don't want to hold this up. But at a minimum I would move the HTTP Code checking in the second step into the first step so it stays with the pyxis logic.

If I move the logic to the first step, then how would I make the decisions based on the HTTP code? Because `onError` is not available in stable tekton, I had to put the logic in the second step, so it's possible to exit as success instead of failing the task as a whole.

Dannyb48:
> > I think we can leave my comment about embed script code for another PR. I don't want to hold this up. But at a minimum I would move the HTTP Code checking in the second step into the first step so it stays with the pyxis logic.
> 
> If I move the logic to the first step, then how would I make the decisions based on the HTTP code? Because `onError` is not available in stable tekton, I had to put the logic in the second step, so it's possible to exit as success instead of failing the task as a whole.

But I would think we would want to fail the task as whole? I don't see a need to move on to the second step since this sanity test is so dependent on the success of that pyxis call and the json output? But maybe I'm missing something that isn't obvious to me right now?

If you move the HTTP code check into the first step where you are making the curl call to pyxis, I believe you can make the decision there.  If it's not a 200 code it should be ok to fail out there and echo out the message. And if that curl command is successful then the task will automatically move onto the second step which would run the conftest to validate the json output to the rego policy.

staticf0x:
> But I would think we would want to fail the task as whole? 

I was following the description in the HACBS-92 story, this line:

> If the above query doesn't return anything (404), end the task without failing

So I don't know what's better. Should the task fail for any code other than 200? It would simplify the task a lot. What is your opinion?

dirgim:
> > But I would think we would want to fail the task as whole?
> 
> I was following the description in the HACBS-92 story, this line:
> 
> > If the above query doesn't return anything (404), end the task without failing
> 
> So I don't know what's better. Should the task fail for any code other than 200? It would simplify the task a lot. What is your opinion?

The issue here is that the external Pyxis endpoint does not show results for unreleased repositories. If we failed the task when getting a 404 for the query, we would block teams with new (unreleased) products from getting their products released in the first place by creating a catch-22.

The compromise solution is not to fail the task when getting a 404, hence the requirement in the HACBS-92 story.

@Dannyb48 let me know if this is acceptable to you.

Dannyb48 (approved):
Now that I understand the context that we are trying to handle the 404 scenario where we want to skip running conftests but NOT fail the conftest. LGTM

staticf0x:
@hongweiliu17 @jsztuka @dirgim @Dannyb48 I updated the policy path after https://github.com/redhat-appstudio/hacbs-test/pull/9 do you think we can merge this?

dirgim:
> @hongweiliu17 @jsztuka @dirgim @Dannyb48 I updated the policy path after [redhat-appstudio/hacbs-test#9](https://github.com/redhat-appstudio/hacbs-test/pull/9) do you think we can merge this?

Yes, I'm fine with merging this, go ahead.

jsztuka:
+1

jsztuka (approved):
lgtm

PR #40, opened 2022-03-31
Add a task for optional labels check

staticf0x:
Add a task for optional labels check using this policy: https://github.com/redhat-appstudio/hacbs-test/pull/3

Based on https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/sanity-label-check.yaml


staticf0x:
@dirgim @hongweiliu17 @jsztuka @Dannyb48 PTAL

tasks/optional-label-check.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: optional-label-check
```
Dannyb48, line 5: Let's prefix `sanity-` to the name

tasks/optional-label-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+
+  steps:
+    - name: inspect-image
+      image: quay.io/rsaar/hacbs-test:latest
+      script: "skopeo inspect docker://$(inputs.params.sourceImage) > /tekton/home/image_inspect.json"
+
+    - name: optional-labels-check
+      image: quay.io/rsaar/hacbs-test:latest
```
Dannyb48, line 18: This is pointing towards your personal repo in quay. We may want to point to the`redhat-appstudio`. So maybe we should merge that PR and push a new build up to quay.io

tasks/optional-label-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+
+  steps:
+    - name: inspect-image
+      image: quay.io/rsaar/hacbs-test:latest
+      script: "skopeo inspect docker://$(inputs.params.sourceImage) > /tekton/home/image_inspect.json"
```
Dannyb48, line 15: I wonder if we should output to a different named file like `tekton/home/optional_label_image_inspect.json` or a quick bash if file exists check don't run the skopeo command. That way if we want to POC parallelizing our sanity tasks we can avoid a race condition with skopeo until hacbs-220 is complete? Thoughts @staticf0x @hongweiliu17 @jsztuka @dirgim

tasks/optional-label-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: optional-label-check
+spec:
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+
+  steps:
+    - name: inspect-image
```
Michkov, line 13: I do not see here reason of splitting this task into two steps. They are using the same image.

tasks/optional-label-check.yaml
Michkov, line 15: No volume is defined which means that `/tekton/home/image_inspect.json` is not persisted for the next step.

Michkov (changes requested):

tasks/optional-label-check.yaml
hongweiliu17: Checking if file exists is a good point. let me add to my hacbs-220

hongweiliu17: It may be saved in workspace to shared in multiple tasks (see https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#workspaces-in-pipelines-and-pipelineruns) if we don't need to always keep the json file.

hongweiliu17: In my image inspection tasks https://github.com/redhat-appstudio/build-definitions/pull/42, I save the json file to workspace

dirgim: We have a bit of a race condition between this PR and #42 - ideally this task would be `runAfter` the inspection task in the pipeline and would just read the inspection JSON from the workspace.

dirgim: If we want to make this task optionally standalone, we'd then first need to check if the skopeo inspection output is available in the workspace and only if it's not there run the inspection part.

tasks/optional-label-check.yaml
staticf0x: Added

tasks/optional-label-check.yaml
staticf0x: Good point, merged

tasks/optional-label-check.yaml
staticf0x: So far I merged those two steps into one, which also fixes that I forgot the shared volume. I guess that I'll wait until #42 to potentially reuse that one step?

staticf0x: I changed the file name, good point @Dannyb48

tasks/sanity-optional-label-check.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sanity-optional-label-check
+spec:
+  params:
+    - name: sourceImage
```
hongweiliu17, line 8: One thing comes to my mind, `sourceImage` here is like `quay.io/org_name/repo_name', but in https://github.com/redhat-appstudio/build-definitions/pull/38/files, it has `registry` and `source-image` and then combine the two params to query in Pyxis. Or use only one param in https://github.com/redhat-appstudio/build-definitions/pull/38/files?
Maybe I am thinking too much.

staticf0x: That's a good point. I think that with Pyxis, we're basically filling two URL params, which then makes sense to have as two separate task params. In this one though, `sourceImage` is more free to be specified as someone wishes. I'd personally leave it as it is. What do others think?

tasks/sanity-optional-label-check.yaml
```diff
@@ -0,0 +1,17 @@
...
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
+
+  steps:
+    - name: optional-label-check
+      image: quay.io/redhat-appstudio/hacbs-test:latest
+      script: |
+        skopeo inspect docker://$(inputs.params.sourceImage) > /tekton/home/optional_label_image_inspect.json
```
Michkov, line 16: nit: you can add to speed up and improve stability `--no-tags`

Michkov, line 16: I would use default workdir, `skopeo inspect --no-tags docker://$(inputs.params.sourceImage > optional_label_image_inspect.json`

tasks/sanity-optional-label-check.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sanity-optional-label-check
+spec:
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/cvpops/test-index:v4.10
```
Michkov, line 10: I would avoid using default when we really want to change it.

tasks/sanity-optional-label-check.yaml
staticf0x: done

staticf0x: done

tasks/sanity-optional-label-check.yaml
staticf0x: done

Michkov:
Please rebase and include tasks/sanity-optional-label-check.yaml in tasks/kustomization.yaml. Also add it into `pipelines/hacbs/hacbs-test.yaml` and test using `./hack/test-builds.sh hacbs`

staticf0x:
> Please rebase and include tasks/sanity-optional-label-check.yaml in tasks/kustomization.yaml. Also add it into `pipelines/hacbs/hacbs-test.yaml` and test using `./hack/test-builds.sh hacbs`

Rebased and added, however `./hack/test-builds.sh hacbs` skips all the jobs for some reason, why is that?

Michkov:
@staticf0x probably because the image is already built, you can remove the by running `oc delete is --all`

staticf0x:
Thanks, that was it. The tests pass now.

tasks/sanity-optional-label-check.yaml
```diff
@@ -0,0 +1,19 @@
...
+  workspaces:
+    - name: workspace
+
+  steps:
+    - name: optional-label-check
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      script: |
+        skopeo inspect --no-tags docker://$(inputs.params.sourceImage) > optional_label_image_inspect.json
+        /usr/bin/conftest test optional_label_image_inspect.json --policy /project/image/optional-labels.rego --output=json
```
Michkov, line 19: the result:
```
step-optional-label-check
+ skopeo inspect --no-tags docker://image-registry.openshift-image-registry.svc:5000/project/single-container-app:37daed8
+ /usr/bin/conftest test optional_label_image_inspect.json --policy /project/image/optional-labels.rego --output=json
[
	{
		"filename": "optional_label_image_inspect.json",
		"namespace": "main",
		"successes": 0
	}
]
```
Having no idea if the image is ok or not.

Probably there should be `--no-fail` as in sanity-label-check. Result should be stored as result -> HACBS_TEST_OUTPUT

staticf0x:
@Michkov As @dirgim mentioned, we can actually reuse `sanity-label-check` and just pass the correct namespace. PTAL now, if the test is enough for HACBS-201 requirements.

dirgim:
LGTM in the new form

Michkov (approved):

PR #42, opened 2022-04-01
Add tasks for image inspection

hongweiliu17:
@dirgim @Dannyb48 @jsztuka @staticf0x How about your ideas? Thank you!

### DIFF
--- a/tasks/sanity-inspect-image.yaml
+++ b/tasks/sanity-inspect-image.yaml
@@ -1,0 +1,48 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sanity-inspect-image
+spec:
+  description: >-
+    Get manifest data for the source image and its base image to workspace
+  params:
+    - name: IMAGE_URL
+      description: the fully qualified image name
+      type: string
+  results:
+    - description: Base image the source image is built from
+      name: BASE_IMAGE
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      IMAGE_INSPECT=/tmp/image_inspect.json
+      BASE_IMAGE_INSPECT=/tmp/base_image_inspect.json
+      RAW_IMAGE_INSPECT=/tmp/raw_image_inspect.json
+
+      echo "Inspecting manifest for source image $(params.IMAGE_URL)"
+      skopeo inspect --no-tags docker://$(params.IMAGE_URL) > $IMAGE_INSPECT
+      skopeo inspect --no-tags --raw docker://$(params.IMAGE_URL) > $RAW_IMAGE_INSPECT
+
+      echo "Getting base image manifest for source image $(params.IMAGE_URL)"
+      BASE_IMAGE_NAME="$(jq -r ".annotations.\"org.opencontainers.image.base.name\"" $RAW_IMAGE_INSPECT)"
+      BASE_IMAGE_DIGEST="$(jq -r ".annotations.\"org.opencontainers.image.base.digest\"" $RAW_IMAGE_INSPECT)"
+      if [ $BASE_IMAGE_NAME == 'null' ]; then
+        echo "Cannot get base image info from 'annotations'"
+        echo "Trying to get base image info from 'Labels'"
+        BASE_IMAGE_NAME="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_INSPECT)"
+        BASE_IMAGE_DIGEST="$(jq -r ".annotations.\"org.opencontainers.image.base.digest\"" $IMAGE_INSPECT)"
+        if [ "$BASE_IMAGE_NAME" == 'null' ]; then
+          echo "Cannot get base image info from 'Labels', please check the source image $(params.IMAGE_URL)"
+          exit 0
+        fi
+      fi
+      if [ -z "$BASE_IMAGE_NAME" ]; then
+        echo "Source image $(params.IMAGE_URL) is built from scratch, so there is no base image"
+        exit 0
+      fi
+      BASE_IMAGE="${BASE_IMAGE_NAME%:*}@$BASE_IMAGE_DIGEST"
+      echo "The base image is $BASE_IMAGE, get its manifest now"
+      skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
+      echo "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
--- a/pipelines/hacbs/hacbs-test.yaml
+++ b/pipelines/hacbs/hacbs-test.yaml
@@ -1,6 +1,17 @@
 - op: add
   path: /spec/tasks/-
   value:
+    name: sanity-inspect-image
+    runAfter:
+      - build-container
+    taskRef:
+      name: sanity-inspect-image
+    params:
+    - name: IMAGE_URL
+      value: $(params.output-image)
+- op: add
+  path: /spec/tasks/-
+  value:
     name: sanity-label-check
     runAfter:
       - build-container
--- a/tasks/kustomization.yaml
+++ b/tasks/kustomization.yaml
@@ -13,3 +13,4 @@ resources:
 - utils-task.yaml
 - cosign-verify.yaml
 - cosign-verify-attestation.yaml
+- sanity-inspect-image.yaml


tasks/sanity-inspect-parent-image.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - name: sourceImage
+    description: the fully qualified image name
+    default: quay.io/hongweiliu/skopeo-inspect-test:latest
+    type: string
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      # 'org.opencontainers.image.base.name' exists for OCI format image only but does not exist in docker format image
+      # So we should check if  when the image is built by docker format image
```
Dannyb48, line 17: It doesn't look like the code to check the docker format image is in place? If I remember correctly, in the first day of the M4 F2F someone suggested we could use this to determine the base image

`skopeo inspect --config --no-tags docker://quay.io/mkovarik/spring-petclinic:dc80a7f | grep FROM`

Although using that example it shows that it was using an image from a secure registry or probably redhat access catalog.

`image-registry.openshift-image-registry.svc:5000/openshift/java:openjdk-11-ubi8`

That being said that example petclinic app image is using label `io.openshift.s2i.build.image` so I wonder if this another label we need to look for to determine the base image?

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,24 @@
...
+  params:
+    - name: sourceImage
+      description: the fully qualified image name
+      default: quay.io/hongweiliu/skopeo-inspect-test:latest
+      type: string
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      if [ -f "$(workspaces.conftest-ws.path)/image_inspect.json" ];then
```
Dannyb48, line 16: Same comment as the deprecated image check around the design patterns. I know it's perfectly legal to embed scripting in a tekton task but I'm leaning towards us putting a lot of this logic in scripts/tools in the hacbs-test repo. So in this case encapsulate all this skopeo logic from both tasks into a lightweight script in our hacbs-test repo. That way we can make the changes against the script rather than against the task. Keeps the tasks cleaner and less change since it's always interfacing with the same entry point. Thoughts?

Dannyb48 (commented):
I'm actually thinking we should merge this two task into a single task sanity-inspect-image. And parameterize the base image check with a boolean to turn the parent image checking on/off.

tasks/sanity-inspect-parent-image.yaml
```diff
@@ -0,0 +1,28 @@
...
+      if [ -f "$(workspaces.conftest-ws.path)/parent_image_inspect.json" ];then
+        echo "$(workspaces.conftest-ws.path)/parent_image_inspect.json exists"
+      else
+        echo "Deriving parent image for source image: $(params.sourceImage)"
+        PARENT_IMAGE="$(skopeo inspect --no-tags --config docker://$(inputs.params.sourceImage) | jq -r ".config.Labels.\"org.opencontainers.image.base.name\"")"
```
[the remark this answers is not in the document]
hongweiliu17: Yes. Update. Thanks.

tasks/sanity-inspect-parent-image.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sanity-inspect-parent-image
```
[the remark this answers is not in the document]
hongweiliu17: Updated

tasks/sanity-inspect-parent-image.yaml
```diff
@@ -0,0 +1,28 @@
...
+    default: quay.io/hongweiliu/skopeo-inspect-test:latest
+    type: string
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      # 'org.opencontainers.image.base.name' exists for OCI format image only but does not exist in docker format image
+      # So we should check if  when the image is built by docker format image
+      if [ -f "$(workspaces.conftest-ws.path)/parent_image_inspect.json" ];then
+        echo "$(workspaces.conftest-ws.path)/parent_image_inspect.json exists"
```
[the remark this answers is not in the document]
hongweiliu17: Updated

tasks/sanity-inspect-parent-image.yaml
hongweiliu17: `skopeo inspect --config --no-tags docker://quay.io/mkovarik/spring-petclinic:dc80a7f | grep FROM` is not deriving from the `From` statement inf Dockerfile, it is reading from comment when building, so it is not real `From`.
```
 % skopeo inspect --config --no-tags docker://quay.io/mkovarik/spring-petclinic:dc80a7f | grep FROM
            "comment": "FROM image-registry.openshift-image-registry.svc:5000/openshift/java:openjdk-11-ubi8",
```

tasks/sanity-inspect-image.yaml
hongweiliu17: Thanks. I understand. I checked the script, should we keep consistent with the tasks in other workstream? Or we may get a agreement for it.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,38 @@
...
+      else
+        echo "Inspecting source image $(params.SOURCE_IMAGE)"
+        if [ -f "$(workspaces.conftest-ws.path)/image_inspect.json" ];then
+          echo "$(workspaces.conftest-ws.path)/image_inspect.json exists"
+        else
+          skopeo inspect --no-tags docker://$(params.SOURCE_IMAGE) > $(workspaces.conftest-ws.path)/image_inspect.json
+        fi
+      fi
+  workspaces:
+  - mountPath: /workspace/conftest-ws
```
Michkov, line 37: no need to define mountPath. It's not used directly.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,38 @@
...
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      if [ $(params.BASE) == 'true' ];then
+        echo "Inspecting base image for source image: $(params.SOURCE_IMAGE)"
+        if [ -f "$(workspaces.conftest-ws.path)/base_image_inspect.json" ];then
+          echo "$(workspaces.conftest-ws.path)/base_image_inspect.json exists"
+        else
+          BASE_IMAGE="$(cat $(workspaces.conftest-ws.path)/image_inspect.json | jq -r ".Labels.\"org.opencontainers.image.base.name\"")"
```
Michkov, line 25: `BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $(workspaces.conftest-ws.path)/image_inspect.json"` -> no need to use `cat`

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,38 @@
...
+      description: is to get base image?
+      default: 'false'
+      type: string
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      if [ $(params.BASE) == 'true' ];then
+        echo "Inspecting base image for source image: $(params.SOURCE_IMAGE)"
+        if [ -f "$(workspaces.conftest-ws.path)/base_image_inspect.json" ];then
```
Michkov, line 22: Put `$(workspaces.conftest-ws.path)/base_image_inspect.json` into variable, rather then putting it everywhere

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: sanity-inspect-image
+spec:
+  params:
+    - name: SOURCE_IMAGE
+      description: the fully qualified image name
+      default: quay.io/hongweiliu/skopeo-inspect-test:latest
+      type: string
+    - name: BASE
+      description: is to get base image?
```
jsztuka, line 13: Maybe change description to something like: **Represents base image**

tasks/sanity-inspect-image.yaml
hongweiliu17: Yes, thank you.

tasks/sanity-inspect-image.yaml
hongweiliu17: Thank you!

tasks/sanity-inspect-image.yaml
hongweiliu17: Yes. Thank you.

tasks/sanity-inspect-image.yaml
hongweiliu17: Thank you, how about changing to `Inspecting base image for source image`?

jsztuka: I think that describing variable with process that happens to it is not necessary. What do you think?

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,40 @@
...
+    script: |
+      IMAGE_INSPECT_FILE=$(workspaces.conftest-ws.path)/image_inspect.json
+      BASE_IMAGE_INSPECT_FILE=$(workspaces.conftest-ws.path)/base_image_inspect.json
+      if [ $(params.BASE) == 'true' ];then
+        echo "Inspecting base image for source image: $(params.SOURCE_IMAGE)"
+        BASE_IMAGE_INSPECT_FILE=$(workspaces.conftest-ws.path)/base_image_inspect.json
+        if [ -f "$BASE_IMAGE_INSPECT_FILE" ];then
+          echo "$BASE_IMAGE_INSPECT_FILE exists"
+        else
+          BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_INSPECT_FILE)"
```
dirgim, line 28: Aren't the `org.opencontainers.image.base.name` and `org.opencontainers.image.base.digest` items supposed to be under `annotations` in the skopeo raw output?

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,40 @@
...
+          skopeo inspect --no-tags docker://$BASE_IMAGE > $BASE_IMAGE_INSPECT_FILE
+        fi
+      else
+        echo "Inspecting source image $(params.SOURCE_IMAGE)"
+        if [ -f "$IMAGE_INSPECT_FILE" ];then
+          echo "$IMAGE_INSPECT_FILE exists"
+        else
+          skopeo inspect --no-tags docker://$(params.SOURCE_IMAGE) > $IMAGE_INSPECT_FILE
```
dirgim, line 36: I think we want to run `skopeo inspect --no-tags --raw docker://$(params.SOURCE_IMAGE)` here in addition to this (so both with and without `--raw` option) to get the image config metadata.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,40 @@
...
+      type: string
+    - name: BASE
+      description: Inspecting base image for source image
+      default: 'false'
+      type: string
+  steps:
+  - name: inspect-image
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    script: |
+      IMAGE_INSPECT_FILE=$(workspaces.conftest-ws.path)/image_inspect.json
```
jsztuka, line 20: I'm not sure if `conftest-ws` fits in this context, since we are not using conftest here at all, maybe sanity-ws makes more sense?

hongweiliu17: Yes. `sanity-ws` makes sense.

tasks/sanity-inspect-image.yaml
hongweiliu17: Yes, it exists in the data returned by `skopeo inspect --no-tags --raw docker://quay.io/mkovarik/spring-petclinic:dc80a7f`.
I would build an oci image by the tasks in this repo and them compare the manifest data the docker and oci

hongweiliu17: I tested the `buildah bud` with our tkn pipeline/tasks file for the same Dockerfile, if the format is oci, there are annotations with `org.opencontainers.image.base.digest` and `org.opencontainers.image.base.name` in the data gotten by `skopeo inspect --raw`, but for docker format image, there are not these info.

tasks/sanity-inspect-image.yaml
hongweiliu17: Yes. I revise the code and use the logic as below, try to get base image from `annotations`, if not found, will try to get base image from `Labels`, if not found, it will fail. Does it make sense?
```
BASE_IMAGE="$(jq -r ".annotations.\"org.opencontainers.image.base.name\"" $IMAGE_RAW_MANIFEST)"
if [ $BASE_IMAGE == 'null' ]; then
  echo "Can not get base image info from 'annotations'"
  echo "Trying to get base image info from 'Labels'"
  BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_MANIFEST)"
  if [ $BASE_IMAGE == 'null' ]; then
    echo "Can not get base image info from 'Labels', please check the source image $(params.SOURCE_IMAGE)"
    exit 1
  fi
fi
```

tasks/sanity-inspect-image.yaml
hongweiliu17: Yes. You are right.
And since the `base image` should exist by default, so I revise the code and will get the the manifest for the source image and base image at the same time without the param. Could you take a look?

deploy.sh
```diff
@@ -0,0 +1,3 @@
+oc apply -f tasks/sanity-inspect-image.yaml
```
dirgim, line 1: I'm guessing this script is meant for local testing, so it can be omitted from the PR.

hongweiliu17: Yes! Thanks. Sorry, let me remove it.

tasks/sanity-inspect-image.yaml
Dannyb48: I think EC has wrapped some of their stuff in wrapper scripts that their task interfaces with https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/cosign-verify-attestation.yaml#L30 which is a script https://github.com/redhat-appstudio/build-definitions/blob/main/appstudio-utils/util-scripts/cosign-verify-attestation.sh

But we don't have to do it as part of this PR. We can just merge it and put a ticket in the backlog to refactor some of these things out into scripts in our hacbs-test repo.

tasks/sanity-inspect-image.yaml
Dannyb48: Yes, I think this looks like a good solution.

tasks/sanity-inspect-image.yaml
hongweiliu17: @Dannyb48 Thank you!

jsztuka:
What is the reason behind six newly added files?

hongweiliu17:
> What is the reason behind six newly added files?

Thanks. These are the files I used to test my code. Deleted.

hongweiliu17:
@dirgim @Dannyb48 @jsztuka I add some code for the image which might be built `From scratch` and pass the `BASE_IMAGE` to task result, then the following tasks can make use of it. Does it make sense?

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+        echo "Can not get base image info from 'annotations'"
+        echo "Trying to get base image info from 'Labels'"
+        BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_MANIFEST)"
+        if [ $BASE_IMAGE == 'null' ]; then
+          echo "Can not get base image info from 'Labels', please check the source image $(params.SOURCE_IMAGE)"
+          exit 1
+        fi
+      fi
+      if [ -z $BASE_IMAGE ]; then
+        echo "Sourc image $(params.SOURCE_IMAGE) is built from scratch, so there is not base image"
```
dirgim, line 41: nit/typo: 'Sourc' _> 'Source'

dirgim:
looks good to me apart from the tiny nit

tasks/sanity-inspect-image.yaml
hongweiliu17: Thank you! Updated.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: Task
+metadata:
+  name: sanity-inspect-image
+spec:
+  description: >-
+    Get manifest data for the source image and its base image to workspace
+  params:
+    - name: SOURCE_IMAGE
+      description: the fully qualified image name
+      default: quay.io/hongweiliu/conftest_oci:latest
```
staticf0x, line 12: Shouldn't we remove a default? Especially if it's your personal repo

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+      IMAGE_RAW_MANIFEST=$(workspaces.sanity-ws.path)/image_raw_manifest.json
+
+      echo "Inspecting manifest for source image $(params.SOURCE_IMAGE)"
+      skopeo inspect --no-tags docker://$(params.SOURCE_IMAGE) > $IMAGE_MANIFEST
+      skopeo inspect --no-tags --raw docker://$(params.SOURCE_IMAGE) > $IMAGE_RAW_MANIFEST
+
+      echo "Getting base image manifest for source image $(params.SOURCE_IMAGE)"
+      BASE_IMAGE="$(jq -r ".annotations.\"org.opencontainers.image.base.name\"" $IMAGE_RAW_MANIFEST)"
+      if [ $BASE_IMAGE == 'null' ]; then
+        echo "Can not get base image info from 'annotations'"
```
staticf0x, line 32:
```suggestion
        echo "Cannot get base image info from 'annotations'"
```

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+        echo "Trying to get base image info from 'Labels'"
+        BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_MANIFEST)"
+        if [ $BASE_IMAGE == 'null' ]; then
+          echo "Can not get base image info from 'Labels', please check the source image $(params.SOURCE_IMAGE)"
```
staticf0x, line 36:
```suggestion
          echo "Cannot get base image info from 'Labels', please check the source image $(params.SOURCE_IMAGE)"
```

tasks/sanity-inspect-image.yaml
hongweiliu17: Thanks. I thought someone would test it with one example. Let me remove it.

tasks/sanity-inspect-image.yaml
hongweiliu17: Thanks.

tasks/sanity-inspect-image.yaml
hongweiliu17: Thanks.

jsztuka (approved):
Lgtm

staticf0x (approved):
LGTM

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+          exit 1
```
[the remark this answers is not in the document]
hongweiliu17: I think it can be gated in former step, we need `BASE_IMAGE` exists, we can still fail it in case nobody figure it out before this step.

hongweiliu17:
@Dannyb48 Could you take a look again? Thank you!

tasks/sanity-inspect-image.yaml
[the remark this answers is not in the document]
dirgim: Based on the latest information from the pipeline service/build teams, we are not supposed to fail any of our tasks. If we can't get the `BASE_IMAGE`, the subsequent tasks that run Conftest can report that the `BASE_IMAGE` is missing.
So, removing this `exit 1` should be all we need to do here.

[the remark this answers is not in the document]
hongweiliu17: Thank you for the info. Code is updated.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,46 @@
...
+      BASE_IMAGE="$(jq -r ".annotations.\"org.opencontainers.image.base.name\"" $IMAGE_RAW_MANIFEST)"
+      if [ $BASE_IMAGE == 'null' ]; then
+        echo "Cannot get base image info from 'annotations'"
+        echo "Trying to get base image info from 'Labels'"
+        BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_MANIFEST)"
+        if [ $BASE_IMAGE == 'null' ]; then
+          echo "Cannot get base image info from 'Labels', please check the source image $(params.SOURCE_IMAGE)"
+        fi
+      fi
+      if [ -z $BASE_IMAGE ]; then
```
Michkov, line 38: must be `[ -z "$BASE_IMAGE" ]`

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,46 @@
...
+      echo "Inspecting manifest for source image $(params.SOURCE_IMAGE)"
+      skopeo inspect --no-tags docker://$(params.SOURCE_IMAGE) > $IMAGE_MANIFEST
+      skopeo inspect --no-tags --raw docker://$(params.SOURCE_IMAGE) > $IMAGE_RAW_MANIFEST
+
+      echo "Getting base image manifest for source image $(params.SOURCE_IMAGE)"
+      BASE_IMAGE="$(jq -r ".annotations.\"org.opencontainers.image.base.name\"" $IMAGE_RAW_MANIFEST)"
+      if [ $BASE_IMAGE == 'null' ]; then
+        echo "Cannot get base image info from 'annotations'"
+        echo "Trying to get base image info from 'Labels'"
+        BASE_IMAGE="$(jq -r ".Labels.\"org.opencontainers.image.base.name\"" $IMAGE_MANIFEST)"
```
Michkov, line 33: must be based on digest, eg. `registry.access.redhat.com/ubi8/ubi:latest` will download newest image.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,46 @@
...
+      fi
+      if [ -z $BASE_IMAGE ]; then
+        echo "Source image $(params.SOURCE_IMAGE) is built from scratch, so there is no base image"
+      else
+        echo "The base image is $BASE_IMAGE, get its manifest now"
+        skopeo inspect --no-tags docker://$BASE_IMAGE > $BASE_IMAGE_MANIFEST
+      fi
+      echo "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
+  workspaces:
+  - name: sanity-ws
```
Michkov, line 46: using workspaces is very limiting, currently the pipelines does not count with `sanity-ws`, also we have access only to RWO pvc so it will not be possible to run tests in parallel. Ideally all test could be done in the task/step.

dirgim: That's very disappointing news - are we then completely limited to Tekton results to communicate between tasks?

Michkov: I am using [prototype-build-compliance](https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/base/prototype-build-compliance.yaml) pipeline as a reference. There are currently only 3 workspaces (only 1 pvc based, others are usually secrets) and tasks are usually complete.

It would be possible to re-use workspace with the source code, but in that case you cannot use it in parallel tasks.

CC @jduimovich @amfred

Michkov:
Please rebase and include tasks/sanity-inspect-image.yaml in tasks/kustomization.yaml. Also add it into pipelines/hacbs/hacbs-test.yaml and test using `./hack/test-builds.sh hacbs`

tasks/sanity-inspect-image.yaml
Dannyb48: @Michkov @jduimovich @amfred Would this feature alleviate the issue?

https://tekton.dev/vault/pipelines-v0.15.2/workspaces/#specifying-workspace-order-in-a-pipeline-and-affinity-assistants

tasks/sanity-inspect-image.yaml
hongweiliu17: Ok, thanks. It work as well without ""

tasks/sanity-inspect-image.yaml
hongweiliu17: Ok. So let me just use the same solution with https://github.com/redhat-appstudio/build-definitions/pull/55 for now.

tasks/sanity-inspect-image.yaml
hongweiliu17: Ok, thanks.

tasks/sanity-inspect-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+          exit 0
+        fi
+      fi
+      if [ -z "$BASE_IMAGE_NAME" ]; then
+        echo "Source image $(params.IMAGE_URL) is built from scratch, so there is no base image"
+        exit 0
+      fi
+      BASE_IMAGE="${BASE_IMAGE_NAME%:*}@$BASE_IMAGE_DIGEST"
+      echo "The base image is $BASE_IMAGE, get its manifest now"
+      skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT
```
Michkov, line 47: When running via s2i builder it's failing since it's trying to inspect internal registry:
```
+ skopeo inspect --no-tags docker://image-registry.openshift-image-registry.svc:5000/openshift/nodejs@sha256:1451b4849c535edfc238b11e85d8bbc9599516fb9d079bb8e116051b532a04a0
time="2022-04-21T08:50:26Z" level=fatal msg="Error parsing image name \"docker://image-registry.openshift-image-registry.svc:5000/openshift/nodejs@sha256:1451b4849c535edfc238b11e85d8bbc9599516fb9d079bb8e116051b532a04a0\": reading manifest sha256:1451b4849c535edfc238b11e85d8bbc9599516fb9d079bb8e116051b532a04a0 in image-registry.openshift-image-registry.svc:5000/openshift/nodejs: manifest unknown: manifest unknown"
```

Probably it could be enough to ignore this right now -> `skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true`

hongweiliu17: Ok, thanks. Let use it for now. Updated.

PR #46, opened 2022-04-08
Add opa cli tool to appstudio-utils docker image

simonbaird:
OPA is the tool used for for validating rego policies such as the
ones that will be used by the Enterprise Contract to validate
pipeline runs.

It's not clear yet whether the final version of the Enterprise
Contract tasks will use this base image, but there's a reasonable
chance they will, and either way it seems like this is generally
useful tool to add alongside the others.

See also:
* https://github.com/open-policy-agent/opa/releases
* https://www.openpolicyagent.org/
* https://issues.redhat.com/browse/HACBS-233
* https://issues.redhat.com/browse/HACBS-235


simonbaird:
I was wondering if anyone thinks we should verify the checksum of the downloaded file when doing a build. Not in this PR, but I could do it if anyone thinks it's worthwhile.

Would be conceptually similar to this:
* https://github.com/hacbs-contract/ec-policies/blob/main/Makefile#L35

simonbaird:
Choose one or other of this and #47 (not both).

simonbaird:
PTAL @Michkov @zregvart @bcaton85 @joejstuart

joejstuart:
> I was wondering if anyone thinks we should verify the checksum of the downloaded file when doing a build. Not in this PR, but I could do it if anyone thinks it's worthwhile.
> 
> Would be conceptually similar to this:
> 
> * https://github.com/hacbs-contract/ec-policies/blob/main/Makefile#L35
> 
> Update: I made an alternative PR that does check the sha: #47 .

I think this is a good idea.

bcaton85 (approved):
LGTM

PR #47, opened 2022-04-08
Add opa to appstudio-utils docker image with sha check

simonbaird:
OPA is the tool used for for validating rego policies such as the
ones that will be used by the Enterprise Contract to validate
pipeline runs.

It's not clear yet whether the final version of the Enterprise
Contract tasks will use this base image, but there's a reasonable
chance they will, and either way it seems like this is generally
useful tool to add alongside the others.

See also:
* https://github.com/open-policy-agent/opa/releases
* https://www.openpolicyagent.org/
* https://issues.redhat.com/browse/HACBS-233
* https://issues.redhat.com/browse/HACBS-235


simonbaird:
Actually the checksum check doesn't work like that, maybe because the filename is different.

simonbaird:
Fixed now - problem was I had the sha for the static version but fetched the non-static version.

simonbaird:
Choose one or other of this and #46 (not both).

simonbaird:
PTAL @Michkov @zregvart @bcaton85 @joejstuart

zregvart:
I'd go with this version, perhaps we could (in a different PR) add hash checks to other bits downloaded.

simonbaird:
Replaced by #48 - sorry for the noise.

PR #48, opened 2022-04-11
Add extra tools to utils image for contract

simonbaird:
* OPA cli tool: Used for for validating rego policies such
  as the ones that will be used by the Enterprise Contract to
  validate pipeline runs. Also verifies the checksum because
  it seems like a good idea.

* Newer jq: Install it directly rather than with dnf. I was hitting
  a weird jq problem when running the contract fetch scripts (see
  upcoming commits) in the container and it went away when I
  upgraded to jq-1.6. The dnf installed version was jq-1.5.

* Git: I want to use git to checkout rego policy files.

See also:
- https://github.com/open-policy-agent/opa/releases
- https://www.openpolicyagent.org/
- https://issues.redhat.com/browse/HACBS-233
- https://issues.redhat.com/browse/HACBS-235

### DIFF
--- a/appstudio-utils/Dockerfile
+++ b/appstudio-utils/Dockerfile
@@ -1,12 +1,19 @@
 FROM registry.access.redhat.com/ubi8/ubi
 
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o /usr/bin/jq && chmod +x /usr/bin/jq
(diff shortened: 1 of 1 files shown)


simonbaird:
FYI @Michkov @jduimovich .

simonbaird:
https://github.com/simonbaird/build-definitions/pull/1 is some WIP that shows why I want to add these to the base image.

simonbaird:
Ack, thanks @Michkov .

PR #49, opened 2022-04-12
Add yamllint checking before pull request is merged

jinqi7:
Add yamllint checking for the tasks and pipelines yaml files in the repo

Signed-off-by: jinqi <jinqi@dell-per740xd-16.lab.eng.pek2.redhat.com>

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -7,6 +7,7 @@ metadata:
     pipelinesascode.tekton.dev/on-event: "pull_request"
     pipelinesascode.tekton.dev/on-target-branch: "main"
     pipelinesascode.tekton.dev/task: "git-clone"
+    pipelinesascode.tekton.dev/task-2: "yaml-lint"
     pipelinesascode.tekton.dev/max-keep-runs: "5"
 spec:
   params:
@@ -32,21 +33,17 @@ spec:
             value: $(params.repo_url)
           - name: revision
             value: $(params.revision)
-      - name: noop-task
+      - name: yaml-lint-check
         runAfter:
           - fetch-repository
+        taskRef:
+          name: yaml-lint
         workspaces:
-          - name: source
+          - name: shared-workspace
             workspace: workspace
-        taskSpec:
-          workspaces:
-            - name: source
-          steps:
-            - name: noop-task
-              image: registry.access.redhat.com/ubi8/ubi-micro:8.4
-              workingDir: $(workspaces.source.path)
-              script: |
-                exit 0
+        params:
+        - name: args
+          value: ["."]
   workspaces:
     - name: workspace
       persistentVolumeClaim:
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,35 @@
+ignore: |
+  /vendor
+
+rules:
+  braces: enable
+  brackets: enable
+  colons: enable
+  commas: enable
+  comments:
+    level: warning
+  comments-indentation:
+    level: warning
+  document-end: disable
+  document-start: disable
+  empty-lines: enable
+  empty-values: enable
+  hyphens: enable
+  key-duplicates: enable
+  key-ordering: disable
+  line-length: disable
+  new-line-at-end-of-file: disable
+  new-lines: enable
+  octal-values: enable
+  quoted-strings: disable
+  trailing-spaces: disable
+  truthy:
+    level: warning
+
+  # accept both     key:
+  #                   - item
+  #
+  # and             key:
+  #                 - item
+  indentation:
+    indent-sequences: whatever
--- a/pipelines/base/docker-build.yaml
+++ b/pipelines/base/docker-build.yaml
@@ -3,9 +3,9 @@ kind: Pipeline
 metadata:
   name: docker-build
   labels:
-    "pipelines.openshift.io/used-by" : "build-cloud"
-    "pipelines.openshift.io/runtime" : "generic"
-    "pipelines.openshift.io/strategy" : "docker"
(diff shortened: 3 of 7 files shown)


.tekton/pull-request.yaml
```diff
@@ -32,9 +32,20 @@ spec:
...
           - fetch-repository
+        taskRef:
+          name: yaml-lint
+        workspaces:
+          - name: source
+            workspace: workspace
+        params:
+        - name: args
+          value: ["."]
+      - name: noop-task
```
Michkov, line 46: noop-task can be removed

tasks/yaml-lint.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: tekton.dev/v1beta1
```
Michkov, line 1: include this task from tekton hub via pipelinesAsCode annotation -> https://pipelinesascode.com/docs/guide/resolver/#remote-task-support

Michkov (changes requested):

tasks/yaml-lint.yaml
jinqi7: Added below line in pull-request.yaml -
pipelinesascode.tekton.dev/task: "yaml-lint"

.tekton/pull-request.yaml
jinqi7: Removed.

Michkov:
@jinqi7 CI is failing on    ` declared workspace "shared-workspace" is required but has not been bound`

jinqi7:
> @jinqi7 CI is failing on ` declared workspace "shared-workspace" is required but has not been bound`
The PR is updated. But I still met "User jinqi7 is not allowed to run CI on this repo" after I joined "redhat-appstudio" org.  Is it as expected? Thanks

.tekton/pull-request.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     pipelinesascode.tekton.dev/on-event: "pull_request"
     pipelinesascode.tekton.dev/on-target-branch: "main"
     pipelinesascode.tekton.dev/task: "git-clone"
+    pipelinesascode.tekton.dev/task: "yaml-lint"
```
Michkov, line 10: to avoid issue with yaml validation, please use  `pipelinesascode.tekton.dev/task-2: "yaml-lint"`

tasks/yaml-lint.yaml
Michkov: This file can be now removed

Dannyb48: @Michkov would it make sense to move this task to a task folder under the pipeline as code directory Since this meant only for CI not to be distributed with hacbs pipelines?

Dannyb48 (changes requested):
I don't think we need to add this task to the kustomize tank file. The kustomize file is used when building the hacks pipelines. This task will only be used for the ci pipeline as code for this repo.

jinqi7:
> I don't think we need to add this task to the kustomize tank file. The kustomize file is used when building the hacks pipelines. This task will only be used for the ci pipeline as code for this repo.

I think the task may not be limited to be used by ci pipeline as code and it may be used by the applications being build. Please correct me if I am wrong. Thanks

jinqi7:
The PR failure is caused by errors like this -
./pipelines/base/nodejs-builder.yaml
 35:37 error too many spaces before colon (colons)
./pipelines/base/java-builder.yaml
 11:37 error too many spaces before colon (colons)

I have a question - shall we fix our existing yaml files to follow the yaml lint rules? Or, change the yaml lint configuration?

tasks/yaml-lint.yaml
Michkov: This task is not used now, `pipelinesascode.tekton.dev/task-2: "yaml-lint"` downloads it from tekton hub.

Michkov: But that's also option to move under .tekton/tasks and then point to it -> `pipelinesascode.tekton.dev/task-2: ".tekton/tasks/yaml-list.yaml"`, that's maybe better since we would have it under our control.

Dannyb48:
> > I don't think we need to add this task to the kustomize tank file. The kustomize file is used when building the hacks pipelines. This task will only be used for the ci pipeline as code for this repo.
> 
> I think the task may not be limited to be used by ci pipeline as code and it may be used by the applications being build. Please correct me if I am wrong. Thanks

At the moment there is no requirement for us to contribute yaml linting to the hacbs pipelines that users would use.And if a user wants to add that they will eventually be allowed to customize the hacbs pipeline.  So for the time being I would remove it.

tasks/yaml-lint.yaml
Dannyb48: @Michkov ah ok. I misread the conversation. So because we are using the upstream tekton task we are deleting this yaml-lint.yaml from tasks directory. 

But yeah maybe for future tekton tasks that aren't up on the tekton catalog, we should probably place tasks specific to the pipeline as code CI in a .tekton/tasks dir

jinqi7:
> At the moment there is no requirement for us to contribute yaml linting to the hacbs pipelines that users would use.And if a user wants to add that they will eventually be allowed to customize the hacbs pipeline. So for the time being I would remove it.
@Dannyb48 OK, I removed it from my PR since it's introduced by me.

Michkov (approved):

jinqi7:
@Dannyb48, can you please help to review the PR again? Thanks

Dannyb48 (approved):
Lgtm

PR #52, opened 2022-04-19
Minor improvements related to Enterprise Contract

simonbaird:


simonbaird:
It's pretty low risk, so maybe I'll merge soon.

appstudio-utils/util-scripts/lib/fetch/rekor.sh
```diff
@@ -79,7 +79,7 @@ rekor-log-entry-save-from-url() {
 shorten-sha() {
   local sha=$1
 
-  echo "$sha" | head -c 11
+  echo "$sha" | sed 's|^sha256:||' | head -c 11
```
zregvart, line 82: Might not be that important, but in case we switch to sha384/sha512 might make it future proof

```suggestion
  echo "$sha" | sed 's|^.+:||' | head -c 11
```

appstudio-utils/util-scripts/lib/fetch.sh
```diff
@@ -20,10 +20,6 @@ POLICIES_DIR=${POLICIES_DIR:-"$EC_WORK_DIR/policies"}
 POLICY_REPO=${POLICY_REPO:-"https://github.com/hacbs-contract/ec-policies.git"}
 POLICY_REPO_REF=${POLICY_REPO_REF:-"main"}
 
-# Pipeline run name
-PR_NAME=${1:-$( tkn pr describe --last -o name )}
```
lcarva, line 24: :heart:

lcarva (approved):

appstudio-utils/util-scripts/lib/fetch/rekor.sh
simonbaird: I did a version of that in new revision, along with some (overly enthuastic?) TDD since I don't trust my sed regex abilities.

simonbaird:
Good enough I think. Also this removes an annoyance from #56 let's merge so @lcarva can rebase.

Thanks @zregvart for the review.

PR #53, opened 2022-04-19
HACBS-306: Better URL parsing in rekor fetch

robnester-rh:
This commit uses perl to parse a given URL and extract the desired
data in the log_index_from_url and rekor_host_from_url functions in the
rekor.sh script.

Signed-off-by: Rob Nester <rnester@redhat.com>

### DIFF
--- a/appstudio-utils/util-scripts/lib/fetch/rekor.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/rekor.sh
@@ -16,21 +16,16 @@ rekor-uuid-entry() {
 }
 
 # Extract the log index from a transparency url
-# Todo: Make it more robust, use perl maybe
 log-index-from-url() {
   local url=$1
-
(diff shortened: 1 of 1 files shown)


appstudio-utils/util-scripts/lib/fetch/rekor.sh
```diff
@@ -16,21 +16,15 @@ rekor-uuid-entry() {
 }
 
 # Extract the log index from a transparency url
-# Todo: Make it more robust, use perl maybe
 log-index-from-url() {
   local url=$1
-
-  # Assume it ends with something like '?logIndex=1234'
```
simonbaird, line 23: Might want to keep this comment but update it, e.g "Assume it has a url param called logIndex"

simonbaird (approved):

simonbaird:
lgtm.

PR #55, opened 2022-04-20
sanity-label-check without workspace

Michkov:
- Update sanity-label-check parameter in pipelines/hacbs/hacbs-test.yaml
- Remove usage of workspace in sanity-label-check

### DIFF
--- a/tasks/sanity-label-check.yaml
+++ b/tasks/sanity-label-check.yaml
@@ -18,24 +18,16 @@ spec:
     - description: Test output
       name: HACBS_TEST_OUTPUT
   steps:
-    # Inspect the image with skopeo
-    - name: inspect-image
-      image: quay.io/redhat-appstudio/hacbs-test:stable
-      script: "skopeo inspect --no-tags docker://$(inputs.params.IMAGE_URL) | tee $(workspaces.sanity-ws.path)/image_inspect.json"
-
-    # Run conftest on the image inspection data
     - name: basic-sanity-checks-required-labels
       image: quay.io/redhat-appstudio/hacbs-test:stable
       script: |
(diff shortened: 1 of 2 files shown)


tasks/sanity-label-check.yaml
```diff
@@ -18,24 +18,16 @@ spec:
...
       image: quay.io/redhat-appstudio/hacbs-test:stable
       script: |
+        skopeo inspect --no-tags docker://$(inputs.params.IMAGE_URL) | tee /tmp/image_inspect.json
+
         echo "Running conftest using $(inputs.params.POLICY_DIR) policy, $(inputs.params.POLICY_NAMESPACE) namespace"
-        /usr/bin/conftest test --no-fail $(workspaces.sanity-ws.path)/image_inspect.json \
+        /usr/bin/conftest test --no-fail /tmp/image_inspect.json \
         --policy $(inputs.params.POLICY_DIR) --namespace $(inputs.params.POLICY_NAMESPACE) \
-        --output=json | tee $(workspaces.sanity-ws.path)/sanity_label_check_output.json
+        --output=json | tee /tmp/sanity_label_check_output.json
```
dirgim, line 29: After the last F2F meeting, we were under the strong impression that we would be able to save this specific detailed test output to the workspace in order to provide proper info on what happened during the test - though it was clear that persisting the workspace would not be coming soon.

Does this mean now that saving/reporting anything larger that 4k bytes is out of the question?

Michkov: last merge in this file broke hacbs bundle, it's possible to re-use workspace with cloned repository data, that will disable parallel execution. The output files are not currently used so that's why I did not store them now.

dirgim: Ok, we can merge this change to unblock the pipelines, but we will need to further discuss workspace usage as this throws the wrench in a number of items that we are working on.

dirgim (approved):
LGTM

PR #56, opened 2022-04-20
Add new task verify-enterprise-contract

lcarva:
Notable differences from the existing enterprise-contract.yaml task:

* The only data being validated is the image attestation. No data
  data is fetched from the cluster.
* No support for inline public key. The parameter PUBLIC_KEY is used
  as is. I didn't want to wrap cosign in a bash script just to support
  this. Realistically, the key is going to be stored in the cluster
  anyways, just use that. Do note that the Secret for the key does not
  have to be the same one used by Chains. In fact, users of this
  task are expected to create their own Secret containing just the
  cosign.pub key in their own namespaces.
* The whole enterprise contract is verified by this task. This is the
  same approach used by the existing enterprise-contract.yaml task,
  but differs from the cosign-verify*.yaml tasks.

### DIFF
--- a/appstudio-utils/util-scripts/verify-attestation-with-policy.sh
+++ b/appstudio-utils/util-scripts/verify-attestation-with-policy.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# Verifies the given attestation passes the given rego policy
+# usage:
+#   verify-attestation-with-policy.sh <attestation file> <output file> <passed file>
+# where:
+#   <in-toto file> file containing the in-toto attestation to be verified
+#   <output file>  where to store the result in JSON format
+#   <passed file>  where to store the overall passing result
+set -euo pipefail
+
+in_toto="$1"
+output="$2"
+passed="$3"
+
+# An in-toto attestation is authenticated metadata about one or more software
+# artifacts, as per the SLSA Attestation Model:
+# https://github.com/slsa-framework/slsa/blob/main/controls/attestations.md
+# It is assumed at this point that the in-toto attestation has been fetched
+# securely and all signature checks have passed, e.g. via cosign. Therefore,
+# a policy is only applied to the embedded SLSA attestation.
+#
+# NOTE: By using "-s", jq will wrap the object in an array. If there is more
+# than one attestation, each will be a distinct object in the array. This is
+# convenient because "cosign verify-attestation" may output a stream containing
+# multiple objects - not wrapped in an array.
+attestations=$(jq -s '[.[].payload | @base64d | fromjson]' ${in_toto})
+
+cd $(dirname $0)
+source lib/fetch.sh
+
+# TODO: If git clone fails, the task does not fail - fix that!
+./fetch-ec-policies.sh
+
+save-policy-config
+
+title Attestations
+# Show a shortened version of the attestations to avoid excessive logging
+echo -n "$attestations" | jq -r '.[].predicate = "..."'
+# Save the attestations for opa processing
+echo -n "${attestations}" > $(json-data-file attestations)
+
+title Violations
+./check-ec-policy.sh | tee "${output}"
+
+title "Passed?"
+./ec-pass-fail.sh "${output}" | tee "${passed}"
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -50,11 +50,8 @@ for TASK in $SCRIPTDIR/../tasks/*.yaml; do
         rm $TASK_TEMP
     fi
     # Replace appstidio-utils placeholder by newly build appstudio-utils image
-    if yq -M -e e ".spec.steps[0].image==\"appstudio-utils\"" $TASK &>/dev/null; then
-        yq -M e ".spec.steps[0].image=\"$APPSTUDIO_UTILS_IMG\"" $TASK >> $TASK_TEMP
-    else
-        cat $TASK >> $TASK_TEMP
-    fi
+    yq -M e ".spec.steps[] |= select(.image == \"appstudio-utils\").image=\"$APPSTUDIO_UTILS_IMG\"" \
+        $TASK >> $TASK_TEMP
     echo --- >> $TASK_TEMP
     REF="$APPSTUDIO_TASKS_REPO:$BUILD_TAG-$PART"
     for file in $PIPELINE_TEMP/*.yaml; do
--- a/hack/start-verify-ec-task.sh
+++ b/hack/start-verify-ec-task.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/bash
+#
+# Rough hacking guide:
+#
+#  * Commit your changes as required. (No need to push to GitHub.)
+#
+#  * Push a new base image and new task bundles like this:
+#      env MY_QUAY_USER=$USER BUILD_TAG=$(git rev-parse HEAD) ./build-and-push.sh
+#
+#    (Assumes you have the required repos created in quay.io, you're signed in there with podman,
+#    and your quay.io username matches $USER)
+#
+#  * Create a Secret in the current namespace containing the cosign.pub key. This should be a
+#    partial copy of the Secret tekton-chains/signing-secrets. Only the public key should be
+#    included in this Secret. By default the name of the Secret is 'cosign-public-key', set
+#    the PUBLIC_KEY_SECRET environment variable to use a different name. For example:
+#    oc -n tekton-chains get secret signing-secrets -o json | jq '.data."cosign.pub" | @base64d' -r > cosign.pub
+#    oc create secret generic cosign-public-key --from-file=cosign.pub
+#
+#  * Make sure you have an image that has already been signed by Chains.
+#
+#  * (Optional) Modify the POLICY_REPO and POLICY_REPO_REF params below as required.
+#
+#  * Run this script:
+#      ./start-verify-ec-task.sh <image-ref>
+#    where <image-ref> is a valid image reference, e.g. quay.io/spam/bacon@sha256:...
+set -euo pipefail
+
+IMAGE_REF="$1"
+
+PUBLIC_KEY_SECRET="${PUBLIC_KEY_SECRET:-cosign-public-key}"
+
+NAMESPACE="$(oc get sa default -o jsonpath='{.metadata.namespace}')"
+
+#
+# Determine which bundle we should use
+#
+# Todo:
+# - How to discover BUNDLE_NUMBER so it isn't hard coded?
+# - Make this generally more robust
+#
+DEFAULT_BUNDLE=$(
+  # ./build-and-push.sh creates this cm in the current project
+  # and iiuc it overrides the default in `-n build-templates`
+  oc get cm build-pipelines-defaults --output=jsonpath='{.data.default_build_bundle}' )
+BUNDLE_NUMBER=2
+USE_BUNDLE="$(
+  echo "$DEFAULT_BUNDLE" | sed 's/build-templates-bundle/appstudio-tasks/' )-$BUNDLE_NUMBER"
+echo "Using bundle $USE_BUNDLE for task"
+
+#
+# Create the taskrun
+#
+# Todo:
+# - Test it with the default bundle
+# - Would it be nicer to use `tkn start`?
+#
+echo "apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: verify-enterprise-contract-
+spec:
+  taskRef:
+    name: verify-enterprise-contract
+    bundle: $USE_BUNDLE
+  params:
+    - name: IMAGE_REF
+      value: $IMAGE_REF
+
+    - name: PUBLIC_KEY
+      value: k8s://$NAMESPACE/$PUBLIC_KEY_SECRET
+
+    # Modify these defaults as required
+    #- name: POLICY_REPO
+    #  value: https://github.com/hacbs-contract/ec-policies.git
+    #- name: POLICY_REPO_REF
+    #  value: main
+" | oc create -f -
+
+#
+# Watch the taskrun that was created
+#
+tkn tr logs -f $( tkn tr describe --last -o name | sed 's|.*/||' )
--- a/tasks/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract.yaml
@@ -1,0 +1,123 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: verify-enterprise-contract
+  annotations:
+    tekton.dev/displayName: Verify Enterprise Contract
+    tekton.dev/pipelines.minVersion: "0.19"
+    tekton.dev/tags: cosign, chains, signature, opa
+  labels:
+    app.kubernetes.io/version: "0.1"
+
+spec:
+  description: Verify the enterprise contract is met
+  params:
+    - name: IMAGE_REF
+      type: string
+      description: Image reference to verify
+
+    - name: PUBLIC_KEY
+      type: string
+      description: >-
+        Public key used to verify signatures. Must be a valid k8s cosign
+        reference, e.g. k8s://my-space/my-secret where my-secret contains
+        the expected cosign.pub attribute.
+
+    - name: COSIGN_EXPERIMENTAL
+      type: string
+      description: Control transparency log lookups. Set to "1" to enabled it.
+      default: "1"
+
+    - name: REKOR_HOST
+      type: string
+      description: Rekor host for transparency log lookups
+      default: https://rekor.sigstore.dev
+
+    # TODO: We should probably use a bundle URL for this:
+    # https://www.openpolicyagent.org/docs/v0.12.2/bundles/#bundle-file-format
+    # Or an OCI image:
+    # https://github.com/open-policy-agent/opa/issues/1413
+    - name: POLICY_REPO
+      type: string
+      description: Git url for rego policies
+      default: https://github.com/hacbs-contract/ec-policies.git
+
+    - name: POLICY_REPO_REF
+      type: string
+      description: Git sha, branch or tag in git repo for rego policies
+      default: main
+
+  results:
+    - name: OUTPUT
+      description: A list of policy violations
+    - name: PASSED
+      description: A string formatted boolean, either true or false
+
+  steps:
+    - name: verify-image-signature
+      image: appstudio-utils
+      command: [cosign]
+      args:
+        - verify
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--rekor-url=$(params.REKOR_HOST)"
+        - $(params.IMAGE_REF)
+      env:
+        # TODO: What if the workspace is not set? Seems to have no effect:
+        #   https://issues.redhat.com/browse/HACBS-316
+        # TODO: Is it SSL_CERT_DIR or SSL_CERTS_DIR ?
+        - name: SSL_CERT_DIR
+          value: $(workspaces.sslcertdir.path)
+        - name: COSIGN_EXPERIMENTAL
+          value: $(params.COSIGN_EXPERIMENTAL)
+      volumeMounts:
+        - name: shared
+          mountPath: /shared
+
+    - name: verify-image-attestation-signature
+      image: appstudio-utils
+      command: [cosign]
+      args:
+        - verify-attestation
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--rekor-url=$(params.REKOR_HOST)"
+        - "--output-file=/shared/image-in-toto-attestation.json"
+        - $(params.IMAGE_REF)
+      env:
+        # TODO: What if the workspace is not set? Seems to have no effect:
+        #   https://issues.redhat.com/browse/HACBS-316
+        # TODO: Is it SSL_CERT_DIR or SSL_CERTS_DIR ?
+        - name: SSL_CERT_DIR
+          value: $(workspaces.sslcertdir.path)
+        - name: COSIGN_EXPERIMENTAL
+          value: $(params.COSIGN_EXPERIMENTAL)
+      volumeMounts:
+        - name: shared
+          mountPath: /shared
+
+    - name: verify-attestation-with-policy
+      image: appstudio-utils
+      command: [/appstudio-utils/util-scripts/verify-attestation-with-policy.sh]
+      args:
+        - /shared/image-in-toto-attestation.json
+        - $(results.OUTPUT.path)
+        - $(results.PASSED.path)
+      env:
+        - name: EC_WORK_DIR
+          value: /shared/ec-work-dir
+        - name: POLICY_REPO_REF
+          value: $(params.POLICY_REPO_REF)
+        - name: POLICY_REPO
+          value: $(params.POLICY_REPO)
+      volumeMounts:
+        - name: shared
+          mountPath: /shared
+
+  workspaces:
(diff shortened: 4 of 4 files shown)


appstudio-utils/util-scripts/lib/fetch.sh
```diff
@@ -20,6 +20,8 @@ POLICIES_DIR=${POLICIES_DIR:-"$EC_WORK_DIR/policies"}
 POLICY_REPO=${POLICY_REPO:-"https://github.com/hacbs-contract/ec-policies.git"}
 POLICY_REPO_REF=${POLICY_REPO_REF:-"main"}
 
+# TODO: Fix this so it can handle the case when there is no cluster access. PR_NAME
+# is never used when we're processing attestations directly.
 # Pipeline run name
 PR_NAME=${1:-$( tkn pr describe --last -o name )}
 PR_NAME=$( echo "$PR_NAME" | sed 's|.*/||' )
```
simonbaird, line 27: Just FYI, I moved this in #52 which isn't merged yet.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,44 @@
...
+# TODO: fix the checks (other than not_useful) to handle the case when there
+# is no cluster access, i.e. no TaskRun yaml
+# When fixed, replace this line with a call to save-policy-config
+echo '{"non_blocking_checks":[
+    "not_useful",
+    "chains_config",
+    "cluster_sanity",
+    "transparency_urls",
+    "transparency_log_attestations"
+    ]}' | jq > $( json-data-file config policy )
```
simonbaird, line 32: Probably won't need this after https://github.com/hacbs-contract/ec-policies/pull/11 is merged, which might happen in a minute or two...

appstudio-utils/util-scripts/lib/fetch.sh
lcarva: Awesome! Just approved that PR.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+  steps:
+    - name: verify-image-signature
+      image: $(params.EXECUTION_IMAGE)
+      command: [cosign]
+      args:
+        - verify
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--output-file=/shared/image-signature.json"
+        - $(params.IMAGE_REF)
```
simonbaird, line 71: Oh that's really clean. Good idea to not have messy bash here in the task definition.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,127 @@
...
+  results:
+    - name: OUTPUT
+      description: A list of policy violations
+    - name: PASSED
+      description: A string formatted boolean, either true or false
+
+  steps:
+    - name: verify-image-signature
+      image: $(params.EXECUTION_IMAGE)
+      command: [cosign]
```
simonbaird, line 66: What are the square brackets for here?

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: Though.. the rego rules in there do make no incorrect assumptions about the location of attestations in the $EC_WORK/data directory.

tasks/verify-enterprise-contract.yaml
simonbaird: Do we use /shared/image-signature.json at any point? It's invisible to opa/rego here, right?

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,44 @@
...
+
+title Predicate
+# Show a shortened version of the predicate to avoid excessive logging
+echo -n $predicate | jq -r '.predicate = "..."'
+# Save the predicate for opa processing
+echo -n ${predicate} > $(json-data-file predicate)
```
simonbaird, line 38: So the predicate *will* be visible to opa/rego under `data.predicate`, right?
This means we can't write a policy to validate the attestation _type or predicateType or the subject.
Not suggesting it's a problem, but just noting it. Is that intentional design choice? My instinct might be to throw in the full attestation in case we do want it visible in rego, but I'm not confident that's wise or correct.

(One of the (placeholder/demo) rego rules currently does look at the `_type` value at the top level of the attestation. We can easily ditch it though.)

tasks/verify-enterprise-contract.yaml
lcarva: Yeah, the file is not used anywhere. I should probably just remove the `--output-file` parameter to avoid confusion.

tasks/verify-enterprise-contract.yaml
lcarva: `command` expects a list, thus the square brackets. The [docs](https://tekton.dev/docs/pipelines/tasks/) do not explain why `command` is a list, but every example does follow the pattern of a single item in the `command` list, and other parts going into the `args` list. I tried using a string instead, and the Tekton rejects it.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
lcarva: True. I can play around with that repo as a way to get this PR further along.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
lcarva: `_type` and `predicateType` are there. They are part of what gets decoded from the payload.

I think looking at these two fields from rego will be important for us.

tasks/verify-enterprise-contract.yaml
lcarva: Remove the parameter.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
lcarva: With that PR, this is no longer needed. Removed it. (I'll still follow up on adjusting ec-policies further)

appstudio-utils/util-scripts/lib/fetch.sh
simonbaird: Merged. You can rebase when ready.

tasks/verify-enterprise-contract.yaml
simonbaird: Ack, thanks.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: As per slack discussion, it makes sense to me now, thanks.

There's some potential for confusion around the terminogy, i.e. payload vs attestation vs predicate vs envelope and etc, but we'll figure it out.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,41 @@
...
+output="$2"
+passed="$3"
+
+# An in-toto attestation is authenticated metadata about one or more software
+# artifacts, as per the SLSA Attestation Model:
+# https://github.com/slsa-framework/slsa/blob/main/controls/attestations.md
+# It is assumed at this point that the in-toto attestation has been fetched
+# securely and all signature checks have passed, e.g. via cosign. Therefore,
+# a policy is only applied to the embedded SLSA attestation.
+attestion=$(jq '.payload | @base64d | fromjson' ${in_toto})
```
simonbaird, line 21: typo: just noticed it's missing an 'a'

    attestation

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: POLICY_REPO_REF
+      type: string
+      description: Git sha, branch or tag in git repo for rego policies
+      default: main
+
+    - name: EXECUTION_IMAGE
+      type: string
+      description: The image used to executed the various steps in this task
+      # TODO: What exactly is this?
```
simonbaird, line 54: Not sure either, other than it comes from https://github.com/redhat-appstudio/build-definitions/blob/main/appstudio-utils/Dockerfile .

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,126 @@
...
+      type: string
+      description: >-
+        Public key used to verify signatures. Must be a valid k8s cosign
+        reference, e.g. k8s://my-space/my-secret where my-secret contains
+        the expected cosign.pub attribute.
+
+    - name: COSIGN_EXPERIMENTAL
+      type: string
+      description: Control transparency log lookups. Set to "1" to enabled it.
+      default: "0"
```
simonbaird, line 30: I'd consider default to "1" here, or even just leave it hard coded down below and remove the param.

simonbaird:
Does it need a version of hack/start-ec-task.sh ?

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,41 @@
...
+# TODO: If git clone fails, the task does not fail - fix that!
+./fetch-ec-policies.sh
+
+save-policy-config
+
+title Predicate
+# Show a shortened version of the attestions to avoid excessive logging
+echo -n $attestion | jq -r '.predicate = "..."'
+# Save the attestion for opa processing
+echo -n ${attestion} > $(json-data-file attestations)
```
simonbaird, line 35: Same thought as in the ec-policies PR. Would there be multiple of them? Or at least, would the content of that file be an array? If not then maybe "attestation" is a better filename here.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: REKOR_HOST
```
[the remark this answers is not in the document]
lcarva: It is now, thanks!

tasks/verify-enterprise-contract.yaml
lcarva: I figure it out. The `build-and-push.sh` script will use this special value to dynamically update the value to a reference of the image built from the Dockerfile you mentioned.

tasks/verify-enterprise-contract.yaml
lcarva: Changed the default to 1... just in case.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
lcarva: This is addressed now. `data/attestations/data.yaml` is now always a list of attestations.

lcarva:
> Does it need a version of hack/start-ec-task.sh ?

@simonbaird, yes, I think so. Added.

hack/build-and-push.sh
```diff
@@ -50,11 +50,8 @@ for TASK in $SCRIPTDIR/../tasks/*.yaml; do
         rm $TASK_TEMP
     fi
     # Replace appstidio-utils placeholder by newly build appstudio-utils image
-    if yq -M -e e ".spec.steps[0].image==\"appstudio-utils\"" $TASK &>/dev/null; then
-        yq -M e ".spec.steps[0].image=\"$APPSTUDIO_UTILS_IMG\"" $TASK >> $TASK_TEMP
-    else
-        cat $TASK >> $TASK_TEMP
-    fi
+    yq -M e ".spec.steps[] |= select(.image == \"appstudio-utils\").image=\"$APPSTUDIO_UTILS_IMG\"" \
```
lcarva, line 53: cc @Michkov

simonbaird: I had to squint pretty hard, but I think I get it now.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,121 @@
...
+      image: appstudio-utils
+      command: [cosign]
+      args:
+        - verify-attestation
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--rekor-url=$(params.REKOR_HOST)"
+        - "--output-file=/shared/image-in-toto-attestation.json"
+        - $(params.IMAGE_REF)
+      env:
+        # TODO: What if the workspace is not set? Seems to have no effect
```
simonbaird, line 87: See also https://issues.redhat.com/browse/HACBS-316 .

hack/start-verify-ec-task.sh
```diff
@@ -0,0 +1,81 @@
...
+#    partial copy of the Secret tekton-chains/signing-secrets. Only the public key should be
+#    included in this Secret. By default the name of the Secret is 'cosign-public-key', set
+#    the PUBLIC_KEY_SECRET environment variable to use a different name.
+#
+#  * Make sure you have an image that has already been signed by Chains.
+#
+#  * (Optional) Modify the POLICY_REPO and POLICY_REPO_REF params below as required.
+#
+#  * Run this script:
+#      ./start-vec-task.sh <image-ref>
```
simonbaird, line 23: Seems like a previous version of the script name here.

tasks/verify-enterprise-contract.yaml
lcarva: Awesome. Added a link to it for the future.

hack/build-and-push.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -xe
```
simonbaird, line 1: Intentional?

hack/build-and-push.sh
lcarva: Given the amount of time I spent on this, it's probably one of the most expensive lines I ever wrote.

hack/start-verify-ec-task.sh
lcarva: Thanks for catching it! Fixed.

hack/build-and-push.sh
lcarva: Negative. Fixed it. Thanks!

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - name: shared
+          mountPath: /shared
+
+    - name: verify-image-attestation-signature
+      image: appstudio-utils
+      command: [cosign]
+      args:
+        - verify-attestation
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--rekor-url=$(params.REKOR_HOST)"
```
simonbaird, line 84: Aah, so I guess there's no way for cosign to learn the rekor host from what in finds in oci.

hack/start-verify-ec-task.sh
```diff
@@ -0,0 +1,81 @@
...
+#  * Push a new base image and new task bundles like this:
+#      env MY_QUAY_USER=$USER BUILD_TAG=$(git rev-parse HEAD) ./build-and-push.sh
+#
+#    (Assumes you have the required repos created in quay.io, you're signed in there with podman,
+#    and your quay.io username matches $USER)
+#
+#  * Create a Secret in the current namespace containing the cosign.pub key. This should be a
+#    partial copy of the Secret tekton-chains/signing-secrets. Only the public key should be
+#    included in this Secret. By default the name of the Secret is 'cosign-public-key', set
+#    the PUBLIC_KEY_SECRET environment variable to use a different name.
```
simonbaird, line 16: Do you want to throw in an example oc command here for the kids in the back row?

lcarva: Good idea. Added a two-liner.

tasks/verify-enterprise-contract.yaml
lcarva: Not that I know of.

hack/start-verify-ec-task.sh
```diff
@@ -0,0 +1,83 @@
...
+#
+#    (Assumes you have the required repos created in quay.io, you're signed in there with podman,
+#    and your quay.io username matches $USER)
+#
+#  * Create a Secret in the current namespace containing the cosign.pub key. This should be a
+#    partial copy of the Secret tekton-chains/signing-secrets. Only the public key should be
+#    included in this Secret. By default the name of the Secret is 'cosign-public-key', set
+#    the PUBLIC_KEY_SECRET environment variable to use a different name. For example:
+#    oc -n tekton-chains get secret signing-secrets -o json | jq '.data."cosign.pub" | @base64d' -r > cosign.pub
+#    oc create secret generic cosign-public-key --from-file=cosign.pub
```
simonbaird, line 18: Awesome.

simonbaird:
Here's the new task running (and failing) in my cluster against an image I built a while ago:
http://pastebin.test.redhat.com/1046547

Here's it is passing with a modified policy:
http://pastebin.test.redhat.com/1046548

I think this could be ready to merge.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,47 @@
...
+# TODO: If git clone fails, the task does not fail - fix that!
+./fetch-ec-policies.sh
+
+save-policy-config
+
+title Attestations
+# Show a shortened version of the attestations to avoid excessive logging
+echo -n $attestations | jq -r '.[].predicate = "..."'
+# Save the attestations for opa processing
+echo -n ${attestations} > $(json-data-file attestations)
```
simonbaird, line 41: Sometimes you can lose the formatting if you use echo without quoting. Maybe we need this:

    echo -n "${attestations}" > $(json-data-file attestations)

lcarva: Good idea. Done.

simonbaird (approved):

simonbaird:
It's good, I think merge when ready, though you could consider seeking one more opinion if you like.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+#
+# Verifies the given attestation passes the given rego policy
+# usage:
+#   verify-attestation-with-policy.sh <attestation file> <output file> <passed file>
+# where:
+#   <in-toto file> file containing the in-toto attestation to be verified
```
robnester-rh, line 7: Lines 5 and 7 should be using the same nomenclature. Either `attestation file` in both, or `in-toto file` in both.

lcarva:
> It's good, I think merge when ready, though you could consider seeking one more opinion if you like.

@simonbaird, I don't have merge permission.

robnester-rh (approved):
Other than a note regarding parameter naming, good work.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: Dang it. Have to fix it later now..

lcarva: Fixed in https://github.com/redhat-appstudio/build-definitions/pull/62

hack/start-verify-ec-task.sh
```diff
@@ -0,0 +1,83 @@
...
+# Create the taskrun
+#
+# Todo:
+# - Test it with the default bundle
+# - Would it be nicer to use `tkn start`?
+#
+echo "apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: verify-enterprise-contract-
```
zregvart, line 61: I find it better to assign a random name, i.e. `name: verify-enterprise-contract-$(openssl rand --hex 10)` and be sure not to trigger a race condition below

hack/start-verify-ec-task.sh
```diff
@@ -0,0 +1,83 @@
...
+    #- name: POLICY_REPO
+    #  value: https://github.com/hacbs-contract/ec-policies.git
+    #- name: POLICY_REPO_REF
+    #  value: main
+" | oc create -f -
+
+#
+# Watch the taskrun that was created
+#
+tkn tr logs -f $( tkn tr describe --last -o name | sed 's|.*/||' )
```
zregvart, line 83: Using `--last` is problematic, we don't know the number of tasks that will run concurrently so we could be tailing the logs of the wrong task here. We should generate a name for the task, like above, store it in a variable and use it here.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,123 @@
...
+    - name: IMAGE_REF
+      type: string
+      description: Image reference to verify
+
+    - name: PUBLIC_KEY
+      type: string
+      description: >-
+        Public key used to verify signatures. Must be a valid k8s cosign
+        reference, e.g. k8s://my-space/my-secret where my-secret contains
+        the expected cosign.pub attribute.
```
zregvart, line 25: Not allowing an inline public key can lead to issues when the signing key is rolled over as we assume that the signature was created with the same key pair that is stored in both secrets: `signing-secrets` for chains controller and the `...-cosign-public-key` secret containing only the public key. There is a window where the signature is created using previous key pair and both secrets have been updated to the newest key. Running and retrying the task can become problematic as the time between signing the image/attestation and verifying the signature increases.

tasks/verify-enterprise-contract.yaml
```diff
@@ -0,0 +1,123 @@
...
+      args:
+        - verify
+        - "--key=$(params.PUBLIC_KEY)"
+        - "--rekor-url=$(params.REKOR_HOST)"
+        - $(params.IMAGE_REF)
+      env:
+        # TODO: What if the workspace is not set? Seems to have no effect:
+        #   https://issues.redhat.com/browse/HACBS-316
+        # TODO: Is it SSL_CERT_DIR or SSL_CERTS_DIR ?
+        - name: SSL_CERT_DIR
```
zregvart, line 70: Should be [`SSL_CERT_DIR`](https://github.com/golang/go/blob/96c8cc7fea94dca8c9e23d9653157e960f2ff472/src/crypto/x509/root_unix.go#L25)

zregvart (commented):
Come comments that might get addressed in future PRs, here added after merge to best link up with the code.

hack/start-verify-ec-task.sh
lcarva: This script is under `hack`, it is meant to be used to assist us in developing and testing our task. It's not meant for production workloads.

I do see your point though. Since this script starts the task, it could very well be more deterministic on which task it follows.

tasks/verify-enterprise-contract.yaml
lcarva: Cool. I found `SSL_CERT_DIR` in the docs as well but was confused because the `cosign-verify*` tasks use `SSL_CERTS_DIR`.

hack/start-verify-ec-task.sh
lcarva: +1 https://github.com/redhat-appstudio/build-definitions/pull/69

hack/start-verify-ec-task.sh
lcarva: Fixed in https://github.com/redhat-appstudio/build-definitions/pull/69

tasks/verify-enterprise-contract.yaml
zregvart: Good catch! That's a typo there, a [pr to fix it](https://github.com/redhat-appstudio/build-definitions/pull/70).

PR #60, opened 2022-04-22
Add test for docker-build pipeline

jinqi7:
Signed-off-by: Jing Qi <jinqi@redhat.com>

Just a test


jinqi7:
Pipelineref is not supported in current CI although it's supported in HAS repo. May try this way later.

PR #61, opened 2022-04-22
Use main workspace for tests, add evaluation finally task

Michkov:
Use workspace in tests, work inside  `$(workspaces.workspace.path)/hacbs/$(context.task.name)`

Add finally task which evaluates tests.

### DIFF
--- a/tasks/hacbs-test-evaluation.yaml
+++ b/tasks/hacbs-test-evaluation.yaml
@@ -1,0 +1,40 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: hacbs-test-evaluation
+spec:
+  description: >-
+    Show summary of HACBS tests
+  workspaces:
+    - name: workspace
+  params:
+    - name: BASE_IMAGE
+  steps:
+    - name: summary  
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      workingDir: $(workspaces.workspace.path)/hacbs
+      script: |
+        #!/usr/bin/env bash
+        EXIT=0
+
+        # sanity-label-check
+        FAILURES=$(jq '.[] | .failures // {} | .[] | .msg' sanity-label-check/sanity_label_check_output.json)
+        if [ -n "$FAILURES" ]; then
+          echo sanity-label-check test FAIL:
+          echo "$FAILURES"
+          EXIT=1
+        else
+          echo sanity-label-check test PASS
+        fi
+        echo -------------
+
+        # sanity-inspect-image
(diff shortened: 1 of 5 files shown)


mbharatk (approved):

hongweiliu17 (approved):

mbharatk (approved):

hongweiliu17 (approved):

tasks/hacbs-test-evaluation.yaml
```diff
@@ -0,0 +1,40 @@
...
+      script: |
+        #!/usr/bin/env bash
+        EXIT=0
+
+        # sanity-label-check
+        FAILURES=$(jq '.[] | .failures // {} | .[] | .msg' sanity-label-check/sanity_label_check_output.json)
+        if [ -n "$FAILURES" ]; then
+          echo sanity-label-check test FAIL:
+          echo "$FAILURES"
+          EXIT=1
```
dirgim, line 25: We don't want to fail the pipeline if any of the test tasks fail - the intent is for the tasks to only report the outcome of each individual test in the `test_name_output.json` file and allow EC to make the decision to release or block the component.

dirgim, line 25: We don't want to fail the pipeline if any of the test tasks fail - the intent is for the tasks to only report the outcome of each individual test in the `test_name_output.json` file and allow EC to make the decision to release or block the component.

Michkov: Based on pipelineRun it will not be possible to see if the tests failed?

Michkov: I understand that we do not want to block other tests, but in this case the evaluation is in finally section -> all standard tasks have been executed already.

dirgim: Well, it would be possible by checking the output of this evaluation task, but the task itself shouldn't fail the pipeline.

dirgim: > I understand that we do not want to block other tests, but in this case the evaluation is in finally section -> all standard tasks have been executed already.

That's an interesting point, but I think the main reason for not failing the test tasks is to allow the release process to continue even if some tests failed, because at that point we don't know if the Enterprise contract for that component actually requires each single test to pass (there could be exceptions or similar).

Michkov: Updated, removed EXIT, always ends with 0

dirgim: Ack, thanks for the change.

dirgim (approved):
LGTM

PR #70, opened 2022-04-25
Fix typo SSL_CERTS_DIR -> SSL_CERT_DIR

zregvart:
We should be using `SSL_CERT_DIR` here.


simonbaird (approved):

simonbaird:
It's good, but why not remove the todo from tasks/verify-enterprise-contract.yaml in the same commit also?

simonbaird:
..the reason might be that it wasn't merged at the time..

lcarva (approved):

PR #71, opened 2022-04-25
Introduce strict mode for ec policy

lcarva:

### DIFF
--- a/appstudio-utils/util-scripts/verify-attestation-with-policy.sh
+++ b/appstudio-utils/util-scripts/verify-attestation-with-policy.sh
@@ -45,3 +45,9 @@ title Violations
 
 title "Passed?"
 ./ec-pass-fail.sh "${output}" | tee "${passed}"
+
+# If strict mode is enabled, fail the script (and the task)  when the policy check fails.
+# Otherwise, complete successfully.
+if [[ "${STRICT_POLICY:-'1'}" == "1" ]]; then
+    [[ "$(cat ${passed})" == "true" ]]
+fi
--- a/tasks/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract.yaml
@@ -26,7 +26,7 @@ spec:
 
     - name: COSIGN_EXPERIMENTAL
       type: string
-      description: Control transparency log lookups. Set to "1" to enabled it.
+      description: Control transparency log lookups. Set to "0" to disable it.
       default: "1"
 
     - name: REKOR_HOST
@@ -47,6 +47,11 @@ spec:
       type: string
       description: Git sha, branch or tag in git repo for rego policies
       default: main
+
+    - name: STRICT_POLICY
+      type: string
+      description: Fail the task if policy fails. Set to "0" to disable it.
+      default: "1"
 
   results:
     - name: OUTPUT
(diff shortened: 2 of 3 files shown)


tasks/verify-enterprise-contract.yaml
```diff
@@ -26,7 +26,7 @@ spec:
 
     - name: COSIGN_EXPERIMENTAL
       type: string
-      description: Control transparency log lookups. Set to "1" to enabled it.
+      description: Control transparency log lookups. Set to "0" to enabled it.
```
lcarva, line 29: Changing the description to mention the opposite value than what is currently in default.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -45,3 +45,7 @@ title Violations
 
 title "Passed?"
 ./ec-pass-fail.sh "${output}" | tee "${passed}"
+
+if [[ "${STRICT_POLICY:-'1'}" = "1" ]]; then
```
simonbaird, line 49: I know it's the same, but maybe `==` is kinder to the reader here..

tasks/verify-enterprise-contract.yaml
simonbaird: Should be `to disable it` iiuc.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: Perhaps it's clear enough, but I would consider a comment here, e.g

    # Return a non-zero exit code if the policy didn't pass

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -45,3 +45,7 @@ title Violations
...
+    [[ "$(cat ${passed})" = "true" ]]
+fi
```
simonbaird, line 51: 1. Not sure how the if it's needed, but maybe this for extra clarity:

```
     else
         true
     fi
```
Update: set -e means it just immediately fails, right so never mind about this..

2. Maybe it's cleaner to move the exit code into ec-pass-fail.sh. You could pass in the param perhaps:
    
```
     ./ec-pass-fail.sh "${output}" "${STRICT_MODE:-1}" | tee "${passed}"
```

But actually the -o pipefail might mess that up so then maybe you'd have to pass the output file in as well, idk.

simonbaird: This is just an idea to consider, not a blocker for merging.

simonbaird (changes requested):

tasks/verify-enterprise-contract.yaml
simonbaird: This one should be a blocker, otherwise it's some half-baked idea sharing...

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: Again I forgot about set -e, maybe:

     # Cause task to fail

lcarva: `==` is a nice idea. I never know how to properly do conditionals in bash. I verified that using `==` works in different cases. Fixed.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
lcarva: It's much nicer now with the comment and the usage of `==`. I'm leaving it as is but am open to suggestions.

simonbaird (approved):
lgtm.

PR #72, opened 2022-04-25
fetch test results

joejstuart:
Search through all TaskRuns in a Pipeline run for test output

For reference the policy PR https://github.com/hacbs-contract/ec-policies/pull/15

### DIFF
--- a/appstudio-utils/util-scripts/fetch-test-data.sh
+++ b/appstudio-utils/util-scripts/fetch-test-data.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ "$#" -ne 2 ]]; then
+  echo "Wrong number of arguments passed."
+  echo "Usage ./fetch-test-data.sh <PipelineRunName> <TestResultName>"
+  exit 1
+fi
+
+PR_NAME=$1
+TEST_NAME=$2
+
+source $(dirname $0)/lib/fetch.sh
+
+# search all task runs in a pipeline
+# write output in format $basdir/data/test/$task_name/data.json
+for tr in $( pr-get-tr-names $PR_NAME ); do
+  data=$( tr-get-result $tr $TEST_NAME )
+  if [[ ! -z "${data}" ]]; then
+      task_name=$( tr-get-task-name ${tr} )
+      echo "${data}" | jq > $( json-data-file test ${task_name} )
+  fi
+done
+
+show-data
--- a/tasks/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract.yaml
@@ -17,6 +17,15 @@ spec:
       type: string
       description: Image reference to verify
 
+    - name: PIPELINERUN_NAME
+      type: string
+      description: The name of the pipelinerun to gather task results from
+
+    - name: HACBS_TEST_OUTPUT
+      type: string
+      description: The name of the tekton result that holds the test output
+      default: HACBS_TEST_OUTPUT
+
     - name: PUBLIC_KEY
       type: string
       description: >-
@@ -97,6 +106,19 @@ spec:
           value: $(workspaces.sslcertdir.path)
         - name: COSIGN_EXPERIMENTAL
           value: $(params.COSIGN_EXPERIMENTAL)
+      volumeMounts:
+        - name: shared
+          mountPath: /shared
+
+    - name: fetch-test-data
+      image: appstudio-utils
+      command: [/appstudio-utils/util-scripts/fetch-test-data.sh]
+      args:
+        - $(params.PIPELINERUN_NAME)
+        - $(params.HACBS_TEST_OUTPUT)
+      env:
+        - name: EC_WORK_DIR
+          value: /shared/ec-work-dir
       volumeMounts:
         - name: shared
           mountPath: /shared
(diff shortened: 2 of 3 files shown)


zregvart (approved):
I've tested by doing this:

```shell
$ hack/test-builds.sh hacbs # ran a few pipelines
$ appstudio-utils/util-scripts/fetch-test-data.sh java-builder-run-hsmrt HACBS_TEST_OUTPUT # ran the fetching from this PR
sha256sum: 51f897329902d5afb879dd2ab428a78f6e5ac3cdb752642b26c35cbba6f12682
/tmp/ecwork/data/test/sanity-label-check/data.json
$ cd ../ec-policies # this is on branch with https://github.com/hacbs-contract/ec-policies/pull/15
$ opa eval --data /tmp/ecwork/data --data policies --data data --format pretty data.hacbs.contract.main.deny
[]
```

Tweaked `/tmp/ecwork/data/test/sanity-label-check/data.json`, i.e. changing the `"result"` from `"SUCCESS"` to `"FAILURE"`, to see the `opa eval` from above resulting in non-empty array.

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -z "$1" ] || [ -z "$2" ]; then
```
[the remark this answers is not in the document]
joejstuart: Good catch. Fixed.

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -0,0 +1,26 @@
...
+
+source $(dirname $0)/lib/fetch.sh
+
+# search all task runs in a pipeline
+# write output in format $basdir/data/test/$task_name/data.json
+for tr in $( pr-get-tr-names $PR_NAME ); do
+  data=$( tr-get-result $tr $TEST_NAME )
+  if [ ! -z "${data}" ]; then
+      task_name=$( tr-get-task-name ${tr} )
+      echo $data > $( json-data-file test ${task_name} )
```
simonbaird, line 22: Sometimes we add a jq here so it's indented:

      echo "$data" | jq > $( json-data-file test ${task_name} )

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
```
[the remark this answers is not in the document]
joejstuart: Updated. I can never keep them straight.

appstudio-utils/util-scripts/fetch-test-data.sh
joejstuart: Good idea.

tasks/verify-enterprise-contract.yaml
```diff
@@ -17,6 +17,15 @@ spec:
       type: string
       description: Image reference to verify
 
+    - name: PIPELINERUN_NAME
+      type: string
+      description: The name of the pipelinerun to gather task results from
```
simonbaird, line 22: Thinking about the implications of demoing here - if the param is required it's harder to demo without a test task in place. Should we consider making the param optional, so we the existing demo works when there is no handy test output?

Or is it easy enough to prepare some hacbs test output for demo purposes?

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -0,0 +1,26 @@
...
+PR_NAME=$1
+TEST_NAME=$2
+
+source $(dirname $0)/lib/fetch.sh
+
+# search all task runs in a pipeline
+# write output in format $basdir/data/test/$task_name/data.json
+for tr in $( pr-get-tr-names $PR_NAME ); do
+  data=$( tr-get-result $tr $TEST_NAME )
+  if [ ! -z "${data}" ]; then
```
simonbaird, line 20: here too?

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -0,0 +1,26 @@
...
+      task_name=$( tr-get-task-name ${tr} )
+      echo $data | jq > $( json-data-file test ${task_name} )
```
simonbaird, line 22: Sorry more nitpicks: Best practice is to quote "$data" here since otherwise it might canonicalize whitespace, e.g consider `echo this   that` vs `echo "this   that"`

tasks/verify-enterprise-contract.yaml
simonbaird: Not a blocker, but something to consider.

simonbaird:
Nice patch. Lgtm, but check my nitpicks.

simonbaird (approved):

appstudio-utils/util-scripts/fetch-test-data.sh
joejstuart: Double brackets?

tasks/verify-enterprise-contract.yaml
joejstuart: This is a good question. Working with the test results was an integral part of the demo, so I didn't consider making it optional, but it does take away from the flexibility.

simonbaird: Yeah, Zoran was thinking about how to address this with an updated demo script. (As per slack discussion.)

simonbaird:
Let's merge.

PR #75, opened 2022-04-26
Add comment to mention deprecated ET task

simonbaird:
Since it was featured in a recent demo, I think it's better not to
delete it immediately, but let's make it very clear that it's out of
date.


simonbaird:
There could be some arguments for deleting them, but here's my reasoning for keeping them a little more:
* If someone watches a demo video and go looking for the task in the demo, it's better if they find it rather than finding nothing there and feeling confused.
* If we delete those, we should also delete code under lib/fetch which is obsoleted, and that's a non-trivial patch that I don't want to work on right this second.

zregvart (approved):

PR #77, opened 2022-04-27
Fetch policy configuration from ConfigMap

zregvart:
For the policy configuration, namely specifying what rules are
non-blocking we can now provide a ConfigMap named `ec-policy` in the
namespace the task is running in. From the ConfigMap a key with the name
`policy.json` needs to hold the JSON file with the policy configuration.
If the ConfigMap doesn't exist the current fixed value is used as a
fallback.
This will most likely be replaced with a EnterpriseContractPolicy custom
resource in the future.

Ref. https://issues.redhat.com/browse/HACBS-244

### DIFF
--- a/appstudio-utils/util-scripts/lib/fetch/cluster.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/cluster.sh
@@ -13,8 +13,17 @@ k8s-save-data() {
   oc get $namespace_opt $kind $name -o json > $file
 }
 
+_policy-config-from-configmap() {
+  oc get configmap ec-policy -o go-template='{{index .data "policy.json"}}' 2>/dev/null
+}
+
+_default-policy-config() {
+  echo '{"non_blocking_checks":["not_useful"]}'
+}
+
 # Placeholder. In future it will likely come from
 # an EnterpriseContractPolicy crd, see HACBS-244
 save-policy-config() {
-  echo '{"non_blocking_checks":["not_useful"]}' | jq > $( json-data-file config policy )
+  # Note: the namespace the task is running in needs to have the ec-policy ConfigMap
+  { _policy-config-from-configmap || _default-policy-config ; } | jq > "$( json-data-file config policy )"
 }
--- a/test/fetch-ec-data_spec.sh
+++ b/test/fetch-ec-data_spec.sh
@@ -2,15 +2,32 @@
 
 eval "$(shellspec -)"
 
+EC_WORK_DIR=$(mktemp -d)
+export EC_WORK_DIR
+
+cleanup() {
+  rm -rf "${EC_WORK_DIR}"
+}
+BeforeEach 'cleanup'
+
+final_cleanup() {
+  if [[ "${SHELLSPEC_KEEP_TMPDIR}" != "1" ]]; then
+    rm -rf "${EC_WORK_DIR}"
+  else
+    echo "
+Last test EC_WORK_DIR=${EC_WORK_DIR} manually remove it"
+  fi
+}
+AfterAll 'final_cleanup'
+
 Include ./appstudio-utils/util-scripts/lib/fetch.sh
 
 Describe 'json-data-file'
   local root_dir=$( git rev-parse --show-toplevel )
-  local data_dir="/tmp/ecwork/data"
 
   It 'produces expected json file paths'
     When call json-data-file some dir 123 foo 456
-    The output should eq "$data_dir/some/dir/123/foo/456/data.json"
+    The output should eq "$EC_WORK_DIR/data/some/dir/123/foo/456/data.json"
   End
 End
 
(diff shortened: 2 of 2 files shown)


appstudio-utils/util-scripts/lib/fetch/cluster.sh
```diff
@@ -16,5 +16,6 @@ k8s-save-data() {
 # Placeholder. In future it will likely come from
 # an EnterpriseContractPolicy crd, see HACBS-244
 save-policy-config() {
-  echo '{"non_blocking_checks":["not_useful"]}' | jq > $( json-data-file config policy )
+  # Note: the namespace the task is running in needs to have the ec-policy ConfigMap
+  { oc get configmap ec-policy -o go-template='{{index .data "policy.json"}}' 2>/dev/null || echo '{"non_blocking_checks":["not_useful"]}'; } | jq > "$( json-data-file config policy )"
```
simonbaird, line 20: Nitpick:
* line is a bit too long
* Just an idea: Could make it nicer by splitting it up maybe, e.g.

```
real-policy() {
  ...
}

dummy-policy() {
  ...
}

save-policy-config()
  { real-policy || dummy-policy } | jq ...
```

Though it might hit some `set -e` issues, not sure. On the plus side the shellspec will let you know right away.

simonbaird:
Fantastic.

simonbaird (approved):

test/fetch-ec-data_spec.sh
```diff
@@ -123,6 +130,39 @@ Describe 'shorten-sha'
...
 
-# Todo: Coverage for lib/fetch/cluster
+Describe 'save-policy-config'
+
+  Describe 'fetches from config map'
+    Mock oc
+      oc_args=$*
+      %preserve oc_args
+      echo '{"from": "cluster"}'
+    End
```
simonbaird, line 140: Neat.

appstudio-utils/util-scripts/lib/fetch/cluster.sh
zregvart: Much nicer

appstudio-utils/util-scripts/lib/fetch/cluster.sh
```diff
@@ -13,8 +13,17 @@ k8s-save-data() {
   oc get $namespace_opt $kind $name -o json > $file
 }
 
+_policy-config-from-configmap() {
```
zregvart, line 16: using `_` to indicate a private function, wouldn't want code outside of this file to depend on this

test/fetch-ec-data_spec.sh
```diff
@@ -2,15 +2,22 @@
 
 eval "$(shellspec -)"
 
+EC_WORK_DIR=$(mktemp -d)
+export EC_WORK_DIR
+
+cleanup() {
+  rm -rf "${EC_WORK_DIR}"
+}
+AfterEach 'cleanup'
```
zregvart, line 11: Perhaps this should be a `BeforeEach` instead :thinking:

appstudio-utils/util-scripts/lib/fetch/cluster.sh
```diff
@@ -13,8 +13,17 @@ k8s-save-data() {
...
+_default-policy-config() {
+  echo '{"non_blocking_checks":["not_useful"]}'
+}
+
 # Placeholder. In future it will likely come from
 # an EnterpriseContractPolicy crd, see HACBS-244
 save-policy-config() {
-  echo '{"non_blocking_checks":["not_useful"]}' | jq > $( json-data-file config policy )
+  # Note: the namespace the task is running in needs to have the ec-policy ConfigMap
+  { _policy-config-from-configmap || _default-policy-config ; } | jq > "$( json-data-file config policy )"
```
lcarva, line 28: Can you elaborate on this syntax? I haven't seen before

lcarva (approved):

appstudio-utils/util-scripts/lib/fetch/cluster.sh
simonbaird: Not sure either - I might have written `(  .... || ..... ) |` is it functionally different to that?

zregvart: `{ ... }` is used to group two commands here so that output of either would be passed to `jq`, otherwise if it was `_policy-config-from-configmap || _default-policy-config | jq` only if `_default-policy-config` branch is executed it'll be piped to `yq`. The `;` is in there because of [SC1083](https://github.com/koalaman/shellcheck/wiki/SC1083)

zregvart: `( ... )` launches a new shell, this is to save on that, one `$PID` and a bit of memory less...

zregvart: & that's why folk usually do `( cd <somewhere>; <do something> )` so when the script continues `$CWD` is not changed...

simonbaird: Cool, thanks.

simonbaird (approved):

PR #78, opened 2022-04-27
Handle ssl certs in verify-ec task

lcarva:


zregvart (approved):

simonbaird:
I don't understand it, but I like it.

simonbaird (approved):
Approved FWIW, but I'd suggest seeking additional reviews.

PR #79, opened 2022-04-28
Yamllint - remove trailing spaces

Michkov:
Remove trailing spaces and add check into yamllint


simonbaird:
Love it! My editor shows them in bright red so it always bugs me.

PR #80, opened 2022-04-28
Sanity label check uses output from skopeo inspect task

hongweiliu17:


hongweiliu17:
@dirgim @Michkov @jsztuka @Dannyb48 PTAL.
It can pass by ./hack/test-builds.sh hacbs
Here is my test pipeline https://console-openshift-console.apps.hongliutester.lab.pnq2.cee.redhat.com/k8s/ns/hongweiliu/tekton.dev~v1beta1~PipelineRun/java-builder-run-2gmst/logs/build-container

tasks/sanity-label-check.yaml
```diff
@@ -23,20 +23,22 @@ spec:
       image: quay.io/redhat-appstudio/hacbs-test:stable
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
-        skopeo inspect --no-tags docker://$(inputs.params.IMAGE_URL) | tee image_inspect.json
+        if [ -f "$(workspaces.workspace.path)/image_inspect.json" ]; then
```
dirgim, line 26: There should be no need for the `if` condition here - if the file is not there, the task will record the error within the `HACBS_ERROR_OUTPUT` value.

dirgim: Hm, we could save the error in an `error.txt` (or similar) file and then add it to the error message like this:
```    
HACBS_ERROR_OUTPUT=$(jq -rc --arg date 1651125649 --arg error "$(cat error.txt)" --null-input \
'{result: "ERROR", timestamp: $date, failures: [$error]}'
```

Dannyb48: @hongweiliu17 @dirgim could we keep a map of error code and generic error message in the hacbs container and use that to look up generic messages to display in the results output? I could potential  `no such file or directory` being a common error message that would be displayed in an error message we would utilize.

Then the more detailed error message can be redirected to a file?

hongweiliu17: +1

hongweiliu17: > Hm, we could save the error in an `error.txt` (or similar) file and then add it to the error message like this:
> 
> ```
> HACBS_ERROR_OUTPUT=$(jq -rc --arg date 1651125649 --arg error "$(cat error.txt)" --null-input \
> '{result: "ERROR", timestamp: $date, failures: [$error]}'
> ```
> 
> Though we'd have to be careful about the error message size because of the 4k size restriction for Tekton task results.

+1

tasks/sanity-inspect-image.yaml
```diff
@@ -20,9 +20,9 @@ spec:
     image: quay.io/redhat-appstudio/hacbs-test:latest
     workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
     script: |
-      IMAGE_INSPECT=image_inspect.json
-      BASE_IMAGE_INSPECT=base_image_inspect.json
-      RAW_IMAGE_INSPECT=raw_image_inspect.json
+      IMAGE_INSPECT=$(workspaces.workspace.path)/image_inspect.json
+      BASE_IMAGE_INSPECT=$(workspaces.workspace.path)/base_image_inspect.json
+      RAW_IMAGE_INSPECT=$(workspaces.workspace.path)/raw_image_inspect.json
```
Michkov, line 25: adding files into root dir of the workspace is not good idea. Hacbs things should not be mixed with source code.

tasks/sanity-label-check.yaml
```diff
@@ -23,20 +23,22 @@ spec:
...
+          echo "Running conftest using $(inputs.params.POLICY_DIR) policy, $(inputs.params.POLICY_NAMESPACE) namespace"
+          /usr/bin/conftest test --no-fail $(workspaces.workspace.path)/image_inspect.json \
```
Michkov, line 28: why not to use relative path -> `../sanity-label-check/image_inspect.json`

Michkov (changes requested):

tasks/sanity-label-check.yaml
```diff
@@ -23,20 +23,21 @@ spec:
...
+        /usr/bin/conftest test --no-fail ../sanity-inspect-image/image_inspect.json \
         --policy $(inputs.params.POLICY_DIR) --namespace $(inputs.params.POLICY_NAMESPACE) \
-        --output=json | tee sanity_label_check_output.json
-
-        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
-          '{result: "ERROR", timestamp: $date}')
+        --output=json 2> stderr.txt | tee sanity_label_check_output.json
+        
+        ERR_MSG=$(cat stderr.txt)
+        if [ ! -z "$ERR_MSG" ]; then
```
dirgim, line 32: I believe that we want to always have the `HACBS_ERROR_OUTPUT` defined. Maybe we can default to `ERR_MSG` being `unknown` if it's empty.

dirgim:
@hongweiliu17 can you also remove the [IMAGE_URL](https://github.com/redhat-appstudio/build-definitions/blob/623087defddfc87c9281a308b0674840f713a7bb/tasks/sanity-label-check.yaml#L8-L9) param from `sanity-label-check.yaml`?
That parameter won't be needed anymore.

tasks/sanity-label-check.yaml
hongweiliu17: > @hongweiliu17 can you also remove the [IMAGE_URL](https://github.com/redhat-appstudio/build-definitions/blob/623087defddfc87c9281a308b0674840f713a7bb/tasks/sanity-label-check.yaml#L8-L9) param from `sanity-label-check.yaml`? That parameter won't be needed anymore.

Updated. Thanks.

tasks/sanity-label-check.yaml
```diff
@@ -23,15 +21,15 @@ spec:
...
-        --output=json | tee sanity_label_check_output.json
-
-        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
-          '{result: "ERROR", timestamp: $date}')
+        --output=json 2> stderr.txt | tee sanity_label_check_output.json
+        
+        ERR_MSG="$(cat stderr.txt)"
+        ERR_MSG="${ERR_MSG:-unknown}"
+        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --arg ERR_MSG "${ERR_MSG: 0: 3000}" --null-input \
+          '{result: "ERROR", timestamp: $date, failures: $ERR_MSG}')
```
dirgim, line 32: Just a tiny change needed here: `failures: $ERR_MSG` -> `failures: [$ERR_MSG]`
The `failures` field needs to be a list.

hongweiliu17: Updated. When I test it, I find there is only one long string so think it can be a string.

dirgim (approved):
LGTM

staticf0x (approved):
LGTM

Michkov (changes requested):
Pull new changes from main, then it will fail on:
`    invalid input params for task sanity-label-check: didn't need these params but they were provided anyway: [IMAGE_URL]`

hongweiliu17:
> Pull new changes from main, then it will fail on: ` invalid input params for task sanity-label-check: didn't need these params but they were provided anyway: [IMAGE_URL]`

It can pass on my local but I found there is only one taskrun call `sanity-label-check` in my code, but more code is merged to main branch and `sanity-label-check` is called by `sanity-label-check` and `sanity-optional-label-check` https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/hacbs/hacbs-test.yaml#L22-L36,

Let me rebase and fix it.

Thanks finding it.

hongweiliu17:
> > Pull new changes from main, then it will fail on: ` invalid input params for task sanity-label-check: didn't need these params but they were provided anyway: [IMAGE_URL]`
> 
> It can pass on my local but I found there is only one taskrun call `sanity-label-check` in my code, but more code is merged to main branch and `sanity-label-check` is called by `sanity-label-check` and `sanity-optional-label-check` https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/hacbs/hacbs-test.yaml#L22-L36,
> 
> Let me rebase and fix it.
> 
> Thanks finding it.

@Michkov The code is updated. Could you review it again? Thank you!

Michkov (approved):

PR #81, opened 2022-04-28
Remove deprecated contract task and related script

simonbaird:


simonbaird:
I feel like we don't have to merge it right away, but here's what the "Remove old junk" PR might look like. I didn't want to figure out which parts of appstudio-utils/util-scripts/lib/fetch/tekton.sh were still useful, so I just left it alone. There are arguments for removing the rekor stuff and its test coverage I guess. What do you think?

joejstuart:
>  There are arguments for removing the rekor stuff and its test coverage I guess. What do you think?
 I vote for leaving them at this point. Just because ec-policies is focused on using attestations, doesn't mean there won't be another policy that needs to validate cluster information. Maybe... And the scripts contain useful utilities.

simonbaird:
Yeah, that's my thinking too. E.g the shorten-sha function - it's too good to throw away.. :)

simonbaird:
Also there's also the future "validate a pipeline definition" functionality that would need to use data from somewhere other than an attestation fetched with cosign, so that's an argument for keeping some handy cluster fetch functions in case they become useful.

lcarva:
+1 to delete deprecated stuff. It's less things we have to worry about. If we get nostalgic, we can always look at the git history :grin:

simonbaird:
It says no conflicts, so let's merge.

PR #82, opened 2022-04-28
Don't fail immediately if hacbs test results are missing

simonbaird:
If the pipeline run is missing or doesn't have the expected test
results, we get the following error and the task fails:

    find: '/shared/ec-work-dir/data': No such file or directory

Actually `mkdir -p $DATA_DIR` alone would be enough to avoid the
error, but I thought it would be nice to write something to the data
dir.

JIRA: [HACBS-377](https://issues.redhat.com/browse/HACBS-377)

### DIFF
--- a/appstudio-utils/util-scripts/fetch-test-data.sh
+++ b/appstudio-utils/util-scripts/fetch-test-data.sh
@@ -15,12 +15,20 @@ source $(dirname $0)/lib/fetch.sh
 
 # search all task runs in a pipeline
 # write output in format $basdir/data/test/$task_name/data.json
+result_found=
 for tr in $( pr-get-tr-names $PR_NAME ); do
   data=$( tr-get-result $tr $TEST_NAME )
   if [[ ! -z "${data}" ]]; then
+      result_found=1
       task_name=$( tr-get-task-name ${tr} )
       echo "${data}" | jq > $( json-data-file test ${task_name} )
   fi
 done
+
+if [[ -z $result_found ]]; then
+  # let's put an an empty hash here to express the the idea
+  # that we looked for test results and found none
+  echo '{}' > $( json-data-file test )
+fi
 
 show-data
--- a/appstudio-utils/util-scripts/lib/fetch.sh
+++ b/appstudio-utils/util-scripts/lib/fetch.sh
@@ -16,6 +16,8 @@ EC_WORK_DIR=${EC_WORK_DIR:-/tmp/ecwork}
 
 DATA_DIR=${DATA_DIR:-"$EC_WORK_DIR/data"}
 POLICIES_DIR=${POLICIES_DIR:-"$EC_WORK_DIR/policies"}
+mkdir -p $DATA_DIR
+mkdir -p $POLICIES_DIR
 
 POLICY_REPO=${POLICY_REPO:-"https://github.com/hacbs-contract/ec-policies.git"}
 POLICY_REPO_REF=${POLICY_REPO_REF:-"main"}


simonbaird:
With this patch applied:
```
[fetch-test-data] Error from server (NotFound): pipelineruns.tekton.dev "asdfasdf" not found
[fetch-test-data] sha256sum: 074a20bf0663f79925cf727061c091fb95431816065ec37339cea65e544a3fd7
[fetch-test-data] /shared/ec-work-dir/data/test/data.json

...continues to next step...
```
Notice the oc get error doesn't cause the task to fail.

Without this patch:
```
[fetch-test-data] Error from server (NotFound): pipelineruns.tekton.dev "asdfasdf" not found
[fetch-test-data] find: '/shared/ec-work-dir/data': No such file or directory
[fetch-test-data] sha256sum: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
[fetch-test-data] find: '/shared/ec-work-dir/data': No such file or directory

[verify-attestation-with-policy] 2022/04/28 21:01:29 Skipping step because a previous step failed
```

simonbaird:
I was wondering if we should fail explicitly with a useful message in this scenario, but since this task will be removed some day, I figure it's not worth spending a lot of time on that. And much more convenient for hacking and demos if it doesn't fail.

joejstuart:
Should `show-data` be patched to handle the missing data directory, so another script doesn't run into this?

simonbaird:
> Should `show-data` be patched to handle the missing data directory, so another script doesn't run into this?

I thought about that, but it seems like the scripts should be able to depend on the dir existing. Perhaps the solution is to ensure it's created early on somehow.

joejstuart:
> > Should `show-data` be patched to handle the missing data directory, so another script doesn't run into this?
> 
> I thought about that, but it seems like the scripts should be able to depend on the dir existing. Perhaps the solution is to ensure it's created early on somehow.

Not super important since this is going away soon, but fetch.sh seems like a good candidate for init/setup work.

simonbaird:
Yeah good idea. I put a couple of `mkdir -p` commands in there so we can stop worrying about the dir not existing.

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -15,12 +15,20 @@ source $(dirname $0)/lib/fetch.sh
...
+      result_found=1
       task_name=$( tr-get-task-name ${tr} )
       echo "${data}" | jq > $( json-data-file test ${task_name} )
   fi
 done
 
+if [[ -z $result_found ]]; then
+  # let's put an an empty hash here to express the the idea
+  # that we looked for test results and found none
+  echo '{}' > $( json-data-file test )
```
simonbaird, line 31: As mentioned, this isn't really needed. I'm about 60% sure I like the idea, so it's in the patch. Let me know if you don't like it.

simonbaird: It does look weird and confusing when show-data gives you a checksum and shows no files. So that's one (albeit weak) argument in favor of doing it this way.

appstudio-utils/util-scripts/lib/fetch.sh
```diff
@@ -16,6 +16,8 @@ EC_WORK_DIR=${EC_WORK_DIR:-/tmp/ecwork}
 
 DATA_DIR=${DATA_DIR:-"$EC_WORK_DIR/data"}
 POLICIES_DIR=${POLICIES_DIR:-"$EC_WORK_DIR/policies"}
+mkdir -p $DATA_DIR
+mkdir -p $POLICIES_DIR
```
simonbaird, line 20: We happen to know there's a mkdir in the git clone method, so technically we don't need to create $POLICIES_DIR but I think it's best to leave this here anyway to be consistent and to reduce our 'wtf per minute' score.

appstudio-utils/util-scripts/fetch-test-data.sh
lcarva: I do think that it makes sense for this script to do this. Its "interface" is that after it runs, test data can be processed. The data may just be empty.

joejstuart (approved):

PR #83, opened 2022-05-02
Update package name to drop `hacbs.contract`

zregvart:
Follow up needed for the change in:

https://github.com/hacbs-contract/ec-policies/pull/22

### DIFF
--- a/appstudio-utils/util-scripts/check-ec-policy.sh
+++ b/appstudio-utils/util-scripts/check-ec-policy.sh
@@ -8,7 +8,7 @@ POLICY_DIR=$EC_WORK_DIR/policies
 
 FORMAT=${1:-pretty}
 
-OPA_QUERY=${OPA_QUERY:-data.hacbs.contract.main.deny}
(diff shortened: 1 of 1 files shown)


lcarva (approved):

appstudio-utils/util-scripts/check-ec-policy.sh
```diff
@@ -8,7 +8,7 @@ POLICY_DIR=$EC_WORK_DIR/policies
 
 FORMAT=${1:-pretty}
 
-OPA_QUERY=${OPA_QUERY:-data.hacbs.contract.main.deny}
+OPA_QUERY=${OPA_QUERY:-data.main.deny}
```
lcarva, line 11: It makes me wonder if we should introduce OPA_QUERY as a task parameter. Probably not at this point...

lcarva (approved):

simonbaird:
Off topic, but now I know there's an `opa check` command, I'm thinking the `check-ec-policy.sh` isn't a great name.

PR #84, opened 2022-05-04
Robust fetching of policies from git

zregvart:
This makes sure that in case of errors when fetching from git don't
proceed with the evaluation of the policy. In addition an error is
outputted to explain the issue.

Ref. https://issues.redhat.com/browse/HACBS-400

### DIFF
--- a/appstudio-utils/util-scripts/lib/fetch/git.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/git.sh
@@ -5,21 +5,28 @@ git-fetch-repo() {
   local directory="$3"
 
   mkdir -p "$directory"
-  cd "$directory"
+  cd "$directory" || return
 
   # Git clone might work but this avoids downloading the entire repo
-  git init -q .
-  git fetch -q --depth 1 --no-tags $repo $ref_or_sha
-  git checkout -q FETCH_HEAD
+  git init -q . && \
+  git fetch -q --depth 1 --no-tags "$repo" "$ref_or_sha" && \
+  git checkout -q FETCH_HEAD && \
 
   git rev-parse FETCH_HEAD
 }
 
 git-fetch-policies() {
   echo "Fetching policies from $POLICY_REPO_REF at $POLICY_REPO"
-  echo "sha: $( git-fetch-repo $POLICY_REPO $POLICY_REPO_REF $POLICIES_DIR )"
+  local sha
+  if ! sha=$( git-fetch-repo "$POLICY_REPO" "$POLICY_REPO_REF" "$POLICIES_DIR" ); then
+    echo "Unable to fetch polices repository from $POLICY_REPO at ref $POLICY_REPO_REF!"
+    return 1
+  fi
+  echo "sha: $sha"
 
-  # Clean up files we don't need including .git
-  cd $POLICIES_DIR
-  rm -rf .git .github .gitignore README.md Makefile scripts
+  if [[ -d "$POLICIES_DIR" ]]; then
+    # Clean up files we don't need including .git
+    cd "$POLICIES_DIR" || return
+    rm -rf .git .github .gitignore README.md Makefile scripts
+  fi
 }
--- a/appstudio-utils/util-scripts/fetch-test-data.sh
+++ b/appstudio-utils/util-scripts/fetch-test-data.sh
@@ -11,24 +11,26 @@ fi
 PR_NAME=$1
 TEST_NAME=$2
 
-source $(dirname $0)/lib/fetch.sh
+source "$(dirname "$0")/lib/fetch.sh"
 
 # search all task runs in a pipeline
 # write output in format $basdir/data/test/$task_name/data.json
 result_found=
-for tr in $( pr-get-tr-names $PR_NAME ); do
-  data=$( tr-get-result $tr $TEST_NAME )
-  if [[ ! -z "${data}" ]]; then
+task_runs=$( pr-get-tr-names "$PR_NAME" )
+for tr in $task_runs; do
+  data=$( tr-get-result "$tr" "$TEST_NAME" )
+  if [[ -n "${data}" ]]; then
       result_found=1
-      task_name=$( tr-get-task-name ${tr} )
-      echo "${data}" | jq > $( json-data-file test ${task_name} )
+      task_name=$( tr-get-task-name "${tr}" )
+      output_file="$( json-data-file test "${task_name}" )"
+      echo "${data}" | jq > "${output_file}"
   fi
 done
 
 if [[ -z $result_found ]]; then
   # let's put an an empty hash here to express the the idea
   # that we looked for test results and found none
-  echo '{}' > $( json-data-file test )
+  echo '{}' > "$( json-data-file test )"
 fi
 
 show-data
--- a/appstudio-utils/util-scripts/lib/fetch/data.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/data.sh
@@ -19,7 +19,7 @@ json-data-file() {
   file="$dir/data.json"
 
   # Better not silently overwrite data
-  [[ -f $file ]] && echo "Name clash for $file!" && exit 1
+  [[ -f $file ]] && echo "ERROR: Name clash for $file!" 1>&2 && return 1
 
   echo "$file"
 }
--- a/test/fetch-ec-data_spec.sh
+++ b/test/fetch-ec-data_spec.sh
@@ -23,11 +23,18 @@ AfterAll 'final_cleanup'
 Include ./appstudio-utils/util-scripts/lib/fetch.sh
 
 Describe 'json-data-file'
-  local root_dir=$( git rev-parse --show-toplevel )
-
   It 'produces expected json file paths'
     When call json-data-file some dir 123 foo 456
     The output should eq "$EC_WORK_DIR/data/some/dir/123/foo/456/data.json"
+  End
+
+  It 'detects file clashes'
+    mkdir -p "${DATA_DIR}/x"
+    touch "${DATA_DIR}/x/data.json"
+
+    When call json-data-file x
+    The error should equal "ERROR: Name clash for ${DATA_DIR}/x/data.json!"
+    The status should be failure
   End
 End
 
--- a/test/fetch-ec-policies_spec.sh
+++ b/test/fetch-ec-policies_spec.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+eval "$(shellspec -)"
+
+# we're sourcing in the specs below, so `dirname $0` will return this script's directory
+dirname() {
+  echo ./appstudio-utils/util-scripts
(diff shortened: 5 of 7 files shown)


simonbaird:
I didn't try it, but it looks good from the test coverage.

simonbaird:
I don't really understand why it didn't fail before. Does git return exit code zero even when things don't work?

appstudio-utils/util-scripts/lib/fetch/git.sh
```diff
@@ -5,21 +5,28 @@ git-fetch-repo() {
   local directory="$3"
 
   mkdir -p "$directory"
-  cd "$directory"
+  cd "$directory" || exit
```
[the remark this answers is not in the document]
zregvart: Yeah, that should be `return` instead, I followed the [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164) too literally. With `return` (or `exit`) if the status is omitted it'll be set to the status of the last command (i.e. `return $?` is the same as `return`).

Now that I changed to `return` I wonder if the task will fail, I guess I need to write a test for that.

And if `mkdir -p` fails, `cd` will fail surely, so there's that.

simonbaird:
>
> I think we might me missing `set -e` when we do it like this:
> 

Hmm, that would be good to confirm. I would have expected it would keep the set -e.

zregvart:
I want to spin up a cluster and see previous vs this version, probably not today, there is no rush in getting this merged.

lcarva:
> With the task bundle image from this PR:
> 
> ```
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Fetching policy files ----
> [ec : verify-attestation-with-policy] Fetching policies from main at bogus
> [ec : verify-attestation-with-policy] fatal: 'bogus' does not appear to be a git repository
> [ec : verify-attestation-with-policy] fatal: Could not read from remote repository.
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] Please make sure you have the correct access rights
> [ec : verify-attestation-with-policy] and the repository exists.
> [ec : verify-attestation-with-policy] Unable to fetch polices repository from bogus at ref main!
> ```
> 
> And the TaskRun/PipelineRun fails.
> 
> With the task bundle image from #85:
> 
> ```
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Fetching policy files ----
> [ec : verify-attestation-with-policy] Fetching policies from main at bogus
> [ec : verify-attestation-with-policy] fatal: 'bogus' does not appear to be a git repository
> [ec : verify-attestation-with-policy] fatal: Could not read from remote repository.
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] Please make sure you have the correct access rights
> [ec : verify-attestation-with-policy] and the repository exists.
> [ec : verify-attestation-with-policy] error: pathspec 'FETCH_HEAD' did not match any file(s) known to git
> [ec : verify-attestation-with-policy] fatal: ambiguous argument 'FETCH_HEAD': unknown revision or path not in the working tree.
> [ec : verify-attestation-with-policy] Use '--' to separate paths from revisions, like this:
> [ec : verify-attestation-with-policy] 'git <command> [<revision>...] -- [<file>...]'
> [ec : verify-attestation-with-policy] sha: FETCH_HEAD
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Policy files ----
> [ec : verify-attestation-with-policy] sha256sum: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Attestations ----
> [ec : verify-attestation-with-policy] [
> [ec : verify-attestation-with-policy]   {
> [ec : verify-attestation-with-policy]     "_type": "https://in-toto.io/Statement/v0.1",
> [ec : verify-attestation-with-policy]     "predicateType": "https://slsa.dev/provenance/v0.2",
> [ec : verify-attestation-with-policy]     "subject": [
> [ec : verify-attestation-with-policy]       {
> [ec : verify-attestation-with-policy]         "name": "quay.io/zregvart_redhat/single-nodejs-app",
> [ec : verify-attestation-with-policy]         "digest": {
> [ec : verify-attestation-with-policy]           "sha256": "f35e3d3ed2fa6bcd6aeea48e7284b4e21cd97995eed30590e7e637f70fc2b722"
> [ec : verify-attestation-with-policy]         }
> [ec : verify-attestation-with-policy]       }
> [ec : verify-attestation-with-policy]     ],
> [ec : verify-attestation-with-policy]     "predicate": "..."
> [ec : verify-attestation-with-policy]   }
> [ec : verify-attestation-with-policy] ]
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Violations ----
> [ec : verify-attestation-with-policy] undefined
> [ec : verify-attestation-with-policy] 
> [ec : verify-attestation-with-policy] 🔗 ---- Passed? ----
> [ec : verify-attestation-with-policy] false
> ```
> 
> Also the TaskRun/PipelineRun fails.

I'm surprised the next step actually runs which makes me there's still something that is skipping errors.

appstudio-utils/util-scripts/lib/fetch/git.sh
```diff
@@ -5,21 +5,28 @@ git-fetch-repo() {
...
 
   mkdir -p "$directory"
-  cd "$directory"
+  cd "$directory" || return
 
   # Git clone might work but this avoids downloading the entire repo
-  git init -q .
-  git fetch -q --depth 1 --no-tags $repo $ref_or_sha
-  git checkout -q FETCH_HEAD
+  git init -q . && \
```
lcarva, line 11: Hmm why is this needed? `set -e` is there so we don't have to do this, right?

lcarva:
Bash never seems to amaze me. I wrote a little reproducer to see what's going on here:
```bash
#!/bin/bash
set -euo pipefail

just-fail() {
    touch /var/run/ahhhh-no-access
}

test-1() {
    echo "TEST 1: "
    local result="$(just-fail)" && echo "FAILED" || echo "PASS"
    echo
}

test-2() {
    echo "TEST 2: "
    result="$(just-fail)" && echo "FAILED" || echo "PASS"
    echo
}

test-3() {
    echo "TEST 3:"
    echo -n "$(just-fail)" && echo "FAILED" || echo "PASS"
    echo
}

test-1
test-2
test-3

```

```bash
$ ./test.sh
TEST 1: 
touch: cannot touch '/var/run/ahhhh-no-access': Permission denied
FAILED

TEST 2: 
touch: cannot touch '/var/run/ahhhh-no-access': Permission denied
PASS

TEST 3:
touch: cannot touch '/var/run/ahhhh-no-access': Permission denied
FAILED
```

It appears that this is the culprit: https://github.com/redhat-appstudio/build-definitions/blob/10c01d70295de33617ec82b266098b49894abd90/appstudio-utils/util-scripts/lib/fetch/git.sh#L20

This is a potential fix:
```diff
diff --git a/appstudio-utils/util-scripts/lib/fetch/git.sh b/appstudio-utils/util-scripts/lib/fetch/git.sh
index 719293f..6901138 100644
--- a/appstudio-utils/util-scripts/lib/fetch/git.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/git.sh
@@ -17,7 +17,8 @@ git-fetch-repo() {
 
 git-fetch-policies() {
   echo "Fetching policies from $POLICY_REPO_REF at $POLICY_REPO"
-  echo "sha: $( git-fetch-repo $POLICY_REPO $POLICY_REPO_REF $POLICIES_DIR )"
+  sha="$( git-fetch-repo $POLICY_REPO $POLICY_REPO_REF $POLICIES_DIR )"
+  echo "sha: ${sha}"
 
   # Clean up files we don't need including .git
   cd $POLICIES_DIR
```

appstudio-utils/util-scripts/lib/fetch/git.sh
zregvart: I wanted to prevent any output to stdout so that the `$sha` below is not set to any value. Without this tests fail with:

```
  1) git-fetch-repo fails on bogus repository
     When call git-fetch-repo bogus main /tmp/tmp.feHVReVNzS

     1.1) WARNING: There was output to stdout but not found expectation

            stdout:FETCH_HEAD
            

          # test/fetch-git_spec.sh:64-69

  2) git-fetch-repo fails on bogus ref
     When call git-fetch-repo /tmp/tmp.f1kpx6bRwP bogus /tmp/tmp.O5JknTWyrf

     2.1) WARNING: There was output to stdout but not found expectation

            stdout:FETCH_HEAD
            

          # test/fetch-git_spec.sh:71-76
```

Which tells me that these commands progress regardless of `set -e`.

zregvart:
@lcarva that's why in the changes proposed here `$sha` is declared and assigned separately. Shellcheck catches that in [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155).

simonbaird:
Copy/paste from a thread above:

It does feel like this patch is too big. Unless git returns exit code zero on failure (which don't think is the case), then why not just figure out how to ensure set -e is present?

Just blowing up failing with whatever git outputs is good enough. We don't want to do any special error handling (which is why the cd ... || return feels weird to me. I would want cd ... | boom die now. Fail hard and early.

zregvart:
The problem is exactly the same as in #94, `set -e` appears to bare no effect when exercising these functions with shellspec. That is this:

```
#!/usr/bin/env bash

set -euo pipefail

eval "$(shellspec -)"

function f() {
    set -e
    local x
    x=$(nonexistant 2>&1)

    echo "Should not be here: $x, status was $?"
}

Describe 'error handling'
    It 'no worky'
        When call f
        The the status should be failure
    End
End
```

Results with:
```
Running: /usr/bin/bash [bash 5.1.16(1)-release]
F

Examples:
  1) error handling no worky
     When call f

     1.1) The the status should be failure

            expected: failure (non-zero)
                 got: success (zero) [status: 0]

          # wat_spec.sh:18

     1.2) WARNING: There was output to stdout but not found expectation

            stdout:Should not be here: main: line 28: nonexistant: command not found, status was 127
            

          # wat_spec.sh:16-19

Finished in 0.13 seconds (user 0.11 seconds, sys 0.03 seconds)
1 example, 1 failure


Failure examples / Errors: (Listed here affect your suite's status)

shellspec wat_spec.sh:16 # 1) error handling no worky FAILED
```

So that leaves me with only the option I did here, i.e. to chain execution of commands in order to test that error handling was performed.

zregvart:
Will most likely conflict with #94, and I'd like #94 to be merged first

zregvart:
I'm going to close this and revisit if any work needs to be done after we evaluate outcome of [HACBS-451](https://issues.redhat.com/browse/HACBS-451)

PR #86, opened 2022-05-05
Add configure-build-csi-secret to get sharedSecret

Michkov:
Used in hacbs-core-service bundle to use shared secret for pushing into quay.io/redhat-appstudio/*

Dependency on https://github.com/redhat-appstudio/infra-deployments/pull/330


gabemontero (approved):
the shared resource  bits at least LGTM

gabemontero:
fwiw was able to run this test-builds off of a PR against my local cluster with app studio bootstrapped via https://github.com/redhat-appstudio/infra-deployments/pull/330 and I see the shared secret created and recognizable to the shared resource csi driver/controller

PR #91, opened 2022-05-10
Fix test-build.sh chains artifacts

lcarva:

### DIFF
--- a/hack/test-build.sh
+++ b/hack/test-build.sh
@@ -28,9 +28,15 @@ if [ -z "$MY_QUAY_USER" ]; then
 else
   if oc get secret redhat-appstudio-staginguser-pull-secret &>/dev/null; then
      PUSH_WORKSPACE="-w name=registry-auth,secret=redhat-appstudio-staginguser-pull-secret"
+     # Ensure that the pipeline service account has access to the secret. Although the
+     # secret is mounted directly on the pipeline, Tekton Chains needs this linkage so
+     # it can push the image signature and attestation to the same OCI repository.
+     oc secrets link pipeline redhat-appstudio-staginguser-pull-secret --for=pull
   else
      echo redhat-appstudio-staginguser-pull-secret is not created, can be created by:
      echo oc create secret docker-registry redhat-appstudio-staginguser-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
+     echo and link it to the pipeline ServiceAccount:
+     echo oc secrets link pipeline redhat-appstudio-staginguser-pull-secret --for=pull
   fi
   IMG=quay.io/$MY_QUAY_USER/$APPNAME:$IMAGE_SHORT_TAG
   echo Building $IMG


lcarva:
Failed CI has to do with a yamllint failure. I cannot view more information. I also have not changed any yaml files in this PR.

hack/test-build.sh
```diff
@@ -28,6 +28,10 @@ if [ -z "$MY_QUAY_USER" ]; then
 else
   if oc get secret redhat-appstudio-staginguser-pull-secret &>/dev/null; then
      PUSH_WORKSPACE="-w name=registry-auth,secret=redhat-appstudio-staginguser-pull-secret"
+     # Ensure that the pipeline service account has access to the secret. Although the
+     # secret is mounted directly on the pipeline, Tekton Chains needs this linkage so
+     # it can push the image signature and attestation to the same OCI repository.
+     oc secrets link pipeline redhat-appstudio-staginguser-pull-secret --for=pull
   else
      echo redhat-appstudio-staginguser-pull-secret is not created, can be created by:
      echo oc create secret docker-registry redhat-appstudio-staginguser-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
```
joejstuart, line 37: After this should the link command be printed?

lcarva: I didn't add because I had assumed that if someone needs to create the secret, they'll do that then re-run the script. But it's probably a good idea to be more explicit. Done.

joejstuart (approved):

Michkov (approved):

hack/test-build.sh
```diff
@@ -28,9 +28,15 @@ if [ -z "$MY_QUAY_USER" ]; then
 else
   if oc get secret redhat-appstudio-staginguser-pull-secret &>/dev/null; then
      PUSH_WORKSPACE="-w name=registry-auth,secret=redhat-appstudio-staginguser-pull-secret"
+     # Ensure that the pipeline service account has access to the secret. Although the
+     # secret is mounted directly on the pipeline, Tekton Chains needs this linkage so
+     # it can push the image signature and attestation to the same OCI repository.
+     oc secrets link pipeline redhat-appstudio-staginguser-pull-secret --for=pull
```
simonbaird, line 34: Does it error if you try to create an existing link? If not, then lgtm.

hack/test-build.sh
```diff
@@ -28,9 +28,15 @@ if [ -z "$MY_QUAY_USER" ]; then
...
   else
      echo redhat-appstudio-staginguser-pull-secret is not created, can be created by:
      echo oc create secret docker-registry redhat-appstudio-staginguser-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
+     echo and link it to the pipeline ServiceAccount:
+     echo oc secrets link pipeline redhat-appstudio-staginguser-pull-secret --for=pull
```
simonbaird, line 39: I'm not sure I follow the logic, since you're adding it automatically on line 34, why tell user to add it manually?

hack/test-build.sh
lcarva: It does not error.

hack/test-build.sh
lcarva: In order to make things more explicitly to the user running this script. It was the outcome of a previous [comment](https://github.com/redhat-appstudio/build-definitions/pull/91#discussion_r869585801).

Given that the way Chains works with secrets is not broadly known/understood, I'd rather the script be extra verbose around this.

simonbaird: Ok.

PR #94, opened 2022-05-11
Use EnterpriseContractPolicy in v-e-c

zregvart:
This adds `POLICY_CONFIGURATION` parameter to the `verify-enterprise-contract` Tekton task used to point to the EnterpriseContractPolicy custom resource with which the policy configuration for executing the task is governed. Namely, how to fetch the policy files and configuration of non-blocking policy rules.

With respect to current demo scripts, backward compatibility is maintained. So configuring non-blocking policy rules via ConfigMap and fetching from Git using `POLICY_REPO` and `POLICY_REPO_REF` is still possible. But note that `POLICY_CONFIGURATION` takes precedence over
those.

Ref. https://issues.redhat.com/browse/HACBS-244


appstudio-utils/util-scripts/lib/fetch/cluster.sh
```diff
@@ -21,9 +21,38 @@ _default-policy-config() {
...
+    echo "-n ${namespace}"
+  fi
+}
+
 # Placeholder. In future it will likely come from
 # an EnterpriseContractPolicy crd, see HACBS-244
 save-policy-config() {
+  local args=($(cr-namespace-argument "$1") $(cr-name "$1"))
+  if ! non_blocking_data=$(kubectl get enterprisecontractpolicies.appstudio.redhat.com "${args[*]}" -o jsonpath='{.spec.exceptions.nonBlocking}'); then
+    echo "ERROR: unable to find the ec-policy EnterpriseContractPolicy in namespace $(kubectl config view --minify -o jsonpath='{..namespace}')" 1>&2
```
lcarva, line 45: The `kubectl config view` command may return an empty string if the default namespace is selected. At least that's what [stackoverflow](https://stackoverflow.com/questions/55853977/how-to-get-the-current-namespace-of-current-context-using-kubectl) says. Maybe it's not a big deal for us. An alternative is to query for the namespace of any of the ServiceAccounts.

appstudio-utils/util-scripts/lib/fetch/git.sh
```diff
@@ -16,6 +16,12 @@ git-fetch-repo() {
 }
 
 git-fetch-policies() {
+  local repository_and_ref
+  if repository_and_ref=$(kubectl get enterprisecontractpolicies.appstudio.redhat.com ec-policy -o jsonpath='{.spec.sources[0].git.repository}#{.spec.sources[0].git.revision}'); then
```
lcarva, line 20: Shouldn't `ec-policy` not be hard coded?

tasks/verify-enterprise-contract.yaml
```diff
@@ -13,6 +13,15 @@ metadata:
 spec:
   description: Verify the enterprise contract is met
   params:
+    - name: POLICY_CONFIGURATION
+      type: string
+      description: |
+        Name of the policy configuration (EnterpriseContractConfiguration
+        object) to use. `namespace/name` or `name` syntax supported. If
+        namespace is omitted the namespace where the task runs is used.
+        `ec-policy` by default.
```
lcarva, line 22: Redundant since the default is defined on the next line?

appstudio-utils/util-scripts/lib/fetch/git.sh
zregvart: Doh

appstudio-utils/util-scripts/lib/fetch/cluster.sh
zregvart: I'm okay with it returning "" the message will be `"... in namespace"` so it'll at least prompt someone to set the namespace. Reworked this a bit it didn't take into consideration if custom namespace was specified.

simonbaird:
I didn't try it, but lgtm. Not sure I love the bash arg regex replace, but if it works it works.

simonbaird (approved):

simonbaird:
Would it be sensible to expect some e2e coverage for this sooner or later?

zregvart:
> Would it be sensible to expect some e2e coverage for this sooner or later?

Created [HACBS-468](https://issues.redhat.com/browse/HACBS-468)

PR #99, opened 2022-05-12
HACBS-223: Add task to run gosec

staticf0x:
- Test on a Go app: `./hack/test-build.sh https://github.com/redhat-appstudio/application-service docker-build`
- Test on a non-Go app: `./hack/test-build.sh https://github.com/jduimovich/single-container-app docker-build`

Both should pass, but `HACBS_TEST_OUTPUT` should only contain data for the Go app

### DIFF
--- a/tasks/sast-go.yaml
+++ b/tasks/sast-go.yaml
@@ -10,8 +10,8 @@ spec:
     - description: Test output
       name: HACBS_TEST_OUTPUT
   steps:
-      image: quay.io/redhat-appstudio/hacbs-test:stable
     - name: sast-go
+      image: quay.io/redhat-appstudio/hacbs-test:feature-sast
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
         /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=gosec_output.json $(workspaces.workspace.path)/... | tee gosec_output.txt
@@ -24,7 +24,7 @@ spec:
         exit $GOSEC_EXIT_CODE
 
     - name: test-format-result
-      image: quay.io/redhat-appstudio/hacbs-test:stable
+      image: quay.io/redhat-appstudio/hacbs-test:feature-sast
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
         HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
--- a/tasks/sast-go.yaml
+++ b/tasks/sast-go.yaml
@@ -11,12 +11,12 @@ spec:
       name: HACBS_TEST_OUTPUT
   steps:
     - name: gosec-check
-      image: quay.io/rsaar/hacbs-test:latest
+      image: quay.io/redhat-appstudio/hacbs-test:stable
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
         /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=gosec_output.json $(workspaces.workspace.path)/...
     - name: test-format-result
-      image: quay.io/rsaar/hacbs-test:latest
+      image: quay.io/redhat-appstudio/hacbs-test:stable
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
         HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
--- a/tasks/sast-go.yaml
+++ b/tasks/sast-go.yaml
@@ -12,12 +12,12 @@ spec:
   steps:
     - name: gosec-check
       image: quay.io/rsaar/hacbs-test:latest
-      workingDir: $(workspaces.workspace.path)
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       script: |
-        ls $(workspaces.workspace.path)
-        /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=$(workspaces.workspace.path)/gosec_output.json $(workspaces.workspace.path)/...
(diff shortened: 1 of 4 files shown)


Michkov (changes requested):
Please provide description, I do not understand why to clone application-service for each build. Why it's not in the testing image. Also why it's applied to all applications?

tasks/sast-go.yaml
```diff
@@ -0,0 +1,39 @@
...
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      script: |
+        /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=gosec_output.json $(workspaces.workspace.path)/... | tee gosec_output.txt
+        GOSEC_EXIT_CODE=$?
+
+        # Test if any package was found
+        # Even with -no-fail, gosec uses exit code 1 for several states,
+        # including when there are no packages found.
+        grep "No packages found$" gosec_output.txt && echo "Exit 0 because no packages found" && exit 0
+        exit $GOSEC_EXIT_CODE
```
Michkov, line 24: it should never end with different exit than 0

tasks/sast-go.yaml
```diff
@@ -0,0 +1,39 @@
...
+  results:
+    - description: Test output
+      name: HACBS_TEST_OUTPUT
+  steps:
+    - name: gosec-check
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      script: |
+        /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=gosec_output.json $(workspaces.workspace.path)/... | tee gosec_output.txt
+        GOSEC_EXIT_CODE=$?
```
Michkov, line 18: by default the script will use `bash -ex` which means that if $? is different then 0 then it will fail.

tasks/sast-go.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+        # Test if any package was found
+        # Even with -no-fail, gosec uses exit code 1 for several states,
+        # including when there are no packages found.
+        grep "No packages found$" gosec_output.txt && echo "Exit 0 because no packages found" && exit 0
+        exit $GOSEC_EXIT_CODE
+
+    - name: test-format-result
```
Michkov, line 26: I do not see reason to have this as separate step. BTW proposing to add this script to quay.io/redhat-appstudio/hacbs-test image so then it will be easier to reuse.

tasks/sast-go.yaml
```diff
@@ -0,0 +1,39 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: gosec-check
```
Michkov, line 5: Please use same name of task as name of the file.

tasks/sast-go.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - name: workspace
+  results:
+    - description: Test output
+      name: HACBS_TEST_OUTPUT
+  steps:
+    - name: gosec-check
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      script: |
+        /usr/local/go/bin/gosec -no-fail -fmt=sarif -out=gosec_output.json $(workspaces.workspace.path)/... | tee gosec_output.txt
```
Michkov, line 17: that `/...` (tree dots, `cd ...` returns `cd: ...: No such file or directory`) is handled by gosec or is it bug?

staticf0x: Yeah I was puzzled by that too, but the [README](https://github.com/securego/gosec#usage) for gosec says this:
>  To recursively scan from the current directory you can supply ./... as the input argument.

tasks/sast-go.yaml
staticf0x: Will it? I couldn't simulate that, is that not the default on either CRC or the quicklab clusters?

dirgim: Doesn't piping to `tee` overrule the exit code, so it should always be zero anyways?

tasks/sast-go.yaml
staticf0x: That's the problem I was solving here, perhaps it's a bit clunky. However, as the comment says, if we run the task for some repo that doesn't contain any Go code (the minimal reproducible case is by running `./hack/test-build.sh`), then even with `-no-fail`, gosec will exit with code 1. It will however print `No packages found` to stdout, which is the only indication (afaik, I might be wrong) for us that we should not fail the task as a whole. So in the case it finds no packages, line 23 will `exit 0` and if there is any package, we'll exit with whatever gosec decided to exit with. That would _usually_ be 0 but also not always. I was reading gosec's code and unfortunately I think this is what the developers left us with.

However now after typing my comment, I have a feeling that you might have meant something else, like that the task itself shouldn't exit with anything else than 0? If that's the case, no problem with reworking this, however I am still a bit confused about your other comment about `bash -ex` and how to handle non-Go repos.

tasks/sast-go.yaml
staticf0x: Done

tasks/sast-go.yaml
staticf0x: @dirgim you're right: 
```bash
$ false | tee
$ echo $?
0
$ false
$ echo $?
1
```

staticf0x:
@Michkov I hope I got everything you mentioned. So now, piping gosec to `tee` will handle other exit codes than 0 so we should always end up with 0. If there are no packages for gosec to check, the `HACBS_TEST_OUTPUT` will contain "SUCCESS" and 0 successes and 0 failures.

Michkov (approved):

srivickynesh (approved):
LGTM

dirgim (approved):
LGTM

PR #103, opened 2022-05-14
Produce results for the build pipelines

lcarva:
In order to attest PipelineRuns, we are working towards a change in
Chains. The likely mechanism to determine which image needs the
attestation should be assocaited with is by producing the type hinting
results on the PipelineRun itself, instead of the TaskRun. Although this
mechanism is pending approval, making this change in the Pipeline
definitions now allows us to more easily experiment with it.


lcarva:
@jduimovich, @Michkov wdyt?

cc @zregvart, @simonbaird, @bcaton85

Michkov (approved):

PR #104, opened 2022-05-16
WIP: Run findsecbugs test for java project

hongweiliu17:


tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,71 @@
...
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      script: |
+        cd ../../
+        if [ -f "pom.xml" ]; then
+          mvn package
+        else
+          echo "pom.xml file doesn't exist"
+        fi
+    - name: java-sec-check
```
Michkov, line 46: I do not see reason to split the task into multiple steps, it's using one image. So three containers are spawned even when there is no pom.xml.

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - name: OTHER_PARAM
+      description: other parameters to run findsecbugs
+      type: string
+      default: ""
+  steps:
+    - name: compile-java
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      script: |
+        cd ../../
```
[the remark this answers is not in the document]
hongweiliu17: Agree. Thanks!

tasks/sast-java-sec-check.yaml
hongweiliu17: So combine the three steps to one?

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,71 @@
...
+        ERR_MSG="$(cat stderr.txt)"
+        ERR_MSG="${ERR_MSG:-unknown}"
+        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --arg ERR_MSG "${ERR_MSG: 0: 3000}" --null-input \
+          '{result: "ERROR", timestamp: $date, failures: [$ERR_MSG]}')
+        HACBS_TEST_OUTPUT=$(jq -rce --arg date $(date +%s) \
+          '{ result: (if (.runs[].results | length > 0) then "FAILURE" else "SUCCESS" end),
+                   timestamp: $date,
+                   namespace: "default",
+                   successes: 0,
+                   failures: (.runs[].results // [])|map(.message.text)
```
hongweiliu17, line 69: @dirgim @Michkov @Dannyb48 When I test this task in my cluster, I meet error `Error while handling results: Termination message is above max allowed size 4096`, I think for failed test result, we can't output it to result and may need to update this `failure`. Don't know if other tests exceed the limitation. How about your idea?

Michkov: So one thing which I shared in different PR, is that this notification step could be using script which could be added directly into image, so then it can be re-used. And that script could also have check of the result size and potentially strip the output.

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,71 @@
...
+      description: other parameters to run findsecbugs
+      type: string
+      default: ""
+  steps:
+    - name: compile-java
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      image: quay.io/redhat-appstudio/hacbs-test:stable
+      script: |
+        cd ../../
+        if [ -f "pom.xml" ]; then
```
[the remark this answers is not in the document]
hongweiliu17: Thanks. Do you mean use it like $(params.path-context)/pom.xml?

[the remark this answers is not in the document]
Michkov: yes

tasks/sast-java-sec-check.yaml
Michkov: I do not see reason of having three containers, especially when step is not allowed to fail.

tasks/sast-java-sec-check.yaml
hongweiliu17: Goo idea.

tasks/sast-java-sec-check.yaml
hongweiliu17: Ok, let me combine them.

hongweiliu17: Done. Thanks!

tasks/sast-java-sec-check.yaml
dirgim: Agreed on having the script built into the hacbs-test image.
The other thing we would need to consider here is to not include the individual failures for SAST tests (such as this and the go one) into the JSON result output. 
The reason for this is that the results are supposed to be reviewed by the ProdSec team and are not supposed to be automatically consumed by the Enterprise contract.

hongweiliu17: @dirgim So how about just output the number of failed `runs[].results`?

dirgim: @hongweiliu17 yeah, that works for me

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+  params:
+    - name: PATH_CONTEXT
+      description: The path to your source code
+      type: string
+      default: .
+    - name: OUTPUT_FORMAT
+      description: the format of findsecbugs output
+      type: string
+      default: "-sarif"
+    - name: OUTPUT_ONLYANALYZE
```
staticf0x, line 21:
```suggestion
    - name: OUTPUT_ONLY_ANALYZE
```

staticf0x, line 21:
```suggestion
    - name: OUTPUT_ONLY_ANALYZE
```

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+      description: |
+        only analyze given classes and packages;
+        end with .* to indicate classes in a package, .- to indicate a package prefix
+      type: string
+      default: ""
+    - name: OTHER_PARAM
```
staticf0x, line 27: I think this could be named `OPTIONAL_ARGS` or `OPTIONAL_FLAGS` for the command to communicate the intention better.

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+          echo "pom.xml file doesn't exist"
+        fi
+        popd
+
+        JAR_PATH=$(workspaces.workspace.path)/$(params.PATH_CONTEXT)/target/*.jar
+        if [ -f "$JAR_PATH" ]; then
+          /home/findsecbugs-cli/findsecbugs.sh $(params.OTHER_PARAM) $(params.OUTPUT_ONLYANALYZE) $(params.OUTPUT_FORMAT) \
+            -output sast_java_sec_output.json $JAR_PATH  2> stderr.txt
+        else
+          echo "jar file doesn't exist" > stderr.txt
```
staticf0x, line 49: I know we're limited by the output size but would it make sense to have the full path displayed? If not in stderr.txt then maybe just echo it to the stdout?

staticf0x, line 49: I know we're limited by the output size but would it make sense to have the full path displayed? If not in stderr.txt then maybe just echo it to the stdout?

staticf0x (changes requested):

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+      name: HACBS_TEST_OUTPUT
+  params:
+    - name: PATH_CONTEXT
+      description: The path to your source code
+      type: string
+      default: .
+    - name: OUTPUT_FORMAT
+      description: the format of findsecbugs output
+      type: string
+      default: "-sarif"
```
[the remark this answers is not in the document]
hongweiliu17: Yes, thanks!

tasks/sast-java-sec-check.yaml
hongweiliu17: +1

tasks/sast-java-sec-check.yaml
hongweiliu17: Good idea, the file path can be displayed in stderr.txt

tasks/sast-java-sec-check.yaml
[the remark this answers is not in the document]
hongweiliu17: +1

staticf0x (approved):
LGTM

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+          mvn package -f $(params.PATH_CONTEXT)/
+        else
+          echo "pom.xml file doesn't exist in $(workspaces.workspace.path)/$(params.PATH_CONTEXT)"
+        fi
+        popd
+
+        JAR_PATH=$(workspaces.workspace.path)/$(params.PATH_CONTEXT)/target/*.jar
+        if [ -f "$JAR_PATH" ]; then
+          /home/findsecbugs-cli/findsecbugs.sh $(params.OPTIONAL_ARGS) $(params.OUTPUT_ONLY_ANALYZE) -$(params.OUTPUT_FORMAT) \
+            -output sast_java_sec_output.json $JAR_PATH  2> stderr.txt
```
jsztuka, line 47: Do we also want to display msg in case of an error to tekton taskrun also?

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+      description: the format of findsecbugs output
+      type: string
+      default: "sarif"
+    - name: OUTPUT_ONLY_ANALYZE
+      description: |
+        only analyze given classes and packages;
+        end with .* to indicate classes in a package, .- to indicate a package prefix
+      type: string
+      default: ""
+    - name: OPTIONAL_ARGS
```
jsztuka, line 27: Is it planned that this parameters together with OUTPUT_FORMAT and OUTPUT_ONLY_ANALYZE are going to be defined in pipeline in a future?

hongweiliu17: I think we don't know if OUTPUT_ONLY_ANALYZE and  OPTIONAL_ARGS will be used in pipeline or not in the future, but we just provide an option. They are not mandatory parameters, the task still works if the two parameters are not defined.

jsztuka (approved):
LGTM

srivickynesh (approved):
LGTM

dirgim (approved):
LGTM

tasks/sast-java-sec-check.yaml
```diff
@@ -0,0 +1,63 @@
...
+      description: optional parameters to run findsecbugs
+      type: string
+      default: ""
+  steps:
+    - name: java-sec-check
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      image: quay.io/redhat-appstudio/hacbs-test:feature-sast
```
Michkov, line 34: this is temporary?

dirgim: We are using that tag/branch for all of the SAST feature tasks while they are being developed. Once we have all the changes in, we will see how well the full container works (mainly because of its size and number of layers). If that looks good, we'll raise a PR to the main branch and cut a release.

Michkov: ok, thanks

PR #105, opened 2022-05-16
use conftest instead of opa

joejstuart:


appstudio-utils/util-scripts/check-ec-policy.sh
```diff
@@ -4,17 +4,20 @@ set -euo pipefail
...
 
-OPA_QUERY=${OPA_QUERY:-data.main.deny}
+CONFTEST_NAMESPACE=${CONFTEST_NAMESPACE:-main}
 
 [[ ! -d $DATA_DIR ]] && echo "Data dir $DATA_DIR not found!" && exit 1
+[[ ! -d $INPUT_DIR ]] && echo "Input dir $INPUT_DIR not found!" && exit 1
 [[ ! -d $POLICY_DIR ]] && echo "Policy dir $POLICY_DIR dir not found!" && exit 1
 
-opa eval \
+INPUT_FILES=$( find "${INPUT_DIR}" -type f -name "*.json"
```
joejstuart, line 18: This is to support multiple files as input. `json-input-file` creates the file path and takes the filename as an arg.

simonbaird: ...which makes me realize it's missing from the ec-policies PR.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
```diff
@@ -38,7 +38,7 @@ title Attestations
 # Show a shortened version of the attestations to avoid excessive logging
 echo -n "$attestations" | jq -r '.[].predicate = "..."'
```
simonbaird, line 39: Might need to tweak the jq query here, e.g.:

    jq -r '.attestations.[].predicate = "..."'

tasks/verify-enterprise-contract.yaml
```diff
@@ -32,7 +32,6 @@ spec:
         Public key used to verify signatures. Must be a valid k8s cosign
         reference, e.g. k8s://my-space/my-secret where my-secret contains
         the expected cosign.pub attribute.
-
```
simonbaird, line 35: Maybe unintentional whitespace change?

simonbaird:
Awesome, nice patch.

simonbaird (approved):

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -21,14 +21,14 @@ for tr in $( pr-get-tr-names $PR_NAME ); do
   if [[ ! -z "${data}" ]]; then
       result_found=1
       task_name=$( tr-get-task-name ${tr} )
-      echo "${data}" | jq > $( json-data-file test ${task_name} )
+      echo "$data" | jq '{ "tests": . }' > json-input-file "$tr-tests"
```
simonbaird, line 24: I think there's a problem here if there are multiple tasks with results. Firstly, each one would write to the same filename, since $tr is the same. Secondly, even if it didn't do that, I think conftest would merge all the values of the top level "tests" key and we'd end up losing all but the last one.

So I think somehow we need to merge the results to the file each time a new task run is added. I expect jq can handle that, but we might need some @lcarva level jq to do it.

At the end we might have something like this:

```
   {
     "tests": {
          "$task_name_1": { ... },
          "$task_name_2": { ... },
     }
   }
```

zregvart: Should not be more complicated than:
```suggestion
      echo "$data" | jq "{ \"tests\": { \"$task_name\": . } }" > json-input-file "$tr-tests"
```

simonbaird: From slack discussion: Let's not move the test result data into the input. Reasoning:
* Avoid more complicated changes incompatible with existing rego
* Avoid spending time on a problem that is likely to go away if we can in future access result data in the pipelinerun attestion

appstudio-utils/util-scripts/check-ec-policy.sh
simonbaird: ..later.. actually I think it's better if we don't put the test results into the input. We could still keep the find anyway, but it might be better to remove it since it makes the reader assume there are multiple files.

appstudio-utils/util-scripts/verify-attestation-with-policy.sh
simonbaird: Correction from your new revision: `jq -r '.attestations[].predicate = "..."'`

appstudio-utils/util-scripts/check-ec-policy.sh
```diff
@@ -4,17 +4,22 @@ set -euo pipefail
...
 
-opa eval \
+
+
+conftest test $INPUT_FILE \
   --data $DATA_DIR \
-  --data $POLICY_DIR \
-  --format $FORMAT \
-  $OPA_QUERY
+  --policy $POLICY_DIR/policies \
```
simonbaird, line 24: I would expect this to be `--policy $POLICY_DIR`

appstudio-utils/util-scripts/fetch-test-data.sh
```diff
@@ -23,8 +23,8 @@ for tr in $( pr-get-tr-names $PR_NAME ); do
   data=$( tr-get-result $tr $TEST_NAME )
   if [[ ! -z "${data}" ]]; then
       result_found=1
-      task_name=$( tr-get-task-name ${tr} )
-      echo "${data}" | jq > $( json-data-file test ${task_name} )
+      input_file=$( json-input-file )
```
simonbaird, line 26: Are you planning to put attestations and tests all in the one input file? That could work, but shorter patch to success here would probably be, something like this:

    # Line 18 or so
    TEST_RESULTS_FILE=$DATA_DIR/test-results.json

    # Line 27
    json-merge-with-key "$data" "$TEST_RESULTS_FILE" $tr test

simonbaird: Putting attestations and tests all in one file could work but it requires some extra work in the json-merge-with-key function. I am planning to do that anyway - will probably push a revision shortlyl

joejstuart: > Are you planning to put attestations and tests all in the one input file? That could work, but shorter path to success here would probably be something like this:
> 
> ```
> # Line 18 or so
> TEST_RESULTS_FILE=$DATA_DIR/test-results.json
> 
> # Line 27
> json-merge-with-key "$data" "$TEST_RESULTS_FILE" $tr test
> ```

No. not planning on that. Using the input file here was a mistake.

simonbaird:
Should be abandoned now in favor of #111 .

PR #106, opened 2022-05-17
Rename identifiers to snake_case in bash scripts

zregvart:
Identifiers containing a dash (`-`) are not allowed in posix compliant shell scripts. So this renames all our dashed identifiers to snake_case instead.

Why should we care about posix compliance? Simple answer is that error handing in bash scripts is better defined when run in posix mode.

See https://mywiki.wooledge.org/BashFAQ/105
See https://fvue.nl/wiki/Bash:_Error_handling


zregvart:
I'm going to close this, while heroic effort to understand bash error handling, we're stepping away from this, so let's not waste time on this.

PR #108, opened 2022-05-18
Clair - do not wait for scan result

Michkov:
Since clair scan could take from 10 minutes to +5 hours then it does not
make sense to block pipelineRun by 5 minutes.


jsztuka (approved):
LGTM

dirgim (approved):
+1

PR #112, opened 2022-05-23
Add auxiliary enterprise contract pipeline

zregvart:
We'll refactor to run pipelines in the end to end tests, so we need a pipeline that contains enterprise contract tasks. This creates a `auxiliary-ec` pipeline and places it in the `hacbs-templates-bundle` image.

Ref. https://issues.redhat.com/browse/HACBS-424


simonbaird:
I don't understand the name choice of "auxilliary". Is it solely for e2e? If so, why not put e2e in its name?

simonbaird (approved):
Approve, but I'm still wondering about the name choice.

robnester-rh (approved):
LGTM

PR #113, opened 2022-05-26
Update buildah task to include SBOM generation

brunoapimentel:
PLNSRVCE-184

### DIFF
--- a/tasks/buildah.yaml
+++ b/tasks/buildah.yaml
@@ -59,6 +59,33 @@ spec:
         $(params.BUILD_EXTRA_ARGS) --format=$(params.FORMAT) \
         --tls-verify=$(params.TLSVERIFY) --no-cache \
         -f $(params.DOCKERFILE) -t $(params.IMAGE) $(params.CONTEXT)
+      buildah push --storage-driver=$(params.STORAGE_DRIVER) $(params.IMAGE) oci:/var/lib/containers/oci
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+    workingDir: $(workspaces.source.path)
+  - image: quay.io/redhat-appstudio/syft:v0.46.2
+    name: sbom-source
+    args:
+      - dir:$(workspaces.source.path)
+      - --file=$(workspaces.source.path)/sbom-source.json
+      - --output=cyclonedx-json
+  - image: quay.io/redhat-appstudio/syft:v0.46.2
+    name: sbom-image
+    args:
+      - oci-dir:/var/lib/containers/oci
+      - --file=$(workspaces.source.path)/sbom-image.json
+      - --output=cyclonedx-json
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+  - image: $(params.BUILDER_IMAGE)
+    name: inject-sbom
+    resources: {}
+    script: |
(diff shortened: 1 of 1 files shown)


tasks/buildah.yaml
```diff
@@ -63,6 +63,33 @@ spec:
...
+      - --output=cyclonedx-json
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+  - image: $(params.BUILDER_IMAGE)
+    name: inject-sbom
+    resources: {}
+    script: |
+      container=$(buildah from --storage-driver=$(params.STORAGE_DRIVER) --pull-never $(params.IMAGE))
+      buildah --storage-driver=$(params.STORAGE_DRIVER) copy $container $(workspaces.source.path)/sbom-source.json /root/
```
gabemontero, line 86: @stuartwdouglas - this is an example of the calling buildah from a script/shell in order to copy content out of an image in mentioned in the hacbs milestone 6 checkpoint today

in theory we could do something similar in the jvm-build-service for accessing content stored in an image (vs. S3 storage)

gabemontero: thanks for the timely example @brunoapimentel :-)

brunoapimentel: I should give the credits to @ejegrova, all I did was copying the code she wrote :D

gabemontero: totally - and I did mention @ejegrova in our meeting this morning

tasks/buildah.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     description: The location of the buildah builder image.
     name: BUILDER_IMAGE
     type: string
-  - default: vfs
+  - default: overlay
```
Michkov, line 20: Reason of changing this? Overlay does not work in openshift (as you cans see in CI).

Michkov (changes requested):

tasks/buildah.yaml
brunoapimentel: Because when I was using vfs, the  image was not being found. Probably because my environment is a Kind cluster :grimacing: 
I was about to test it in a real cluster berfore marking the PR as ready to review.

gabemontero: yeah only privileged or run as root users can use overlay

tasks/buildah.yaml
```diff
@@ -59,6 +59,34 @@ spec:
...
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+  - image: $(params.BUILDER_IMAGE)
+    name: inject-sbom
+    resources: {}
+    script: |
+      container=$(buildah from --storage-driver=$(params.STORAGE_DRIVER) --pull-never $(params.IMAGE))
+      buildah --storage-driver=$(params.STORAGE_DRIVER) copy $container $(workspaces.source.path)/sbom-source.json /root/
+      buildah --storage-driver=$(params.STORAGE_DRIVER) copy $container $(workspaces.source.path)/sbom-image.json /root/
```
Michkov, line 88: nit: you can set `workingDir: $(workspaces.source.path)` and then copy both files together -> `copy $container sbom-image.json sbom-source.json  /root/` (not tested). Probably it's not bad to add `--add-history` flag.

tasks/buildah.yaml
brunoapimentel: The error was that I was not setting the flag in the push command. It is working now.

PR #114, opened 2022-05-26
Produce materials for the build pipelines

lcarva:
When Tekton Chains attests a PipelineRun, it looks for the presence of
either results, or parameters, named CHAINS-GIT_URL and
CHAINS-GIT_COMMIT. If found, this information is then propagated to the
.predicate.materials section of the SLSA Provenance attestation.

This commit makes the build pipelines produce this information in their
results. Although using parameters is also a possibility, it has a
stronger impact on the pipeline usability since users would have to
specify a different parameter for the git url and commit.

### DIFF
--- a/pipelines/base/template-build/template-build.yaml
+++ b/pipelines/base/template-build/template-build.yaml
@@ -111,6 +111,10 @@ spec:
       value: "$(tasks.build-container.results.IMAGE_URL)"
     - name: IMAGE_DIGEST
       value: "$(tasks.build-container.results.IMAGE_DIGEST)"
+    - name: CHAINS-GIT_URL
+      value: "$(tasks.clone-repository.results.url)"
+    - name: CHAINS-GIT_COMMIT
+      value: "$(tasks.clone-repository.results.commit)"
   workspaces:
     - name: workspace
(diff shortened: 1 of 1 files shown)


joejstuart (approved):
Looks good.

robnester-rh (approved):
LGTM

pipelines/base/template-build/template-build.yaml
```diff
@@ -111,6 +111,10 @@ spec:
       value: "$(tasks.build-container.results.IMAGE_URL)"
     - name: IMAGE_DIGEST
       value: "$(tasks.build-container.results.IMAGE_DIGEST)"
+    - name: CHAINS-GIT_URL
```
Michkov, line 114: it's generic result not related to CHAINS, so it could be just GIT_URL

lcarva: Yes, but Chains specifically looks for results with these names. There's some [upstream conversation](https://github.com/tektoncd/community/pull/697) in making this better, but that is far off. We need these results to be named this way in the foreseeable future.

Michkov (approved):

PR #116, opened 2022-05-27
Add snyk code test task for doing SAST check

jinqi7:
Signed-off-by: Jing Qi <jinqi@redhat.com>

Test example:  ./test-build.sh https://github.com/jduimovich/single-container-app docker-build

### DIFF
--- a/pipelines/hacbs/hacbs-test.yaml
+++ b/pipelines/hacbs/hacbs-test.yaml
@@ -104,6 +104,17 @@
 - op: add
   path: /spec/tasks/-
   value:
+    name: sast-snyk-check
+    runAfter:
+      - clone-repository
+    taskRef:
+      name: sast-snyk-check
+    workspaces:
+    - name: workspace
+      workspace: workspace
+- op: add
+  path: /spec/tasks/-
+  value:
     name: clamav-scan
     runAfter:
       - build-container
--- a/tasks/kustomization.yaml
+++ b/tasks/kustomization.yaml
@@ -21,6 +21,7 @@ resources:
 - verify-enterprise-contract.yaml
 - sast-go.yaml
 - sast-java-sec-check.yaml
+- sast-snyk-check.yaml
 - clamav-scan.yaml
 
 patches:
--- a/tasks/sast-snyk-check.yaml
+++ b/tasks/sast-snyk-check.yaml
@@ -1,0 +1,58 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sast-snyk-check
+spec:
+  description: >-
+    Static code security test with snyk
+  results:
+    - description: Test output
+      name: HACBS_TEST_OUTPUT
+  params:
+    - name: SHARED_SECRET
+      default: test-team-snyk
+    - name: ARGS
+      type: string
+      description: extra args needs to append
+      default: "--all-projects --exclude=test*,vendor,deps"
+  volumes:
+    - name: snyk-secret
+      csi:
+        readOnly: true
+        driver: csi.sharedresource.openshift.io
+        volumeAttributes:
+          sharedSecret: $(params.SHARED_SECRET)
+  steps:
+    - name: sast-snyk-check
+      image: quay.io/redhat-appstudio/hacbs-test:feature-sast
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -z $SNYK_TOKEN ]]; then
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' need to be created in test-team namespace" | tee stdout.txt
+          exit 0
+        fi
+        export SNYK_TOKEN
+        SNYK_EXIT_CODE=0
+        snyk code test $(params.ARGS) ../.. --sarif-file-output=sast_snyk_check_out.json 1>&2>> stdout.txt || SNYK_EXIT_CODE=$?
+        if [[ "$SNYK_EXIT_CODE" -eq 0 ]] || [[ "$SNYK_EXIT_CODE" -eq 1 ]]; then
+          HACBS_TEST_OUTPUT=$(jq -rce --arg date $(date +%s) \
+            '{ result: (if (.runs[].results | length > 0) then "FAILURE" else "SUCCESS" end),
+                   timestamp: $date,
+                   namespace: "default",
+                   successes: 0,
+                   failures: (.runs[].results // [])|map(.message.text)
+                 }' sast_snyk_check_out.json || true)
+        else
+          ERR_MSG="$(cat stdout.out)"
+          HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --arg ERR_MESSAGE "${ERR_MSG}" --null-input \
+            '{result: "ERROR", timestamp: $date, failures: [$ERR_MESSAGE]}')
+        fi
+        echo "${HACBS_TEST_OUTPUT:-${HACBS_ERROR_OUTPUT}}" | tee $(results.HACBS_TEST_OUTPUT.path)
+  workspaces:
+  - name: workspace


tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: snyk-code-test
+spec:
+  description: >-
+    Static code security test with snyk 
+  workspaces:
+    - name: workspace
+      optional: true
```
Michkov, line 10: workspaces defined twice, should not be optional

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+        volumeAttributes:
+          sharedSecret: $(params.shared-secret)
+  steps:
+    - name: snyk-code-test 
+      image: quay.io/jinqi/hacbstest_jinqi_1:latest
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly:true
```
Michkov, line 35: missing space

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
```
Michkov, line 38: nit: handle situation directly when SNYK_TOKEN is empty.

jinqi7: It can be handled by snyk cmd and  it will be reported as an error in the end of the task.

Michkov: it depends what it will report if it will be "No password provided" then it's fine, but if it will return "Unauthorized" then it's not clear if the secret is provided properly.

jinqi7: The  error message is "`snyk` requires an authenticated account. Please run `snyk auth` and try again." and updated it to prompt a message for creating a secret 'test-team-snyk' .

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,60 @@
...
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -e $SNYK_TOKEN ]]; then
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' is needed to be created in test_team namespace for it" 
+          exit 1
```
Michkov, line 38: the task should not fail, `exit 0`

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,60 @@
...
+      image: quay.io/redhat-appstudio/hacbs-test:feature-sast
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -e $SNYK_TOKEN ]]; then
```
Michkov, line 36: `-e FILE` - FILE exists. $SNYK_TOKEN is not file, it's string. so it should be `[ -z "$SNYK_TOKEN" ]`

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,60 @@
...
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' is needed to be created in test_team namespace for it" 
+          exit 1
+        fi
+        export SNYK_TOKEN
+        pushd $(workspaces.workspace.path)
```
Michkov, line 41: workingDir is not used, so it can be set to workingDir instead of pushd.

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,60 @@
...
+        SNYK_EXIT_CODE=0
+        snyk code test $(params.args) --sarif-file-output=snyk_code_test_out.json 1>&2> stdout.out || SNYK_EXIT_CODE=$?
```
Michkov, line 43: the output should go into test folder not to workspace root dir.

jinqi7: Updated

tasks/snyk-code-test.yaml
jinqi7: Updated

tasks/snyk-code-test.yaml
jinqi7: Updated

tasks/snyk-code-test.yaml
jinqi7: Updated

pipelines/hacbs/hacbs-test.yaml
```diff
@@ -87,3 +87,14 @@
...
     - name: workspace
       workspace: workspace
+- op: add
+  path: /spec/tasks/-
+  value:
+    name: snyk-code-test
+    runAfter:
+      - build-container
+    taskRef:
+      name: sny-code-test
```
Michkov, line 97: typo

jinqi7: Fixed

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -z $SNYK_TOKEN ]]; then
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' need to be created in test-team namespace"
+          exit 0
+        fi
+        export SNYK_TOKEN
+        SNYK_EXIT_CODE=0
+        mkdir test
```
Michkov, line 42: `test` is very generic, so here is high risk that the folder will exist in the source code (that's in this workspace). Rather use `$(workspaces.workspace.path)/hacbs/$(context.task.name)` as in other tasks.

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+metadata:
+  name: snyk-code-test
+spec:
+  description: >-
+    Static code security test with snyk
+  results:
+    - description: Test output
+      name: HACBS_TEST_OUTPUT
+  params:
+    - name: shared-secret
```
dirgim, line 12: nit: I'd suggest using the upper case and underscore separated format for the params, e.g. `SHARED_SECRET` in this case.

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+      default: test-team-snyk
+    - name: args
```
dirgim, line 14: nit: I'd suggest using the upper case and underscore separated format for the params, e.g. `ARGS` in this case.

jinqi7: Updated. Thanks

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,60 @@
...
+      workingDir: $(workspaces.workspace.path)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -z $SNYK_TOKEN ]]; then
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' need to be created in test-team namespace"
```
dirgim, line 37: I think we want to capture this error in the `HACBS_TEST_OUTPUT` result, like on line 55.

dirgim, line 37: I think we want to capture this error in the `HACBS_TEST_OUTPUT` result, like on line 55.

jinqi7: OK. Updated

tasks/snyk-code-test.yaml
```diff
@@ -0,0 +1,58 @@
...
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+      script: |
+        #!/usr/bin/env bash
+        SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
+        if [[ -z $SNYK_TOKEN ]]; then
+          echo "SNYK_TOKEN is empty and a secret 'test-team-snyk' which includes 'snyk_token' need to be created in test-team namespace" > stdout.txt
```
srivickynesh, line 37: Can we create the secret if it fails? This needs to be handled within our tests.

jinqi7: We are not sure the user of the task should have the permission to create the secret. So left it as a prework .

srivickynesh: If anyone tries to run it on their own cluster. It should not fail with SNYK_TOKEN empty.

dirgim: We can't create a secret if it's not available within the user's workspace. The current idea is to have a shared secret available for Red Hat builds, but customers would need to set their their own snyk token. If it's not there, we should report an `ERROR` in the task result (but not fail the task itself of course).

jinqi7: I tried in my cluster with latest infra-deployment & created the shared-secret of "test-team-snyk", the task can get the  snyk_token successfully.  It's a shared-secret and not supposed to be created by every user of task.

srivickynesh: Okay. thanks for the update.

pipelines/hacbs/hacbs-test.yaml
```diff
@@ -101,3 +101,14 @@
     workspaces:
     - name: workspace
       workspace: workspace
+- op: add
+  path: /spec/tasks/-
+  value:
+    name: snyk-code-test
```
dirgim, line 107: I think we should go with the convention to prefix the task name with `sast-`, so I'm proposing naming this task `sast-synk-check`

jinqi7: Updated

dirgim (approved):
LGTM

tasks/snyk-code-test.yaml
jinqi7: @Michkov , can you please help to review it again? Thanks

Michkov (approved):

hongweiliu17 (approved):
LGTM

jsztuka (approved):
LGTM

PR #119, opened 2022-05-31
Update buildah task to merge both generated SBOMs 

brunoapimentel:
PLNSRVCE-184

To improve Syft's report, both the image and the source code were scanned and two distinct SBOM's were generated. This PR will merge both contents into a single file to make it more concise.

### DIFF
--- a/tasks/buildah.yaml
+++ b/tasks/buildah.yaml
@@ -79,12 +79,40 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+  - image: registry.redhat.io/ubi8/python-39:1-51
+    name: merge-sboms
+    script: |
+      #!/bin/python3
+      import json
+
+      with open("./sbom-image.json") as f:
+        image_sbom = json.load(f)
+
+      with open("./sbom-source.json") as f:
+        source_sbom = json.load(f)
+
+      def get_identifier(component):
+        return component["name"] + '@' + component["version"]
+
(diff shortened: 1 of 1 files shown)


tasks/buildah.yaml
```diff
@@ -79,12 +79,38 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+  - image: registry.fedoraproject.org/fedora:35
```
brunoapimentel, line 82: I tried using `registry.access.redhat.com/ubi9/python-39` as a base image for this, but since it uses the `default` user, I was getting a permission error on L105. Not sure if there's a better image to use to run this Python script.

gabemontero: is that a typo @brunoapimentel or did you really use `ubi9` instead of `ubi8` ?

That aside, yeah, the official python image shipped with the openshift samples is `registry.redhat.io/ubi8/python-39:latest` 

Next, by `default` user do you mean the `default` service account?  If so, then yeah, I would think the pipeline runs that use this task will need to use the `pipeline` service account in order to have sufficient permissions.

Otherwise, I would need to look at pipeline run / pod yaml, events, and pod logs.

But let's see if @Michkov @ejegrova have anything to add.

brunoapimentel: Same issue happens on ubi8, btw.

Michkov:
```
      securityContext:
        runAsUser: 0
```
Should fix it.

PR #125, opened 2022-06-02
[HACBS-426]: Integrate the deprecated base image check

staticf0x:
- Checks the base image using `org.opencontainers.image.source`
- Add the check to `hacbs-test` pipeline

### DIFF
--- a/pipelines/hacbs/hacbs-test.yaml
+++ b/pipelines/hacbs/hacbs-test.yaml
@@ -49,7 +49,7 @@
     - name: IMAGE_REGISTRY
       value: registry.access.redhat.com
     - name: IMAGE_REPOSITORY
-      value: $(params.git-url)
+      value: $(tasks.sanity-inspect-image.results.BASE_IMAGE_REPOSITORY)
     workspaces:
     - name: sanity-ws
       workspace: workspace
--- a/pipelines/hacbs/hacbs-test.yaml
+++ b/pipelines/hacbs/hacbs-test.yaml
@@ -47,7 +47,7 @@
       name: deprecated-image-check
     params:
     - name: IMAGE_REGISTRY
-      value: quay.io
+      value: registry.access.redhat.com
     - name: IMAGE_REPOSITORY
       value: $(params.git-url)
     workspaces:
--- a/pipelines/hacbs/hacbs-test.yaml
+++ b/pipelines/hacbs/hacbs-test.yaml
@@ -40,6 +40,22 @@
 - op: add
   path: /spec/tasks/-
   value:
+    name: deprecated-base-image-check
+    runAfter:
+      - sanity-inspect-image
+    taskRef:
+      name: deprecated-image-check
+    params:
+    - name: IMAGE_REGISTRY
+      value: quay.io
+    - name: IMAGE_REPOSITORY
+      value: $(params.git-url)
+    workspaces:
+    - name: sanity-ws
+      workspace: workspace
+- op: add
+  path: /spec/tasks/-
+  value:
     name: get-clair-results
     runAfter:
       - build-container
--- a/tasks/sanity-inspect-image.yaml
+++ b/tasks/sanity-inspect-image.yaml
@@ -52,6 +52,4 @@ spec:
       skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
       echo "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
 
-      BASE_IMAGE_INSPECT_RAW=$(skopeo inspect --no-tags --raw docker://$BASE_IMAGE)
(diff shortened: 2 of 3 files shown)


pipelines/hacbs/hacbs-test.yaml
```diff
@@ -37,6 +37,22 @@
...
+  path: /spec/tasks/-
+  value:
+    name: deprecated-base-image-check
+    runAfter:
+      - sanity-inspect-image
+    taskRef:
+      name: deprecated-image-check
+    params:
+    - name: IMAGE_REGISTRY
+      value: quay.io
```
dirgim, line 50: I think we can leave the `IMAGE_REGISTRY` param with the default `registry.access.redhat.com` value, since there should be no entries in Pyxis for `quay.io` images.

staticf0x: Good point, fixed

pipelines/hacbs/hacbs-test.yaml
```diff
@@ -37,6 +37,22 @@
...
+    name: deprecated-base-image-check
+    runAfter:
+      - sanity-inspect-image
+    taskRef:
+      name: deprecated-image-check
+    params:
+    - name: IMAGE_REGISTRY
+      value: registry.access.redhat.com
+    - name: IMAGE_REPOSITORY
+      value: $(params.git-url)
```
hongweiliu17, line 52: Is this the source code url? But I think `IMAGE_REPOSITORY` should point to image repo, is it?

tasks/sanity-inspect-image.yaml
```diff
@@ -49,3 +51,7 @@ spec:
       echo "The base image is $BASE_IMAGE, get its manifest now"
       skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
       echo "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
+
+      BASE_IMAGE_INSPECT_RAW=$(skopeo inspect --no-tags --raw docker://$BASE_IMAGE)
+      BASE_IMAGE_REPOSITORY="$(echo $BASE_IMAGE_INSPECT_RAW | jq -r ".annotations.\"org.opencontainers.image.source\"")"
```
hongweiliu17, line 56: According to https://github.com/opencontainers/image-spec/blob/main/annotations.md, this `org.opencontainers.image.source` is `URL to get source code for building the image`, not a image repo.

pipelines/hacbs/hacbs-test.yaml
dirgim: @staticf0x yes, it should be `BASE_IMAGE_REPOSITORY` from the sanity-inspect-image task, so something like 
```
$(tasks.sanity-inspect-image.results.BASE_IMAGE_REPOSITORY)
```

dirgim (approved):
LGTM, thanks!

jsztuka (approved):
LGTM

tasks/sanity-inspect-image.yaml
```diff
@@ -49,3 +51,5 @@ spec:
       echo "The base image is $BASE_IMAGE, get its manifest now"
       skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
       echo "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
+
+      jq -r ".Name" $BASE_IMAGE_INSPECT | cut -d"/" -f2,3 | tee $(results.BASE_IMAGE_REPOSITORY.path)
```
hongweiliu17, line 55: If we don't make sure the registry is always `registry.access.redhat.com`, so should we also expose the registry to result? Otherwise the `IMAGE_REGISTRY` in deprecated-image-check.yaml is useless.

hongweiliu17 (commented):
Just one comment, and others look good to me, thanks!

tasks/sanity-inspect-image.yaml
hongweiliu17: Actually `BASE_IMAGE` is like `quay.io/hongweiliu/test_image:tag`, you may remove `BASE_IMAGE` and expose base image registry as well. Thanks!

dirgim: There is an issue with the entries for Red Hat images - in Pyxis, they are always defined as `registry.access.redhat.com`, even though the image could be hosted in `registry.redhat.io` or similar. 
Another issue is that Pyxis contains only the entries to Red Hat images and nothing else. This is why the plan is to attempt to test the image repository _as if_ it was hosted in `registry.access.redhat.com`. If we don't find the entry in Pyxis, we will exit gracefully from the task.

hongweiliu17: I see now. Thanks.

PR #128, opened 2022-06-02
Split buildah task to make the SBOM generation steps reusable

brunoapimentel:
This is a proposal to have the sbom generation steps from the buildah task split into a new task. But in order to do that, the push step will also need to be separated into a new task. 

Here's how the docker-build pipeline will look like after the split:
image

The code still needs some clean up, but I created this draft so I could get early comments on the proposal.

### DIFF
--- a/hack/test-build.sh
+++ b/hack/test-build.sh
@@ -42,4 +42,4 @@ else
   echo Building $IMG
 fi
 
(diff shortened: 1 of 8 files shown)


brunoapimentel:
Closing it because we decided to go with the approch in #131

PR #130, opened 2022-06-02
Add converter from cyclonedx to purl format

ejegrova:
PLNSRVCE-202

### DIFF
--- a/tasks/patches/add-sbom-and-push.yaml
+++ b/tasks/patches/add-sbom-and-push.yaml
@@ -50,8 +50,29 @@
 
       image_sbom["components"].sort(key=lambda c: get_identifier(c))
 
-      with open("./sbom.json", "w") as f:
+      with open("./sbom-cyclonedx.json", "w") as f:
         json.dump(image_sbom, f)
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
+
+- op: add
+  path: /spec/steps/-
+  value:
+    image: registry.redhat.io/ubi8/python-39:1-51
+    name: convert-cyclonedx-to-purl
+    script: |
+      #!/bin/python3
+      import json
+
+      with open("./sbom-cyclonedx.json") as f:
+          data = json.load(f)
+
+      purls = [{"purl": component["purl"]} for component in data["components"] if "purl" in component]
+      purl_content = {"image_contents": {"dependencies": purls}}
+
+      with open("sbom-purl.json", "w") as output_file:
+          json.dump(purl_content, output_file)
     workingDir: $(workspaces.source.path)
     securityContext:
       runAsUser: 0
@@ -64,7 +85,7 @@
     resources: {}
     script: |
       container=$(buildah from --pull-never $(params.IMAGE))
-      buildah copy $container sbom.json /root/
+      buildah copy $container sbom-cyclonedx.json sbom-purl.json /root/
       buildah commit $container $(params.IMAGE)
     volumeMounts:
     - mountPath: /var/lib/containers


tasks/buildah.yaml
```diff
@@ -113,12 +113,30 @@ spec:
...
+        data = json.load(f)
+
+      purls = [component["purl"] for component in data["components"] if "purl" in component]
+
+      with open("sbom-purl", "w") as output_file:
+        for purl in purls:
+          output_file.write(f"{purl}\n")
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
```
gabemontero, line 133: do we need this for all uses of the python image, or just when we were using buidlah with it (I thought it was the latter) @brunoapimentel @Michkov @ejegrova

Michkov: for all tasks which creates files, since git is cloned with user 0 (root), when you want to write into dir then you need to be root as well.

tasks/buildah.yaml
```diff
@@ -113,12 +113,30 @@ spec:
...
+    name: convert-cyclonedx-to-purl
+    script: |
+      #!/bin/python3
+      import sys
+      import json
+
+      with open("./sbom.json") as f:
+        data = json.load(f)
+
+      purls = [component["purl"] for component in data["components"] if "purl" in component]
```
Michkov, line 126: it's returning just list of packages, purl format is json.

Michkov (changes requested):

tasks/buildah.yaml
```diff
@@ -113,12 +113,30 @@ spec:
...
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
   - image: $(params.BUILDER_IMAGE)
     name: inject-sbom
     resources: {}
     script: |
       container=$(buildah from --storage-driver=$(params.STORAGE_DRIVER) --pull-never $(params.IMAGE))
-      buildah --storage-driver=$(params.STORAGE_DRIVER) copy $container sbom.json /root/
+      buildah --storage-driver=$(params.STORAGE_DRIVER) copy $container sbom.json sbom-purl /root/
```
brunoapimentel, line 139: I think it would be good to change it to `sbom-purl.json`, since it's still a json file.

tasks/buildah.yaml
gabemontero: ah that makes sense then, and I was unaware of how the git clone was occurring ... thanks for the clarification @Michkov

gabemontero: marking resolved

tasks/buildah.yaml
brunoapimentel: You should use `json.dump` when you write the file.

brunoapimentel: Also, be mindful of the format we use in Cachito nowadays. I know we won't be able to make it 100% equal, but we can at least maintain the structure like:
```
{
    "image_contents": {
        "dependencies": [
            {
                "purl": "pkg:alpine/alpine-baselayout@3.2.0-r18?arch=x86_64&upstream=alpine-baselayout&distro=alpine-3.15.4"
            },
            {
                "purl": "pkg:alpine/alpine-keys@2.4-r1?arch=x86_64&upstream=alpine-keys&distro=alpine-3.15.4"
            },
            //...
         ]
     }
}
```

tasks/buildah.yaml
brunoapimentel: And maybe change the other to `sbom-cyclonedx.json`?

ejegrova: changed both names

tasks/buildah.yaml
ejegrova: used the format from @brunoapimentel  suggestion, please check

PR #131, opened 2022-06-03
Share sbom-generation and push steps between tasks

Michkov:


mbharatk (approved):

Michkov:
@brunoapimentel Parameters removed since we do not want to allow to change them.

brunoapimentel (approved):

PR #132, opened 2022-06-03
Always build with HACBS pipelines

Michkov:
Switch default parameter for rebuild to 'true' in hacbs pipelines.


jduimovich (approved):
LGTM. Good solutions

PR #135, opened 2022-06-04
Adds default value for PUBLIC_KEY

davidmogar:
Right now, a call to the verify-enterprise-contract task requires to pass an image (which will be replaced by an ApplicationSnapshot) and a public key which would look something like k8s://task_namespace/name_of_the_secret.

Ideally, managers of the managed workspace shouldn't have to worry about passing this value in the ReleaseStrategy. To avoid this, we can agree on a fixed name for the secret (cosign-public-key) and get the actual path with variable substitutions.

This commit sets a default value to PUBLIC_KEY which will get the namespace of the running task and append the name `cosign-public-key`. With this change, managers will be only forced to create a secret with this specific name, but ReleaseStrategies will remain valid even when they are deployed in a different workspace.

Signed-off-by: David Moreno García <damoreno@redhat.com>


simonbaird:
ltgm

simonbaird (approved):

lcarva:
Seem like a reasonable default value. LGTM!

PR #139, opened 2022-06-06
sast-go: increase memory limit to 8Gi

Michkov:
4Gi is not enough to build KCP.

For container building 2Gi is enough but for building project like kcp we need more than 4 GB.


dirgim (approved):
lgtm

PR #143, opened 2022-06-07
Fix fallback when missing ECP resource

lcarva:
This fixes the e2e tests by allowing the fallback to be used if an ECP name was provided but does not exist which is now the case since `POLICY_CONFIGURATION` has a default value and is actually used in the bash scripts.


lcarva:
@redhat-appstudio/enterprise-contract

zregvart (approved):

lcarva:
@zregvart, @simonbaird, merge? :pray:

PR #144, opened 2022-06-07
Tweak verify-pipeline so future violations work

simonbaird:
There's a rego bug in ec-policies that means the future policy
violations are not handled correctly if data.config is missing
entirely.

This is a quick workaround for the demo. Hopefully the rego bug will
be fixed later properly and the workaround won't be needed.


simonbaird:
I'll merge it for demo in 40-ish minutes.

PR #145, opened 2022-06-08
Jq for color highlights in verify-pipeline script

simonbaird:


simonbaird:
CI is probably irrelevant since this is a one line change in a bash script in ./hack.

PR #149, opened 2022-06-17
PLNSRVCE-382: Enhance s2i-java and buildah tasks to support jvm-build…

Michkov:
…-service

- removing hacbs-java-builder
- adding jvm-build-service proxy
- merging steps to avoid quota issues
- getting sbom from image dir instead of pushing to oci

Todo:
- [x] enhance buildah builder - detect java and set mirror in dockerfile


brunoapimentel:
Regarding the changes to the SBOM generation, it LGTM.

PR #151, opened 2022-06-22
sast-go: ignore return code of gosec

Michkov:
Gosec could return different return code than 0 which aborts
pipelinerun. Before there was 'tee' command used which hide this
problem.


jinqi7 (approved):

jinqi7:
LGTM

PR #154, opened 2022-07-06
Update ec-policies on new Pipeline Bundles

lcarva:
https://issues.redhat.com/browse/HACBS-748

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -85,6 +85,37 @@ spec:
               done
         taskRef:
           name: update-infra-deployments
+      - name: update-ec-policies
+        runAfter:
+          - build-bundles
+        params:
+          - name: ORIGIN_REPO
+            value: $(params.git-url)
+          - name: REVISION
+            value: $(params.revision)
+          - name: TARGET_GH_REPO
+            value: hacbs-contract/ec-policies
+          - name: GIT_IMAGE
+            value: "$(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)"
+          - name: SCRIPT
+            value: |
+              TAG='$(params.revision)'
+
+              ec track bundle \
+                --input data/acceptable_tekton_bundles.yml \
+                --type task-bundles \
+                --replace \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-1" \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-2" \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-3"
+
+              ec track bundle \
+                --input data/acceptable_tekton_bundles.yml \
+                --type pipeline-bundles \
+                --replace \
+                --bundle "quay.io/redhat-appstudio/build-templates-bundle:${TAG}"
+        taskRef:
+          name: update-infra-deployments
     workspaces:
       - name: workspace
   workspaces:
--- a/tasks/update-infra-deployments.yaml
+++ b/tasks/update-infra-deployments.yaml
@@ -9,6 +9,12 @@ spec:
       description: Script for changing the infra-deployments
     - name: ORIGIN_REPO
     - name: REVISION
+    - name: TARGET_GH_REPO
+      description: GitHub repository of the infra-deployments code
+      default: redhat-appstudio/infra-deployments
+    - name: GIT_IMAGE
+      description: Image reference containing the git command
+      default: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:af7dd5b3b1598a980f17d5f5d3d8a4b11ab4f5184677f7f17ad302baa36bd3c1
     - name: shared-secret
       default: infra-deployments-pr-creator
   volumes:
@@ -25,7 +31,7 @@ spec:
 
   steps:
     - name: update-infra-deployments
-      image: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:af7dd5b3b1598a980f17d5f5d3d8a4b11ab4f5184677f7f17ad302baa36bd3c1
+      image: $(params.GIT_IMAGE)
       volumeMounts:
         - name: infra-deployments-pr-creator
           mountPath: /secrets/deploy-key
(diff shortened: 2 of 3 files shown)


tasks/update-infra-deployments.yaml
```diff
@@ -23,7 +26,7 @@ spec:
           # 'appstudio-staging-ci' - private key for appstudio-staging-ci Github app
   steps:
     - name: update-infra-deployments
-      image: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:af7dd5b3b1598a980f17d5f5d3d8a4b11ab4f5184677f7f17ad302baa36bd3c1
+      image: appstudio-utils
```
lcarva, line 29: Using this image so I can leverage the `yq` command in the `SCRIPT`.

lcarva:
cc @redhat-appstudio/enterprise-contract

tasks/update-infra-deployments.yaml
simonbaird: Does appstudio-utils have enough python installed?

tasks/update-infra-deployments.yaml
```diff
@@ -38,14 +41,17 @@ spec:
...
         REPO_NAME=$(echo $(params.ORIGIN_REPO) | cut -f5 -d/)
 
-        git clone git@github.com:redhat-appstudio/infra-deployments.git
+        git clone "git@github.com:${TARGET_GH_REPO}" infra-deployments
         cd infra-deployments
         git checkout -b $REPO_NAME
 
         $(params.SCRIPT)
 
         git commit -a -m "$REPO_NAME update" -m "Automatic update"
```
simonbaird, line 50: Maybe an idea for a future PR, but maybe we could link to the change that generated this PR in the commit message.

.tekton/push.yaml
```diff
@@ -85,6 +85,30 @@ spec:
...
+            value: hacbs-contract/ec-policies
+          - name: SCRIPT
+            value: |
+              export DIGEST='$(results.build-container.IMAGE_DIGEST)'
+              export BUNDLE='quay.io/redhat-appstudio/build-templates-bundle'
+              export TAG='$(params.revision)'
+              # 720h == 30 days
+              export EFFECTIVE_ON="$(yq --null-input 'now + "720h" | tz("UTC")')"
+
+              yq '."pipeline-bundles"[env(BUNDLE)] |=
```
lcarva, line 106: @simonbaird, since this is now using the `appstudio-utils` image, I could turn this into a call to ec-cli. Do you think it makes sense to introduce this logic to it?

I'm thinking we may want a more complex behavior when dealing with Task bundles since a repo has multiple "streams", e.g. `-1`, `-2`, `-3`, etc.

Something like this:

```
$ ec track bundles --file data/acceptable_tekton_bundles.yml \
    quay.io/redhat-appstudio/build-templates-bundle:ad4c34b71d468a991450b0091cd96765471ed82d \
    quay.io/redhat-appstudio/appstudio-tasks:ad4c34b71d468a991450b0091cd96765471ed82d-1 \
    quay.io/redhat-appstudio/appstudio-tasks:ad4c34b71d468a991450b0091cd96765471ed82d-2 \
    quay.io/redhat-appstudio/appstudio-tasks:ad4c34b71d468a991450b0091cd96765471ed82d-3 \
```

lcarva: Also, the digest of each bundle is not tracked anywhere in this push pipeline. ec-cli could resolve it.

lcarva: Done with latest update.

lcarva:
@zregvart and I were able to test this in a fork. The pipeline successfully created this PR: https://github.com/hacbs-contract/ec-policies/pull/64

zregvart (approved):
Looks good. @redhat-appstudio/build-team we tested on a cluster we setup and a forked [build-definitions](https://github.com/zregvart/build-definitions) repository and we got the expected [pull request](https://github.com/hacbs-contract/ec-policies/pull/64) created. Can someone have a look to see if it's all good from your end as well?

tasks/update-infra-deployments.yaml
```diff
@@ -23,7 +29,7 @@ spec:
           # 'appstudio-staging-ci' - private key for appstudio-staging-ci Github app
```
Michkov, line 29: There is an issue with sharedSecret, you will need to create custom one, since it requires deploy key must be unique for each project. Another (better) option is to use [github api contents](https://docs.github.com/en/rest/repos/contents#create-or-update-file-contents) , so then we would not need deploy key.

.tekton/push.yaml
```diff
@@ -85,6 +85,37 @@ spec:
...
+            value: |
+              TAG='$(params.revision)'
+
+              ec track bundle \
+                --input data/acceptable_tekton_bundles.yml \
+                --type task-bundles \
+                --replace \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-1" \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-2" \
+                --bundle "quay.io/redhat-appstudio/appstudio-tasks:${TAG}-3"
```
Michkov, line 110: nit/future: would be nice to output these images in https://github.com/redhat-appstudio/build-definitions/blob/main/.tekton/pull-request.yaml#L91 to avoid hardcoding number of bundles.

tasks/update-infra-deployments.yaml
Michkov: having it in progress

.tekton/push.yaml
lcarva: Yes, I was working on a patch to make the `build-bundle` task generate such file but I didn't want it to increase this PR too much. I can definitely tackle that shortly after this PR.

tasks/update-infra-deployments.yaml
lcarva: Oh interesting. I didn't know about that approach. Will that create a PR to be reviewed? Or just push a commit directly?

Michkov: So basically there are two secrets one is deploy key which is project related, used to create branch and push the commits (using ssh key) and second is secret for github app to be able to create Pull Request.
Now I am trying to rewrite code to use github api contents instead of standard git commands. Prototype done, trying to inject to the task.

Michkov (approved):

tasks/update-infra-deployments.yaml
lcarva: PR is rebased on the change that implements the new option.

PR #155, opened 2022-07-06
PLNSRVCE-452: parameterize hacbs dependency analyser image so it can be overwritten in PR tests

gabemontero:
Unlike with the infra-deployments/preview.sh pattern, the only way I see where we can override the image used for the dependency analyser produced by jvm-build-service PRs is to

a) make the image a parameter in the task
b) when we eventually have e2e tests for jvm-build-service, those tests look for the env vars already being set by the ci config in openshift/release and *somehow* specify an override for this parameter when a build is invoked

Not knowing yet how b) will be implemented, this PR does a), which pending a better idea, is the first step regardless.

@mmorhun @mbharatk @dwalluck FYI

@sbose78 - what do you think?  Right approach, or should I pivot to another idea you may have?

@Michkov @stuartwdouglas @psturc FYI when you get back from PTO

### DIFF
--- a/tasks/buildah.yaml
+++ b/tasks/buildah.yaml
@@ -22,6 +22,10 @@ spec:
     description: The location of the buildah builder image.
     name: BUILDER_IMAGE
     type: string
+  - default: quay.io/redhat-appstudio/hacbs-jvm-dependency-analyser:3f4824c3a29836a33a5bca3bcad386930b4197bc
+    description: The location of the HACBS dependency analyser image.
+    name: DEPENDENCY_ANALYSER_IMAGE
+    type: string
   - default: ./Dockerfile
     description: Path to the Dockerfile to build.
     name: DOCKERFILE
--- a/tasks/patches/add-sbom-and-push.yaml
+++ b/tasks/patches/add-sbom-and-push.yaml
@@ -33,7 +33,7 @@
   path: /spec/steps/-
   value:
     name: analyse-dependencies-java-sbom
-    image: quay.io/redhat-appstudio/hacbs-jvm-dependency-analyser:3f4824c3a29836a33a5bca3bcad386930b4197bc
+    image: $(params.DEPENDENCY_ANALYSER_IMAGE)
     script: |
       if [ -f /var/lib/containers/java ]; then
         /opt/jboss/container/java/run/run-java.sh path $(cat /workspace/container_path)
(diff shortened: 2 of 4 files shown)


gabemontero:
ok I found https://issues.redhat.com//browse/PLNSRVCE-141 and PR 106 from the application service repo and refreshed my memory on this one (one of the first items I reviewed when joining the team).

So while this flow is great for development flow, where someone has access to quay.io/MY_USERNAME in order to push updated bundles, and then tell application service to use that bundle via the configmap, at first blush I don't see where the destination is for push the updated bundle image to.  AFAIK e2e jobs cannot push images to quay.io/redhat-appstudio.  Nor is there a quay.io ORG available for CI jobs.

Pending a mistaken assumption there, it is time to start discussing this situation with 

/assign @sbose78 

I'll see if I can get some time with him tomorrow.

sbose78 (approved):

gabemontero:
OK @sbose78 and I had a lengthy discussion on this (it ended up being on a private slack thread so I cannot post a reference).

The net is that while he (per https://github.com/redhat-appstudio/build-definitions/pull/155#issuecomment-1180644086), myself (and @stuartwdouglas for that matter) think having to rebuild the bundle (and leverage the confirmap support) to change the dependency analyzer image for e2e tests is overkill, we are not comforatble with surfacing this parameter, as it invariably leaks this tuning knob to the end user, which is not what we want.

So, before we give up and go down the "rebuild the bundle path", I agreed to explore what defining an env var on the controllers, where we could specify image overrides for tasks, etc. defined in the bundles, with the rest of the build team, starting with standup tomorrow with @mmorhun, @psturc, @brunoapimentel , @mbharatk while Michal is out.  If we feel "good" about it, I'll pursue this week.  Otherwise, we'll wait until Michal comes back next week and let him make the final call.

gabemontero:
/hold

for now, not closing yet, in case this parameter helps the controller override based on env var setting path

sbose78:
+1 on the summary

gabemontero:
so yeah @mmorhun reminded me today that refs to tasks/pipelines/etc. defined in bundle is totally handled by tekton (duh!) so there is not intercept point available in the controller.

@brunoapimentel has been offering to help with jvm-build-serivce, and given he wrote the config map override stuff, updating jvm-build-service CI to leverage this and rebuilt the bundle sounds like a good work item for him to get his feet wet.

I'll tee him up with that today and we'll open an item off of https://issues.redhat.com/browse/PLNSRVCE-452 to track the work.

PR #163, opened 2022-07-18
sast-java-sec-check: remove newlines from error messages

dirgim:


tasks/sast-java-sec-check.yaml
```diff
@@ -62,7 +62,7 @@ spec:
 
         ERR_MSG="$(cat stderr.txt)"
         ERR_MSG="${ERR_MSG:-unknown}"
-        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --arg MSG "${ERR_MSG: 0: 3000}"
+        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --arg MSG "${ERR_MSG//[$'\t\r\n']: 0: 3000}"
```
Michkov, line 65: is it working?
```
$ ERR_MSG='my
> cool
> error
> '
$ echo "$ERR_MSG"
my
cool
error

$ echo "${ERR_MSG//[$'\t\r\n']: 0: 3000}"
my
cool
error

$

Comment (dirgim) [reply tasks/sast-java-sec-check.yaml:65]: Looks like I over-optimized, it doesn't seem to work when combined with substring :D

Comment (dirgim) [reply tasks/sast-java-sec-check.yaml]: Reverted to the original solution, thanks for double checking @Michkov

**Review by Michkov — APPROVED** [review]

File: tasks/sast-java-sec-check.yaml [hunk 2]
```diff
@@ -62,6 +62,7 @@ spec:
 
         ERR_MSG="$(cat stderr.txt)"
```
hongweiliu17, line 63: Thanks for making the fix, I tried and it looks changing this line to `ERR_MSG="$(echo -n $(cat stderr.txt))"` can remove the blank lines as well.

hongweiliu17 (approved):

tasks/sast-java-sec-check.yaml
dirgim: Ack, updated

PR #168, opened 2022-07-18
remove eval command from ec-cli refactor

joejstuart:


tasks/verify-enterprise-contract.yaml
```diff
@@ -140,4 +140,4 @@ spec:
 
   volumes:
     - name: shared
-      emptyDir: {}
+      emptyDir: {}
```
robnester-rh, line 143: newline needed

robnester-rh (approved):
one nit.

tasks/verify-enterprise-contract-v2.yaml
```diff
@@ -70,15 +70,14 @@ spec:
...
-        - "--input"
+        - validate
+        - image
+        - "--json-input"
         - "$(params.IMAGES)"
         - "--policy"
         - "$(params.POLICY_CONFIGURATION)"
         - "--public-key"
         - "$(params.PUBLIC_KEY)"
-        - "--rekor-url"
```
lcarva, line 80: The default value for this parameter in the ec-cli is `https://rekor.sigstore.dev/`. Removing this will force the public instance of rekor to be used, always.

How about we change the default value of REKOR_HOST in this task to an empty string?

PR #169, opened 2022-07-18
refactor to use updated validate command

joejstuart:


lcarva (approved):
I think this will do.

robnester-rh (approved):
LGTM

PR #171, opened 2022-07-19
verify-ec-v2: fix display and strict

lcarva:
This change ensures the policy report is always displayed, regardless of its status.

It also fixes the STRICT parameter so it can be disabled.

https://issues.redhat.com/browse/HACBS-881
https://issues.redhat.com/browse/HACBS-882


lcarva:
@redhat-appstudio/enterprise-contract ptal

zregvart (approved):
Neat :+1:

PR #173, opened 2022-07-20
Remove `cosign-verify-*` tasks

zregvart:
The problem is with the `cosign-verify-attestation` task that attempts
to store the attestation JSON into `VERIFY_ATTESTATION_JSON` Tekton
result. Given that today we're creating PipelineRun attestations, the
size of which is >4096 (around 20KB for a buildah build), we want to
stop using these. The `verify-enterprise-contract` (v1) task will run
`cosign` to validate and fetch the attestation and the attestation JSON
will be stored on a shared volume, so we don't have this issue there.


zregvart:
cc @redhat-appstudio/enterprise-contract, I'm also looking into removing this from the e2e tests, IIRC we assert this there

lcarva (approved):

lcarva:
This should only be merged after https://github.com/redhat-appstudio/e2e-tests/pull/146 is merged, I think.

zregvart:
https://github.com/redhat-appstudio/e2e-tests/pull/151 should be merged before we proceed to merge this

zregvart:
@Michkov I think [this change](https://github.com/redhat-appstudio/build-definitions/pull/173/commits/12936a403e7f46b49bb680c1294a94286ea74461) I included here would benefit from your review.

Michkov (approved):

PR #174, opened 2022-07-20
Dynamically determine built bundles

lcarva:
Changed hack/build-and-push.sh to record the Tekton bundles that it
produces. Then, use that output to call the "ec track" command.

This handles the case where the number of bundles changes unexpectedly.

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -1,5 +1,24 @@
 #!/bin/bash -e
 
+# Helper function to record the image reference from the output of
+# the "tkn bundle push" command into a given file.
+# Params:
+#   1. Image reference including the tag
+#   2. Output file
+function save_ref() {
+    local output
+    output="$(< /dev/stdin)"
+    echo "${output}"
+    local digest
+    digest="$(echo "${output}" | grep -Po '@\K(sha256:[a-f0-9]*)')"
+
+    local tagRef
+    tagRef="$1"
+    local refFile
+    refFile="$2"
+    echo "${tagRef}@${digest}" >> "${refFile}"
+}
+
 # local dev build script
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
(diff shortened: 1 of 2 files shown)


lcarva:
@Michkov, this addresses your comment from the initial implementation. I don't have a good way to perform an end-to-end test on these changes. I can attest that I have 
* Tested the changes to `hack/build-and-push.sh` by running the script locally with different environment variables.
* Tested the embedded script of the task `update-ec-policies` in `.tekton/push.yaml` by copying and pasting the commands on my terminal.

Michkov (approved):

hack/build-and-push.sh
```diff
@@ -34,6 +53,10 @@ HACBS_BUNDLE_IMG=quay.io/$MY_QUAY_USER/${TEST_REPO_NAME:-hacbs-templates-bundle}
 HACBS_BUNDLE_LATEST_IMG=quay.io/$MY_QUAY_USER/${TEST_REPO_NAME:-hacbs-templates-bundle}:${TEST_REPO_NAME:+hacbs-}latest
 HACBS_CORE_BUNDLE_IMG=quay.io/$MY_QUAY_USER/${TEST_REPO_NAME:-hacbs-core-service-templates-bundle}:${TEST_REPO_NAME:+hacbs-core-}latest
 
+OUTPUT_TASK_BUNDLE_LIST="${OUTPUT_TASK_BUNDLE_LIST-task-bundle-list}"
```
lcarva, line 56: @zregvart, taking https://github.com/hacbs-contract/ec-cli/pull/109 into account. I can change this to produce a single list of bundle images. Then, change the build-bundle task to make a single call to `ec track`. Thoughts?

zregvart: That would mean less bash, so :+1:

lcarva: Opted to keeping two lists but only calling ec-cli once with the combined list. This script produces other bundles that we do not care for (at least now). I don't want to accidentally start allowing them in the ec policy.

robnester-rh (approved):
LGTM

PR #177, opened 2022-07-27
PLNSRVCE-426 Expose Digest of Base Images

tkdchen:

### DIFF
--- a/tasks/patches/add-sbom-and-push.yaml
+++ b/tasks/patches/add-sbom-and-push.yaml
@@ -102,6 +102,9 @@
     name: inject-sbom-and-push
     resources: {}
     script: |
+      # Expose base image digests
+      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' | grep -v $(params.IMAGE) > $(results.BASE_IMAGES_DIGESTS.path)
+
       base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $(params.IMAGE))
       base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $(params.IMAGE))
       container=$(buildah from --pull-never $(params.IMAGE))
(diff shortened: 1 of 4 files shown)


tasks/patches/add-sbom-and-push.yaml
```diff
@@ -119,6 +119,24 @@
...
 
+- op: add
+  path: /spec/steps/-
+  value:
+    image: $(params.BUILDER_IMAGE)
+    name: expose-base-images-digests
+    resources: {}
+    script: |
+      base_images=()
+      for base_image in $(grep "^FROM " $1 | cut -d" " -f2); do
```
Michkov, line 130: whati si `$1` ?

I would just print all image which were used -> `buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' >$(results.BASE_IMAGES_DIGESTS.path)`. And I would put it at the beginning of `inject-sbom-and-push` step, to reduce number of steps (due to quota limits.)

tasks/patches/add-sbom-and-push.yaml
```diff
@@ -102,6 +102,9 @@
     name: inject-sbom-and-push
     resources: {}
     script: |
+      # Expose base image digests
+      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' >$(results.BASE_IMAGES_DIGESTS.path)
```
gabemontero, line 106:
```suggestion
      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' > $(results.BASE_IMAGES_DIGESTS.path)
```

simple readability suggestion

tasks/patches/add-sbom-and-push.yaml
```diff
@@ -102,6 +102,9 @@
     name: inject-sbom-and-push
     resources: {}
     script: |
+      # Expose base image digests
+      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' > $(results.BASE_IMAGES_DIGESTS.path)
```
Michkov, line 106: we should probably ignore temporary image which we created -> `| grep -v $(params.IMAGE)`

Michkov (approved):

PR #179, opened 2022-07-28
PLNSRVCE-539 move to workspace based caches

stuartwdouglas:
This requires https://github.com/redhat-appstudio/jvm-build-service/pull/105 to also be merged.

### DIFF
--- a/tasks/patches/add-sbom-and-push.yaml
+++ b/tasks/patches/add-sbom-and-push.yaml
@@ -158,9 +158,9 @@
   - image: quay.io/redhat-appstudio/hacbs-jvm-sidecar:70a63556e607e02fe6a19a3f13bb35cedc1f3043
     env:
       - name: QUARKUS_REST_CLIENT_CACHE_SERVICE_URL
-        value: "http://hacbs-jvm-cache.jvm-build-service.svc.cluster.local"
+        value: "http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local"
       - name: QUARKUS_S3_ENDPOINT_OVERRIDE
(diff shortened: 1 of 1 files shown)


tasks/patches/add-sbom-and-push.yaml
```diff
@@ -126,9 +126,9 @@
     imagePullPolicy: Always
     env:
       - name: QUARKUS_REST_CLIENT_CACHE_SERVICE_URL
-        value: "http://hacbs-jvm-cache.jvm-build-service.svc.cluster.local"
+        value: "http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local"
```
gabemontero, line 129: ah @psturc, re: our test discussion earlier today, yes, I *knew* wokspace based caches were on the horizon

I don't *think* this change will have a direct bearing on the component / hacbs tekton bundle test case we discussed, assuming there are no functional issues when we pivot to it, but certainly follow this PR/story, and as we discussed, we should have a common conversation with @stuartwdouglas at some point around making sure the build is accessing caches/proxies as desired.

gabemontero:
bump @stuartwdouglas - do we want to promote this from draft status, rebase, and on the path for merging?

gabemontero:
ok the jvm build service on the staging cluster as the per namespace launch cache pod logic in place now for a few days

in repo e2e's have been using the analogous setting at https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/base/maven-v0.2.yaml#L199-L200 in the dev flow based e2e's likewise for a bit now

let's merge this

@psturc as discussed previously I *believe* your switch of e2e's to component based flows can resume once this is merged and the requisite update to the bundles up a quay.io/redhat-appstudio occurs.

PR #181, opened 2022-07-29
Update ec-policies bundle update

zregvart:
With hacbs-contract/ec-cli#109 we removed the `--type` parameter from
the `track bundle` subcommand, and replaced this with detecting the
content of the bundle by examining the bundle image. This updates to
remove the `--type` parameter from the PipelineRun that is used to
update the bundles stored in hacbs-contract/ec-policies repository.


lcarva:
Can we do https://github.com/redhat-appstudio/build-definitions/pull/174 instead?

PR #182, opened 2022-07-29
Fix fetching policy configuration

zregvart:
When the EnterpriseContractPolicy custom resource has `nonBlocking` key
at all the result of piping an empty string to `jq` renders empty string
and the `config.json` ends up being empty file.
This makes sure that in that case the `config.json` doesn't end up being
empty by setting `non_blocking_data` to `[]` when fetching the
configuration from the cluster.

### DIFF
--- a/appstudio-utils/util-scripts/lib/fetch/cluster.sh
+++ b/appstudio-utils/util-scripts/lib/fetch/cluster.sh
@@ -55,6 +55,9 @@ save-policy-config() {
     local namespace=${namespace_arg#-n }
     echo "ERROR: unable to find the ec-policy EnterpriseContractPolicy in namespace ${namespace:-$(kubectl config view --minify -o jsonpath='{..namespace}')}" 1>&2
   else
+    if [[ -z "${non_blocking_data}" ]]; then
+      non_blocking_data='[]'
+    fi
     # Save the config data from the ECP
     echo "$non_blocking_data" | jq '{"config": {"policy": {"non_blocking_checks": . }}}' > "${config_file}"
     return 0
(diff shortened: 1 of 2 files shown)


appstudio-utils/util-scripts/lib/fetch/cluster.sh
```diff
@@ -55,6 +55,9 @@ save-policy-config() {
     local namespace=${namespace_arg#-n }
     echo "ERROR: unable to find the ec-policy EnterpriseContractPolicy in namespace ${namespace:-$(kubectl config view --minify -o jsonpath='{..namespace}')}" 1>&2
   else
+    if [[ -z "${non_blocking_data}" ]]; then
```
lcarva, line 58: I'm very disappointed jq doesn't have a way of handling this. I tried the alternative operator, `//`, but jq seems to just do nothing if the input is empty.

lcarva (approved):

appstudio-utils/util-scripts/lib/fetch/cluster.sh
lcarva: Ah! A bit wonky but works:

```
🐚 echo '["foo"]' | jq -s '.[0] // [] | {"config": {"policy": {"non_blocking_checks": . }}}'
echo '["foo"]' | jq -s '.[0] // [] | {"config": {"policy": {"non_blocking_checks": . }}}'
{
  "config": {
    "policy": {
      "non_blocking_checks": [
        "foo"
      ]
    }
  }
}

🐚 echo | jq -s '.[0] // [] | {"config": {"policy": {"non_blocking_checks": . }}}'
{
  "config": {
    "policy": {
      "non_blocking_checks": []
    }
  }
}

```

What you have is fine and probably more readable. I just couldn't let this go :grin:

simonbaird (approved):

PR #187, opened 2022-08-03
[HACBS-959]-skip clamav scan for attestation images

jsztuka:
There is currently issue with attestation images causing errors during **oc image extract** command:

`error: unable to extract layer sha256:570dc024d08eb5044a0c29d41d4f777931b910ce30080b583e4aabdbc0eb43ff from quay.io/redhat-appstudio/user-workload:sha256-96aad71edef114ed6486116a66d9e2e1f6e469ca0f6e156ec0b062b3c98922a0.att: archive/tar: invalid tar header`

This change will be rolled-back after the attestation images are fixed.

In case of attestation image both clamav and parsing is skipped (exit 0)

### DIFF
--- a/tasks/clamav-scan.yaml
+++ b/tasks/clamav-scan.yaml
@@ -32,6 +32,12 @@ spec:
         # strip new-line escape symbol from parameter and save it to variable
         imageanddigest=$(echo $imagewithouttag@'$(params.image-digest)')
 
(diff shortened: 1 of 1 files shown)


dirgim (approved):
LGTM

srivickynesh (approved):
LGTM

Michkov (approved):

PR #189, opened 2022-08-08
feat[HACBS-959]-catch oc image extract output

jsztuka:
This prevents running clamscan on images that were unable to extract.

### DIFF
--- a/tasks/clamav-scan.yaml
+++ b/tasks/clamav-scan.yaml
@@ -41,7 +41,12 @@ spec:
         [ -f /workspace/registry-auth/.dockerconfigjson ] && REGISTRY_ARGS="-a /workspace/registry-auth/.dockerconfigjson"
         mkdir content
         cd content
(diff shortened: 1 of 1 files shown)


Dannyb48 (approved):
LGTM. I'm glad we are capturing more information about the extraction!

tasks/clamav-scan.yaml
```diff
@@ -41,7 +41,12 @@ spec:
         [ -f /workspace/registry-auth/.dockerconfigjson ] && REGISTRY_ARGS="-a /workspace/registry-auth/.dockerconfigjson"
         mkdir content
```
srivickynesh, line 42:
```suggestion
        mkdir content
```
create only if it doesn't exists

jsztuka: This folder is initiating place where the image will be extracted, so it will be created each time as new(it does not exist before each run).

Michkov (approved):

PR #191, opened 2022-08-11
Create jvm build config and default secret for pull

Michkov:
- Ensure that jvm-build-config is created so cache service is deployed and can be used
- Ensure redhat-appstudio-registry-pull-secret is created so the build task will not be blocked in PaC workflow

### DIFF
--- a/tasks/init.yaml
+++ b/tasks/init.yaml
@@ -11,6 +11,9 @@ spec:
     - name: rebuild
       description: Rebuild the image if exists
       default: "false"
+    - name: jvm-build-service-init
+      description: Set jvm build service config
+      default: "false"
   results:
     - name: build
   steps:
@@ -26,4 +29,10 @@ spec:
           echo -n "true" > $(results.build.path)
         else
           echo -n "false" > $(results.build.path)
+        fi
+    - name: hacbs-init
+      image: registry.redhat.io/openshift4/ose-cli:v4.10
+      script: |
+        if [ "$(params.jvm-build-service-init)" == "true" ] && ! oc get configmap jvm-build-config &>/dev/null; then
+          oc create configmap jvm-build-config --from-literal enable-rebuilds="false"
         fi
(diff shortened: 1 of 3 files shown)


tasks/init.yaml
```diff
@@ -27,3 +30,13 @@ spec:
...
+    - name: hacbs-init
+      image: registry.redhat.io/openshift4/ose-cli:v4.10
+      script: |
+        if [ "$(params.jvm-build-service-init)" == "true" ] && ! oc get configmap jvm-build-config &>/dev/null; then
+          oc create configmap jvm-build-config --from-literal enable-rebuilds="false"
+        fi
+        # Create empty secret which is now hardcoded in PaC Pipelinerun template
+        if ! oc get secret redhat-appstudio-registry-pull-secret &>/dev/null; then
+          oc create secret generic redhat-appstudio-registry-pull-secret
+        fi
```
psturc, line 42: just tested it out on my cluster, works fine 👍

tasks/init.yaml
```diff
@@ -27,3 +30,13 @@ spec:
         else
           echo -n "false" > $(results.build.path)
         fi
+    - name: hacbs-init
+      image: registry.redhat.io/openshift4/ose-cli:v4.10
+      script: |
+        if [ "$(params.jvm-build-service-init)" == "true" ] && ! oc get configmap jvm-build-config &>/dev/null; then
+          oc create configmap jvm-build-config --from-literal enable-rebuilds="false"
```
psturc, line 37: this configmap is also created in cases when I run `docker-build` or `nodejs-builder` pipelines, because that `jvm-build-service-init` parameter is added to those pipelines as well:
Screenshot 2022-08-11 at 12 22 06
I guess we need it only in `java-builder` pipeline?

psturc (commented):
One question about the `jvm-build-service-init` param. Otherwise lgtm

tasks/init.yaml
Michkov: not for nodejs-builder, but for docker-build yes since there could be java build in dockerfile

Michkov: updated, so now the patch is applyed pipelines with label -> `pipelines.openshift.io/runtime == java` and `pipelines.openshift.io/runtime == generic`

psturc (approved):

stuartwdouglas (approved):

PR #193, opened 2022-08-15
fix[HACBS-951]-result size limit

jsztuka:
Tekton **result** capabilities are not sufficient for our use cases. We needed to avoid duplicities in gosec log.


dirgim (approved):
LGTM

srivickynesh (approved):
LGTM

PR #196, opened 2022-08-23
Automatically resolve short-names

Michkov:
Buildah started to require full image names which breaks current image
builds.

Setting short-name-mode to disabled -> Use all unqualified-search registries without prompting.


psturc:
👀

psturc:
I have verified that these changes work with short/full image names. However I couldn't reproduce the issue seen on a staging cluster - the build step worked with short image name on my cluster even with the old bundle

```bash
STEP-BUILD

STEP 1/1: FROM alpine
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
```

PR #200, opened 2022-08-26
feat: Add new enterprise-contract pipeline

lcarva:
This pipeline is meant to be used by the integration controller to
provider faster feedback to users.

https://issues.redhat.com/browse/HACBS-1005

### DIFF
--- a/pipelines/base/enterprise-contract.yaml
+++ b/pipelines/base/enterprise-contract.yaml
@@ -1,0 +1,64 @@
+# The purpose of this pipeline is to execute the verify-enterprise-contract-v2 task for container
+# images that are built but not automatically released in order to provide early feedback to users.
+# When auto release is enabled, the task is executed by the release pipeline immediately after the
+# container images are built, thus it is not necessary to execute the task via this pipeline.
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: enterprise-contract
+  labels:
+    "skip-hacbs-test": "true"
+spec:
+  params:
+    - name: SNAPSHOT
+      type: string
+      description: |
+        Spec section of an ApplicationSnapshot resource. Not all fields of the
+        resource are required. A minimal example:
+          {
+            "components": [
+              {
+                "containerImage": "quay.io/example/repo:latest"
+              }
+            ]
+          }
+        Each "containerImage" in the "components" array is validated.
+    - name: POLICY_CONFIGURATION
+      type: string
+      description: |
+        Name of the policy configuration (EnterpriseContractConfiguration
+        object) to use. `namespace/name` or `name` syntax supported. If
+        namespace is omitted the namespace where the task runs is used.
+      default: ec-policy
+    - name: SSL_CERT_DIR
+      type: string
+      description: |
+        Path to a directory containing SSL certs to be used when communicating
+        with external services. This is useful when using the integrated registry
+        and a local instance of Rekor on a development cluster which may use
+        certificates issued by a not-commonly trusted root CA. In such cases,
+        "/var/run/secrets/kubernetes.io/serviceaccount" is a good value. Multiple
+        paths can be provided by using the ":" separator.
+      default: ""
+  results:
+    - name: REPORT
+      value: "$(tasks.verify.results.REPORT)"
+  tasks:
+    - name: verify
+      params:
+        - name: POLICY_CONFIGURATION
+          value: "$(params.POLICY_CONFIGURATION)"
+        - name: IMAGES
+          value: "$(params.SNAPSHOT)"
+        - name: SSL_CERT_DIR
+          value: "$(params.SSL_CERT_DIR)"
+        # Tekton only creates results for successful tasks. Since the goal of this pipeline
+        # is not to gate an operation, but to provide feedback to users, do not fail if there
+        # are violations. This ensures users can view the full validation report.
+        - name: STRICT
+          value: "false"
+        # Public Key should come from the Policy Configuration
+        - name: PUBLIC_KEY
+          value: ""
+      taskRef:
+        name: verify-enterprise-contract-v2
--- a/pipelines/base/kustomization.yaml
+++ b/pipelines/base/kustomization.yaml
@@ -6,3 +6,4 @@ resources:
 - nodejs-builder
 - noop.yaml
 - prototype-build-compliance.yaml
+- enterprise-contract.yaml
--- a/pipelines/hacbs/e2e-ec.yaml
+++ b/pipelines/hacbs/e2e-ec.yaml
@@ -21,8 +21,16 @@ spec:
     - name: IMAGES
       type: string
       description: |
-        Spec section of an ApplicationSnapshot. The containerImage of each
-        component will be verified.
+        Spec section of an ApplicationSnapshot resource. Not all fields of the
+        resource are required. A minimal example:
+          {
+            "components": [
+              {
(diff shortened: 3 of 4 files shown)


lcarva:
@dirgim, ptal :pray:

pipelines/base/enterprise-contract.yaml
```diff
@@ -0,0 +1,57 @@
...
+  labels:
+    "skip-hacbs-test": "true"
+spec:
+  params:
+    - name: IMAGES
+      type: string
+      description: |
+        Spec section of an ApplicationSnapshot. The containerImage of each
+        component will be verified.
+    - name: POLICY_CONFIGURATION
```
dirgim, line 18: Is this something that the Integraton service will be able to figure out based on the Application's configuration or should we use the default in most cases?

pipelines/base/enterprise-contract.yaml
```diff
@@ -0,0 +1,57 @@
...
+  name: enterprise-contract
+  labels:
+    "skip-hacbs-test": "true"
+spec:
+  params:
+    - name: IMAGES
+      type: string
+      description: |
+        Spec section of an ApplicationSnapshot. The containerImage of each
+        component will be verified.
```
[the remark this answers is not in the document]
lcarva: Hmm. It is parsed as an ApplicationSnapshot Spec by the ec-cli. I don't know if we want to hide that from users as it may cause confusion (maybe?). How about adding an example to the description?

pipelines/base/enterprise-contract.yaml
lcarva: Good question. The Policy Configuration looks like this:
```
apiVersion: appstudio.redhat.com/v1alpha1
kind: EnterpriseContractPolicy
metadata:
  name: ec-policy
spec:
  description: Red Hat's enterprise requirements
  exceptions:
    nonBlocking: []
  publicKey: |
    -----BEGIN PUBLIC KEY-----
    ...
    -----END PUBLIC KEY-----
  rekorUrl: https://my.rekor.example.com
  sources:
  - git:
      repository: https://github.com/hacbs-contract/ec-policies.git
      revision: main
```

For the RedHat org, I believe all the fields will be the same. Someone, however, may want to add more policy repos under `.spec.sources` for additional validations.

For now, I think it's safe to assume that it'll always be the same ECP resource, but it's probably a good idea to plan for it to be configurable in the future (by whatever mechanism).

Wdyt?

pipelines/base/enterprise-contract.yaml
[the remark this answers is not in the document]
lcarva: Updated the description. Let me know what you think.

pipelines/base/enterprise-contract.yaml
```diff
@@ -0,0 +1,65 @@
...
+        # Tekton only creates results for successful tasks. Since the goal of this pipeline
+        # is not to gate an operation, but to provide feedback to users, do not fail if there
+        # are violations. This ensures users can view the full validation report.
+        - name: STRICT
+          value: "false"
+        # Public Key should come from the Policy Configuration
+        - name: PUBLIC_KEY
+          value: ""
+      taskRef:
+        # Does this get automatically linked to the bundle? If not how do other pipelines work?
```
dirgim, line 64: In the general case referring to the task by name shouldn't work if the task definition isn't installed first. Our approach for example integration test pipelines was to refer the task by bundle image since we can't assume that the correct task definition would be present.

@Michkov do you know if there's any replacement happening here or does the task definition need to be present on the cluster?

lcarva: My apologies, I just realized that this is a leftover comment. The replacement happens when running the script `hack/build-and-push.sh`. 

When I run this locally, this is what it turns into:
```
$ jq '.spec.tasks[].taskRef'
{
  "bundle": "quay.io/lucarval/appstudio-tasks:2022-08-30-102651-3",
  "name": "verify-enterprise-contract-v2"
}
```

So if you run the Pipeline from the pipeline bundle that is automatically generated for every commit to main in this repo, it should be all set. Here's how I run this Pipeline:
```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: enterprise-contract-
spec:
  params:
  - name: SNAPSHOT
    value: |
      {
        "components": [
          {"containerImage": "quay.io/lucarval/tekton-test:ede69244d7"}
        ]
      }
  - name: SSL_CERT_DIR
    value: /var/run/secrets/kubernetes.io/serviceaccount
  pipelineRef:
    bundle: quay.io/lucarval/build-templates-bundle:2022-08-30-102651@sha256:b0dee279a793ee1d4f3842ec33f7154c3bfaf21674cdbcc23eaed72547156d3b
    name: enterprise-contract
```

You don't have to know which image to use in taskRef, but you do need to know the bundle in which the pipeline definition lives. Does that make sense?

pipelines/base/enterprise-contract.yaml
dirgim: Ok, so if I understand correctly we should look for the ECP resource in the application's workspace and add that under this parameter?

pipelines/base/enterprise-contract.yaml
dirgim: Yes, that makes sense - the bundle with the pipeline is what needs to be set in the `IntegrationTestScenario` CR. Do you know if this will be a set image with a tag?

lcarva: The repo will always be `quay.io/redhat-appstudio/build-templates-bundle`. There isn't a "latest" tag. Images are tagged based on git commit ID. For instance, right now, the latest image is `quay.io/redhat-appstudio/build-templates-bundle:510fa6e99f1fa1f816c96354bbaf1ad155c6d9c3`

This is something that would have to be constantly updated. There's some gitops workflows triggered by this repo that creates pull requests in other repos when new images are available. Maybe it's something we can look into for this case, too.

Dannyb48: Could we tag it latest at the quay.io level using the quay api post bundle push and just keeping moving the latest tag on every bundle push?

lcarva: There has been push back with such approach as it makes things nondeterministic, which outweighs its convenience. I would say that a "latest" tag would be good for trying things out, but I would be skeptic about using it in any official workflow.

@Michkov, are you opposed to the idea of also tagging the image in `quay.io/redhat-appstudio/build-templates-bundle` with a latest tag?

Michkov: Maybe we can use different tag eg. `dev`

dirgim: Yeah, I second adding a tag, otherwise [DEVHAS-161](https://issues.redhat.com/browse/DEVHAS-161) might get complicated to implement.

lcarva: Ack. Let's aim for that then. I'd like to do that as a separate PR if possible.

dirgim: Agreed, we can start testing the current form by using the sha. Thanks for all the inputs everybody.

dirgim (approved):
LGTM

Dannyb48 (approved):
LGTM

lcarva:
I don't have access to merge this PR. Could a generous soul help me out? :pray:

Michkov:
@lcarva you should have access now

lcarva:
Thanks, @Michkov!

PR #201, opened 2022-08-31
Add the gosec-output.json contents the sast-go task logs

dirgim:


Dannyb48 (approved):
LGTM

Let's file other PRs to make sure we are doing the same thing for our other scanners

PR #207, opened 2022-09-13
Generate HACBS_ERROR_OUTPUT when correct repository_data.json is not …

hongweiliu17:
1. Generate HACBS_ERROR_OUTPUT when correct repository_data.json is not found in task deprecated-image-check
2. Try to fix 000 return_code by add `-k`, referring to https://stackoverflow.com/questions/21412642/i-am-getting-http-code-as-000-or-time-out-on-server-for-the-curl-in-shell-script

### DIFF
--- a/tasks/deprecated-image-check.yaml
+++ b/tasks/deprecated-image-check.yaml
@@ -29,7 +29,7 @@ spec:
       script: |
         #!/usr/bin/env bash
         echo "Querying Pyxis for $(params.IMAGE_REPOSITORY)..."
-        http_code=$(curl -s -o $(workspaces.sanity-ws.path)/repository_data.json -w '%{http_code}' "https://catalog.redhat.com/api/containers/v1/repositories/registry/$(params.IMAGE_REGISTRY)/repository/$(params.IMAGE_REPOSITORY)")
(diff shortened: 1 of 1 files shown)


tasks/deprecated-image-check.yaml
```diff
@@ -46,29 +46,34 @@ spec:
...
+          cat $(workspaces.sanity-ws.path)/stderr.txt
         fi
 
     # Format the test results and save them as the HACBS_TEST_OUTPUT Tekton result
     - name: test-format-result
       image: quay.io/redhat-appstudio/hacbs-test:latest
       script: |
-        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
-          '{result: "ERROR", timestamp: $date}')
+        ERR_MSG="$(cat $(workspaces.sanity-ws.path)/stderr.txt)"
```
dirgim, line 65: We could add the `echo -n` to this command to prevent newline issues, like we did [in the Java sast check](https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/sast-java-sec-check.yaml#L64).

hongweiliu17: Yes, thanks! Updated.

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #208, opened 2022-09-15
fix(HACBS-1072): Increase memory requests for SAST tasks

dirgim:


Michkov (approved):

Dannyb48 (approved):
LGTM

srivickynesh (approved):
LGTM

dirgim:
@skabashnyuk the SAST tasks tend to use more memory when they need to inspect more complex codebases. 
In this case we've increased the `requests` of the task definition to ensure that the pod that's executing the task gets scheduled on a node that has a minimum of requested RAM. 
This is a balancing act because setting too high of a request would mean that the SAST tasks would not be possible to execute on clusters with nodes that don't satisfy their request requirements, while having it too low would lead to using nodes that don't have enough memory to run the task, leading to it erroring out.

PR #210, opened 2022-09-26
Include image digest in task references

zregvart:
When the pipeline images are built `taskRef`'s `bundle` field is set with the reference to the bundle image containing the task identified by it's name. That image reference does not include the task bundle image digest.

This adds the task bundle image digest to the `bundle` field as well.

Fixes https://issues.redhat.com/browse/HACBS-1086

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -5,6 +5,8 @@
 # Params:
 #   1. Image reference including the tag
 #   2. Output file
+# Returns the piped in standard input followed by a line containing the image
+# reference including tag and digest
 function save_ref() {
     local output
     output="$(< /dev/stdin)"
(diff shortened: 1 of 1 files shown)


lcarva:
Something is up. When I built this locally, I see that the `enterprise-contract` Pipeline refers to the `-2` task Tekton bundle, it should be `-3`. Also, any pipeline that uses one of the tasks from `-1` task Tekton bundle, appear to have an empty bundle, e.g.: `{"bundle": "", "name": "buildah"}`

Off by one error?

zregvart:
> Off by one error?

Thanks! Fixed. I think.

lcarva:
> > Off by one error?
> 
> Thanks! Fixed. I think.

I still see the issue :(

zregvart:
@lcarva can you have another look, third time's the charm

lcarva (approved):
Works great!

Michkov (approved):

zregvart:
Thanks for the review!

PR #219, opened 2022-10-25
feat(HACBS-1213): add task sbom-json-check.yaml

hongweiliu17:

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -226,6 +226,22 @@ spec:
       workspaces:
       - name: registry-auth
         workspace: registry-auth
+    - name: sbom-json-check
+      when:
+      - input: $(params.hacbs)
+        operator: in
+        values: ["true"]
+      runAfter:
+        - build-container
+      taskRef:
+        name: sbom-json-check
+        version: "0.1"
+      params:
+      - name: IMAGE_URL
+        value: $(tasks.build-container.results.IMAGE_URL)
+      workspaces:
+      - name: workspace
+        workspace: workspace
 
   finally:
     - name: show-summary
--- a/task/sbom-json-check/0.1/sbom-json-check.yaml
+++ b/task/sbom-json-check/0.1/sbom-json-check.yaml
@@ -1,0 +1,66 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: sbom-json-check
+spec:
+  description: >-
+    Check the syntax of the sbom-cyclonedx.json file which should be found
+    in the /root/buildinfo/content_manifests/ directory
+  params:
+    - name: IMAGE_URL
+      description: the fully qualified image name to be verified
+      type: string
+  results:
+    - description: Test output
+      name: HACBS_TEST_OUTPUT
+  workspaces:
+    - name: workspace
+  steps:
+  - name: sbom-json-check
+    image: quay.io/redhat-appstudio/hacbs-test:latest
+    workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+    securityContext:
+      runAsUser: 0
+      capabilities:
+        add:
+          - SETFCAP
+    volumeMounts:
+      - mountPath: /shared
+        name: shared
+    script: |
+      cd /shared
+      touch fail_result.txt
+      if [ -f "sbom-cyclonedx.json" ]
+      then
+        result=$(echo -n $(cyclonedx-linux-x64 validate --input-file sbom-cyclonedx.json))
+        if [[ ! $result =~ "BOM validated successfully" ]]
+        then
+          echo "sbom-cyclonedx.json: $result" > fail_result.txt
+        fi
+      else
+        echo "cannot access 'sbom-cyclonedx.json': No such file or directory" > fail_result.txt
+      fi
+
+      FAIL_RESULTS="$(cat fail_result.txt)"
+      if [[ -z $FAIL_RESULTS ]]
+      then
+        HACBS_TEST_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
+          '{result: "SUCCESS",timestamp: $date, namespace: "default", successes: 1, failures: 0}')
+      else
+        echo "Fail to verify sbom-cyclonedx.json for image $(params.IMAGE_URL) with reason: $FAIL_RESULTS"
+        HACBS_ERROR_OUTPUT=$(jq -rc --arg date $(date +%s) --null-input \
+          '{result: "FAILURE", timestamp: $date, successes: 0, failures: 1}')
+      fi
+
+      echo "${HACBS_TEST_OUTPUT:-${HACBS_ERROR_OUTPUT}}" | tee $(results.HACBS_TEST_OUTPUT.path)
+  sidecars:
+  - image: $(params.IMAGE_URL)
+    name: copy-the-tested-file
+    volumeMounts:
+    - mountPath: /shared
+      name: shared
+    script: cp -R /root/buildinfo/content_manifests/* /shared/
+  volumes:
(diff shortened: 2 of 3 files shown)


task/sbom-json-check/0.1/sbom-json-check.yaml
```diff
@@ -0,0 +1,57 @@
...
+    workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+    securityContext:
+      runAsUser: 0
+      capabilities:
+        add:
+          - SETFCAP
+    script: |
+      cd /tmp
+      curl -OL https://github.com/CycloneDX/cyclonedx-cli/releases/download/latest/cyclonedx-linux-x64
+      dnf -y install libicu jq 2> fail_result.txt
```
Michkov, line 31: no way, to install anything in the task. How do you know that curl or dnf are part of the image?

hongweiliu17: Good question. I also feel they might be missing in image.

So I wanted to test by running `./hack/test-builds.sh hacbs`, but it doesn't work because https://github.com/redhat-appstudio/build-definitions/blob/main/hack/test-build/kustomization.yaml#L5 doesn't work for the new folder `task`. After I change the folder to `task`, new issue comes out.

@Michkov Do you know the right way to make `./hack/test-builds.sh hacbs` working?

Michkov: Updating scripts to work with new format.

I would create sidecar from params.IMAGE_URL which would have `cp -r /root/buildinfo/content_manifests/ /shared` where shared would be shared volume used by main script which would be run on image which is including cyclonedx

hongweiliu17: Wow.. Good idea! I thought to extract that folder from image file but feel it is not good...

hongweiliu17: @Michkov Updated. Thank you for the good idea!

task/sbom-json-check/0.1/sbom-json-check.yaml
```diff
@@ -0,0 +1,65 @@
...
+      fi
+
+      echo "${HACBS_TEST_OUTPUT:-${HACBS_ERROR_OUTPUT}}" | tee $(results.HACBS_TEST_OUTPUT.path)
+  sidecars:
+  - image: $(params.IMAGE_URL)
+    name: copy-the-tested-file
+    volumeMounts:
+    - mountPath: /shared
+      name: shared
+    script: cp -R /root/buildinfo/content_manifests/* /shared/
```
zregvart, line 62: It's probably a fringe scenario it could be that the built image doesn't have a `/usr/bin/cp` command in it, e.g. the file based catalog images or images built from scratch. Should we be concerned with this?

hongweiliu17:
@Michkov @zregvart @dirgim 
When I went through the code about sbom, I didn't find which code is using the output `cyclonedx-json` generated by https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L108-L110. Do you know it? Thanks.

I can see https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L164-L166 generate `sbom-cyclonedx.json` and then it is copied to `/root/buildinfo/content_manifests/` https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L206

Michkov:
@hongweiliu17 output of HACBS-303

dirgim:
@hongweiliu17 I was also confused by the `--output=cyclonedx-json` in https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L108-L110 but it turns out it's the [file format option](https://github.com/anchore/syft#output-formats) and not the output filename as I initially thought. 
The sbom output from the above commands is stored in `sbom-source.json` and `sbom-image.json` and the [information from those files](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L135-L162) is used to construct and save the final sbom files in [cyclonedx](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L165-L166) and [purl](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L172-L173) formats.

hongweiliu17:
> @hongweiliu17 output of HACBS-303
Thank you!

hongweiliu17:
> @hongweiliu17 I was also confused by the `--output=cyclonedx-json` in https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L108-L110 but it turns out it's the [file format option](https://github.com/anchore/syft#output-formats) and not the output filename as I initially thought. The sbom output from the above commands is stored in `sbom-source.json` and `sbom-image.json` and the [information from those files](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L135-L162) is used to construct and save the final sbom files in [cyclonedx](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L165-L166) and [purl](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L172-L173) formats.

Thank you!

task/sbom-json-check/0.1/sbom-json-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+          - SETFCAP
+    volumeMounts:
+      - mountPath: /shared
+        name: shared
+    script: |
+      cd /shared
+      touch fail_result.txt
+      if [ -f "sbom-cyclonedx.json" ]
+      then
+        result=$(echo -n $(cyclonedx-linux-x64 validate --fail-on-errors --input-file sbom-cyclonedx.json))
```
dirgim, line 36: Do we need the `--fail-on-errors` option set if we're checking the output of the validation in the `$result`?

hongweiliu17: Yes, we can remove it

hongweiliu17: Updated.

dirgim: Thanks!

dirgim (approved):
LGTM

hongweiliu17:
I rebase the PR to resolve the conflict, especially for the change https://github.com/redhat-appstudio/build-definitions/pull/219/files#diff-073dcc2fd850949d01892f5daa50e15d7ba496f9603e783cdd356ae6027e4622

jsztuka (approved):
lgtm

pipelines/template-build/template-build.yaml
```diff
@@ -226,7 +226,20 @@ spec:
       workspaces:
       - name: registry-auth
         workspace: registry-auth
-
+    - op: add
+      path: /spec/tasks/-
```
Michkov, line 230: it's task directly, not jsonpatch format

Michkov (changes requested):

pipelines/template-build/template-build.yaml
hongweiliu17: Sorry....., I just copy it, not read it carefully

task/sbom-json-check/0.1/sbom-json-check.yaml
Michkov: I suppose that the pipeline will not contain this task and many others.

Michkov:
Just note that currently the e2e-test is not testing HACBS, we are working on implementation in build-service.

Michkov (approved):

PR #228, opened 2022-11-03
feat[HACBS-1038]: Merge clair scan with conftest clair in build-definitions

Josh-Everett:


task/conftest-clair/0.1/conftest-clair.yaml
```diff
@@ -36,5 +36,32 @@ spec:
                    failures: (.failures // [])|map(.metadata.details.name) | unique
                  }' $(workspaces.conftest-ws.path)/clair-vulnerabilities.json || true)
         echo "${HACBS_TEST_OUTPUT:-${HACBS_ERROR_OUTPUT}}" | tee $(results.HACBS_TEST_OUTPUT.path)
-  workspaces:
-    - name: conftest-ws
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
```
Michkov, line 41: it should be merged on step level, there should not be multiple tasks in one task dir.

Michkov (changes requested):

task/conftest-clair/0.1/conftest-clair.yaml
```diff
@@ -36,5 +36,32 @@ spec:
...
+      script: |
+        #!/usr/bin/env bash
+
+        imagewithouttag=$(echo '$(params.image-url)' | sed "s/\(.*\):.*/\1/" | tr -d '\n')
+        # strip new-line escape symbol from parameter and save it to variable
+        imageanddigest=$(echo $imagewithouttag@'$(params.image-digest)')
+
+        [ -f /workspace/registry-auth/.dockerconfigjson ] && REGISTRY_ARGS="/workspace/registry-auth/"
+
+        clair-action report --image-ref=$imageanddigest --db-path=/tmp/matcher.db --format=quay --docker-config-dir=$REGISTRY_ARGS > $(workspaces.clair-ws.path)/clair-result.json
```
jsztuka, line 67: Since the workspace(**conftest-ws** and **clair-ws**) is not required anymore, we can drop the part mentioning it: 
`$(workspaces.clair-ws.path)/clair-result.json`
Its also important to mention that get-clair-scan step, must preceeds **conftest-clair**.
One and only namespace that will remain is **registry-auth** since it is needed to get the report form private repos.

jsztuka: I recommend using tekton native fs for sharing the output of `get-clair-scan`. You can use /tekton/home folder for this purpose.

task/conftest-clair/0.1/conftest-clair.yaml
```diff
@@ -36,5 +36,32 @@ spec:
...
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "appstudio, hacbs"
+  name: get-clair-scan
```
dirgim, line 48: For after the tasks are merged into a single task: I think it would probably be best to rename the resulting task to something like `clair-scan`

jsztuka:
One more thing that needs to be added.
Since we are using `[ -f /workspace/registry-auth/.dockerconfigjson ]` we need to reference this workspace for the task.
To achieve that, we just need this at the very bottom of the task:
```
  workspaces:
    - name: registry-auth
      optional: true
```

task/conftest-clair/0.1/conftest-clair.yaml
```diff
@@ -7,22 +7,39 @@ metadata:
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "appstudio, hacbs"
-  name: conftest-clair
```
hongweiliu17, line 10: From my point of view, naming the folder/file as the same name with task name `clair-scan` can make it easier to maintain/find the task.

task/conftest-clair/0.1/conftest-clair.yaml
```diff
@@ -34,7 +51,8 @@ spec:
                    namespace,
                    successes,
                    failures: (.failures // [])|map(.metadata.details.name) | unique
```
hongweiliu17, line 53: FYI, https://github.com/redhat-appstudio/build-definitions/blob/main/task/conftest-clair/0.1/conftest-clair.yaml is updated to use integer for failures field, so rebase is needed to load the change.

hongweiliu17:
> One more thing that needs to be added. Since we are using `[ -f /workspace/registry-auth/.dockerconfigjson ]` we need to reference this workspace for the task. To achieve that, we just need this at the very bottom of the task:
> 
> ```
>   workspaces:
>     - name: registry-auth
>       optional: true
> ```
> 
> Also I think that we can keep the empty README.md file in the directory.

We need to keep OWNERS file in this directory as well.

dirgim:
@Josh-Everett the PR is getting close, but there are conflicts that need to be resolved

jsztuka (approved):
lgtm

dirgim:
@Josh-Everett the PR looks good to me code-wise, can you squash the commits so it's a single one?

pipelines/template-build/template-build.yaml
```diff
@@ -188,31 +188,16 @@ spec:
       runAfter:
         - build-container
       taskRef:
-        name: get-clair-scan
+        name: clair-scan
```
Michkov, line 191: would be nice to rename the taskrun name (few lines above) get-clair-results -> clair-scan

Josh-Everett: ack , like this suggestion

dirgim (approved):
LGTM

PR #229, opened 2022-11-03
fix: build-and-push: mac os compatibility

psturc:


hack/build-and-push.sh
```diff
@@ -28,12 +34,6 @@ function save_ref() {
     echo "${tagRef}@${digest}"
 }
 
-if [[ $(uname) = Darwin ]]; then
-    CSPLIT_CMD="gcsplit"
-else
-    CSPLIT_CMD="csplit"
```
psturc, line 34: csplit is not used anymore

hack/build-and-push.sh
```diff
@@ -2,10 +2,16 @@
 
 set -e -o pipefail
 
+if [[ $(uname) = Darwin ]]; then
+    MKTEMP_CMD="gmktemp"
```
tkdchen, line 6: What is gmktemp?

PR #236, opened 2022-11-07
Pass task and pipeline bundles list via workspace

tkdchen:
Since the task and pipeline bundle are built and pushed per one respectively, the task-bundle-list and pipeline-bundle-list may exceed the limit size of tekton task result.

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -60,11 +60,6 @@ spec:
           params:
             - name: revision
               type: string
-          results:
-            - name: TASK_BUNDLES
-              description: A "\n" separated list of created task bundles
-            - name: PIPELINE_BUNDLES
-              description: A "\n" separated list of created pipeline bundles
           steps:
             - name: build-bundles
               image: quay.io/redhat-appstudio/appstudio-utils:{{ revision }}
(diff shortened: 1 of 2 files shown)


.tekton/push.yaml
```diff
@@ -124,8 +117,8 @@ spec:
           - name: SCRIPT
             value: |
               BUNDLES=(
-                $(tasks.build-bundles.results.TASK_BUNDLES)
-                $(tasks.build-bundles.results.PIPELINE_BUNDLES)
+                $(workspaces.source.path)/task-bundle-list
```
Michkov, line 120: should not be `cat $(workspaces.source.path)/task-bundle-list`?

.tekton/push.yaml
```diff
@@ -134,8 +127,9 @@ spec:
                 --input data/acceptable_tekton_bundles.yml \
                 --replace \
                 "${BUNDLES_PARAM[@]}"
-        taskRef:
-          name: update-infra-deployments
+        workspaces:
+          - name: source
+            workspace: workspace
```
Michkov, line 132: task update-infra-deployments does not have workspace defined

PR #240, opened 2022-11-10
Use quay API to check if the task exists in current version

Michkov:
Check if task with tag '${version}-{file-git-sha}' exists, if so then skip pushing of the task since it's already there.


hack/build-and-push.sh
```diff
@@ -81,17 +81,22 @@ do
...
-
-    # copy task to new tag pointing to commit where the file was changed lastly, so that image persists
-    # even when original tag is updated
+    repository=${TEST_REPO_NAME:-task-${task_name}}
+    tag=${TEST_REPO_NAME:+${task_name}-}${task_version}
+    task_bundle=quay.io/$MY_QUAY_USER/${repository}:${tag}
     task_file_git_sha=$(git log -n 1 --pretty=format:%H -- ${task_file})
-    skopeo copy "docker://${task_bundle}" "docker://${task_bundle}-${task_file_git_sha}"
-
+    digest=$(curl -s https://quay.io/api/v1/repository/$MY_QUAY_USER/$repository/tag/?specificTag=${tag}-${task_file_git_sha} | yq '.tags[0].manifest_digest')
```
lcarva, line 88: Interesting! I was not aware of this API. I would've gone with something like:
```
skopeo inspect --no-tags "docker://${task_bundle}" | yq .Digest
```

lcarva (approved):
It's beautiful!

PR #245, opened 2022-11-29
Fix the build pipeline

zregvart:
[Make sure hermetic-build result is produced](https://github.com/redhat-appstudio/build-definitions/commit/dd729340099ee5440bc05f3a95d6a9088bd9f465)

If the Tekton Task result is not produced any when expression using that result will evaluate to `false`. The `hermetic-build` result of the `git-clone` Task would be set to `true` if and only if the `cachi2.params` exists in the workspace, if the file does not exist it will not be set.
The current default is that a when expression evaluating to false skips the task where the when expression is defined, but also all tasks that `runAfter` it. See[1],`scope-when-expressions-to-task` feature flag[2] and TEP-0059[3].
Without the `cachi2.params` file in the workspace, the when condition on the `prefetch-dependencies` would evaluate to false, and the `build-container` step that has `runAfter: prefetch-dependencies` would be skipped as well, cascading to every other step in the pipeline other than `show-summary` and `sast-snyk-check` tasks.
With this change the `hermetic-build` result is always set to either `"true"` or `"false"`.

[1] https://tekton.dev/docs/pipelines/pipelines/#guard-task-execution-using-when-expressions
[2] https://tekton.dev/docs/operator/tektonpipeline/
[3] https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md

[Use resource dependencies](https://github.com/redhat-appstudio/build-definitions/commit/b238b09f1ec6258d11d3d3e30075ffaff9d5006e)

The parameters git and image URLs should be taken from the `clone-repository` and `build-container` tasks not from the user input.

This creates a resource dependency between the `sanity-inspect-image` task and the `build-container` task, so if the `build-container` task is skipped the `sanity-inspect-image` task will not be run. Not creating a resource dependency leads to `sanity-inspect-image` task running with an image reference pointing to an image that could not have been built in the pipeline when the `build-container` task is skipped.

Similarly for the `show-summary` task, as it now depends on `clone-repository` and `build-container` tasks running. And in addition to that the `show-summary` now includes git revision that was cloned, and since the image SHA-256 digest is passed the exact image that was built.

[1] https://tekton.dev/docs/pipelines/pipelines/#guard-task-execution-using-when-expressions

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -129,7 +129,7 @@ spec:
         version: "0.1"
       params:
       - name: IMAGE_URL
-        value: $(params.output-image)
+        value: "$(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)"
       workspaces:
       - name: workspace
         workspace: workspace
@@ -270,9 +270,9 @@ spec:
       - name: pipeline-run-name
         value: "$(context.pipelineRun.name)"
       - name: git-url
-        value: "$(params.git-url)"
+        value: "$(tasks.clone-repository.results.url)?rev=$(tasks.clone-repository.results.commit)"
       - name: image-url
-        value: $(params.output-image)
+        value: "$(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)"
   results:
     - name: IMAGE_URL
       value: "$(tasks.build-container.results.IMAGE_URL)"
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -193,6 +193,8 @@ spec:
       printf "%s" "${PARAM_URL}" > "$(results.url.path)"
       if [ -e cachi2.params ]; then
         printf "true" > "$(results.hermetic-build.path)"
+      else
+        printf "false" > "$(results.hermetic-build.path)"
       fi
   workspaces:
   - description: The git repo will be cloned onto the volume backing this Workspace.
--- a/task/sanity-inspect-image/0.1/sanity-inspect-image.yaml
+++ b/task/sanity-inspect-image/0.1/sanity-inspect-image.yaml
@@ -36,11 +36,18 @@ spec:
       BASE_IMAGE_INSPECT=base_image_inspect.json
       RAW_IMAGE_INSPECT=raw_image_inspect.json
 
-      echo "Inspecting manifest for source image $(params.IMAGE_URL)"
-      skopeo inspect --no-tags docker://$(params.IMAGE_URL) > $IMAGE_INSPECT
-      skopeo inspect --no-tags --raw docker://$(params.IMAGE_URL) > $RAW_IMAGE_INSPECT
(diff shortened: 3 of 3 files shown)


zregvart:
The `sanity-inspect-image` step fails with this change with:
```
Inspecting manifest for source image quay.io/redhat-appstudio/test-images:14d68911e6d941f997e8693bc6a5d92c@sha256:902ed9a2410e9b58dfe7b0259af41e13193310b2bd82b848a21f128fa701db52
time="2022-11-30T10:05:34Z" level=fatal msg="Error parsing image name \"docker://quay.io/redhat-appstudio/test-images:14d68911e6d941f997e8693bc6a5d92c@sha256:902ed9a2410e9b58dfe7b0259af41e13193310b2bd82b848a21f128fa701db52\": Docker references with both a tag and digest are currently not supported"
```

Michkov:
sanity-inspect-image failing on:
```
Inspecting manifest for source image quay.io/redhat-appstudio/test-images:14d68911e6d941f997e8693bc6a5d92c@sha256:902ed9a2410e9b58dfe7b0259af41e13193310b2bd82b848a21f128fa701db52
time="2022-11-30T10:05:34Z" level=fatal msg="Error parsing image name \"docker://quay.io/redhat-appstudio/test-images:14d68911e6d941f997e8693bc6a5d92c@sha256:902ed9a2410e9b58dfe7b0259af41e13193310b2bd82b848a21f128fa701db52\": Docker references with both a tag and digest are currently not supported"
```

zregvart:
@redhat-appstudio/build-team @redhat-appstudio/test-team please take a look

dirgim (approved):
LGTM

Michkov (approved):

pipelines/template-build/template-build.yaml
```diff
@@ -270,9 +270,9 @@ spec:
       - name: pipeline-run-name
         value: "$(context.pipelineRun.name)"
       - name: git-url
-        value: "$(params.git-url)"
+        value: "$(tasks.clone-repository.results.url)?rev=$(tasks.clone-repository.results.commit)"
       - name: image-url
-        value: $(params.output-image)
+        value: "$(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)"
```
lcarva, line 275: I thought the `show-summary` task was meant to always run (which is why it's in `finally`). There are cases where the pipeline skips the build on purpose because it determines that a build is not needed. Do we still want the `show-summary` task to run in such cases? With this change it will not.

zregvart: Correct, that is the side effect of this change. If the `build-container` step is skipped, `show-summary` will be skipped as well.

If we wish to have `show-summary` run always, this needs to be reverted. Obviously the intent with `finally` was to have it run always, that changes now because I wanted to have correct information presented. That's my reasoning at least, I could easily revert this.

WDYT @Michkov?

zregvart:
I'm going to merge this as it is blocking another PR. Easy to revert or touch up on in a followup PR.

pipelines/template-build/template-build.yaml
Michkov: Thinking of removing show-summary task completely

PR #255, opened 2022-12-16
Update cachi2 image to fix env vars used for build

chmeliik:
The currently used cachi2 image has a bug ([cachi2#45](https://github.com/containerbuildsystem/cachi2/pull/45)) where it will never generate any environment variables for the build. That means the GOMODCACHE environment variable won't be set => the build will not use the pre-fetched dependencies and will instead (try to) download them again.

Update to a fixed version.

### DIFF
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -19,7 +19,7 @@ spec:
     name: flags
     default: ""
   steps:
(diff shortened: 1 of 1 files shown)


chmeliik:
@brunoapimentel @Michkov @tkdchen PTAL

We had a pretty nasty bug that would have broken hermetic builds if buildah's --network=none actually worked

But we have integration tests now, so it won't happen again :smile:

Michkov (approved):

PR #258, opened 2022-12-19
Fully integrate Cachi2 into the build pipelines

brunoapimentel:
With the current integration, Cachi2 was limited to Go repositories that had the main module defined in the root of the repository. It was also dependent on the existence of a cachi2.params file that lived in the root of the repository.

Considering the addition of new package manager support to Cachi2 and the discouraging of having additional configuration files in the user's repository, this PR improves the current integration so that:

- The user can pass all configuration needed to Cachi2 using the `prefetch-input` pipeline parameter
- The preftech step no longer relies on the existence of a cachi2.params file
- The user can configure the network isolation using the `hermetic` parameter

### How to test this PR

```
tkn pipeline start docker-build \
-w name=workspace,volumeClaimTemplateFile=workspace-template.yaml \
-w name=registry-auth,secret=redhat-appstudio-staginguser-pull-secret \
-p revision="3c2735a005453932ececda0a17cfe13ef4a9c03d" \
-p git-url="https://github.com/chmeliik/dashdotdb"  \
-p output-image=quay.io/myuser/dashdotdb \
-p hermetic=true \
-p prefetch-input='[{"type":"pip"}]' \
--use-param-defaults
```

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -44,7 +44,11 @@ spec:
     type: string
   - default: "false"
     description: Determines if build will be executed without network access.
-    name: HERMETIC_BUILD
+    name: HERMETIC
+    type: string
+  - default: ""
+    description: In case it is not empty, the prefetched content should be made available to the build.
+    name: PREFETCH_INPUT
     type: string
   results:
   - description: Digest of the image just built
@@ -63,6 +67,10 @@ spec:
       value: oci
     - name: STORAGE_DRIVER
       value: vfs
+    - name: HERMETIC
+      value: $(params.HERMETIC)
+    - name: PREFETCH_INPUT
+      value: $(params.PREFETCH_INPUT)
   steps:
   - image: $(params.BUILDER_IMAGE)
     name: build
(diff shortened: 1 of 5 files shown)


task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -10,25 +10,26 @@ metadata:
...
+      fi
+
       cachi2 fetch-deps \
       --source=$(workspaces.source.path) \
       --output=$(workspaces.source.path)/cachi2/output \
-      --package '{
-        "type": "$(params.package-type)",
-        "path": "$(params.package-path)"
-      }'
+      $(params.input)
```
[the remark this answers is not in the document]
brunoapimentel: I turned the param into an env variable, following the same approach they use in the git-clone task.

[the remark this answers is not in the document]
chmeliik: Nice, could we do it consistently everywhere? (At least in the places that this PR touches, I might file a story for the rest of the pipeline)

brunoapimentel:
Currently, we can use the `prefetch-input` parameter to pass a JSON string, such as:

```sh
-p prefetch-input='[{"type":"pip"}]'
```

But if the string contains any spaces, the shell is still mangling it:
```sh
# this input
-p prefetch-input='[{"type": "pip"}]'
# will convert to
'[{"type": '
```
I still couldn't find a way to fix it, though.

Michkov (approved):

task/buildah/0.1/buildah.yaml
```diff
@@ -44,7 +44,11 @@ spec:
     type: string
   - default: "false"
     description: Determines if build will be executed without network access.
-    name: HERMETIC_BUILD
+    name: HERMETIC
+    type: string
+  - default: ""
+    description: Build dependencies that were previously prefetched.
+    name: PREFETCHED_CONTENT
```
chmeliik, line 51: Just a nitpick for the name and description - does this reflect what the param actually is?

chmeliik (approved):
LGTM, small nitpick

brunoapimentel:
I also tested this PR using a full StoneSoup install. This pipeline definition worked just fine: https://github.com/brunoapimentel/retrodep/blob/master/.tekton/retrodep-push.yaml#L27-L30

task/buildah/0.1/buildah.yaml
brunoapimentel: I was also not satisfied with this. At first I wanted to somehow make this into a boolean value, but I couldn't find a reasonable way to do it. I tried to make this parameter dependent on a result emitted from the prefetch-dependencies task, but then the buildah task was be skipped altogether every time the prefetch-dependencies task was skipped :face_exhaling: 

I changed the parameter to actually reflect what it is, although I think it is slightly confusing when you look at the `buildah` task apart from the whole pipeline.

chmeliik: > I tried to make this parameter dependent on a result emitted from the prefetch-dependencies task, but then the buildah task was be skipped altogether every time the prefetch-dependencies task was skipped :face_exhaling:

Oh, right :disappointed: 

The new name seems fine to me

PR #265, opened 2023-01-05
feat(HACBS-1436): patch fbc-label-check to fbc-builder pipeline

hongweiliu17:


hongweiliu17:
@jsztuka has included the fbc-label-check in https://github.com/redhat-appstudio/build-definitions/pull/254 and merged it It is not needed any more.

PR #267, opened 2023-01-09
Update version of images from redhat repositories

Michkov:
Trying to avoid issues with pulling from registry.redhat.io


Michkov:
Failed on `Pipeline build-templates-e2e/docker-build can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "sanity-label-check": yaml: control characters are not allowed` but the task is not touched in this PR and it's possible to download and apply.

psturc (approved):

PR #270, opened 2023-01-09
fix: overriding pipeline bundle used in e2e

psturc:
Since we've started using [build pipeline selector](https://github.com/redhat-appstudio/build-service/pull/84) instead of build-pipelines configmap, we need to update the logic in the pull request pipeline to install the custom build pipeline selector to `build-templates-e2e` namespace, where PR e2e tests are running.

Thanks to this we will be able to test changes to this repo per PR (again).

### DIFF
--- a/hack/util-install-bundle.sh
+++ b/hack/util-install-bundle.sh
@@ -6,11 +6,24 @@ NAMESPACE=${2:-$(oc project -q)}
 if [  -z "$BUNDLE" ]; then 
     echo "No Bundle Name"
     exit 1 
-fi   
+fi
 
-oc create configmap build-pipelines-defaults --from-literal=default_build_bundle=$BUNDLE -o yaml --dry-run=client | oc -n "$NAMESPACE" apply -f-
+cat << EOF | oc -n "$NAMESPACE" apply -f-
+apiVersion: appstudio.redhat.com/v1alpha1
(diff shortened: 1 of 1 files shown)


MartinBasti:
Could be a commit msg be more descriptive and explain what and why please?

psturc:
@MartinBasti it needs more work. I will add a description to the PR later

hack/util-install-bundle.sh
```diff
@@ -6,11 +6,24 @@ NAMESPACE=${2:-$(oc project -q)}
...
+      when:
+        dockerfile: true
+EOF
 
-echo "Default Pipelines configured to come from the namespace 'build-templates':"
-oc get cm build-pipelines-defaults -n build-templates -o jsonpath='{.data}{"\n"}'
-echo "Override Pipelines configured to come from the namespace '$NAMESPACE':"
-oc get cm build-pipelines-defaults -n "$NAMESPACE" -o jsonpath='{.data}{"\n"}'
+echo "Pipeline selectors configured to come from the namespace 'build-service':"
+oc get buildpipelineselector build-pipeline-selector -n build-service -o yaml | yq '.spec.selectors'
```
Michkov, line 27: this is not working since you do not have permission, but I think that this information is not important.

Michkov (approved):

PR #273, opened 2023-01-12
Run the git-clone task as root

chmeliik:
Currently, other tasks in the pipeline assume that the source was cloned as root. Make sure the step will indeed run as root even if the default USER in the git-init image is non-root.

Otherwise, for example the prefetch-dependencies task (which runs as root) may fail to fetch git tags when processing Go modules:

    cmdline: git fetch -v --force --tags origin
    stderr: 'fatal: detected dubious ownership in repository at '/workspace/source''

### DIFF
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -128,6 +128,8 @@ spec:
     image: $(params.gitInitImage)
     name: clone
     resources: {}
+    securityContext:
+      runAsUser: 0
     script: |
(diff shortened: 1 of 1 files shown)


mkosiarc (approved):

chmeliik:
Will try to figure out why the tests failed

chmeliik:
> Will try to figure out why the tests failed

Ah, fixed on main, thanks @Michkov for the tip

Michkov (approved):

brunoapimentel (approved):

PR #278, opened 2023-01-19
fix(STONE-355): clair-scan must not fail the pipelinerun

jsztuka:
As Title clarifies, preventing both clair steps from failing.


Michkov (approved):

MartinBasti (approved):

dirgim (approved):
+1

PR #283, opened 2023-01-26
fix(HACBS-1684): Fix sast_snyk_check results creation

MartinBasti:
Fixing typo in filename "sast_snyk_check_out.json" which prevented to generate result.

Also create ERROR result before ito cover case when creating standard result fails

Signed-off-by: Martin Basti <mbasti@redhat.com>


jsztuka (approved):
lgtm

Josh-Everett (approved):
nice, +1

PR #288, opened 2023-01-30
use DOCKER_CONFIG env for auth files and push sbom via cosign

Michkov:
DOCKER_CONFIG env provides way how to globally set container registries auth - podman, buildah, docker, cosign are using it. For cosign it's only way how to configure auth. The auth file is composed in configure-build step which is taking secrets from serviceaccount and merge them with shared secret. So when shared secret will be removed we can just remove the task and use only secrets provided by serviceaccount (handled by tekton).

Instead of manual creation of sbom image use cosign to upload sbom - in future it could be function of tekton-chains. To receive the sbom `cosign download sbom $IMAGE` eg. `cosign download sbom quay.io/mkovarik/single-container-app:62c06bf`

### DIFF
--- a/pipelines/docker-build/patch.yaml
+++ b/pipelines/docker-build/patch.yaml
@@ -22,10 +22,8 @@
     value: $(params.dockerfile)
   - name: CONTEXT
     value: $(params.path-context)
-  - name: BUILD_EXTRA_ARGS
-    value: "$(tasks.appstudio-configure-build.results.buildah-auth-param)"
-  - name: PUSH_EXTRA_ARGS
-    value: "$(tasks.appstudio-configure-build.results.buildah-auth-param)"
+  - name: DOCKER_AUTH
+    value: "$(tasks.appstudio-configure-build.results.docker-auth)"
   - name: HERMETIC
     value: "$(params.hermetic)"
   - name: PREFETCH_INPUT
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -258,16 +258,6 @@ spec:
       cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
       echo -n "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
 
-      # Create and push sbom image
-      sbom_container=$(buildah from scratch)
-      sbom_container_mountpath=$(buildah mount $sbom_container)
-      cp sbom-cyclonedx.json sbom-purl.json $sbom_container_mountpath
-      buildah umount $sbom_container
-      buildah commit working-container $(params.IMAGE).sbom
-      buildah push \
-        --tls-verify=$(params.TLSVERIFY) \
-        $(params.IMAGE).sbom
-
     securityContext:
       runAsUser: 0
       capabilities:
@@ -276,6 +266,18 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+    workingDir: $(workspaces.source.path)
+
+  - name: upload-sbom
+    image: quay.io/redhat-appstudio/cosign:v1.13.1
+    args:
+      - attach
+      - sbom
+      - --sbom
+      - sbom-cyclonedx.json
+      - --type
+      - cyclonedx
+      - $(params.IMAGE)
     workingDir: $(workspaces.source.path)
 
   volumes:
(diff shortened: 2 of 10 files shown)


task/s2i-java/0.1/s2i-java.yaml
```diff
@@ -237,15 +237,6 @@ spec:
       cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
       echo -n "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
 
-      # Create and push sbom image
-      sbom_container=$(buildah from scratch)
-      sbom_container_mountpath=$(buildah mount $sbom_container)
-      cp sbom-cyclonedx.json sbom-purl.json $sbom_container_mountpath
```
chmeliik, line 242: Does this mean we are getting rid of the sbom-purl.json format?

Michkov: that stays in the container image, but it's not pushed to the registry as separate image.

chmeliik: ah, got it

task/buildah/0.1/buildah.yaml
```diff
@@ -281,6 +268,18 @@ spec:
       name: varlibcontainers
     workingDir: $(workspaces.source.path)
 
+  - name: upload-sbom
+    image: ghcr.io/sigstore/cosign/cosign:03468a1498876ccfb72e06f63fdb4f1c562dbca1
```
mmalina, line 272: Where does the image spec come from? I found this: https://github.com/sigstore/cosign#dockerfile

And they use this in an example: `gcr.io/projectsigstore/cosign:v1.13.0`

In any case, I know it's not commonly done, but I think it would be nice to have a comment to say where that image comes from.

pipelines/docker-build/patch.yaml
```diff
@@ -22,10 +22,8 @@
     value: $(params.dockerfile)
   - name: CONTEXT
     value: $(params.path-context)
-  - name: BUILD_EXTRA_ARGS
-    value: "$(tasks.appstudio-configure-build.results.buildah-auth-param)"
-  - name: PUSH_EXTRA_ARGS
-    value: "$(tasks.appstudio-configure-build.results.buildah-auth-param)"
+  - name: DOCKER_AUTH
```
lcarva, line 25: I wonder if passing the auth info as a parameter will cause it to be recorded in plain text in the PipelineRun/TaskRun resource which may be added as plain text to the image attestation.

Michkov: this is path to the folder containing config.yaml, so it does not show the secret in the plaintext.

task/buildah/0.1/buildah.yaml
Michkov: I took it from https://github.com/sigstore/cosign/pkgs/container/cosign%2Fcosign

pipelines/docker-build/patch.yaml
lcarva: beautiful!

lcarva (approved):

chmeliik (approved):
LGTM :+1:

lcarva:
@Michkov, I was hitting authentication issues when the newly added tkn-bundle task. 

I saw the following error in the logs:
```
warning: unsuccessful cred copy: ".docker" from "/tekton/creds" to "/home/nonroot": unable to create destination directory: mkdir /home/nonroot: permission denied
```

You may see this in OpenShift if the Task/ServiceAccount doesn't have the right permissions. Presumably, this is why we don't see it on the buildah task.

There's a Red Hat solutions article about this: https://access.redhat.com/solutions/6956010

We faced this issue a while back when the release team tried to run the verify-ec task in a more lockdown environment with a custom ServiceAccount.

The fix from the solution is pretty simple. We still use it today, see [here](https://github.com/hacbs-contract/ec-cli/blob/b06ee374f8526fa8ea7039b8d650388983994a58/task/0.1/verify-enterprise-contract.yaml#L91-L94).

With that fix in place, all the secrets linked to the ServiceAccount (for mount) are automatically mounted on the pod for my task. Most container tooling is aware of this behavior and will find the secrets without requiring any additional configuration or environment variable.

Mentioning it here in case that simplifies the approach from this PR.

Michkov:
@lcarva that's different issue, please link the secret without `--for=pull`, that's wrong and I need to fix https://github.com/redhat-appstudio/build-definitions/blob/main/hack/test-build.sh#L42 (tomorrow I promise).

PR #292, opened 2023-01-31
STONEBLD-682: hermetic build with unshare

Michkov:
- Enable network-isolation on new linux namespace level.
- Prefetch images used by dockerfile
- Inject sourcing of cachi2.env

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -96,8 +96,8 @@ spec:
         echo "Cannot find Dockerfile $(params.DOCKERFILE)"
         exit 1
       fi
-      if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] && grep -q '^RUN \(./\)\?mvn' "$dockerfile_path"; then
-        sed -i -e "s|^RUN \(\(./\)\?mvn\(.*\)\)|RUN echo \"<settings><mirrors><mirror><id>mirror.default</id><url>http://$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR/v1/cache/default/0/</url><mirrorOf>*</mirrorOf></mirror></mirrors></settings>\" > /tmp/settings.yaml; \1 -s /tmp/settings.yaml|g" "$dockerfile_path"
+      if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] && grep -q '^\s*RUN \(./\)\?mvn' "$dockerfile_path"; then
+        sed -i -e "s|^\s*RUN \(\(./\)\?mvn\(.*\)\)|RUN echo \"<settings><mirrors><mirror><id>mirror.default</id><url>http://$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR/v1/cache/default/0/</url><mirrorOf>*</mirrorOf></mirror></mirrors></settings>\" > /tmp/settings.yaml; \1 -s /tmp/settings.yaml|g" "$dockerfile_path"
         touch /var/lib/containers/java
       fi
 
@@ -110,7 +110,7 @@ spec:
       if [ "${HERMETIC}" == "true" ]; then
         BUILDAH_ARGS="--pull=never"
         UNSHARE_ARGS="--net"
-        for image in $(grep -i '^FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}'); do
+        for image in $(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}'); do
           unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
         done
         echo "Build will be executed with network isolation"
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -117,9 +117,9 @@ spec:
       fi
 
       if [ -n "${PREFETCH_INPUT}" ]; then
-        VOLUME_MOUNTS="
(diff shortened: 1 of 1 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -107,29 +107,36 @@ spec:
       sed -i 's/^short-name-mode = .*/short-name-mode = "disabled"/' /etc/containers/registries.conf
 
       if [ "${HERMETIC}" == "true" ]; then
-        NETWORK="--network none"
+        BUILDAH_ARGS="--pull=never"
+        UNSHARE_ARGS="--net"
+        for image in $(grep -i '^FROM' "$dockerfile_path" | awk '{print $2}'); do
```
chmeliik, line 112: Not too long ago we had a bug filed in dockerfile-parse for `FROM --platform=...`: https://github.com/containerbuildsystem/dockerfile-parse/issues/124

Should we try to deal with those cases?

task/buildah/0.1/buildah.yaml
```diff
@@ -107,29 +107,36 @@ spec:
...
         echo "Build will be executed with network isolation"
       fi
 
-      if [ ! -z "${PREFETCH_INPUT}" ]; then
+      if [ -n "${PREFETCH_INPUT}" ]; then
         VOLUME_MOUNTS="
           --volume $(realpath ./cachi2/output):/cachi2/output:U \
           --volume $(realpath ./cachi2/cachi2.env):/cachi2/cachi2.env:U"
-
+        sed -i 's|^\s*run |RUN source /cachi2/cachi2.env \&\& \\\n    |i' "$dockerfile_path"
```
chmeliik, line 122: Maybe the leading whitespace in `^\s*run` doesn't need to be there

Technically buildah will consider this Dockerfile valid, but the rest of the code in this task does not accept leading whitespace

```dockerfile
FROM alpine:latest

    RUN echo hello there
```

Alternatively, other commands in this task would have have to be updated to accept leading whitespace

task/buildah/0.1/buildah.yaml
```diff
@@ -107,29 +107,36 @@ spec:
...
         echo "Prefetched content will be made available"
       fi
 
-      buildah bud \
+      # Creating new namespace to run buildah
+      echo 'root:1:4294967294' | tee -a /etc/subuid >> /etc/subgid
+      unshare -Ufp $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah bud \
```
chmeliik, line 128: What are the numbers based on? I went through the discussion in STONEBLD-660, but it's still not clear to me

4294967294 is some magic number?

65536 is the "number of UIDs and GIDs available"? How do we know how many are available?

chmeliik: @nalind you might have some more context for the magic number?

chmeliik: Ah nevermind, it's the maximum possible number isn't it
```python
In [3]: 2**32 - 2
Out[3]: 4294967294
```

nalind: UIDs and GIDs are currently 32-bit unsigned numbers.  42949672945, as you've noted, is more or less what you see when you cast -1 to an unsigned 32-bit number, and the kernel uses that as a magic "invalid UID/GID" value, so the kernel won't allow it to be included in mappings.
65536 is what we usually allocate in /etc/subuid and /etc/subgid when a user account is created with `useradd`, so that's an educated guess.  If we're already UID 0 and we don't actually want to change the current UID/GID mappings, a more robust implementation could look at /proc/self/uid_map and /proc/self/gid_map to see which UIDs and GIDs are mapped in the parent user namespace, and use the values found there to map the ranges it finds there onto themselves.

chmeliik: Ack, thanks for the explanation :+1:

task/buildah/0.1/buildah.yaml
Michkov: added

chmeliik (approved):
LGTM :+1: 

@psturc it would be good to remove the explicit `source /cachi2/cachi2.env` from the Dockerfile used for the integration test to test that the injecting works

Michkov:
hmm, I removed sys-chroot but it does not work properly right now with the injection of volumes

Michkov:
```
Error: building at STEP "RUN source /cachi2/cachi2.env &&             pip3 install --no-cache-dir --use-pep517 . &&             pip3 install --no-cache-dir gunicorn": resolving mountpoints for container "717413c23383f3ae1b1cdec7a5136f5859d347b0ab9693b815416972db9e2e69": failed to chown recursively host path: lchown /workspace/source/cachi2/cachi2.env: operation not permitted
```
`-m` flag does not help. It's strange I had successful run but I cannot reproduce now...

chmeliik (commented):
LGTM

minor nitpick for the last commit message:
> Do not add cachi2 to container

This is about making sure that the ./cachi2 directory won't be copied when the user does `COPY . .` in the Dockerfile, right? Could that be explained in a bit more detail in the commit message?

chmeliik (approved):

task/buildah/0.1/buildah.yaml
```diff
@@ -70,7 +70,7 @@ spec:
     - name: DOCKER_CONFIG
       value: $(params.DOCKER_AUTH)
   steps:
-  - image: $(params.BUILDER_IMAGE)
+  - image: quay.io/redhat-appstudio/buildah:v1.28
```
brunoapimentel, line 73: Should we change the default image in the BUILDER_IMAGE parameter instead of here? It seems like the `prototype-build-compliance` pipeline is still setting this parameter.

In case we want to force the use of buildah:v1.28, we should get rid of the BUILDER_IMAGE param altogether.

Michkov: `prototype-build-compliance` can be ignored. The BUILDER_IMAGE is used in next steps of this task since new image requires additional "SYS_CHROOT" capabality or running in `unshare`, so I kept old image for other tasks, since we need the 1.28 only for build step.

brunoapimentel (approved):

PR #295, opened 2023-02-02
Fixes to the tekton-bundle-builder pipeline

lcarva:
This commit includes various small fixes that are essential for executing the tekton-bundle-builder pipeline:

1. Pass the pipeline path-context parameter to the CONTEXT parameter of the tkn-bundle task.
2. Pass the pipeline output-image parameter to the IMAGE parameter of the tkn-bundle task.
3. Remove certain tasks from the pipeline that assume the built image is binary. Tekton bundles are not.
4. Add new HOME parameter to the tkn-bundle task to avoid permission issues when mounting linked secrets. For details, see: https://access.redhat.com/solutions/6956010
5. Change the IMAGE_URL result from the tkn-bundle task to not include the digest. Other tasks, assume this value does not have a digest, and blindly add the IMAGE_DIGEST to it which creates an image reference that contains the digest twice.
6. Change the IMAGE_URL result from the tkn-bundle task to include the tag of the image. This matches the current behavior of the other build tasks.
7. Change the IMAGE_URL and IMAGE_DIGEST results from the tkn-bundle task to no longer include an unexpected trailing new-line. Update tests to catch this in the future.

Required for https://issues.redhat.com/browse/HACBS-1557

### DIFF
--- a/pipelines/tekton-bundle-builder/kustomization.yaml
+++ b/pipelines/tekton-bundle-builder/kustomization.yaml
@@ -15,6 +15,18 @@ patches:
       value:
         name: tkn-bundle
         version: "0.1"
+    - op: add
+      path: /spec/tasks/4/params
+      value:
+      - name: IMAGE
+        value: $(params.output-image)
+      - name: CONTEXT
+        value: $(params.path-context)
+    # Remove tasks that assume a binary image
+    - op: remove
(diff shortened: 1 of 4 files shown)


zregvart (approved):
Thanks!

PR #299, opened 2023-02-06
STONEBLD-31: restrict hermetic build for FBC

Michkov:
Do not allow download external definitions in Dockerfile

### DIFF
--- a/pipelines/fbc-builder/patch.yaml
+++ b/pipelines/fbc-builder/patch.yaml
@@ -24,6 +24,8 @@
     value: $(params.path-context)
   - name: DOCKER_AUTH
     value: "$(tasks.init.results.container-registry-secret)"
+  - name: HERMETIC
+    value: "true"
 - op: add
   path: /spec/tasks/-
   value:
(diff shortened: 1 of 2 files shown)


tkdchen (approved):

pipelines/fbc-builder/patch.yaml
```diff
@@ -24,6 +24,8 @@
     value: $(params.path-context)
   - name: DOCKER_AUTH
     value: "$(tasks.appstudio-configure-build.results.docker-auth)"
+  - name: HERMETIC
+    value: "true"
```
brunoapimentel, line 28: Just a heads-up that this will make the prefetch task run and [print](https://github.com/redhat-appstudio/build-definitions/blob/main/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L27): 

```
Build will be executed with network isolation, but no content was configured to be prefetched.
This will likely result in a build failure.
```

Maybe we should get rid of the last sentence to avoid user confusion.

PR #301, opened 2023-02-07
fix(HACBS-1746, sbom-json-check): remove unsafe usage of sidecars

MartinBasti:
Running image as sidecar is insecure, an newly built image can contain malicious cp command. Also, image may not contain cp command at all.

Instead of mounting image content via sidecar, extract it directly via oc image extract

Signed-off-by: Martin Basti <mbasti@redhat.com>


dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #302, opened 2023-02-07
Cachi2: bump image to update docs links, call inject-files subcommand

chmeliik:
(see commit messages)


chmeliik:
Will wait with this until the e2e test for Go is done

chmeliik:
PTAL, 2 minor fixes for cachi2 Pip

* currently, docs links in error message are broken because we've updated the docs upstream (the new image has updated links)
* and call a missing subcommand that enables pip offline install to work even for dependencies like `foo @ https://github.com/...archive/1.0.0.tar.gz

Michkov (approved):
e2e-test passing, enough for me :-)

brunoapimentel (approved):

PR #306, opened 2023-02-09
STONEBLD-10: set -e to task script properly

tkdchen:

### DIFF
--- a/task/configure-build/0.1/configure-build.yaml
+++ b/task/configure-build/0.1/configure-build.yaml
@@ -33,6 +33,7 @@ spec:
           mountPath: /secret/default-push-secret
       script: |
         #!/usr/bin/env bash
+        set -e
         echo "App Studio Configure Build"
         mkdir /workspace/source/.docker
         DEST=/workspace/source/.docker/config.json
--- a/task/summary/0.1/summary.yaml
+++ b/task/summary/0.1/summary.yaml
@@ -22,6 +22,7 @@ spec:
       image: registry.redhat.io/openshift4/ose-cli:v4.11
       script: |
         #!/usr/bin/env bash
+        set -e
         echo
         echo "App Studio Build Summary:"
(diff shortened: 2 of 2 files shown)


Michkov:
Test tasks are not supposed to fail when there is an issue.

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -31,6 +31,7 @@ spec:
           cpu: 10m
       script: |
         #!/usr/bin/env bash
+        set -e
```
MartinBasti, line 34: We cannot fail script, checks must always return `HACBS_TEST_OUTPUT`

Any error handling must happen within the script

MartinBasti (changes requested):

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
tkdchen: Remove.

task/configure-build/0.1/configure-build.yaml
```diff
@@ -33,6 +33,7 @@ spec:
           mountPath: /secret/default-push-secret
       script: |
         #!/usr/bin/env bash
+        set -e
```
MartinBasti, line 36: maybe `set -euo pipefail` to have all important failures covered?

tkdchen: Looks unnecessarily for now in this script I think. `-e` is enough for catching potential problems as described in the JIRA issue. Option `pipefail` is added for some other script that has pipe operation, and `-u` is helpful along with the size of script is growing.

MartinBasti: okay

MartinBasti (approved):

PR #308, opened 2023-02-13
STONEBLD-694: Add e2e test to verify cachi2 pip support

Michkov:
Use https://github.com/cachito-testing/pip-e2e-test for testing hermetic build with pip.

### DIFF
--- a/hack/util-install-bundle.sh
+++ b/hack/util-install-bundle.sh
@@ -18,7 +18,7 @@ metadata:
   name: build-pipeline-selector
 spec:
   selectors:
-    - name: Hermetic build (docker)
+    - name: Hermetic build - golang
       pipelineParams:
         - name: hermetic
           value: "true"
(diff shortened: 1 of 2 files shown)


hack/util-install-bundle.sh
```diff
@@ -31,6 +31,19 @@ spec:
...
+        - name: hermetic
+          value: "true"
+        - name: prefetch-input
+          value: "pip"
+      pipelineRef:
+        name: docker-build
+        bundle: ${BUNDLE}
+      when:
+        dockerfile: true
+        language: Python
```
chmeliik, line 45: "python" is lowercase in https://github.com/cachito-testing/pip-e2e-test/blob/main/devfile.yaml#L3, not sure if that matters

chmeliik (commented):
changes LGTM, probably won't pass until #302 is in though

hack/util-install-bundle.sh
Michkov: does not matter

chmeliik (approved):
LGTM :+1:

Michkov:
test will fail on:
```
Pipeline build-templates-e2e/docker-build can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "sanity-inspect-image": GET https://quay.io/v2/auth?scope=repository%3Aredhat-appstudio-tekton-catalog%2Fpull-request-builds%3Apull&service=quay.io: unexpected status code 502 Bad Gateway: <html>
```

PR #310, opened 2023-02-15
STONEBLD-716: recover pipelines for core services

tkdchen:
Build pipelines are recovered for Stonesoup core services, which are used to build core services in CI with an additional ability of allowing to update infra-deployments.

A major difference than before is that one pipeline per bundle is built and pushed to Quay.io. Bundle image has format:

    pipeline-core-services-pipeline_name:latest

e.g. pipeline-core-services-docker-build:latest

A new annotation appstudio.openshift.io/core-services-ci is introduced so that such pipelines can be identified easily to be handled in the script.


pipelines/core-services/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../docker-build
```
Michkov, line 4: it's enough to create docker-build pipeline.

pipelines/core-services/csi-secret.yaml
```diff
@@ -0,0 +1,5 @@
+- op: add
+  path: /spec/tasks/2/params
+  value:
+    - name: shared-secret
+      value: redhat-appstudio-staginguser
```
Michkov, line 5: this is for wrong task should be 0 now, also this is not needed at all. We will rather link secret to serviceaccount instead of using shared secret. So please just remove this file (and link in kustomization.yaml)

Michkov:
testing integration on new cluter

/retest

Michkov (approved):

PR #314, opened 2023-02-16
Run Shellspec tests via GitHub Actions

zregvart:
The test we have in `task/tkn-bundle/0.1/spec` is now exercised on
GitHub Actions by running `hack/test-shellspec.sh`. There is support for
GitHub reporting via annotations and incremental test runs -- i.e. skip
running change-unrelated tests.

Ref. https://issues.redhat.com/browse/HACBS-1740

### DIFF
--- a/task/tkn-bundle/0.1/spec/tkn-bundle_spec.sh
+++ b/task/tkn-bundle/0.1/spec/tkn-bundle_spec.sh
@@ -8,9 +8,19 @@ eval "$(shellspec - -c) exit 1"
 
 Describe "tkn-bundle task"
   setup() {
+    if ! command -v kubectl &> /dev/null; then
+      echo "ERROR: Please install kubectl"
+      return 1
+    fi
+
     # Create Kind cluster if it doesn't exist already
     CLUSTER_NAME="test-tkn-bundle"
-    kind get clusters -q | grep -q "${CLUSTER_NAME}" || kind create cluster -q --name="${CLUSTER_NAME}"
+    if ! command -v kind &> /dev/null; then
+      curl https://i.jpillora.com/kubernetes-sigs/kind | bash
+      mv kind "$HOME/.local/bin"
+    fi
+    kind get clusters -q | grep -q "${CLUSTER_NAME}" || kind create cluster -q --name="${CLUSTER_NAME}" || { echo 'ERROR: Unable to create a kind cluster'; return 1; }
+    kubectl cluster-info 2>&1 || { echo 'ERROR: Failed to access the cluster'; return 1; }
 
     # Install Tekton Pipeline, proceed with the rest of the test of the setup
     kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.42.0/release.yaml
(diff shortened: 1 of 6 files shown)


zregvart:
TIL, GitHub Actions are not enabled on this repository. What do we think about enabling them? cc @Michkov

Michkov:
@zregvart I do not have opinion here. Do I need to do anything on project level to set it? (seems to be enabled by default)

zregvart:
I guess I expected the workflow to trigger, perhaps I made a mistake somewhere. Let me see on my fork.

zregvart:
Found an issue in the workflow, need to fix the definition.

MartinBasti:
> I guess I expected the workflow to trigger, perhaps I made a mistake somewhere. Let me see on my fork.


IIRC, if it's first workflow, it must be merged first into main and then GH will activate them

zregvart:
The workflow [passed](https://github.com/zregvart/build-definitions/actions/runs/4233020812) on my fork, and [failed](https://github.com/zregvart/build-definitions/actions/runs/4232986527) when a deliberate error was introduced. This required quite a bit of troubleshooting, there is an environmental difference between Fedora and Ubuntu that's at play here and I can't pinpoint it. Lots of workarounds were added to make this work.

zregvart:
I don't know if the settings allow running a workflow from pull requests. @Michkov if you see [Approve and run](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) at the bottom here then please click on it so I can make sure that the workflow succeeds here. I would suspect that there can't be any differences from my fork and here.

Once this is merged the only restriction that could prevent the workflow from running would be the [Actions permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository). If it is currently set to to allow only certain actions this needs to be expanded to include the ones that are used in this workflow (`actions/checkout` and `jerop/tkn`).

zregvart:
@Michkov can you have a look at this?

task/tkn-bundle/0.1/spec/tkn-bundle_spec.sh
```diff
@@ -10,7 +10,17 @@ Describe "tkn-bundle task"
...
     CLUSTER_NAME="test-tkn-bundle"
-    kind get clusters -q | grep -q "${CLUSTER_NAME}" || kind create cluster -q --name="${CLUSTER_NAME}"
+    if ! command -v kind &> /dev/null; then
+      curl https://i.jpillora.com/kubernetes-sigs/kind | bash
+      mv kind "$HOME/.local/bin"
+    fi
+    kind get clusters -q | grep -q "${CLUSTER_NAME}" || kind create cluster -q --name="${CLUSTER_NAME}" || { echo 'ERROR: Unable to create a kind cluster'; return 1; }
+    kubectl cluster-info 2>&1 || { echo 'ERROR: Failed to access the cluster'; return 1; }
+
+    if ! command -v kubectl &> /dev/null; then
```
psturc, line 20: should this be checked before the command above?

Michkov (approved):

PR #323, opened 2023-02-22
feat(HACBS-1670): change result format for clair scan

hongweiliu17:
* Save ERROR result to HACBS_TEST_OUTPUT only
* Add number of vulnerability critical/high/medium/low to a new result CLAIR_SCAN_RESULT

The content of CLAIR_SCAN_RESULT is like:
```
{
  "vulnerabilities": {
    "critical": 1,
    "high": 0,
    "medium": 1,
    "low":0
}
```

### DIFF
--- a/task/clair-scan/0.1/clair-scan.yaml
+++ b/task/clair-scan/0.1/clair-scan.yaml
@@ -20,6 +20,8 @@ spec:
   results:
     - name: HACBS_TEST_OUTPUT
       description: test output
+    - name: CLAIR_SCAN_RESULT
+      description: clair scan result
   steps:
     - name: get-vulnerabilities
       image: quay.io/redhat-appstudio/clair-in-ci:latest  # explicit floating tag, daily updates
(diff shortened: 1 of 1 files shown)


task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -20,6 +20,8 @@ spec:
   results:
     - name: HACBS_TEST_OUTPUT
```
hongweiliu17, line 21: At first I drop HACBS_TEST_OUTPUT from the task since CLAIR_SCAN_RESULT is used to store the vulnerability number of clair scan result, but later I realize we need to save an ERROR result in a tekton result once there is any error, so I add HACBS_TEST_OUTPUT back.

MartinBasti:
I'm puzzled about checks consistency here, aren't all checks supposed to always return  HACBS_TEST_OUTPUT ?

hongweiliu17:
> I'm puzzled about checks consistency here, aren't all checks supposed to always return HACBS_TEST_OUTPUT ?

Good question, `Maybe it's best to drop this result from this task` is mentioned in comment https://issues.redhat.com/browse/HACBS-1670?focusedCommentId=21754377&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21754377,  I wanted to drop this HACBS_TEST_OUTPUT but later I feel we need to expose the ERROR result, so I want EC to confirm if they can consume this task.

jsztuka:
Looking god so far, I think its our best bet to preserve both results. I suppose EC can process more than one result for single task.

task/clair-scan/0.1/clair-scan.yaml
arewm: You can always report the HACBS_TEST_OUTPUT as well in case there are any other services which are depending on this result pattern. When reporting success, you can also indicate a `note` that `CLAIR_SCAN_RESULT` has the actual scan result summary to help users get to that point if they are looking.

hongweiliu17: Thanks! Updated.

MartinBasti (approved):
:rocket:

jsztuka (approved):
lgtm

arewm (approved):
Thanks!

PR #325, opened 2023-02-22
fix(HACBS-1821): update fbc-validation to correctly validate fbc fragment

jsztuka:
Fixes vital checks for FBC fragment

- presence of opm binary
- presence of grpc_health_probe
- runs opm validate on catalog
- makes sure there is only one opm.package present

### DIFF
--- a/task/fbc-validation/0.1/fbc-validation.yaml
+++ b/task/fbc-validation/0.1/fbc-validation.yaml
@@ -19,10 +19,13 @@ spec:
     - name: workspace
   steps:
     - name: extract-and-check-binaries
-      image: quay.io/redhat-appstudio/hacbs-test:v1.0.9@sha256:e3714c1ec1bd7bc754a510cba6892052a74bf523a185cf75a59d84059375444f
+      image: quay.io/redhat-appstudio/hacbs-test:v1.0.10@sha256:ae276785973aaffb4554c1798cf629df33e7058b30c32b75bd8ada54b2afe42f
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
       securityContext:
         runAsUser: 0
+        capabilities:
+          add:
+            - SETFCAP
       resources:
         limits:
           memory: 4Gi
@@ -32,9 +35,11 @@ spec:
           cpu: 10m
       script: |
         #!/usr/bin/env bash
+        set -o pipefail
         source /utils.sh
+
         ### Try to extract binaries with configs > check binaries functionality > check opm validate ###
-        if [ ! -f ../sanity-inspect-image/image_inspect.json ] || [ -z $(cat ../sanity-inspect-image/image_inspect.json) ]; then
+        if [ ! -s ../sanity-inspect-image/image_inspect.json ]; then
           echo "File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image"
           HACBS_TEST_OUTPUT="$(make_result_json -r ERROR -t 'File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image!')"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
@@ -42,14 +47,29 @@ spec:
         fi
 
         conffolder=$(cat ../sanity-inspect-image/image_inspect.json | jq -r '.Labels ."operators.operatorframework.io.index.configs.v1"')
-
+        if [ $? -ne 0 ]; then
+          echo "Could not get labels from sanity-inspect-image/image_inspect.json, make sure the file exists and contains this label: operators.operatorframework.io.index.configs.v1."
+          HACBS_TEST_OUTPUT="$(make_result_json -r ERROR)"
+          echo "${HACBS_TEST_OUTPUT}" | tee "$(results.HACBS_TEST_OUTPUT.path)"
+          exit 0
+        fi
+        mkdir -p /tmp/image-content
+        pushd /tmp/image-content
         image_with_digest="$(params.IMAGE_URL)@$(params.IMAGE_DIGEST)"
 
-        mkdir confdir
-        if ! oc image extract "${image_with_digest}" --file /bin/opm --file /bin/grpc_health_probe --path $conffolder/*:confdir/ ; then
+        if ! oc image extract "${image_with_digest}" ; then
           echo "Unable to extract image! Skipping checking binaries!"
           HACBS_TEST_OUTPUT="$(make_result_json -r ERROR -t 'Unable to extract image! Skipping checking binaries!')"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
+          popd
+          exit 0
+        fi
+
+        if [ -z "$(ls -A .$conffolder)" ]; then
+          echo "$conffolder is empty, missing catalog file."
+          HACBS_TEST_OUTPUT="$(make_result_json -r ERROR)"
+          echo "${HACBS_TEST_OUTPUT}" | tee "$(results.HACBS_TEST_OUTPUT.path)"
+          popd
           exit 0
         fi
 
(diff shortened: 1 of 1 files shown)


task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -57,32 +63,35 @@ spec:
...
-        if ! ./opm render confdir | jq -en 'reduce (inputs | select(.schema == "olm.package")) as $obj (0; .+1) == 1'; then
+        if ! opm render ."${conffolder}" | jq -en 'reduce (inputs | select(.schema == "olm.package")) as $obj (0; .+1) == 1'; then
           echo "!FAILURE! - more than one olm.packages are not permitted in a FBC fragment"
           failure_num=`expr $failure_num + 1`
           TESTPASSED=false
         fi
         if [ $TESTPASSED == false ]; then
           HACBS_ERROR_OUTPUT="$(make_result_json -r FAILURE -f $failure_num -s `expr $check_num - $failure_num`)"
           echo "${HACBS_ERROR_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
+          cd ..
```
hongweiliu17, line 90: We may delete the file out of the  `if..else..` so we can have only one `cd & rm -fr`, but it is not a big deal.

hongweiliu17 (approved):

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Thats a good point. Thanks!

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -34,19 +37,22 @@ spec:
         #!/usr/bin/env bash
         source /utils.sh
         ### Try to extract binaries with configs > check binaries functionality > check opm validate ###
-        if [ ! -f ../sanity-inspect-image/image_inspect.json ] || [ -z $(cat ../sanity-inspect-image/image_inspect.json) ]; then
+
+        if [[ ! -f ../sanity-inspect-image/image_inspect.json ]] || [[ -z $(cat ../sanity-inspect-image/image_inspect.json) ]]; then
```
hongweiliu17, line 41: See https://github.com/redhat-appstudio/build-definitions/pull/326#discussion_r1115431258, they are similar

hongweiliu17 (approved):

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -34,19 +37,22 @@ spec:
...
           echo "File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image"
           HACBS_TEST_OUTPUT="$(make_result_json -r ERROR -t 'File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image!')"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
           exit 0
         fi
 
         conffolder=$(cat ../sanity-inspect-image/image_inspect.json | jq -r '.Labels ."operators.operatorframework.io.index.configs.v1"')
+        image_with_digest=$(params.IMAGE_URL)@$(params.IMAGE_DIGEST)
```
MartinBasti, line 49: Why this change?

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -34,19 +37,22 @@ spec:
...
 
-        image_with_digest="$(params.IMAGE_URL)@$(params.IMAGE_DIGEST)"
+        #make temporary folder for image content which will be deleted at the end of this task
```
MartinBasti, line 51:
```suggestion
        # make temporary folder for image content which will be deleted at the end of this task
```

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -34,19 +37,22 @@ spec:
...
+        mkdir to-delete
```
MartinBasti, line 52: Could we name it like catalog-fragment-content to keep semantic of what's inside?

MartinBasti: I see working dir is a workspace, shouldn't, why do we need copy content to workspace volume? Couldn't we just use local dir or emtpyDir? Copying to shared mount will make it much slower, we don't need it

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -86,3 +91,4 @@ spec:
           HACBS_TEST_OUTPUT="$(make_result_json -r SUCCESS -s $check_num)"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
         fi
+        cd .. && rm -fr to-delete
```
MartinBasti, line 94: bonus pints for making `to-delete` a constant instead of copy-pasting the same name 3 times

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -57,24 +63,23 @@ spec:
...
         fi
-        if [ ! -f "grpc_health_probe" ]; then
-          echo "!FAILURE! - grpc_health_probe binary check failed"
+        if [ ! $(find . -name grpc_health_probe ) ]; then
+          echo "!FAILURE! - grpc_health_probe binary presence check failed"
           failure_num=`expr $failure_num + 1`
           TESTPASSED=false
         fi
-        if ! ./opm validate confdir; then
+        if ! opm validate ."${conffolder}"; then
```
MartinBasti, line 77: Did we validate that conffolder is non-empty to avoid weird errors?

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -34,19 +37,22 @@ spec:
...
-        if [ ! -f ../sanity-inspect-image/image_inspect.json ] || [ -z $(cat ../sanity-inspect-image/image_inspect.json) ]; then
+
+        if [[ ! -f ../sanity-inspect-image/image_inspect.json ]] || [[ -z $(cat ../sanity-inspect-image/image_inspect.json) ]]; then
           echo "File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image"
           HACBS_TEST_OUTPUT="$(make_result_json -r ERROR -t 'File ../sanity-inspect-image/image_inspect.json is not generated correctly, please check HACBS_TEST_OUTPUT of task sanity-inspect-image!')"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
           exit 0
         fi
 
         conffolder=$(cat ../sanity-inspect-image/image_inspect.json | jq -r '.Labels ."operators.operatorframework.io.index.configs.v1"')
```
MartinBasti, line 48: IMO we should check if this command failed and print appropriate error message

MartinBasti (changes requested):

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Lets put back the quotes, don't know why they were stripped.

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: We are using workspace same as from previous step to get: 
../sanity-inspect-image/image_inspect.json

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Check added.

task/fbc-validation/0.1/fbc-validation.yaml
MartinBasti: yes, for `../sanity-inspect-image/image_inspect.json` it's ok, but why put there some temp data from oc extract?

task/fbc-validation/0.1/fbc-validation.yaml
MartinBasti: uff, sorry I haven't noticed :D so change it then

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -32,21 +35,39 @@ spec:
...
+        if [ $? -ne 0 ]; then
+          echo "Could not get labels from sanity-inspect-image/image_inspect.json, make sure the file exists and contains this label: operators.operatorframework.io.index.configs.v1."
+          HACBS_TEST_OUTPUT="$(make_result_json -r FAILURE -f 1)"
```
MartinBasti, line 52:
```suggestion
          HACBS_TEST_OUTPUT="$(make_result_json -r ERROR)"
```

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Changed that to /tmp

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Updated

task/fbc-validation/0.1/fbc-validation.yaml
MartinBasti: @jsztuka so can we apply the same fix here?

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -32,49 +35,67 @@ spec:
...
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
           exit 0
         fi
 
+        if [ -z "$(ls -A .$conffolder)" ]; then
+          echo "$conffolder is empty, missing catalog file."
+          HACBS_TEST_OUTPUT="$(make_result_json -r ERROR)"
+          echo "${HACBS_TEST_OUTPUT}" | tee "$(results.HACBS_TEST_OUTPUT.path)"
+          popd
+          rm -fr /tmp/*
```
MartinBasti, line 74: the same here

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: We do, updated.

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -86,3 +107,5 @@ spec:
           HACBS_TEST_OUTPUT="$(make_result_json -r SUCCESS -s $check_num)"
           echo "${HACBS_TEST_OUTPUT}" | tee $(results.HACBS_TEST_OUTPUT.path)
         fi
+        popd
+        rm -fr /tmp/*
```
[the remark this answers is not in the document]
jsztuka: Alright, /tmp content will not be removed in this task.

task/fbc-validation/0.1/fbc-validation.yaml
jsztuka: Same as above.

MartinBasti (approved):

PR #327, opened 2023-02-23
hermetic: use . instead source

Michkov:
'source' command is bash build-in, so it's not working for images with sh only. '.' is more universal.


chmeliik (approved):
Oh, good catch. LGTM :+1:

brunoapimentel (approved):

PR #328, opened 2023-02-23
Use older e2e-test image to fix e2e-tests issue

Michkov:
using older e2e-tests image which does not include https://github.com/redhat-appstudio/e2e-tests/blob/266c7898952a16e22471f0b928ddb3fed71ec0f8/pkg/framework/framework.go#L40


psturc (approved):

jsztuka (approved):
+1

PR #329, opened 2023-02-23
Change STRICT param to true in the EC pipeline

simonbaird:
See commentary for explanation. I'm not sure if this will require a change in the tests, but I'll push the PR and see how it goes.

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -50,11 +50,12 @@ spec:
           value: "$(params.SNAPSHOT)"
         - name: SSL_CERT_DIR
           value: "$(params.SSL_CERT_DIR)"
-        # Tekton only creates results for successful tasks. Since the goal of this pipeline
-        # is not to gate an operation, but to provide feedback to users, do not fail if there
-        # are violations. This ensures users can view the full validation report.
+        # It's confusing for users to see a passing taskrun that represents a failing EC test.
+        # For that reason let's have the taskrun fail when there are EC violations. Also, if
+        # this is set to false (IIUC), it's not possible to have the IntegrationTest gate the
+        # deploy to the devel environment work, which is what users expect to be able to do.
         - name: STRICT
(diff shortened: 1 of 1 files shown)


robnester-rh (approved):
lgtm

pipelines/enterprise-contract.yaml
```diff
@@ -50,11 +50,12 @@ spec:
           value: "$(params.SNAPSHOT)"
         - name: SSL_CERT_DIR
           value: "$(params.SSL_CERT_DIR)"
-        # Tekton only creates results for successful tasks. Since the goal of this pipeline
-        # is not to gate an operation, but to provide feedback to users, do not fail if there
-        # are violations. This ensures users can view the full validation report.
+        # It's confusing for users to see a passing taskrun that represents a failing EC test.
+        # For that reason let's have the taskrun fail when there are EC violations. Also, if
+        # this is set to false (IIUC), it's not possible to have the IntegrationTest gate the
```
lcarva, line 55: +1 to having a safer default. But I also think the integration folks should be explicit about which strict level they want. My understanding at the time was that the verify-ec task in the integration pipeline should behave like the tests in the build pipeline. They always pass, but produce a report. It is, of course, a different approach than what the release pipeline does. Have you checked with the integration folks?

lcarva (approved):

pipelines/enterprise-contract.yaml
simonbaird: https://issues.redhat.com/browse/STONE-574 talks about the bigger picture, including the new UX that uses the HACBS_TEST_OUTPUT results to show status to the user. So I think this is okay as a short term thing.

simonbaird:
The merge button is green now, but I'm not sure if I should merge. @Michkov FYI.

Michkov:
@simonbaird CI is currently broken since the prod/stage cluster is not ready. Feel free to merge. It will be validated once more in infra-deployments.

PR #332, opened 2023-02-28
fix(HACBS-1817): fix the potentially invalid use of skopeo inspect

jinqi7:


MartinBasti (changes requested):
This doesn't work, as I mentioned in ticket, --config doesn't provide annotations and fix won't be so easy as just using --config option. This will be much more complicated to address and probably we cannot use skopeo for it
```
skopeo inspect --no-tags --raw --config docker://quay.io/martinbasti/single-container-app:test | jq -r ".annotations.\"org.opencontainers.image.base.digest\""
null

```

MartinBasti:
can we close this?

PR #335, opened 2023-03-03
STONEBLD-884: Quick fix for running on Openshift 1.9

Michkov:
Fixing issue `invalid pipelineresults [JAVA_COMMUNITY_DEPENDENCIES], the referred results don't exist`

There is still issue with java pipelines with JVM build service enabled.

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -79,6 +79,7 @@ spec:
         memory: 512Mi
         cpu: 10m
     script: |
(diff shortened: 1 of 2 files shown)


psturc (approved):

lcarva:
@Michkov, is this needed because that particular result is conditional and may not be emitted in some cases?

Michkov:
@lcarva yes, seems that there is new validation that the results were created in Openshift Pipelines 1.9.

PR #339, opened 2023-03-08
Adapts the incremental support for GitHub

zregvart:
The script no longer fetches into `main` but to a temp branch. This should fix the:

```
fatal: refusing to fetch into branch 'refs/heads/main' checked out at '/home/runner/work/build-definitions/build-definitions'
```

error.

The script now takes into account running on GitHub Actions and takes the relevant git change commit id and base from the available environment variables.


zregvart:
> Seems reasonable

We shall see

PR #343, opened 2023-03-14
STONEBLD-683: Avoid having params in script section

Michkov:
Based on
https://tekton.dev/docs/pipelines/tasks/#substituting-in-script-blocks we should avoid to pass $(params.X) directly to the script section.

PR #345, opened 2023-03-15
STONEBLD-683: Fix of update-infra-deployments

Michkov:
Fixing update-infra-deployments which was broken by #343


tisutisu (approved):
LGTM

PR #347, opened 2023-03-15
HACBS-1821-support dual catalog format

jsztuka:
catalog.yaml could contain json format data, so its not possible to parse it by yq, thats why jq is needed there.


Michkov:
json is subset of yaml it should be working with `yq`

MartinBasti:
what's the error you encounter?

jsztuka:
> what's the error you encounter?

This file, generated by OPM
https://github.com/scoheb/fbc-test/blob/main/4.13/catalog/kubevirt-hyperconverged/catalog.yaml

jsztuka:
Reference to the discussion:
https://redhat-internal.slack.com/archives/C030SGP2VB9/p1678818363603369

MartinBasti:
it's even worse, it's concatenated json, it even need parsers with that support. jq can do it, python json cannot.

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -32,10 +32,13 @@ spec:
 
         relImgs="$(yq -r '.relatedImages[]?.image' $catalog)"
         if [ $? -ne 0 ]; then
-          echo "Could not get related images, make sure catalog.yaml exists in FBC fragment image and has valid yaml format."
-          HACBS_TEST_OUTPUT="$(make_result_json -r FAILURE -f 1)"
-          echo "${HACBS_TEST_OUTPUT}" | tee "$(results.HACBS_TEST_OUTPUT.path)"
+          relImgs="$(jq -r '.relatedImages[]?.image' $catalog)"
```
MartinBasti, line 35: can we provide an explanatory comment here, why we are doing such magic with jq?

jsztuka: Sounds about right.

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -30,12 +30,19 @@ spec:
         FAILEDIMAGES=""
         catalog="$(find $(workspaces.workspace.path)/hacbs/fbc-validation/ -name catalog.yaml)"
 
+        # OPM generates catalog file in a way that yaml file could contain json type of data
```
lcarva, line 33: I think this comment is a bit misleading. Anyone reading this will immediately think something is off because JSON is a subset of YAML.

The problem here is three fold:
1. `opm` produces a stream of json objects - not an array
2. `catalog.yaml` could also be in yaml format (although I don't fully understand that scenarios that would cause this)
3. `yq` cannot properly handle a stream of objects in a dynamic way

Starting with version 4.27.1, you can force `yq` to read the input as json via the `-p` flag, e.g. `yq -p json $catalog`. However, if the catalog is in YAML format, then this will fail. It would be nice if `yq` could handle this in a more dynamic way. It might be worth filing an issue upstream.

To avoid having to repeat the expression between `yq` and `jq`, as well as, ensuring these tools behave the same way (they don't for more advanced expressions), consider fist converting the file into either json or yaml if needed, then using one of the tools to actually extract values from it, i.e. normalize the input then operate on it instead of trying to operate directly on the different formats.

jsztuka: Yes, we can arrange around this in many ways. OR we can recieve actual standardized format and then the solution would be much simpler and easier to work with.
Currently there is two kinds of catalogs formats AFAIK. Yaml file with yaml data, yaml file with json data.
I'm not aware of the way that catalogs are going to be used anyplace else, but this double file/data format will not cause pain only on our side.

jsztuka:
Since the format does not allow to use only `yq` or `jq` I think the best way of handling this is filling the issue upstream for `yq` so it can handle usch situation.
It is not clear when the implementation will take place, so I advise I create issue in StoneIntegration Jira to track the progress of the ticket and in a meantime lets apply those changes in this PR so FBC-related-image check gets unblocked. Mention in comment that such think will need to change in future and that it is being tracked.

dirgim (approved):
LGTM

PR #349, opened 2023-03-17
STONEBLD-960: summary - do not show image if build failed

Michkov:
Do not show generated image in summary when build-container task failed.


psturc (approved):
lgtm

PR #354, opened 2023-03-22
intg-13: let user know why yq failed

jsztuka:


task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -36,6 +36,7 @@ spec:
 
         relImgs="$(yq -r '.relatedImages[]?.image' $catalog)"
         if [ $? -ne 0 ]; then
+          echo "yq failed because catalog.yaml contains stream of JSON objects, proceeding with jq..."
```
MartinBasti, line 39: we cannot tell that is contains stream of JSON, it can be just random data file at this point

jsztuka: That is true, I will rephrase that.

dirgim (approved):
LGTM

MartinBasti (approved):

PR #358, opened 2023-03-23
Let renovate bot pin ec-cli image

lcarva:


lcarva:
@MartinBasti , @Michkov ptal

MartinBasti:
we don't want to pin digest for everything
https://github.com/redhat-appstudio/build-definitions/blob/main/task/clamav-scan/0.1/clamav-scan.yaml#L147  this must stay floating tag (the same for Clair task) because it's updated daily

lcarva:
I wonder if I just put a digest now, then renovate bot will automatically keep that up to date :thinking:

MartinBasti:
IIrC it will, it does pinning for cachi2 image

lcarva:
Updated the approach.

MartinBasti (approved):

PR #363, opened 2023-03-28
fix(STONEINTG-373): false-positive clair report

jsztuka:
This prevents reporting false positive reports by clair in case there are 0 vulnerabilities image.

### DIFF
--- a/task/clair-scan/0.1/clair-scan.yaml
+++ b/task/clair-scan/0.1/clair-scan.yaml
@@ -63,8 +63,8 @@ spec:
         #!/usr/bin/env bash
         . /utils.sh
 
-        if [[ ! -f /tekton/home/clair-vulnerabilities.json ]] || [[ "$(jq '.[] | has("failures")' /tekton/home/clair-vulnerabilities.json)" == "false" ]]; then
-          note="Task $(context.task.name) failed: /tekton/home/clair-vulnerabilities.json did not generate correctly. For details, check Tekton task log."
+        if [[ ! -f /tekton/home/clair-vulnerabilities.json ]]; then
+          note="Task $(context.task.name) failed: /tekton/home/clair-vulnerabilities.json did not generate. For details, check Tekton task log."
           HACBS_TEST_OUTPUT=$(make_result_json -r "ERROR" -t "$note")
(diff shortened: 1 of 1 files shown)


dheerajodha:
The branch is out-of-date

dirgim (approved):
LGTM

dirgim:
> The branch is out-of-date, should I click on "Update branch" on GitHub UI?

I don't think there's any need to do that in this case as there are no conflicts and the change doesn't affect the other tasks within the pipeline.

dheerajodha (approved):

MartinBasti (changes requested):
Actually no :D this will break later on jq if `failures` is not there
`Cannot iterate over null (null)`

Can we update vulnerabilities-check.rego to always provide `failures` as and empty list

dirgim:
> Actually no :D this will break later on jq if `failures` is not there `Cannot iterate over null (null)`
> 
> Can we update vulnerabilities-check.rego to always provide `failures` as and empty list

Actually I don't think we can :/ Afaik Conftest is in charge of constructing that field.

dirgim:
Something like this e.g.:
```
critical: (.[] | .failures? // [] | map(select(.metadata.details.name=="clair_critical_vulnerabilities").metadata."vulnerabilities_number" // 0)| add // 0)
```
could work to address when there are no failures in the `clair-vulnerabilities.json`, wdyt?

jsztuka:
Still getting `Cannot iterate over null (null)` even when trying above solution.

jsztuka:
Currently getting following output:
```
STEP-CONFTEST-VULNERABILITIES

[
	{
		"filename": "/tekton/home/clair-result.json",
		"namespace": "required_checks",
		"successes": 4
	}
]
STEP-TEST-FORMAT-RESULT

{"vulnerabilities":{"critical":0,"high":0,"medium":0,"low":0}}
{"result":"SUCCESS","timestamp":"1680084268","note":"Task clair-scan completed: Refer to Tekton task result CLAIR_SCAN_RESULT for vulnerabilities scanned by Clair.","namespace":"default","successes":0,"failures":0,"warnings":0}
```

MartinBasti (approved):

PR #367, opened 2023-03-30
feat(clair-ci): provide clair output to logs

jsztuka:
This behaviour was discussed and agreed on. We want to provide more explicit information to custommer regarding clair-scans. Thanks to this, they will be able to search for specific CVEs and how to fix them or version that fixes the CVEs.


MartinBasti (approved):

dirgim (approved):
+1

PR #369, opened 2023-03-30
add label for enterprise contract pipeline

joejstuart:
This will be used to identify it's the enterprise contract pipeline by the UI

https://issues.redhat.com/browse/HACBS-2062


lcarva (approved):

lcarva:
/ok-to-test

Not sure if that's needed here? I can't tell if the CI is actually running.

PR #379, opened 2023-04-11
STONEBLD-1046: Require MIGRATION.md for new versions

Michkov:
- Start requiring new version when there is breaking change in the task -> when task update requires change of pipeline.
- Add check of existance of MIGRATION.md in new versions of tasks.
- Create version 0.2 of deprecated-image-check with MIGRATION.md - due to change from STONEINTG-233

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,14 @@ Shellspec tests can be run by invoking `hack/test-shellspec.sh`.
 
 Release is done by setting env variable `MY_QUAY_USER=redhat-appstudio`, `BUILD_TAG=$(git rev-parse HEAD)` and running `hack/build-and-push.sh`.
 
+### Versioning
+
+When the task update changes the interface (eg. change of parameters, workspaces or results names) then a new version of the task should be created. The folder with the new version must contain `MIGRATION.md` with instructions on how to update the current pipeline file in user's `.tekton` folder.
+
+Adding a new parameter with a default value does not require the task version increase.
+
+Task version increase must be approved by Project Manager.
+
 ## Testing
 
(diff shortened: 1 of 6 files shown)


dirgim:
Thanks for opening this @Michkov! Do we need to revert the changes to the [0.1 version of the task](https://github.com/redhat-appstudio/build-definitions/blob/main/task/deprecated-image-check/0.1/deprecated-image-check.yaml) so it references the old workspace?

Michkov:
@dirgim I would not revert that, since some users are already using that version (0.1) with test-ws, so potentially they can be confused by migration note but the test will not be failing for them.

dirgim:
> @dirgim I would not revert that, since some users are already using that version (0.1) with test-ws, so potentially they can be confused by migration note but the test will not be failing for them.

Ok, that sounds reasonable, thanks. LGTM in that case!

dirgim (approved):

README.md
```diff
@@ -45,6 +45,14 @@ Shellspec tests can be run by invoking `hack/test-shellspec.sh`.
 
 Release is done by setting env variable `MY_QUAY_USER=redhat-appstudio`, `BUILD_TAG=$(git rev-parse HEAD)` and running `hack/build-and-push.sh`.
 
+### Versioning
+
+When the task update changes the interface (eg. change of parameters, workspaces or results names) then a new version of the task should be created. The folder with the new version must contain `MIGRATION.md` with instructions on how to update the current pipeline file in user's `.tekton` folder.
+
+Adding a new parameter with a default value does not require the task version increase.
+
+Task version increase must be approved by Project Manager.
```
tkdchen, line 54: I think whether to bump a new version or not can be determined during grooming and be part of the acceptance criteria as a result. How does Project Manager get involved into this process?

PR #381, opened 2023-04-13
Add CI job to check partner task submission

tkdchen:
This commit also introduces a directory structure for partner tasks, that should follow the same structure of existing tasks with versions.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -83,7 +83,7 @@ spec:
               workingDir: $(workspaces.source.path)
               script: |
                 #!/usr/bin/env bash
-                .tekton/scripts/check-partner-tasks.sh
+                ./hack/check-partner-tasks.sh
           workspaces:
             - name: source
         workspaces:
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -73,6 +73,22 @@ spec:
         workspaces:
           - name: source
             workspace: workspace
+      - name: check-partner-tasks
+        runAfter:
+          - build-container
+        taskSpec:
+          steps:
+            - name: check-task-structure
+              image: quay.io/redhat-appstudio/pull-request-builds:build-definitions-utils-{{revision}}
+              workingDir: $(workspaces.source.path)
+              script: |
+                #!/usr/bin/env bash
+                .tekton/scripts/check-partner-tasks.sh
+          workspaces:
+            - name: source
+        workspaces:
+          - name: source
+            workspace: workspace
       - name: build-bundles
         params:
           - name: revision
--- a/.tekton/scripts/check-partner-tasks.sh
+++ b/.tekton/scripts/check-partner-tasks.sh
@@ -105,13 +105,35 @@ check_task_schema() {
     return 0
 }
 
-check_dir_structure || exit 1
-
 exitcode=0
-if ! check_task_schema; then
-    exitcode=$((exitcode+1))
+check_dir_structure_status=Fail
+check_task_schema_status="n/a "
+check_privilege_use_status="n/a "
+
+if check_dir_structure; then
+    check_dir_structure_status=Pass
+    if check_task_schema; then
+        check_task_schema_status=Pass
+    else
+        check_task_schema_status=Fail
+        exitcode=$((exitcode+1))
+    fi
+
+    if check_privilege_use; then
+        check_privilege_use_status=Pass
+    else
+        check_privilege_use_status=Fail
+        exitcode=$((exitcode+1))
+    fi
+else
+    exitcode=1
 fi
-if ! check_privilege_use; then
-    exitcode=$((exitcode+1))
-fi
+
+echo "
+|        Check         | Status |
+|----------------------+--------|
+| Directory structure  | $check_dir_structure_status   |
+| Task YAML definition | $check_task_schema_status   |
+| Privilege use        | $check_privilege_use_status   |"
+
 exit $exitcode
--- a/.tekton/scripts/check-partner-tasks.sh
+++ b/.tekton/scripts/check-partner-tasks.sh
@@ -7,6 +7,7 @@ cd "$SCRIPTDIR/../.." || exit 1
 # task/partners/<task name>/<version>/<task file>
 
 if [ ! -e "partners/" ]; then
+    echo "No partners directory exists. Skip checks against partner tasks."
     exit 0
 fi
 
--- a/.tekton/scripts/check-partner-tasks.sh
+++ b/.tekton/scripts/check-partner-tasks.sh
@@ -73,7 +73,7 @@ privileged:true:securityContext.privileged is not allowed to be true."
     done <<<"$check_cases"
 
     if [ -s "$check_result" ]; then
-        echo "Task $task_file uses unexpected privileges:"
+        echo "Task $task_file has unexpected privileges:"
         cat "$check_result"
         echo
         echo >"$check_result"
--- a/.tekton/scripts/check-partner-tasks.sh
+++ b/.tekton/scripts/check-partner-tasks.sh
@@ -9,14 +9,23 @@ if [ ! -e "partners/" ]; then
 fi
 
 check_dir_structure() {
-    issue_found=0
+    resultf=$(mktemp)
+    check_result=$(mktemp)
+
     for task_dir in partners/*; do
+        owners_file="$task_dir/OWNERS"
+        if [ ! -e "$owners_file" ]; then
+            echo "error: missing file $owners_file" >>"$check_result"
+            echo "issue found" >"$resultf"
+        fi
+
         task_name=$(basename "$task_dir")
-        for version_dir in "$task_dir"/*; do
+        find "$task_dir" -mindepth 1 -maxdepth 1 -type d | \
+        while read -r version_dir; do
             task_version=$(basename "$version_dir")
             if ! echo "$task_version" | grep "^[0-9]\+\.[0-9]\+$" >/dev/null 2>&1; then
-                echo "error: $version_dir is not a version directory. A version directory name only contains digits and dot, e.g. 0.1, 1.2"
-                issue_found=1
+                echo "error: $version_dir is not a version directory. A version directory name only contains digits and dot, e.g. 0.1, 1.2" >>"$check_result"
+                echo "issue found" >"$resultf"
             else
                 # Ignore empty directory whatever it is a version directory or not.
                 if [ -n "$(ls "$version_dir")" ]; then
@@ -30,19 +39,26 @@ check_dir_structure() {
                         fi
                     done
                     if [ -z "$found" ]; then
-                        echo "error: no task file is found under $version_dir. A task file must be named with the same task name. For example, task name is task1, then task file must have name task1.yaml"
-                        issue_found=1
+                        echo "error: no task file is found under $version_dir. A task file must be named with the same task name. For example, task name is task1, then task file must have name task1.yaml" >>"$check_result"
+                        echo "issue found" >"$resultf"
                     fi
                     # Check README
                     readme_file="$version_dir/README.md"
                     if [ ! -s "$readme_file" ]; then
-                        echo "warning: it is recommended to provide an informative document for task $task_name in $readme_file"
+                        echo "warning: it is recommended to provide an informative document for task $task_name in $readme_file" >>"$check_result"
                     fi
                 fi
             fi
         done
     done
-    return $issue_found
+
+    if [ -s "$check_result" ]; then
+        cat "$check_result"
+        echo
+    fi
+
+    [ -s "$resultf" ] && return 1
+    return 0
 }
 
 check_privilege_use() {
--- a/.tekton/scripts/check-partner-tasks.sh
+++ b/.tekton/scripts/check-partner-tasks.sh
@@ -72,31 +72,31 @@ privileged:true:securityContext.privileged is not allowed to be true."
 
     find partners/*/*/*.yaml | awk -F '/' '{ print $0, $2, $4 }' | \
     while read -r task_file task_name yaml_file; do
-    if [ "${task_name}.yaml" != "$yaml_file" ]; then
-        continue
-    fi
+        if [ "${task_name}.yaml" != "$yaml_file" ]; then
+            continue
+        fi
 
-    while IFS=: read -r sc_config check_value err_msg; do
-        task_idx=0
-        # When no securityContext config is set, yq outputs message "Error: no matches found" to stderr.
-        # Redirect this message into stdout in order to suppress these potential Errors, and such error
-        # messages does not match any config value obviously.
-        for value in $(yq -e ".spec.steps[].securityContext.$sc_config" "$task_file" 2>&1); do
-            if [ "$value" == "$check_value" ]; then
-                echo "found" >>"$resultf"
-                step_name=$(yq -e ".spec.steps[$task_idx].name" "$task_file")
-                echo "error: in step $step_name, $err_msg" >>"$check_result"
-            fi
-            task_idx=$((task_idx+1))
-        done
-    done <<<"$check_cases"
+        while IFS=: read -r sc_config check_value err_msg; do
+            task_idx=0
+            # When no securityContext config is set, yq outputs message "Error: no matches found" to stderr.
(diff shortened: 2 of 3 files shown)


.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,85 @@
...
+          echo "found" >"$resultf"
+          step_name=$(yq -e ".spec.steps[$task_idx].name" "$task_file")
+          echo "error: in step $step_name, $err_msg Task $task_file"
+        fi
+        task_idx=$((task_idx+1))
+      done
+    done <<EOF
+allowPrivilegeEscalation:true:allowPrivilegeEscalation is not allowed to be true to request more privileges.
+runAsUser:0:root(0) is not allowed to be set to securityContext.runAsUser.
+privileged:true:ecurityContext.privileged is not allowed to be true.
```
Michkov, line 73: ecurityContext

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,85 @@
...
+  rv=0
+  if [ -n "$(cat "$resultf")" ]
+  then
+    rv=1
+  fi
+  return $rv
+}
+
+check_dir_structure || exit 1
+check_privilege_use || exit 1
```
Michkov, line 85: when check_dir_structure will fail then check_privilege_use will not be executed

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,85 @@
...
+        task_idx=$((task_idx+1))
+      done
+    done <<EOF
+allowPrivilegeEscalation:true:allowPrivilegeEscalation is not allowed to be true to request more privileges.
+runAsUser:0:root(0) is not allowed to be set to securityContext.runAsUser.
+privileged:true:ecurityContext.privileged is not allowed to be true.
+EOF
+  done
+  rv=0
+  if [ -n "$(cat "$resultf")" ]
```
Michkov, line 77: `[ -s "$resultf" ]`
man test
```
       -s FILE
              FILE exists and has a size greater than zero
```

.tekton/scripts/check-partner-tasks.sh
tkdchen: Yes, this is intended. An expected directory structure is a prerequisite so that `check_privilege_use` can find out the task files.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+check_dir_structure() {
+  issue_found=0
+  for task_dir in partners/*; do
+    task_name=$(basename "$task_dir")
+    for version_dir in "$task_dir"/*; do
+      task_version=$(basename "$version_dir")
+      if ! echo "$task_version" | grep "^[0-9.]\+$" >/dev/null 2>&1; then
+        echo "error: $version_dir is not a version directory. A version directory name only contains digits and dot, e.g. 0.1, 0.2.1"
```
tisutisu, line 20: I am wondering .. Do we expect the partners to look into the PR check failure logs to figure out what failed or we can communicate the the error messages to the partner submitter through the comments in the PR.

tkdchen: Good question. What you are thinking is an ideal solution for partner to check out the potential failure messages. That would be good to be implemented in a separate JIRA story, and it is not only for the partner task submission, but it can also be an general solution for showing pipeline summary information to users.

tkdchen:
An example of output:

```
Task schema validation failed: partners/task1/0.1/task1.yaml
error: no objects passed to apply

Task schema validation failed: partners/task1/0.2/task1.yaml
Error from server (BadRequest): error when creating "partners/task1/0.2/task1.yaml": admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "capbilities"

Task partners/task1/0.2/task1.yaml uses unexpected privileges:
error: in step further-task-2, allowPrivilegeEscalation is not allowed to be true to request more privileges.
error: in step sast-go, root(0) is not allowed to be set to securityContext.runAsUser.
error: in step further-task-1, securityContext.privileged is not allowed to be true.
```

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,117 @@
...
+check_dir_structure || exit 1
+
+exitcode=0
+if ! check_task_schema; then
+    exitcode=$((exitcode+1))
+fi
+if ! check_privilege_use; then
+    exitcode=$((exitcode+1))
+fi
+exit $exitcode
```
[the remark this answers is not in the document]
tkdchen: See the latest commit

.tekton/scripts/check-partner-tasks.sh
tisutisu: Related to my question, I see this statement `CI tests do the due diligence on the changes proposed in the PR. Success/Failures are reported in a way that the PR author can take reasonable action to resolve the issue.` in the ADR.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,139 @@
...
+    exit 0
+fi
+
+check_dir_structure() {
+    issue_found=0
+    for task_dir in partners/*; do
+        task_name=$(basename "$task_dir")
+        for version_dir in "$task_dir"/*; do
+            task_version=$(basename "$version_dir")
+            if ! echo "$task_version" | grep "^[0-9.]\+$" >/dev/null 2>&1; then
```
tisutisu, line 19: If someone by mistake name a version directory as `1..2` (with double . in the middle), they will not get this error right?

tisutisu: Do we need to consider `semantic versioning` regex here?

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$SCRIPTDIR/../.." || exit 1
+
+# Expected partner task directory structure:
+# task/partners/<task name>/<version>/<task file>
+
+if [ ! -e "partners/" ]; then
+    exit 0
```
tisutisu, line 10: Will it be a good idea to add a message here like `SKIPPED .. since PR is not related to partner task submission`

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$SCRIPTDIR/../.." || exit 1
+
```
tisutisu, line 5: Will it be good to consider the fact that partner users can also run the script locally before sending the PR , it will reduce the number of pipeline runs (if partner users try trail and error method to fix the issues in the task), one approach I think maybe by passing the task directory as an argument to the script. WDYT?

tkdchen: User is able to run `.tekton/scripts/check-partner-tasks.sh` directly. I'd like to keep the simplicity for now.

.tekton/scripts/check-partner-tasks.sh
tkdchen: A new commit is added to show a message.

.tekton/scripts/check-partner-tasks.sh
tisutisu: Do you mean without modifying the script?

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,140 @@
...
+fi
+
+echo "
+|        Check         | Status |
+|----------------------+--------|
+| Directory structure  | $check_dir_structure_status   |
+| Task YAML definition | $check_task_schema_status   |
+| Privilege use        | $check_privilege_use_status   |"
+
+exit $exitcode
```
tisutisu, line 140: Apart from checking status, do you think user might be looking for `task-bundle reference` to use in their custom PipelineRun ?

tkdchen: What do you mean?

.tekton/scripts/check-partner-tasks.sh
tkdchen: Yes, for now, not add an argument to allow specifying a directory to check.

.tekton/scripts/check-partner-tasks.sh
tisutisu: As a partner user I would like to know the taskRef for my submitted task through PR to use in my custom tekton pipeline, in that case do I have to search through the logs to get the reference?

.tekton/scripts/check-partner-tasks.sh
tkdchen: The task directory structure was added as a catalog by referencing the tekton catalog. Version is not specific to a semantic version scheme, and so far I also don't see any requirement of this "complicated" versioning methodology. But, I still modified the regex to validate the version that should be in format N.M and updated the error message to avoid ambiguity.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,140 @@
...
+                echo "found" >>"$resultf"
+                step_name=$(yq -e ".spec.steps[$task_idx].name" "$task_file")
+                echo "error: in step $step_name, $err_msg" >>"$check_result"
+            fi
+            task_idx=$((task_idx+1))
+        done
+    done <<<"$check_cases"
+
+    if [ -s "$check_result" ]; then
+        echo "Task $task_file uses unexpected privileges:"
```
tisutisu, line 76:
```suggestion
        echo "Task $task_file has unexpected privileges:"
```

tkdchen: Updated.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$SCRIPTDIR/../.." || exit 1
+
+# Expected partner task directory structure:
+# task/partners/<task name>/<version>/<task file>
+
+if [ ! -e "partners/" ]; then
```
tisutisu, line 9: Do you think naming the directory to `partner_task` would be a good idea instead of `partners`?

.tekton/scripts/check-partner-tasks.sh
tisutisu: okay got it. Thanks.

.tekton/scripts/check-partner-tasks.sh
tkdchen: I would use `partners`. Word "task" inside `partner_task` looks redundant.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,159 @@
...
+fi
+
+check_dir_structure() {
+    resultf=$(mktemp)
+    check_result=$(mktemp)
+
+    for task_dir in partners/*; do
+        owners_file="$task_dir/OWNERS"
+        if [ ! -e "$owners_file" ]; then
+            echo "error: missing file $owners_file" >>"$check_result"
```
tisutisu, line 18: Do you think separating the `OWNERS` file check with the directory structure will be good idea? Also if the checking if the OWNERS file is empty or not?

tkdchen: No separation.

> Also if the checking if the OWNERS file is empty or not?

I have no idea. But, in order to keep the consistency, this file should exist at least.

tisutisu: I think the partner engineer should atleast provide their github user id inside the OWNERS file so that in case any future communication required, we can use that id. Empty OWNERS file only for the sake of passing the validation does not make sense to me. WDYT?

tkdchen: I don't want to make this assumption.

tisutisu: I was trying to think of what could be the use case for OWNERS file in case of partner task submission. Should contain some information that will be useful for us?

.tekton/scripts/check-partner-tasks.sh
tisutisu: Since we created a improvement story to address this. Closing this for this PR.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,159 @@
...
+            echo "issue found" >"$resultf"
+        fi
+
+        task_name=$(basename "$task_dir")
+        find "$task_dir" -mindepth 1 -maxdepth 1 -type d | \
+        while read -r version_dir; do
+            task_version=$(basename "$version_dir")
+            if ! echo "$task_version" | grep "^[0-9]\+\.[0-9]\+$" >/dev/null 2>&1; then
```
Michkov, line 26:
```
error: partners/test/1.42.2 is not a version directory. A version directory name only contains digits and dot, e.g. 0.1, 1.2
```
Any reason why limit only to major.minor?

tkdchen: No special reason, just referenced the version used in tekton catalog back to the time when catalog was introduced to build-definitions. In practice, when task API is changed, it works well by increasing either major or minor version, that should depend on the preference of task maintainer. `major.minor` looks enough for the task version maintenance purpose.

tkdchen: Oh, if there is a reason, that is to keep consistency with the version scheme used for predefined tasks under `task/`, at where version scheme is convention by task maintainers and not forced.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,159 @@
...
+    return 0
+}
+
+check_privilege_use() {
+    resultf=$(mktemp)
+    check_result=$(mktemp)
+
+    check_cases="\
+allowPrivilegeEscalation:true:allowPrivilegeEscalation is not allowed to be true to request more privileges.
+runAsUser:0:root(0) is not allowed to be set to securityContext.runAsUser.
```
Michkov, line 70: currently some of our tasks are using runAsUser so it should be probably allowed. It's usually needed when PVC is used.

.tekton/pull-request.yaml
```diff
@@ -73,6 +73,22 @@ spec:
...
+        runAfter:
+          - build-container
+        taskSpec:
+          steps:
+            - name: check-task-structure
+              image: quay.io/redhat-appstudio/pull-request-builds:build-definitions-utils-{{revision}}
+              workingDir: $(workspaces.source.path)
+              script: |
+                #!/usr/bin/env bash
+                .tekton/scripts/check-partner-tasks.sh
```
Michkov, line 86: should be rather in ./hack/check-partner-tasks.sh -> so that user can run it manually.

.tekton/scripts/check-partner-tasks.sh
```diff
@@ -0,0 +1,159 @@
...
+}
+
+exitcode=0
+check_dir_structure_status=Fail
+check_task_schema_status="n/a "
+check_privilege_use_status="n/a "
+
+if check_dir_structure; then
+    check_dir_structure_status=Pass
+    if check_task_schema; then
```
Michkov, line 135: for manual execution it could run `oc whoami` -> if non-zero return code then it's not logged and check_task_schema should be skipped.

.tekton/pull-request.yaml
tkdchen: Done

.tekton/scripts/check-partner-tasks.sh
tkdchen: Done

Michkov (approved):

.tekton/scripts/check-partner-tasks.sh
tisutisu: Okay .. got it

PR #382, opened 2023-04-13
artifacts workspace is referenced in push.yaml

joejstuart:
This is to fix a CI error `workspace binding "artifacts" does not match any declared workspace`


task/update-infra-deployments/0.1/update-infra-deployments.yaml
```diff
@@ -295,3 +295,7 @@ spec:
 
         if __name__ == '__main__':
             main()
+
+  workspaces:
+    - name: artifacts
+      description: Workspace containing arbitrary artifacts used during the task run.
```
Michkov, line 301:
```suggestion
      description: Workspace containing arbitrary artifacts used during the task run.
      optional: true
```

Michkov (approved):

PR #388, opened 2023-04-24
STONEBLD-1213: return partner task result as comment

Michkov:


tkdchen:
Screenshot_20230426_102511

Should we change it to some other bot name?

tkdchen (commented):
LGTM. Just considering the bot name might be something else of build-definitions itself.

.tekton/pull-request.yaml
```diff
@@ -83,9 +83,43 @@ spec:
               workingDir: $(workspaces.source.path)
               script: |
                 #!/usr/bin/env bash
-                ./hack/check-partner-tasks.sh
+                ./hack/check-partner-tasks.sh > partner-tasks.out
+                echo $? > partner-tasks-code.out
+                cat partner-tasks.out
+            - name: create-comment
+              image: quay.io/chmouel/github-app-token@sha256:b4f2af12e9beea68055995ccdbdb86cfe1be97688c618117e5da2243dc1da18e
```
tisutisu, line 90: Would it be good to move the image under `redhat-appstudio` org?

tisutisu:
> Screenshot_20230426_102511
> 
> Change it to some other bot name?

Looks like `No partners directory exists. Skip checks against partner tasks.` is going to added as comment to every PR which does not touch the partners directory. Will it be good to handle it so that it won't create noise for other PRs?

tkdchen:
> Looks like No partners directory exists. Skip checks against partner tasks. is going to added as comment to every PR which does not touch the partners directory. Will it be good to handle it so that it won't create noise for other PRs?

So, script should probably keep quiet if no partners directory exist.

tisutisu:
> > Looks like No partners directory exists. Skip checks against partner tasks. is going to added as comment to every PR which does not touch the partners directory. Will it be good to handle it so that it won't create noise for other PRs?
> 
> So, script should probably keep quiet if no partners directory exist.

This log `No partners directory exists. Skip checks against partner tasks` is helpful when checking taskrun log in the pipelinerun, instead of blank it conveys that we ave skipped this task step since the PR is not touching partners directory. Will it be good to use filtering the log before commenting, if it does not contain `errors` or `warnings`, skip commenting?

.tekton/pull-request.yaml
Michkov: moved

.tekton/pull-request.yaml
```diff
@@ -83,9 +83,43 @@ spec:
...
+                - name: REPO_OWNER
+                  value: "{{repo_owner}}"
+                - name: REPO_NAME
+                  value: "{{repo_name}}"
+                - name: PULL_REQUEST_NUMBER
+                  value: "{{pull_request_number}}"
+              script: |
+                #!/usr/bin/env bash
+                exit_code=$(cat partner-tasks-code.out)
+                if [ $exit_code != "0" ]; then
```
tisutisu, line 113: Will it be good to communicate a success message through the comments in case partner task passed all validation like `Thank you for submitting your Task. It passed all the checks and contains no errors. We will take a look.`

PR #393, opened 2023-04-28
Bump ec task bundle version

mbestavros:
This should hopefully fix a regression introduced earlier today.

PR #394, opened 2023-04-28
Update to latest bundle for Enterprise Contract task

simonbaird:
Needed in a hurry for the change to the task result name, i.e. it's TEST_OUTPUT instead of HACBS_TEST_OUTPUT.


simonbaird:
Dupe of #393. Abandoning.

PR #395, opened 2023-05-03
STONEBLD-1157: Allow to set expiration on image tags

Michkov:
- Allow to set expiration of image tag
- by default do not expire

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -45,6 +45,10 @@ spec:
     description: In case it is not empty, the prefetched content should be made available to the build.
     name: PREFETCH_INPUT
     type: string
+  - default: ""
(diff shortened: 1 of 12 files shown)


mmorhun (approved):

MartinBasti:
That single commit have 2 totally unrelated stories, can we keep unrelated changes in separate commits pls? I would like to be able see clear changes in git log and don't die from git bisect as mutliple teams are using this repo

task/buildah/0.1/buildah.yaml
```diff
@@ -45,6 +45,10 @@ spec:
     description: In case it is not empty, the prefetched content should be made available to the build.
     name: PREFETCH_INPUT
     type: string
+  - default: ""
+    description: Delete image tag after specified time. Empty means to keep the image tag.
```
tisutisu, line 49: Does adding some examples for the values in the description will help like `time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively`?

PR #398, opened 2023-05-03
update docs for jvm build service

sbose78:


chmeliik:
Outdated and JVM build-service is not relevant anymore

PR #402, opened 2023-05-11
STONEBLD-867: Remove shared secret

sbose78:
https://issues.redhat.com/browse/STONEBLD-867


Michkov:
Converted to draft since we will need first update e2e-tests to stop using push shared secret.

Michkov (approved):

PR #406, opened 2023-05-22
fix(STONEINTG-418): inspect-image task should fail better

MartinBasti:
Retry to get image metadata with skopeo to give time to succesfully show image in the registry.

Report failures in better way than just crashing bash commands

### DIFF
--- a/task/inspect-image/0.1/inspect-image.yaml
+++ b/task/inspect-image/0.1/inspect-image.yaml
@@ -122,11 +122,24 @@ spec:
         echo "Source image ${IMAGE_URL} is built from scratch, so there is no base image."
         exit 0
       fi
+
       BASE_IMAGE="${BASE_IMAGE_NAME%:*}@$BASE_IMAGE_DIGEST"
-      echo "The base image is $BASE_IMAGE. Fetching its manifest."
-      skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
       echo "Detected base image: $BASE_IMAGE"
       echo -n "$BASE_IMAGE" > $(results.BASE_IMAGE.path)
+
+      for run in $(seq 1 $max_run); do
+        status=0
+        [ "$run" -gt 1 ] && sleep $sleep_sec  # skip last sleep
+        echo "Inspecting base image ${BASE_IMAGE} (try $run/$max_run)."
+        skopeo inspect --no-tags "docker://$BASE_IMAGE"  > $BASE_IMAGE_INSPECT && break || status=$?
+      done
+      if [ "$status" -ne 0 ]; then
+          echo "Failed to inspect base image ${$BASE_IMAGE}"
+          note="Task $(context.task.name) failed: Encountered errors while inspecting image. For details, check Tekton task log."
+          TEST_OUTPUT=$(make_result_json -r ERROR -t "$note")
+          echo "${TEST_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
+          exit 0
+      fi
 
       BASE_IMAGE_REPOSITORY="$(jq -r '.Name | sub("[^/]+/"; "") | sub("[:@].*"; "")' "$BASE_IMAGE_INSPECT")"
       echo "Detected base image repository: $BASE_IMAGE_REPOSITORY"
--- a/task/inspect-image/0.1/inspect-image.yaml
+++ b/task/inspect-image/0.1/inspect-image.yaml
@@ -70,18 +70,38 @@ spec:
       if [[ "${IMAGE_URL}" == *":"*"@"* ]]; then
         IMAGE_URL="${IMAGE_URL/:*@/@}"
       fi
-      echo "Inspecting manifest for source image ${IMAGE_URL}."
-      skopeo inspect --no-tags docker://"${IMAGE_URL}" > $IMAGE_INSPECT 2> stderr.txt || true
-      skopeo inspect --no-tags --raw docker://"${IMAGE_URL}" > $RAW_IMAGE_INSPECT 2>> stderr.txt || true
 
-      if [ ! -z $(cat stderr.txt) ]; then
-        echo "Skopeo inspect failed, inspect-image test encountered the following error:"
-        cat stderr.txt
-        note="Task $(context.task.name) failed: Encountered errors while inspecting image. For details, check Tekton task log."
-        TEST_OUTPUT=$(make_result_json -r ERROR -t "$note")
-        echo "${TEST_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
-        exit 0
+      status=-1
+      max_run=5
+      sleep_sec=10
+      for run in $(seq 1 $max_run); do
+        status=0
+        [ "$run" -gt 1 ] && sleep $sleep_sec  # skip last sleep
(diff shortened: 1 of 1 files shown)


task/inspect-image/0.1/inspect-image.yaml
```diff
@@ -70,24 +70,49 @@ spec:
...
+        $(skopeo inspect --no-tags docker://"${IMAGE_URL}" > $IMAGE_INSPECT)
+        status=$?
+        if [ "$status" -eq 0 ]; then
+          break
+        fi
+      done
+      if [ "$status" -ne 0 ]; then
+          # image may not exist, we cannot recover from this error, return exit error code
+          echo "Failed to inspect image ${IMAGE_URL}"
+          exit 1
```
dirgim, line 89: This does conflict with [ADR 14. to let pipelines proceed](https://redhat-appstudio.github.io/book/ADR/0014-let-pipelines-proceed.html) but I can definitely see the point of not being able to inspect the image being a breaking issue. 
I still think we'll need to discuss it some more with the other RHTAP developers before committing to a change like this.

MartinBasti: I can change it to never-fail, but it gives no sense to me, if image cannot be pulled, all subsequent tasks will fail anyway.

ADR 14 says it about `linting and testing` tasks. IMHO this is neither, it's just task that fetches metadata.

dirgim: Yes, but the linting and testing tasks depend on this image inspection task. I think that the spirit of ADR 14 is to allow for users to configure their Enterprise Contract to not require the built-in linting/testing/inspection.
Failing the inspect task would then block their entire component build irrespective of what EC they have configured.

MartinBasti: okay, I'll update it to neverfailing task

MartinBasti: updated

dirgim (approved):
LGTM

task/inspect-image/0.1/inspect-image.yaml
dirgim: Thanks. We can follow up with further discussion on ADR 14 in the future

task/inspect-image/0.1/inspect-image.yaml
```diff
@@ -70,24 +70,51 @@ spec:
...
-        cat stderr.txt
-        note="Task $(context.task.name) failed: Encountered errors while inspecting image. For details, check Tekton task log."
-        TEST_OUTPUT=$(make_result_json -r ERROR -t "$note")
-        echo "${TEST_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
-        exit 0
+      status=-1
+      max_run=5
+      sleep_sec=10
+      for run in $(seq 1 $max_run); do
+        [ "$run" -gt 1 ] && sleep $sleep_sec  # skip last sleep
```
dheerajodha, line 78: aren't we skipping the first sleep?

task/inspect-image/0.1/inspect-image.yaml
```diff
@@ -70,24 +70,51 @@ spec:
...
+        echo "Inspecting manifest for source image ${IMAGE_URL} (try $run/$max_run)."
+        $(skopeo inspect --no-tags docker://"${IMAGE_URL}" > $IMAGE_INSPECT)
+        status=$?
+        if [ "$status" -eq 0 ]; then
+          break
+        fi
+      done
+      if [ "$status" -ne 0 ]; then
+          note="Failed to inspect image ${IMAGE_URL}"
```
dheerajodha, line 87: wouldn't the Docs people be upset if we change the note?

task/inspect-image/0.1/inspect-image.yaml
MartinBasti: eh, yes and no :D 
it's written like to avoid sleep after last try, but yes it skips first sleep

task/inspect-image/0.1/inspect-image.yaml
```diff
@@ -70,24 +70,51 @@ spec:
...
-        TEST_OUTPUT=$(make_result_json -r ERROR -t "$note")
-        echo "${TEST_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
-        exit 0
+      status=-1
+      max_run=5
+      sleep_sec=10
+      for run in $(seq 1 $max_run); do
+        [ "$run" -gt 1 ] && sleep $sleep_sec  # skip last sleep
+        echo "Inspecting manifest for source image ${IMAGE_URL} (try $run/$max_run)."
+        $(skopeo inspect --no-tags docker://"${IMAGE_URL}" > $IMAGE_INSPECT)
```
dheerajodha, line 80: I read about `$(...)`syntax in Bash which is called command substitution, you now add this. How did substitution work correctly so far without this?

dheerajodha (approved):

dheerajodha (approved):
/lgtm just unsure about my comment on the "note" field

jsztuka (approved):
lgtm

task/inspect-image/0.1/inspect-image.yaml
MartinBasti: There was `|| true` before, so it always was true and never failed.
I used `$()` which runs in subshell so I could collect return error code

task/inspect-image/0.1/inspect-image.yaml
MartinBasti: I'll update it

task/inspect-image/0.1/inspect-image.yaml
dheerajodha: Thanks for the explanation :+1:

MartinBasti: you are right, it's wrong. Tekton by default injects `set -e`
I'm going to rework this

MartinBasti: done

task/inspect-image/0.1/inspect-image.yaml
```diff
@@ -99,13 +122,28 @@ spec:
...
-      skopeo inspect --no-tags docker://$BASE_IMAGE  > $BASE_IMAGE_INSPECT || true
-      echo -n "$BASE_IMAGE" | tee $(results.BASE_IMAGE.path)
+      echo "Detected base image: $BASE_IMAGE"
+      echo -n "$BASE_IMAGE" > $(results.BASE_IMAGE.path)
+
+      for run in $(seq 1 $max_run); do
+        status=0
+        [ "$run" -gt 1 ] && sleep $sleep_sec  # skip last sleep
+        echo "Inspecting base image ${BASE_IMAGE} (try $run/$max_run)."
+        $(skopeo inspect --no-tags "docker://$BASE_IMAGE"  > $BASE_IMAGE_INSPECT) && status=$?; break || status=$?
```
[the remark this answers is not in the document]
MartinBasti: yes, I'm just stupid and cannot do the same 3 times correctly

dirgim (approved):
LGTM

dheerajodha (approved):
/lgtm

chmeliik (approved):

MartinBasti:
rebased

PR #408, opened 2023-05-23
Add icr-scan task to RHTAP

openrefactory:


partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Server-URL
+      description: The URL of the Server where data will be sent
+
+    - name: iCR-Username
+      description: The Username for the iCR Server
+      default: admin
+
+    - name: iCR-Password
+      description: The Password for the iCR Server 
+      default: admin
```
sbose78, line 35: would this actually work ?

partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Git-Revision
+      description: Branch of the Git Repository
+      default: master
+
+    - name: Server-URL
+      description: The URL of the Server where data will be sent
+
+    - name: iCR-Username
+      description: The Username for the iCR Server
+      default: admin
```
sbose78, line 31: Would this actually work?

sbose78 (commented):
Thank you for you submission! :)

partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+      default: admin   
+
+  steps:
+    - name: icr-scan
+      image: docker.io/munahaf/redhat-tekton@sha256:d47534a332af1c410ae9651d728d689d9808bc6dd0f54092cc4252d7288f270f
+
+      resources:
+        requests:
+          memory: 8Gi
+          cpu: 1000m
```
amfred, line 44: These requests are high.  Why do you need 8Gi and a full CPU reserved to start the task?

partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+        limits:
+          memory: 8Gi
+          cpu: 2500m
```
amfred, line 47: I'm also wondering why the CPU limit is so high.  We thought this was calling out to an external service.  Our default/typical limit would be closer to 500m or even less if it's not CPU-heavy.

amfred, line 47: I'm also wondering why the CPU limit is so high.  We thought this was calling out to an external service.  Our default/typical limit would be closer to 500m or even less if it's not CPU-heavy.

partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+      description: The Username for the iCR Server
+      default: admin
+
+    - name: iCR-Password
+      description: The Password for the iCR Server 
+      default: admin   
+
+  steps:
+    - name: icr-scan
+      image: docker.io/munahaf/redhat-tekton@sha256:d47534a332af1c410ae9651d728d689d9808bc6dd0f54092cc4252d7288f270f
```
sbose78, line 39: Is there an openfactory org owned image ?

partners/icr-scan/0.1/icr-scan.yaml
```diff
@@ -0,0 +1,64 @@
...
+      default: master
+
+    - name: Server-URL
+      description: The URL of the Server where data will be sent
+
+    - name: iCR-Username
+      description: The Username for the iCR Server
+      default: admin
+
+    - name: iCR-Password
```
MartinBasti, line 33: isn't this a security flaw to pass password via parameters? Shouldn't ba password stored in Openshift Secrets and  only secret referenced here?

lcarva: BTW parameters appear in clear text on the image attestation

partners/icr-scan/0.1/icr-scan.yaml
openrefactory: This is just a placeholder, the default password and username for any icr server is admin. We expect that someone who will use the task will specify the username and password using encryption instead of plain text.

partners/icr-scan/0.1/icr-scan.yaml
openrefactory: The analysis does the heavy work and we are over provisioning, but definitiely there is some significant resource that will be needed. We had a recent conversation with Brian Cook in which he is suggesting some other way to provision it. We will make that change soon.

partners/icr-scan/0.1/icr-scan.yaml
openrefactory: This is just a placeholder. We expect that the users will use secrets instead of plaintext

MartinBasti:
Can we close this PR?

chmeliik:
Abandoned

PR #409, opened 2023-05-23
Hardcode the default "snyk-secret"

sbose78:
* The name of the secret would always be `snyk-secret`
image

* The volume is set to 'optional: true' which means the absence of the `secret` wouldn't cause a failure ?

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -200,17 +200,11 @@ spec:
       - input: $(params.skip-checks)
         operator: in
         values: ["false"]
-      - input: $(params.snyk-secret)
-        operator: notin
-        values: [""]
       runAfter:
         - clone-repository
       taskRef:
         name: sast-snyk-check
         version: "0.1"
-      params:
-        - name: SNYK_SECRET
-          value: $(params.snyk-secret)
       workspaces:
       - name: workspace
         workspace: workspace
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -50,10 +50,6 @@ spec:
       name: java
       type: string
       default: "false"
-    - description: Snyk Token Secret Name
-      name: snyk-secret
-      type: string
-      default: ""
     - description: Image tag expiration time, time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
       name: image-expires-after
       default: ""
--- a/task/sast-snyk-check/0.1/sast-snyk-check.yaml
+++ b/task/sast-snyk-check/0.1/sast-snyk-check.yaml
@@ -46,7 +46,7 @@ spec:
         if [[ -z $SNYK_TOKEN ]]; then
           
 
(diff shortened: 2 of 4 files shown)


task/sast-snyk-check/0.1/sast-snyk-check.yaml
```diff
@@ -16,7 +16,7 @@ spec:
   params:
     - name: SNYK_SECRET
       description: Name of secret which contains Snyk token.
-      default: ""
+      default: snyk-secret
```
enarha, line 19: maybe snyk-shared-secret?

enarha:
I guess we can also remove the pipelines/core-services/snyk-secret.yaml patch?

dirgim:
If we want to always attempt to run the `sast-synk-check` task, I think we will need to silence the error message that happens when trying to read the secret [here](https://github.com/redhat-appstudio/build-definitions/blob/c31f6ad4939f52b30b10331097a21a0d03ca1745/task/sast-snyk-check/0.1/sast-snyk-check.yaml#L45) because that seemed to be a point of confusion for users.

sbose78:
@dirgim Maybe we should update the message here to reflect that this is an opt-in
"echo "SNYK_TOKEN not defined. Create $SNYK_SECRET secret with key containing Snyk token." | tee stdout.txt"

sbose78:
This also means we'd have to update the name of the secret used by the services on infra-deployments ?

sbose78:
To make things easier, I added the name of the secret "snyk-secret" in the message.

task/sast-snyk-check/0.1/sast-snyk-check.yaml
```diff
@@ -44,7 +44,7 @@ spec:
         . /utils.sh
         SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
         if [[ -z $SNYK_TOKEN ]]; then
-          echo "SNYK_TOKEN not defined. Create $SNYK_SECRET secret with key containing Snyk token." | tee stdout.txt
+          echo "If you wish to use the Snyk code SAST task, please create a secret named snyk-secret with the key containing the Snyk token." | tee stdout.txt
```
Michkov, line 47: please provide exact key name.

Michkov (changes requested):
Should be also https://github.com/redhat-appstudio/build-definitions/blob/a383e661bbf64c1500efc1cfe2b01f4936ec2a8b/pipelines/template-build/template-build.yaml#L53 updated or removed?

pipelines/template-build/template-build.yaml
```diff
@@ -50,10 +50,6 @@ spec:
       name: java
       type: string
       default: "false"
-    - description: Snyk Token Secret Name
-      name: snyk-secret
-      type: string
-      default: ""
```
Michkov, line 56: With deletion of this you need to delete references -> lines 203 and 213.

Michkov (approved):

dirgim (approved):
LGTM

PR #413, opened 2023-05-29
STONEBLD-1320: Remove pipelineRun secret

Michkov:
Remove secret which was used for merging of secret generated by tekton in `~/.docker/config.json` and shared secret for accessing user-workload repository. User-workload repository has been replaced by image-controller. `~/.docker/config.json` provided by tekton should be used directly so that we can later remove create/update/delete permission on the appstudio-pipeline serviceaccount.


dirgim:
@Michkov the changes look good - does the [README.md for clamav-scan](https://github.com/redhat-appstudio/build-definitions/blob/main/task/clamav-scan/0.1/README.md) also need to be modified?

PR #414, opened 2023-05-31
STONEBLD-1250: fix permission on /var/lib/containers

Michkov:
Some images are having issue with custom group permission on volume created by emptyDir, updating folder group to root.

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -120,6 +120,8 @@ spec:
         touch /var/lib/containers/java
       fi
 
+      # Fixing group permission on /var/lib/containers
+      chown root:root /var/lib/containers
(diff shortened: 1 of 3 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -120,6 +120,8 @@ spec:
         touch /var/lib/containers/java
       fi
 
+      # Fixing grop permission on /var/lib/containers
```
psturc, line 123:
```suggestion
      # Fixing group permission on /var/lib/containers
```

Michkov: updated thanks

psturc (approved):

mmorhun (approved):

PR #416, opened 2023-05-31
Fix the tkn-bundle pipeline

lcarva:
The build-container task moved from 4th to 3rd index.


joejstuart (approved):

lcarva:
I don't think the e2e test failure is related, but retesting just in case.

PR #421, opened 2023-06-01
chore(renovate): improve renovate experience

MartinBasti:
* Adding grouping per team and shared images.
* Reviewers will be assigned by renovatebot automatically.
* Shared images will be updated only twice a week to prevent too many updates in PR and noise

Note: it's not possible to add rule "*" to match everything else, that's why all images/prefixes are enumerated

Note: if there is no matching rule, a separate PR will be created

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,8 +10,47 @@
   },
   "packageRules": [
     {
-      "matchPackagePatterns": ["*"],
-      "groupName": "all"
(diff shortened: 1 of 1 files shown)


lcarva (approved):
:tada:

dirgim (approved):
LGTM

zregvart (approved):

PR #424, opened 2023-06-02
setting aside more resources for buildah

amfred:
We're hoping that by setting aside more cpu resources for buildah tasks, the scheduler will a) distribute them more evenly across nodes and b) auto-scale to more nodes sooner. The hope is that this would improve our ability to handle spikes of concurrent PipelineRuns by recruiting more compute resources.

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -91,8 +91,8 @@ spec:
         memory: 4Gi
         cpu: 2
       requests:
-        memory: 512Mi
-        cpu: 10m
+        memory: 1Gi
+        cpu: 250m
     script: |
       if [ -e "$CONTEXT/$DOCKERFILE" ]; then
         dockerfile_path="$CONTEXT/$DOCKERFILE"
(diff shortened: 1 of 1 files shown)


amfred:
See also: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1685704688375289

task/buildah/0.1/buildah.yaml
```diff
@@ -92,7 +92,7 @@ spec:
         cpu: 2
       requests:
         memory: 512Mi
-        cpu: 10m
+        cpu: 250m
```
gabemontero, line 95: In a bit of cross referencing with other tekton based projects that use builds, like the shipwright project that @sbose78 @adambkaplan and I used to be more involved with, they are also setting memory reqeusts to 250m

See https://github.com/shipwright-io/build/blob/f76244975148458a4afb4b3bd2d855710c29281a/samples/buildstrategy/buildah/buildstrategy_buildah_shipwright_managed_push_cr.yaml#L175

IBM's used of shipwright, with tekton under the covers, is also a similar service based offering which they have been running at scale.

Hence I would consider this value vetted even beyond what @amfred and the performance folks saw today.

And when comparing 250m vs. the 2 full core limit, that is still  a conservative start IMO.

Let's go with this.

gabemontero (approved):
/lgtm

jhutar (approved):

Michkov (approved):

PR #425, opened 2023-06-06
STONEBLD-29: add slack-webhook-notification

Michkov:
Add new task for sending messages via slack using webhooks. Integrate the task with core-services pipeline so that teams can be notified when their push pipelinerun failed.

### DIFF
--- a/pipelines/core-services/slack-notification.yaml
+++ b/pipelines/core-services/slack-notification.yaml
@@ -1,0 +1,24 @@
+- op: add
+  path: /spec/params/-
+  value:
+    name: slack-webhook-notification-team
+    default: ""
+- op: add
+  path: /spec/finally/-
+  value:
+    name: slack-webhook-notification
+    taskRef:
+      name: slack-webhook-notification
+      version: "0.1"
+    when:
+    - input: $(params.slack-webhook-notification-team)
+      operator: notin
+      values: [""]
+    - input: $(tasks.status)
+      operator: in
+      values: ["Failed"]
+    params:
+    - name: message
+      value: Tekton pipelineRun $(context.pipelineRun.name) failed
+    - name: key-name
+      value: $(params.slack-webhook-notification-team)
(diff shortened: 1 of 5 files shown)


task/slack-webhook-notification/0.1/README.md
```diff
@@ -0,0 +1,11 @@
+# slack-webhook-notification task
+
+Sends message to slack using incoming webhook
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|message|Message to be sent||true|
+|secret-name|Secret which should contain at least one key where the value is webhook URL for slack.|slack-webhook-notification-secret|false|
```
tkdchen, line 9: Does the content look like:
```
key1=value1
key2=value2
...
```

But from the curl command inside task script, it doesn't.

tkdchen (approved):

task/slack-webhook-notification/0.1/slack-webhook-notification.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: slack-webhook-notification
+spec:
+  description: >-
+    Sends message to slack using incoming webhook
+  params:
+    - name: message
+      description: Message to be sent
+    - name: secret-name
+      description: |
+        Secret with least one key where value is webhook URL for slack.
```
psturc, line 18:
```suggestion
        Secret with at least one key where value is webhook URL for slack.
```

pipelines/core-services/slack-notification.yaml
```diff
@@ -0,0 +1,24 @@
...
+    when:
+    - input: $(params.slack-webhook-notification-team)
+      operator: notin
+      values: [""]
+    - input: $(tasks.status)
+      operator: in
+      values: ["Failed"]
+    params:
+    - name: message
+      value: Tekton pipelineRun $(context.pipelineRun.name) failed
```
psturc, line 22: would it make sense to add more useful information, like the namespace where the pipelinerun failed? or would this be enough for people to know where to look when they got this message

Michkov: that's tricky since we do not have much information inside the pipeline, I can add namespace but hardcoding url is probably not right.

PR #430, opened 2023-06-08
Create variations of the ec pipeline

lcarva:
Copies of the enterprise-contract.yaml are now available with a different default value for the POLICY_CONFIGURATION parameter. This is a workaround until the UI can provide sufficient controls to allow users to provide parameters.

We cannot use kustomize to make these variations because these pipelines will be used with Tekton's git resolver. This requires the pipeline definition to appear in full in git.

https://issues.redhat.com/browse/HACBS-2294


pipelines/enterprise-contract-everything.yaml
```diff
@@ -0,0 +1,65 @@
+# This is a copy of enterprise-contract.yaml with a different default value for the
+# POLICY_CONFIGURATION parameter. This is a workaround until the UI can provide sufficient
+# controls to allow users to provide parameters. https://issues.redhat.com/browse/HACBS-2294
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: enterprise-contract-all
```
joejstuart, line 7: Should this match the filename?

joejstuart (approved):

pipelines/enterprise-contract-everything.yaml
lcarva: Hawk eye! Fixed.

PR #432, opened 2023-06-13
Update images of init and summary

Michkov:
Select smaller images which provides tools which are needed.

As bonus "fix" [RHTAPBUGS-397](https://issues.redhat.com//browse/RHTAPBUGS-397)


flacatus (approved):

flacatus:
Maybe minor problem i see here is that are 4.13 images. Otherwise LGTM to me

PR #435, opened 2023-06-14
STONEBLD-1305: Example of partner task

Michkov:
Add example partner task and document way how to include in user's pipelinerun.


Michkov:
I am following current implementation, tests and README.md uses structure `partners/<task_name>/<task_version>/<task_name>.yaml`. Change of this is out-of-scope for this Pull-request.

PR #442, opened 2023-06-21
feat: Try to increase requests here to improve situation about unfair scheduling these pods

jhutar:
Related Slack discussion: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1686211516841129


jhutar:
We have https://github.com/redhat-appstudio/build-definitions/pull/520

PR #443, opened 2023-06-21
Add temporary timestamps to heviest tasks

jhutar:
This is how basic pipelines can look like on Stage:

image

We can see `buildah` is a longest task. For advanced pipelines, `clamav` is also very long running.

That is why I'm trying to add these timestamps - to get more insight into why.

Slightly related Slack thread: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1686211516841129


task/buildah/0.1/buildah.yaml
```diff
@@ -159,23 +165,30 @@ spec:
...
         $VOLUME_MOUNTS \
         $BUILDAH_ARGS \
         ${LABELS[@]} \
         --tls-verify=$TLSVERIFY --no-cache \
         --ulimit nofile=4096:4096 \
+        --log-level=debug \
         -f "$dockerfile_path" -t $IMAGE $CONTEXT
 
+      echo "$( date --utc -Ins ) Start buildah from"
       container=$(buildah from --pull-never $IMAGE)
```
gabemontero, line 179: need to add a `--log-level=debug` for the `buildah from` invocation as well .... in the slowest build from the last run in stage @jhutar I saw the time needed for this increase from 1 second with the fastest build to 10 seconds with the slowest build

gabemontero: I suspect it is disk IO related like the other delays I've seen, but it would be good to confirm

jhutar: Oh, right! Adding and rerunning

PR #445, opened 2023-06-22
Add memory and CPU request to git-clone task

stuartwdouglas:
This is the first task that is run using the PVC, so when this task is scheduled to an AZ all other tasks in the PipelineRun that use the workspace will be scheduled to the same AZ.

If this task has no resource requests the scheduler can over assign tasks to the same AZ.

### DIFF
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -124,7 +124,10 @@ spec:
       value: $(workspaces.basic-auth.path)
     image: $(params.gitInitImage)
     name: clone
-    resources: {}
+    resources: #we request the same amount of memory as the buildah task, as this task is run first it will determine the AZ the pipeline runs in
+      requests:
+        memory: 512Mi
+        cpu: 250m
     securityContext:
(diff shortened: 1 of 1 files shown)


sbose78 (approved):

stuartwdouglas:
I don't think this failure is related, but I think it is a real problem. I just imported a sample in the UI and it created multiple components, so 'Expected length of the detected Components was not 1' for component detection could indicate the underlying problem.

Michkov:
Yop, strange, there is really CDQ with three components detected...
/retest

I would like to have this change validated by performance team.

For testing this should be created before creating component in the namespace:
```
apiVersion: appstudio.redhat.com/v1alpha1
kind: BuildPipelineSelector
metadata:
  name: build-pipeline-selector
spec:
  selectors:
    - name: all-pipelines
      pipelineRef:
        name: docker-build
        bundle: quay.io/redhat-appstudio-tekton-catalog/pull-request-builds:docker-build-dd8dd0bf734d23d51172bbd86364ac1a6b78d724
```

PR #449, opened 2023-06-26
Rename step to sbom-syft-generate

Michkov:
Rename sbom-get to sbom-syft-generate to cleary name the step based on what it's doing.


tisutisu (approved):
LGTM

PR #458, opened 2023-07-03
Add check for symlinks pointing outside of repo

mkosiarc:
If the check is enabled and the cloned repository contains symlinks pointing outside of the repo, the build will fail.

[STONEBLD-1328](https://issues.redhat.com//browse/STONEBLD-1328)

### DIFF
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -75,13 +75,19 @@ spec:
       the gitInitImage param with an image containing custom user configuration.
     name: userHome
     type: string
+  - default: "true"
+    description: |
+      Check symlinks in the repo. If they're pointing outside of the repo, the build will fail.
+    name: enableSymlinkCheck
+    type: string
   results:
   - description: The precise commit SHA that was fetched by this Task.
     name: commit
   - description: The precise URL that was fetched by this Task.
     name: url
   steps:
-  - env:
+  - name: clone
+    env:
     - name: HOME
       value: $(params.userHome)
     - name: PARAM_URL
@@ -123,7 +129,6 @@ spec:
     - name: WORKSPACE_BASIC_AUTH_DIRECTORY_PATH
       value: $(workspaces.basic-auth.path)
     image: $(params.gitInitImage)
-    name: clone
     resources: {}
     securityContext:
       runAsUser: 0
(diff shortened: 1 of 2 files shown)


task/git-clone/0.1/git-clone.yaml
```diff
@@ -188,6 +207,15 @@ spec:
       if [ "${EXIT_CODE}" != 0 ] ; then
         exit "${EXIT_CODE}"
       fi
+
+      if [ "${PARAM_ENABLE_SYMLINK_CHECK}" = "true" ] ; then
+        check_symlinks
+        EXIT_CODE="$?"
+        if [ "${EXIT_CODE}" != 0 ] ; then
+          exit "${EXIT_CODE}"
+        fi
```
MartinBasti, line 216: Isn't this behavior alredy enforced by
```
set -eu
```

MartinBasti: and because you are using pipe in your code, it should be probably extended to
```
set -euo pipefail
```

task/git-clone/0.1/git-clone.yaml
```diff
@@ -165,6 +172,18 @@ spec:
         fi
       }
 
+      check_symlinks() {
+        find $CHECKOUT_DIR -type l -print | \
+          while read symlink
+          do
+            target=$(readlink -f "$symlink")
+            if ! [[ "$target" =~ ^$CHECKOUT_DIR ]]; then
+              echo "The cloned repository contains symlink pointing outside of the cloned repository: $symlink"
```
Michkov, line 181: would be better to aggregate all symlinks pointing outside and print all of them not just first one.

task/git-clone/0.1/git-clone.yaml
```diff
@@ -165,6 +172,18 @@ spec:
         fi
       }
 
+      check_symlinks() {
+        find $CHECKOUT_DIR -type l -print | \
```
tisutisu, line 176: Parent image does not contain the find command .. so this command fails
`+ find /workspace/output/ -type l -print
/tekton/scripts/script-0-7m9xq: line 39: find: command not found`

task/git-clone/0.1/git-clone.yaml
mkosiarc: changed

task/git-clone/0.1/git-clone.yaml
mkosiarc: Updated code and changed sh to bash so I can use process substitution.

task/git-clone/0.1/git-clone.yaml
mkosiarc: now the check is in different step that uses different image that has the find command available.

mkosiarc:
tested this with the following repo contaiting two symlinks pointing outside of the directory and one valid symlink https://github.com/mkosiarc-rhtap/symlink-check-test

mkosiarc:
> It's hard to review with all the indentation changes

I am aware, but since I added another step I had to indent it another level

Michkov:
> > It's hard to review with all the indentation changes
> 
> I am aware, but since I added another step I had to indent it another level

I do not agree, you have not added another level, just added new item to list. If you want to update indentation then please do that in separate commit.

mkosiarc:
> > > It's hard to review with all the indentation changes
> > 
> > 
> > I am aware, but since I added another step I had to indent it another level
> 
> I do not agree, you have not added another level, just added new item to list. If you want to update indentation then please do that in separate commit.

You are right, I misused my local linter. I apologize for the confusion and now the commit is updated.

task/git-clone/0.1/git-clone.yaml
```diff
@@ -190,6 +195,42 @@ spec:
       fi
       printf "%s" "${RESULT_SHA}" > "$(results.commit.path)"
       printf "%s" "${PARAM_URL}" > "$(results.url.path)"
+  - name: symlink-check
+    image: registry.redhat.io/ubi8:8.8-1009
```
Michkov, line 199: probably better to use ubi9.

task/git-clone/0.1/git-clone.yaml
```diff
@@ -190,6 +195,42 @@ spec:
...
+    env:
+    - name: PARAM_ENABLE_SYMLINK_CHECK
+      value: $(params.enableSymlinkCheck)
+    - name: PARAM_SUBDIRECTORY
+      value: $(params.subdirectory)
+    - name: WORKSPACE_OUTPUT_PATH
+      value: $(workspaces.output.path)
+    resources: {}
+    securityContext:
+      runAsUser: 0
```
Michkov, line 209: nit: this is not needed since ubi is using root(uid 0) by default.

task/git-clone/0.1/git-clone.yaml
mkosiarc: changed

task/git-clone/0.1/git-clone.yaml
mkosiarc: changed

Michkov (approved):

PR #464, opened 2023-07-10
RHTAPBUGS-464,STONEBLD-1494: unify memory usage

Michkov:
Set memory request and limit to same value to ensure that scheduled node has enough memory.

Make memory of build task configurable via PipelineRun parameter `build_task_memory_request`


Michkov:
This approach does not work there is different way to configure memory at runtime https://github.com/tektoncd/community/blob/main/teps/0094-configuring-resources-at-runtime.md

RHTAPBUGS-464 part is in #466

PR #466, opened 2023-07-11
RHTAPBUGS-464: sync memory limit and request

Michkov:
Avoid issue with failing build because of running on node with enough memory availably.


tkdchen (approved):
LGTM

PR #473, opened 2023-07-13
prefetch-dependencies: use semver for cachi2 image

chmeliik:
STONEBLD-1270

The cachi2 image gets rebuilt almost everyday, but the changes are often not relevant. To prevent renovatebot from spamming the repository with irrelevant version bumps, use the recently released 0.1.0 semantic version. Renovatebot should now bump cachi2 only when it gets a new release.


MartinBasti (approved):
<3

PR #481, opened 2023-07-19
Merge Cachi2's SBOM components into the final SBOM

brunoapimentel:
Adds a new step to the buildah task that will insert the components reported by Cachi2 into the SBOM generated by Syft, while also performing deduplication.

See the [merge script](https://github.com/containerbuildsystem/cachi2/blob/main/utils/merge_syft_sbom.py) in the Cachi2 repo for more info.

[STONEBLD-1357](https://issues.redhat.com//browse/STONEBLD-1357)

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -208,12 +208,12 @@ spec:
       name: varlibcontainers
     securityContext:
       runAsUser: 0
+
   - name: merge-syft-sboms
     image: registry.access.redhat.com/ubi9/python-39:1-125
     script: |
       #!/bin/python3
       import json
-      import os
 
       # load SBOMs
       with open("./sbom-image.json") as f:
@@ -238,14 +238,6 @@ spec:
       # write the CycloneDX unified SBOM
       with open("./sbom-cyclonedx.json", "w") as f:
         json.dump(image_sbom, f, indent=4)
-
-      # create and write the PURL unified SBOM
-      purls = [{"purl": component["purl"]} for component in image_sbom["components"] if "purl" in component]
-      purl_content = {"image_contents": {"dependencies": purls}}
-
-      with open("sbom-purl.json", "w") as output_file:
-        json.dump(purl_content, output_file, indent=4)
-
     workingDir: $(workspaces.source.path)
     securityContext:
       runAsUser: 0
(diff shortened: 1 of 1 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -146,6 +146,7 @@ spec:
       if [ -n "${PREFETCH_INPUT}" ]; then
         mv cachi2 /tmp/
         chmod -R go+rwX /tmp/cachi2
+        cp /tmp/cachi2/output/bom.json ./sbom-cachi2.json
```
chmeliik, line 149: Which directory is `.`? Are we putting the SBOM in the root of the user's repo?

The user could accidentally `COPY . /foo` the SBOM into their image, can we avoid that?

task/buildah/0.1/buildah.yaml
```diff
@@ -233,14 +234,38 @@ spec:
...
 
-      # create and write the PURL unified SBOM
-      purls = [{"purl": component["purl"]} for component in image_sbom["components"] if "purl" in component]
+  - name: merge-cachi2-sbom
+    image: quay.io/redhat-appstudio/cachi2:0.2.0@sha256:4bbc8bb086522568c76aa71a0d0c0a1a891ed5a683ffb1e3d696f3a02914ecaa
+    script: |
+      if [ -n "${PREFETCH_INPUT}" ]; then
+        echo "Merging contents of sbom-cachi2.json into sbom-cyclonedx.json"
+        /src/utils/merge_syft_sbom.py sbom-cachi2.json sbom-cyclonedx.json > sbom-temp.json
+        mv sbom-temp.json sbom-cyclonedx.json
```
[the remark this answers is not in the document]
chmeliik: Ok, it does https://tekton.dev/docs/pipelines/tasks/#running-scripts-within-steps

But I'd feel safer with an explicit `#!/usr/bin/env bash` + `set -o errexit -o nounset -o pipefail`

task/buildah/0.1/buildah.yaml
brunoapimentel: `.` is the [workspace](https://github.com/brunoapimentel/build-definitions/blob/merge-cachi2-sbom/task/buildah/0.1/buildah.yaml#L251) that gets mounted in the image (same one that got the cloned sources).

Note that this workspace is not added to the final image, which is already built at this point. This is [how](https://github.com/brunoapimentel/build-definitions/blob/merge-cachi2-sbom/task/buildah/0.1/buildah.yaml#L283) we inject the SBOMs to the image (to the usual `/root/buildinfo/content_manifests`)

The SBOM is also being [pushed](https://github.com/brunoapimentel/build-definitions/blob/merge-cachi2-sbom/task/buildah/0.1/buildah.yaml#L303-L313) to the registry.

task/buildah/0.1/buildah.yaml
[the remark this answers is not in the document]
brunoapimentel: I don't mind setting the shebang and options, but I noticed the other scripts in this Task are not setting either of them. Should we make this the odd one?

task/buildah/0.1/buildah.yaml
chmeliik: This part is before the build, `buildah build` happens at the end of this same script (though the build command is hidden in lots of `unshare` dark magic)

brunoapimentel: Oh, good catch, I confused this with the `merge-cachi2-sbom` step.

I believe the `source` workspace is the only way to persist the content between steps as of now. I can try to make this copy after the build is done, it should solve the issue.

lkolacek (approved):
LGTM :+1:

chmeliik (approved):
LGTM :+1:

tkdchen (approved):

PR #482, opened 2023-07-20
STONEBLD-1494: add variation of buildah with different memory

Michkov:
Use kustomize to allow generate multiple buildah task version with different memory configuration.

### DIFF
--- a/.tekton/scripts/check-task-pipeline-bundle-repos.sh
+++ b/.tekton/scripts/check-task-pipeline-bundle-repos.sh
@@ -21,9 +21,12 @@ has_missing_repo=
 echo "Checking existence of task and pipeline bundle repositories ..."
 
 # tasks
-for task_file in task/*/*/*.yaml; do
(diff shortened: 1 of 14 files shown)


zregvart (approved):
LGTM (the e0eb412fcbb7eac82dc39a707dce2f93450ecaa5 bit)

gabemontero (approved):
So just to document in comment, we are going with
- default 4Gi
- then can set 6, 8, or 10 Gi

/lgtm

@Michkov @gbenhaim even it this is "short term" until https://issues.redhat.com/browse/PLNSRVCE-1368 , in support of this PR, does it make sense to get these options documented over in say https://redhat-appstudio.github.io/infra-deployments/components/build-service/README.html or wherever is more appropriate?

If documenting is already considered part of https://issues.redhat.com/browse/STONEBLD-1494 apologies I couldn't tell.

Thanks.

tkdchen (approved):

PR #484, opened 2023-07-21
fix(hack): link to right service account

MartinBasti:
There was a change from `pipeline` to `appstudio-pipeline`. Reflecting this change in hack script.

Without this change, custom secret is not linked to PLRs.

### DIFF
--- a/hack/test-build.sh
+++ b/hack/test-build.sh
@@ -27,10 +27,10 @@ if [ -z "$MY_QUAY_USER" ]; then
   echo MY_QUAY_USER env variable is not set, pushing to $IMG
 else
   if oc get secret redhat-appstudio-staginguser-pull-secret &>/dev/null; then
-     # Ensure that the pipeline service account has access to the secret. Although the
+     # Ensure that the appstudio-pipeline service account has access to the secret. Although the
      # secret is mounted directly on the pipeline, Tekton Chains needs this linkage so
      # it can push the image signature and attestation to the same OCI repository.
-     oc secrets link pipeline redhat-appstudio-staginguser-pull-secret
+     oc secrets link appstudio-pipeline redhat-appstudio-staginguser-pull-secret
   else
      echo "Secret redhat-appstudio-staginguser-pull-secret is not created, can be created by:"
      echo "Docker:"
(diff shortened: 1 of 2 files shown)


Michkov:
thinking if we should do this in smart way, basically the build-definitions pipelines are independent, they are executable on cluster with tekton, where default SA for pipelineruns is pipeline

MartinBasti:
So something like this?

```diff
diff --git a/hack/test-build.sh b/hack/test-build.sh
index d996032..f98fb62 100755
--- a/hack/test-build.sh
+++ b/hack/test-build.sh
@@ -50,4 +50,4 @@ if [ "$SKIP_CHECKS" == "1" ]; then
   SKIP_CHECKS_PARAM="-p skip-checks=true"
 fi
 
-tkn pipeline start $PIPELINE_NAME -w name=workspace,volumeClaimTemplateFile=$SCRIPTDIR/test-build/workspace-template.yaml $SKIP_CHECKS_PARAM -p git-url=$GITREPO -p output-image=$IMG --use-param-defaults
+tkn pipeline start $PIPELINE_NAME -s pipeline -w name=workspace,volumeClaimTemplateFile=$SCRIPTDIR/test-build/workspace-template.yaml $SKIP_CHECKS_PARAM -p git-url=$GITREPO -p output-image=$IMG --use-param-defaults

```


I'm not sure though, infra-deployment prepare cluster for the new SA, is it a good idea to use different SA? It may have different permissions and on hardened systems it's not even created.

MartinBasti:
^ @Michkov WDYT?

Michkov:
The issue is that 'infra-deployments' does not create proper serviceaccount. It's handled by host-member operator, so the appstudio-pipeline SA is created only for namespaces created via registration service.
So when we create eg test namespace `oc new-project test` then we do not have pipeline or appstudio-pipeline service account.

So we can maybe call https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/build/setup-namespace.sh and then use the appstudio-pipeline SA.

Michkov (approved):

jsztuka:
Rebase and lets merge this one.

jsztuka (approved):

PR #487, opened 2023-07-21
feat(STONEINTG-503): check for trusted OPM images

MartinBasti:
fbc-validation now checks for base image source.
If image is not in list of approved images, validation will fail.

### DIFF
--- a/pipelines/fbc-builder/patch.yaml
+++ b/pipelines/fbc-builder/patch.yaml
@@ -73,3 +73,8 @@
   value:
   - name: POLICY_NAMESPACE
     value: fbc_checks
+- op: add
+  path: /spec/tasks/12/params/-
+  value:
+    name: BASE_IMAGE
+    value: $(tasks.inspect-image.results.BASE_IMAGE)
--- a/task/fbc-validation/0.1/fbc-validation.yaml
+++ b/task/fbc-validation/0.1/fbc-validation.yaml
@@ -16,6 +16,8 @@ spec:
       description: Fully qualified image name.
     - name: IMAGE_DIGEST
       description: Image digest.
+    - name: BASE_IMAGE
+      description: Fully qualified base image name.
   results:
     - name: TEST_OUTPUT
       description: Tekton task test output.
@@ -30,6 +32,8 @@ spec:
           value: $(params.IMAGE_URL)
         - name: IMAGE_DIGEST
           value: $(params.IMAGE_DIGEST)
+        - name: BASE_IMAGE
+          value: "$(params.BASE_IMAGE)"
       securityContext:
         runAsUser: 0
         capabilities:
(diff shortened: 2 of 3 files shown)


MartinBasti:
example:
```
STEP-EXTRACT-AND-CHECK-BINARIES

Base image {quay.io/operator-framework/opm@sha256:0b10c3c3f713f11c805505e0aa686ce2a347b814814c5e110a084903e513babe} is not allowed for the file based catalog image. Allowed images: registry.redhat.io/openshift4/ose-operator-registry
{"result":"FAILURE","timestamp":"1689956221","note":"Task fbc-validation failed: Base image {quay.io/operator-framework/opm@sha256:0b10c3c3f713f11c805505e0aa686ce2a347b814814c5e110a084903e513babe} is not allowed for the file based catalog image. For details, check Tekton task logs","namespace":"default","successes":0,"failures":1,"warnings":0}
```

pipelines/fbc-builder/patch.yaml
```diff
@@ -73,3 +73,8 @@
   value:
   - name: POLICY_NAMESPACE
     value: fbc_checks
+- op: add
```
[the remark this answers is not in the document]
MartinBasti: `-` was needed

lcarva:
@MartinBasti, should this be done by EC? I think all the information is already available there.

MartinBasti:
> @MartinBasti, should this be done by EC? I think all the information is already available there.

Yes, but no. This is first part of bigger epic, we have to validate if opm comes from the righ repo, because we want to extract it from base-image and use it for validation.

MartinBasti:
@arewm Shouldn't be also this image trusted `quay.io/operator-framework/opm` ?

arewm:
Until it is requested by some user, we will continue to not consider the operator-framework images as being "trusted."

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -47,6 +51,37 @@ spec:
         set -o pipefail
         source /utils.sh
 
+        declare -a ALLOWED_BASE_IMAGES=("registry.redhat.io/openshift4/ose-operator-registry")
```
MartinBasti, line 54: JFTR, I still don't like fact that those images should be hardcoded here

MartinBasti: I was suggesting to have it as parameter, this can be default value. So if users really want to, they can update it.

MartinBasti: but on the other hand, malicous image could be an attack vector, because we are going to run opm from the image in this task

MartinBasti:
https://github.com/Michkov/simple-fbc/pull/1 needs this fix for successful run

task/fbc-validation/0.1/fbc-validation.yaml
arewm: Right -- that attack vector is why the proposal was made. It is unfortunate that this required a Red Hat authorization to be able to run, but if we hear feedback for wanting to build without the authentication then we can consider broadening the restriction.

jsztuka (approved):
lgtm

dirgim (approved):
LGTM, just needs a rebase

PR #499, opened 2023-07-28
feat(stoneintg-507): use base image opm binary for fbc validation

14rcole:
After verifying that the base image is trusted (see [STONEINTG-503](https://issues.redhat.com//browse/STONEINTG-503)) we need to use the opm binary from the base image to run fbc validation. Because Tekton tasks do not easily support running nested containers, this change extracts the base image and use the opm binary within it to validate the target image.

### DIFF
--- a/task/fbc-validation/0.1/fbc-validation.yaml
+++ b/task/fbc-validation/0.1/fbc-validation.yaml
@@ -121,27 +121,45 @@ spec:
         # copy content of conffolder to confdir - will be used in next task - related image check
         cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        EXTRACT_DIR="/extracted_base_img"
+        mkdir "${EXTRACT_DIR}"
+        if ! oc image extract ${BASE_IMAGE} --path /:"${EXTRACT_DIR}"; then
+          echo "Unable to extract opm binary"
+          note="Task $(context.task.name) failed: Failed to extract base image with oc extract command, so it cannot validate extracted binaries.  For details, check Tekton task log."
+          ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+          exit 0
+        fi
+
+        OPM_BINARIES="$(find "${EXTRACT_DIR}" -type f -name opm)"
+        BINARIES_COUNT=$(wc -l <<< "${OPM_BINARIES}")
+        if [[ $BINARIES_COUNT -ne "1" ]]; then
+            note="Task $(context.task.name) failed: Expected exactly one opm binary in base image.  For details, check Tekton task log"
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            echo "found $BINARIES_COUNT opm binaries:"
+            echo "${OPM_BINARIES}"
+            exit 0
+        fi
+        OPM_BINARY=$(echo "${OPM_BINARIES}" | head -n 1)
+        echo "OPM_BINARY: '${OPM_BINARY}'"
+        chmod 775 "$OPM_BINARY"
+
+        # We have totally 3 checks here currently
+        check_num=3
         failure_num=0
         TESTPASSED=true
 
-        if [[ ! $(find . -name "opm") ]]; then
-          echo "!FAILURE! - opm binary presence check failed."
-          failure_num=`expr $failure_num + 1`
-          TESTPASSED=false
-        fi
         if [[ ! $(find . -name "grpc_health_probe") ]]; then
           echo "!FAILURE! - grpc_health_probe binary presence check failed."
           failure_num=`expr $failure_num + 1`
           TESTPASSED=false
         fi
-        if ! opm validate ."${conffolder}"; then
+        if ! ${OPM_BINARY} validate ."${conffolder}"; then
           echo "!FAILURE! - opm validate check failed."
           failure_num=`expr $failure_num + 1`
           TESTPASSED=false
         fi
-        if ! opm render ."${conffolder}" | jq -en 'reduce (inputs | select(.schema == "olm.package")) as $obj (0; .+1) == 1'; then
+        if ! ${OPM_BINARY} render ."${conffolder}" | jq -en 'reduce (inputs | select(.schema == "olm.package")) as $obj (0; .+1) == 1'; then
           echo "!FAILURE! - More than one olm.packages is not permitted in a FBC fragment."
           failure_num=`expr $failure_num + 1`
           TESTPASSED=false


task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -121,27 +121,32 @@ spec:
         # copy content of conffolder to confdir - will be used in next task - related image check
         cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        EXTRACT_DIR="/extracted_base_img"
+        mkdir "${EXTRACT_DIR}"
+        oc image extract ${BASE_IMAGE} "${EXTRACT_DIR}"
+        OPM_BINARIES="$(find ~+${EXTRACT_DIR} -type f -name opm)"
+        if [[ $(wc -l <<< "$temp") = "1" ]]; then
```
jsztuka, line 128: What happens when opm_binaries are not found?

jsztuka, line 128: Just out of curiosity, what does `<<< "$temp"` do? I tried it with different(random) name of the variable and it always returns 1.

jsztuka: Is **$temp** defined natively?

14rcole: That was a leftover from writing the code.  I've replaced `$temp` with `${OPM_BINARIES}`

14rcole: That should also be not equal to one.  I've fixed that too

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -121,27 +121,32 @@ spec:
         # copy content of conffolder to confdir - will be used in next task - related image check
         cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        EXTRACT_DIR="/extracted_base_img"
+        mkdir "${EXTRACT_DIR}"
+        oc image extract ${BASE_IMAGE} "${EXTRACT_DIR}"
+        OPM_BINARIES="$(find ~+${EXTRACT_DIR} -type f -name opm)"
+        if [[ $(wc -l <<< "$${OPM_BINARIES}") -ne "1" ]]; then
```
jsztuka, line 128: $$ returns process id, so the line is always 1, I think changing it to "${OPM_BINARIES}" will do the trick.

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -121,27 +121,32 @@ spec:
...
         cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        EXTRACT_DIR="/extracted_base_img"
+        mkdir "${EXTRACT_DIR}"
+        oc image extract ${BASE_IMAGE} "${EXTRACT_DIR}"
+        OPM_BINARIES="$(find "${EXTRACT_DIR}" -type f -name opm)"
+        if [[ $(wc -l <<< "$${OPM_BINARIES}") -ne "1" ]]; then
+            echo "expected exactly one opm binary in base image."
```
MartinBasti, line 129: In case of error, you should return error OUTPUT:
```
          ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
          echo "${ERROR_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
```
where `$note` is user friendly message about what went wrong

MartinBasti, line 129: Also, we should print out which opm binaries were found for easier debugging

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -121,27 +121,32 @@ spec:
         # copy content of conffolder to confdir - will be used in next task - related image check
         cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        EXTRACT_DIR="/extracted_base_img"
+        mkdir "${EXTRACT_DIR}"
+        oc image extract ${BASE_IMAGE} "${EXTRACT_DIR}"
```
MartinBasti, line 126: There is that thing, that checks cannot fail, this script may exit with 1 when image extract fail, we should make sure to fetch exit code and then fail using TEST_OUTPUT with ERROR result (exit 0 in bash after)

MartinBasti: https://github.com/redhat-appstudio/book/blob/main/ADR/0013-integration-service-api-contracts.md

MartinBasti (changes requested):

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -111,47 +134,84 @@ spec:
...
+
+        cp -r .$conffolder/*
+        $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
+
+
+        EXTRACT_DIR="/extracted_base_img"
+
+        mkdir "${EXTRACT_DIR}"
+
+        oc image extract ${BASE_IMAGE} --path /:"${EXTRACT_DIR}"
```
dirgim, line 157: I think we should handle possible failures here in much the same way as it's done when extracting the conffolder [here](https://github.com/redhat-appstudio/build-definitions/blob/49ec9460ea5248229d89fad18814a36c4d1be32d/task/fbc-validation/0.1/fbc-validation.yaml#L105-L112).

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -47,13 +47,19 @@ spec:
...
 
-        declare -a ALLOWED_BASE_IMAGES=("registry.redhat.io/openshift4/ose-operator-registry")
+
+        declare -a
+        ALLOWED_BASE_IMAGES=("registry.redhat.io/openshift4/ose-operator-registry")
+
 
         ### FBC base image check
+
         if [ -z "${BASE_IMAGE}" ]; then
```
MartinBasti, line 63: Why all those style changes?
I'm not even sure if `declare -a` would work now

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -111,47 +134,84 @@ spec:
...
           echo "${TEST_OUTPUT}" | tee $(results.TEST_OUTPUT.path)
         fi
+
         popd
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
+        runAsUser: 0
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
```
MartinBasti, line 217: Why do we need PVC for working dir and just local dir? It makes operations slower.

task/fbc-validation/0.1/fbc-validation.yaml
14rcole: This was unintentional.  I was doing some editing in the openshift yaml editor when I was testing the task.  It seems to have done some weird things with the formatting.  I'll go back through and fix things.

task/fbc-validation/0.1/fbc-validation.yaml
14rcole: I think that was also stuff I accidentally copied back over from OpenShift when I was testing.  I'll remove it.

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -91,7 +93,7 @@ spec:
           exit 0
         fi
 
-        conffolder=$(cat ../inspect-image/image_inspect.json | jq -r '.Labels ."operators.operatorframework.io.index.configs.v1"')
```
MartinBasti, line 94: is this intended change?

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -118,30 +121,52 @@ spec:
           popd
           exit 0
         fi
-        # copy content of conffolder to confdir - will be used in next task - related image check
-        cp -r .$conffolder/* $(workspaces.workspace.path)/hacbs/$(context.task.name)/confdir
 
-        # We have totally 4 checks here currently
-        check_num=4
+        # copy content of conffolder to confdir - will be used in next task -
+        related image check
```
MartinBasti, line 126: Have you tested this script? does it even work?

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -62,6 +62,7 @@ spec:
           exit 0
         fi
 
+
```
MartinBasti, line 65: extra newlines, not needed I'd say

MartinBasti: it was marked as resolved, but I still see newline here

MartinBasti:
All my comments were marked as resolved, but I see no changed, have you forgotten to push cahnges?

14rcole:
> All my comments were marked as resolved, but I see no changed, have you forgotten to push cahnges?

Apparently I had Friday brain last week.  Yes, changes were made and never pushed.  They were, however, verified in quickcluster.

dirgim (approved):
LGTM

PR #520, opened 2023-08-08
Bigger resources for git clone: 250m/512Mi

jhutar:
OCP creates a PV for a first TaskRun in PipelineRun that needs it. Once created, all the other tasks in the PipelineRun needs to run on the same OCP node, as PV is tied to that node.

By setting substitutional requests for that first task (git-init) we make sure OCP scheduler will have a chance to schedule it on node that has enough capacity.

Also git-init is usually very short when compared with buildah, so the resources should not be blocked for too long.

This was discussed on this mtg: https://docs.google.com/document/d/1UXjkG7NUwrPJMba7FSQjT57ICDH0nBT3gWujhNX-p1c/edit#heading=h.jyt53srvyj4h


jhutar:
Hello @simonbaird . Please could you review?

PR #524, opened 2023-08-09
fix: only run push or pull-request pipelineruns when there is a related change

tisutisu:
This PR will address story: https://issues.redhat.com/browse/STONEBLD-1581

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -4,9 +4,8 @@ kind: PipelineRun
 metadata:
   name: build-definitions-pull-request
   annotations:
-    pipelinesascode.tekton.dev/on-event: "pull_request"
     pipelinesascode.tekton.dev/on-cel-expression: |
-            target_branch == "main" && (".tekton/***".pathChanged() || "appstudio-utils/***".pathChanged() || "hack/***".pathChanged() || "partners/***".pathChanged() || "pipelines/***".pathChanged() || "task/***".pathChanged())
+            event == "pull_request" && target_branch == "main" && (".tekton/***".pathChanged() || "appstudio-utils/***".pathChanged() || "hack/***".pathChanged() || "partners/***".pathChanged() || "pipelines/***".pathChanged() || "task/***".pathChanged())
     pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/yaml-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.1/sast-snyk-check.yaml]"
     pipelinesascode.tekton.dev/task-2: "yaml-lint"
     pipelinesascode.tekton.dev/max-keep-runs: "5"
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: |
-            event == "pull_request" && target_branch == "main" && (".tekton/***".pathChanged() || "appstudio-utils/***".pathChanged() || "hack/***".pathChanged() || "partners/***".pathChanged() || "pipelines/***".pathChanged() || "task/***".pathChanged())
+            event == "pull_request" && target_branch == "main" && (".tekton/***".pathChanged() || "appstudio-utils/***".pathChanged() || "hack/***".pathChanged() || "partners/***".pathChanged() || "pipelines/***".pathChanged() || "task/***".pathChanged() || ".yamllint".pathChanged())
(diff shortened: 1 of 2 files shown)


MartinBasti:
How often is doc folder updated that we have to deal with this issue?

I see that this is PoC that should be applied for other repos as well, I find this as error prone as you have to enumerate all directories/files you want to check and we may forgot to update on-cel-expression when we add new file/dir.

In good world, also README/docs changes should be verified by CI (markdown linters, etc..)

How about `.yamllint` and `syft.Dockerfile`, will changes there trigger CI?

tisutisu:
> How often is doc folder updated that we have to deal with this issue?

Agree that we don't change docs often, but in a single [PR](https://github.com/redhat-appstudio/build-definitions/pull/501) with README change in this repo, the pull-request pipeline got triggered multiple times (once while opening the PR and other runs subsequently for addressing review comments). 

> I see that this is PoC that should be applied for other repos as well, I find this as error prone as you have to enumerate all directories/files you want to check and we may forgot to update on-cel-expression when we add new file/dir.

We are not saying needs to be applied every other repo of RHTAP, but where it makes sense. (Where we are releasing README file as part of the release, there this change doesn't make sense)

> In good world, also README/docs changes should be verified by CI (markdown linters, etc..)

Markdown linters are not used in this repo, so where we are using markdown linters there also the change is not applicable.

> How about `.yamllint` and `syft.Dockerfile`, will changes there trigger CI?

`.yamllint` changes should trigger the CI, will address this. I am not sure about `syft.Dockerfile` usage. @Michkov Could you please help me if we should include `syft.Dockerfile` change?

Michkov:
@tisutisu syft.Dockerfile should not be included, but @mkosiarc could use the same approach (and create new pipelinerun) to watch only changes in syft.Dockerfile to run tests.

MartinBasti:
I still think that this is premature optimization. I can see benefit of running a specific pipeline for specific path, but this "include all except one" approach looks error prone to me.

lcarva:
I see that the image was signed/attested properly by Tekton Chains:

```
🐚 cosign tree quay.io/redhat-user-workloads/build-templates-e2e/test-app-524/retrodep-7gct@sha256:b9bb66163ad4790be77196546a47ce549614eeb77b4ddb7a31d8bdb974761a01
📦 Supply Chain Security Related artifacts for an image: quay.io/redhat-user-workloads/build-templates-e2e/test-app-524/retrodep-7gct@sha256:b9bb66163ad4790be77196546a47ce549614eeb77b4ddb7a31d8bdb974761a01
└── 💾 Attestations for an image tag: quay.io/redhat-user-workloads/build-templates-e2e/test-app-524/retrodep-7gct:sha256-b9bb66163ad4790be77196546a47ce549614eeb77b4ddb7a31d8bdb974761a01.att
   └── 🍒 sha256:ec8c512ddbd92b5c77cd1043cc765f34fb565cf3315bd54517ce5bfeac484bd8
└── 🔐 Signatures for an image tag: quay.io/redhat-user-workloads/build-templates-e2e/test-app-524/retrodep-7gct:sha256-b9bb66163ad4790be77196546a47ce549614eeb77b4ddb7a31d8bdb974761a01.sig
   └── 🍒 sha256:2ec026e7eef4db50e924ae23364002b2fc8878fa9d287efd07eadaa3f38772d0
└── 📦 SBOMs for an image tag: quay.io/redhat-user-workloads/build-templates-e2e/test-app-524/retrodep-7gct:sha256-b9bb66163ad4790be77196546a47ce549614eeb77b4ddb7a31d8bdb974761a01.sbom
   └── 🍒 sha256:b1633281d81b06143fff8ad2b6d1b4cdaa4fa6071054de23aed04c9870dc1b33

```

I have also verified that the public key in the EC report matches the public key of the prod RH01 cluster.

I have also verified that both cosign and ec can validate the image with that key just fine :confused: 

From the report, it looks like the image signature check passed, but the image attestation check didn't. This tells me that the public key is in fact correct.

I wonder if this is a timing issue. If the Chains controller was busy, it could've taken longer than usual for the attestation to be created. Since the signature happens when the TaskRun completes, and the attestation only when the PipelineRun completes, this seems like a probably cause.

Would it be possible to change the test to wait for the annotation `chains.tekton.dev/signed: "true"` on the PipelineRun is set before proceeding? IIRC, the integration folks also had to do something like this to avoid timing issues.

lcarva:
/retest

Let's see if my theory has any merit...

lcarva:
Let's see if this helps: https://github.com/redhat-appstudio/e2e-tests/pull/696

chmeliik (commented):
Will we notice when we add a new file and it doesn't trigger CI?

tisutisu:
> Will we notice when we add a new file and it doesn't trigger CI?

PaC controller will ignore the pipelinerun when it does not match the CEL expression. So we will not see the pipelinerun status in the checks of PR.

tisutisu:
Closing this for now. Currently PaC does not support simpler way to skip pipeline-runs as discussed [here](https://github.com/openshift-pipelines/pipelines-as-code/issues/1417).

PR #526, opened 2023-08-10
fix(STONEINTG-31): use versions with clamDB

MartinBasti:
Using major version will allow us to do breaking changes in future without breaking user workflows.

Explicitly disable renovatebot from updating tags, we want to keep the same floating tag there.

Note: we also keep latest tag for compat reasons (but not forever)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -51,6 +51,13 @@
       ],
       "schedule": ["on monday and wednesday"],
       "groupName": "shared"
+    },
+    {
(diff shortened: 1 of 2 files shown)


MartinBasti:
needs https://github.com/redhat-appstudio/hacbs-test/pull/124

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #531, opened 2023-08-15
DO-NOT-MERGE: mount in host container storage for performance comparison

gabemontero:


gabemontero:
I think more is needed to mount the actual host containers storage (i.e. the /var/lib/container empty dir) but I'm building toward that incrementally

PR #558, opened 2023-09-04
change workspace to /tekton/home for deprecated-image-check

sonam1412:
As the deprecated-image-check is not sharing any info between the different tasks and only using its own working directory, it doesn't seem to need any PVC to be used from the workspace. 
Hence, the workspace can be removed/replaced with the /tekton/home directory in the build-definitions deprecated-image-check task.

### DIFF
--- a/task/deprecated-image-check/0.3/deprecated-image-check.yaml
+++ b/task/deprecated-image-check/0.3/deprecated-image-check.yaml
@@ -1,0 +1,107 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "appstudio, hacbs"
+  name: deprecated-image-check
+spec:
+  description: >-
+    Identifies the unmaintained and potentially insecure deprecated base images.
+    Pyxis API collects metadata from image repository, and Conftest applies supplied policy to identify the deprecated images using that metadata.
+  params:
+    - name: POLICY_DIR
+      description: Path to directory containing Conftest policies.
+      default: "/project/repository/"
+    - name: POLICY_NAMESPACE
+      description: Namespace for Conftest policy.
+      default: "required_checks"
+    - name: BASE_IMAGES_DIGESTS
+      description: Digests of base build images.
+
+  results:
+    - name: PYXIS_HTTP_CODE
+      description: HTTP code returned by Pyxis API endpoint.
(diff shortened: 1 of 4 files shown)


jsztuka (approved):
lgtm

sonam1412:
clair-scan test results from local testing 
image

dirgim (approved):
LGTM

task/deprecated-image-check/0.3/deprecated-image-check.yaml
```diff
@@ -0,0 +1,107 @@
...
+          value: $(params.BASE_IMAGES_DIGESTS)
+      script: |
+        #!/usr/bin/env bash
+        readarray -t IMAGE_ARRAY < <(echo -n "$BASE_IMAGES_DIGESTS" | sed 's/\\n/\'$'\n''/g')
+        for BASE_IMAGE in ${IMAGE_ARRAY[@]};
+        do
+          IFS=:'/' read -r IMAGE_REGISTRY IMAGE_WITH_TAG <<< $BASE_IMAGE; echo "[$IMAGE_REGISTRY] [$IMAGE_WITH_TAG]"
+          IMAGE_REPOSITORY=`echo $IMAGE_WITH_TAG | cut -d ":" -f1`
+          IMAGE_REGISTRY=${IMAGE_REGISTRY//registry.redhat.io/registry.access.redhat.com}
+          export IMAGE_REPO_PATH=/tekton/home/${IMAGE_REPOSITORY}
```
14rcole, line 45: `/tekton/home` is referenced a number of times through the code.  Can you assign it to a variable and reference that instead?  This will make it easier to change if we ever need to do so.

sonam1412:
deprecated-image-check results from the local testing
image

task/deprecated-image-check/0.3/deprecated-image-check.yaml
sonam1412: Thats a good point. This seems to be a pattern in all the other tasks as well ex: [clair-scan](https://github.com/sonam1412/build-definitions/blob/update-workspace-depracated-image-check/task/clair-scan/0.1/clair-scan.yaml#L43), [clamav-scan](https://github.com/sonam1412/build-definitions/blob/update-workspace-depracated-image-check/task/clamav-scan/0.1/clamav-scan.yaml#L67). I think this should be standardized in all the tasks. Shall we cover it in a separate PR ?

sonam1412: Discussed on the standup and confirmed to handle this change in a different PR

14rcole (approved):

PR #559, opened 2023-09-05
Update update-infra-deployments.yaml script

tkdchen:


task/update-infra-deployments/0.1/update-infra-deployments.yaml
```diff
@@ -202,46 +204,43 @@ spec:
                 return json_output
 
             def create_reset_branch(self):
-                branch = os.environ.get('ORIGIN_REPO').split('/')[-1]
-                target_gh_repo = os.environ.get('TARGET_GH_REPO')
-                target_branch = self._request("GET", f"/repos/{target_gh_repo}/git/refs/heads/{branch}").json()
-                main_branch_sha = self._request("GET", f"/repos/{target_gh_repo}/git/refs/heads/main").json()['object']['sha']
+                branch = ORIGIN_REPO.split('/')[-1]
```
mkosiarc, line 207: What are the values or ORIGIN_REPO here?

Here the **ORIGIN_REPO.split('/')[-1]** refers to branch, but on line 187, we have the same code where it refers to repo name

```
repo_name = ORIGIN_REPO.split('/')[-1]
```

task/update-infra-deployments/0.1/update-infra-deployments.yaml
```diff
@@ -120,7 +120,10 @@ spec:
 
         EXPIRE_MINUTES_AS_SECONDS = int(os.environ.get('GITHUBAPP_TOKEN_EXPIRATION_MINUTES', 10)) * 60
         # TODO support github enteprise
-        GITHUB_API_URL = os.environ.get('GITHUB_API_URL')
+        GITHUB_API_URL = os.environ["GITHUB_API_URL"]
```
mkosiarc, line 123: In case these required env variables are missing, do we want just to fail with just the KeyError?

tkdchen: The error is for us only to know something is misconfigured. KeyError should be enough to let us know.

task/update-infra-deployments/0.1/update-infra-deployments.yaml
tkdchen: Each RHTAP component has its own branch to hold updates in the target repository, which is generally the infra-deployments, and in build-definitions CI, update-infra-deployments task is also used to update enterprise-contract/ec-policies.

mkosiarc: I see, thanks for the explanation.

mkosiarc (approved):

rcerven (approved):

PR #566, opened 2023-09-07
Create OPA data bundle of acceptable tkn bundles

lcarva:
EC has the concept of acceptable tekton bundles. These are OCI image references that specify which tekton bundles are acceptable to build artifacts in RHTAP. Today, these are tracked in git. The push.yaml pipeline of this repo is responsible for making a pull request in the relevant EC repo update the list. However, this is subject to timing issues when the pull request is not merged fast enough, causing some of the references to be lost.

This commit introduces a new approach by tracking the list of acceptable tekton bundles in an OPA data bundle, which is just another type of OCI image. By doing this, the changes are deployed immediately, significantly lowering the risk of timing issues.

For now, the list is maintained in two places. This facilitate the transition. See the epic https://issues.redhat.com/browse/HACBS-2554 for more details.

Tracker for this change:
https://issues.redhat.com/browse/HACBS-1796

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -134,6 +134,59 @@ spec:
         workspaces:
           - name: artifacts
             workspace: workspace
+      - name: build-acceptable-bundles
+        runAfter:
+          - build-bundles
+        workspaces:
+          - name: artifacts
+            workspace: workspace
+        taskSpec:
+          workspaces:
+            - name: artifacts
+              description: Workspace containing arbitrary artifacts used during the task run.
+          volumes:
+          - name: quay-secret
+            secret:
+              secretName: redhat-appstudio-tekton-catalog-build-definitions-pull-secret
+          steps:
+            - name: build-bundles
+              image: quay.io/redhat-appstudio/appstudio-utils:{{ revision }}
+              script: |-
+                #!/usr/bin/env bash
+                set -euo pipefail
+
+                # TODO: Eventually, the input for the ec-track-bundle command will be the OPA data
+                # bundle itself. Use the data from git instead until the OPA data bundle has been
+                # initially populated, see https://issues.redhat.com/browse/HACBS-2562.
+                BUNDLES_FILE=acceptable_tekton_bundles.yml
+                curl -L "https://raw.githubusercontent.com/enterprise-contract/ec-policies/main/data/${BUNDLES_FILE}" \
+                  -o "${BUNDLES_FILE}"
+
+                BUNDLES=(
+                  $(workspaces.artifacts.path)/task-bundle-list
+                  $(workspaces.artifacts.path)/pipeline-bundle-list
+                )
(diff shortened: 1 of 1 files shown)


.tekton/push.yaml
```diff
@@ -134,6 +134,58 @@ spec:
...
+                  $(workspaces.artifacts.path)/pipeline-bundle-list
+                )
+                BUNDLES_PARAM=($(cat ${BUNDLES[@]} | awk '{ print "--bundle=" $0 }'))
+
+                # The OPA data bundle is tagged with the current timestamp. This has two main
+                # advantages. First, it prevents the image from accidentally not having any tags,
+                # and getting garbage collected. Second, it helps us create a timeline of the
+                # changes done to the data over time.
+                TAG="$(date '+%s')"
+                DATA_BUNDLE_REPO='quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles'
```
lcarva, line 175: I think someone needs to create this first before merging.

.tekton/push.yaml
```diff
@@ -134,6 +134,58 @@ spec:
...
+
+                # Update the OPA data bundle.
+                ec track bundle \
+                  --input "${BUNDLES_FILE}" \
+                  --output "oci:${DATA_BUNDLE_REPO}:${TAG}" \
+                  ${BUNDLES_PARAM[@]}
```
zregvart, line 181: I'd probably go with xargs rather than awk, makes it a bit clearer IMHO

```suggestion
                cat \
                $(workspaces.artifacts.path)/task-bundle-list \
                $(workspaces.artifacts.path)/pipeline-bundle-list \
                | xargs xargs printf -- '--bundle=%s\n' \
                | xargs ec track bundle \
                  --input "${BUNDLES_FILE}" \
                  --output "oci:${DATA_BUNDLE_REPO}:${TAG}"
```

zregvart (approved):
LGTM

.tekton/push.yaml
```diff
@@ -134,6 +134,58 @@ spec:
...
+              description: Workspace containing arbitrary artifacts used during the task run.
+          volumes:
+          - name: quay-secret
+            secret:
+              secretName: redhat-appstudio-tekton-catalog-build-definitions-pull-secret
+          steps:
+            - name: build-bundles
+              image: quay.io/redhat-appstudio/appstudio-utils:{{ revision }}
+              script: |-
+                set -euo pipefail
```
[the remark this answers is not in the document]
lcarva: I don't know if that makes a difference here, but added anyways.

.tekton/push.yaml
lcarva: That does not look clearer to me :sweat_smile: 

I'll keep it as is unless you feel very strongly about this.

.tekton/push.yaml
lcarva: This has been done: https://quay.io/repository/redhat-appstudio-tekton-catalog/data-acceptable-bundles?namespace=redhat-appstudio-tekton-catalog

PR #568, opened 2023-09-08
Handle no acceptable bundle changes

lcarva:
If there are no new task/pipeline tekton bundles to be added to the list of acceptable bundles, exit the build-acceptable-bundles task successfully.

This may happen if a commit is added that doesn't touch the task or pipeline definitions, e.g. update README.md.


robnester-rh (approved):
lgtm

PR #570, opened 2023-09-12
feat(STONEINTG-222) set euo to shell to fail on error early

hongweiliu17:
Signed-off-by: Hongwei Liu <hongliu@redhat.com>


task/deprecated-image-check/0.2/deprecated-image-check.yaml
```diff
@@ -60,7 +63,9 @@ spec:
           value: $(params.POLICY_NAMESPACE)
       script: |
         #!/usr/bin/env sh
+        set -euo pipefail
         source /utils.sh
+        trap cleanup ERR
```
jsztuka, line 69: Shouldn't this script also have: 
` trap 'echo "Line:$LINENO"' ERR`
Since it uses pipe below? Line 85

task/deprecated-image-check/0.3/deprecated-image-check.yaml
```diff
@@ -60,7 +63,9 @@ spec:
           value: $(params.POLICY_NAMESPACE)
       script: |
         #!/usr/bin/env sh
+        set -euo pipefail
         source /utils.sh
+        trap cleanup ERR
```
jsztuka, line 68: I might be undestanding this wrongly, but why this script doesn't include
` trap 'echo "Line:$LINENO"' ERR`
I see pipe in script below.

hongweiliu17: That cleanup is defined in https://github.com/redhat-appstudio/hacbs-test/pull/137/files#diff-b17e8443b3f0dd1a2f1701af585a3a6222f2631b717c934c6fede3c8b16837ecR161 we import and call it.

task/deprecated-image-check/0.2/deprecated-image-check.yaml
hongweiliu17: `cleanup` will add the error info to `TEST_OUTPUT`

task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -153,6 +159,8 @@ spec:
       name: database
       script: |
         #!/usr/bin/env bash
+        set -euo pipefail
+        trap 'echo "Script errored at line ${LINENO}: ${BASH_COMMAND}"' ERR
```
MartinBasti, line 163: we don't need this line

task/deprecated-image-check/0.1/deprecated-image-check.yaml
```diff
@@ -36,6 +36,9 @@ spec:
           value: $(params.BASE_IMAGES_DIGESTS)
       script: |
         #!/usr/bin/env bash
+        set -euo pipefail
+        trap 'echo "Script errored at line ${LINENO}: ${BASH_COMMAND}"' ERR
```
MartinBasti, line 40: Should be trap handle_error here?

task/deprecated-image-check/0.2/deprecated-image-check.yaml
```diff
@@ -36,6 +36,9 @@ spec:
           value: $(params.BASE_IMAGES_DIGESTS)
       script: |
         #!/usr/bin/env bash
+        set -euo pipefail
+        trap 'echo "Script errored at line ${LINENO}: ${BASH_COMMAND}"' ERR
```
MartinBasti, line 40: Should be handle_error here?

task/deprecated-image-check/0.3/deprecated-image-check.yaml
```diff
@@ -36,6 +36,9 @@ spec:
           value: $(params.BASE_IMAGES_DIGESTS)
       script: |
         #!/usr/bin/env bash
+        set -euo pipefail
+        trap 'echo "Script errored at line ${LINENO}: ${BASH_COMMAND}"' ERR
```
MartinBasti, line 40: Handle error?

task/sast-snyk-check/0.1/sast-snyk-check.yaml
```diff
@@ -41,7 +41,9 @@ spec:
           value: $(params.ARGS)
       script: |
         #!/usr/bin/env bash
+        set -euo pipefail
         . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' ERR
         SNYK_TOKEN="$(cat /etc/secrets/snyk_token)"
```
MartinBasti, line 47: This will fail with -e, we should check emptyness of file instead?

MartinBasti: https://github.com/redhat-appstudio/build-definitions/pull/587/commits/4281c2f1d2f0f93ccdc7f0530bfebbff5d8ea99f I addressed it here

task/deprecated-image-check/0.1/deprecated-image-check.yaml
hongweiliu17: This task references registry.access.redhat.com/ubi8/ubi-minimal, not our hacbs-test image.

hongweiliu17: Actually we don't need it because we should not fail here

task/deprecated-image-check/0.2/deprecated-image-check.yaml
hongweiliu17: Actually we don't need it because we should not fail in this step

task/deprecated-image-check/0.3/deprecated-image-check.yaml
hongweiliu17: Actually we don't need it because we should not fail in this step

task/deprecated-image-check/0.1/deprecated-image-check.yaml
MartinBasti: Then we probably should use hacbs-test image.

> Actually we don't need it because we should not fail here

Why? there is many codelines that could fail on random unexpected error

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -28,8 +28,10 @@ spec:
           cpu: 10m
       script: |
         #!/usr/bin/env bash
-        set -o pipefail
+        set -euo pipefail
         source /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
         FAILEDIMAGES=""
         catalog="$(find $(workspaces.workspace.path)/hacbs/fbc-validation/ -name catalog.yaml)"
```
MartinBasti, line 37: I cannot comment there, but in line 40, 42; this will stop work
```
        if [ $? -ne 0 ]; then
```
shuold be changed to
``` 
        status=0
        relImgs="$(yq -r '.relatedImages[]?.image' $catalog || staus=$?)"
        if [ $? -ne $status ]; then
```

task/fbc-validation/0.1/fbc-validation.yaml
```diff
@@ -49,8 +49,9 @@ spec:
           cpu: 10m
       script: |
         #!/usr/bin/env bash
-        set -o pipefail
+        set -euo pipefail
         source /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
 
         declare -a ALLOWED_BASE_IMAGES=("registry.redhat.io/openshift4/ose-operator-registry")
```
MartinBasti, line 57: on line 95, 96; this won't work anymore
```
        conffolder=$(cat ../inspect-image/image_inspect.json | jq -r '.Labels ."operators.operatorframework.io.index.configs.v1"')
        if [ $? -ne 0 ]; then
```
should be replaced to

```
        status=0
        conffolder=$(jq -f ../inspect-image/image_inspect.json -r '.Labels ."operators.operatorframework.io.index.configs.v1"') || status=$?
        if [ $status -ne 0 ]; then
```
As I pointed out several times here:
https://redhat-internal.slack.com/archives/C034UFMUSV6/p1695224381051229?thread_ts=1694681340.976039&cid=C034UFMUSV6

MartinBasti (changes requested):
set -e changes behavior and few code parts must be udpated, commented inline.

Also, waiting for updated image, I'd like to do a deep testing as well

task/deprecated-image-check/0.1/deprecated-image-check.yaml
hongweiliu17: The image from `ubi-minimal` can be enough and the next step will generate `TEST_OUTPUT` according to `results.PYXIS_HTTP_CODE.path` https://github.com/redhat-appstudio/build-definitions/blob/main/task/deprecated-image-check/0.1/deprecated-image-check.yaml#L68

MartinBasti: We are trying to handle unexpected failures and silent hiding of errors. I don't know how you can be so sure, that with any change in future this will be working smoothly forever.

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
MartinBasti: EDIT: `||` must be outside of `$()`

hongweiliu17: I am thinking, maybe we should remove `set -euo pipefail` since we will have that `trap EXIT`.

MartinBasti: without `euo pipefail` faulty code won't trigger exit, so having trap EXIT without this doesn't do much

task/deprecated-image-check/0.1/deprecated-image-check.yaml
hongweiliu17: I mean `ubi-minimal` have all command used in that step, so it is enough. So let me change to use hacbs-test

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
hongweiliu17: Fixed. Thanks.

task/fbc-validation/0.1/fbc-validation.yaml
hongweiliu17: Fixed. Thanks.

hongweiliu17:
> set -e changes behavior and few code parts must be udpated, commented inline.
> 
> Also, waiting for updated image, I'd like to do a deep testing as well

Yes, thanks for the reminder.

MartinBasti (approved):
Changes looks good, but please test is carefully

task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -45,6 +45,11 @@ spec:
           memory: 512Mi
           cpu: 10m
       script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
```
MartinBasti, line 51: I realized, we cannot do this in every step, it would emit result multiple times, that might be confusing.

We may need to have different handlers
handle_error only for steps not generating TEST_OUTPUT, it will only print error and exit 0
handle_error_with_test_output, which is current implementation of handle_error

MartinBasti (changes requested):

task/clamav-scan/0.1/clamav-scan.yaml
MartinBasti: image
on the other hands it doesn't look bad, but still if task is not supposed to generate TEST_OUTPUT

hongweiliu17: If there is an failure in a step of a tekton task, won't the steps stop immediately? Will the next steps of this task be still run?

hongweiliu17: It is interesting:
When there is no `trap` in tasks, it runs as below: 
```
STEP-STEP1

/tekton/scripts/script-0-2scgd: line 3: undefined1: unbound variable
STEP-STEP2

2023/10/09 08:25:30 Skipping step because a previous step failed
```
But if we have `trap` in steps, it runs as below, I think the `trap` makes the step finally look succeed, so that we get multiple `TEST_OUTPUT`:
```
STEP-STEP1

/tekton/scripts/script-0-c6ws4: line 5: undefined1: unbound variable
{"result":"ERROR","timestamp":"1696840096","note":"Unexpected error: Script errored at command: echo $undefined1.","namespace":"default","successes":0,"failures":0,"warnings":0}
STEP-STEP2

/tekton/scripts/script-1-66n4h: line 5: undefined2: unbound variable
{"result":"ERROR","timestamp":"1696840097","note":"Unexpected error: Script errored at command: echo $undefined2.","namespace":"default","successes":0,"failures":0,"warnings":0}
```

hongweiliu17: @MartinBasti @dirgim I am thinking to add the real failure to `note` of `TEST_OUTPUT`, maybe we can add some code before [this line](https://github.com/redhat-appstudio/hacbs-test/blob/main/test/utils.sh#L162): Once the `TEST_OUTPUT` exists, we can only print some message to tell users the task experienced error in the previous step but skip generating `TEST_OUTPUT` again?

MartinBasti: > If there is an failure in a step of a tekton task, won't the steps stop immediately?


But that's whole purpose of why we are adding TRAP, check tasks cannot fail according ADR, exit code must be 0

hongweiliu17: I mean, if one step fails with unexpected shell error, then generate `TEST_OUTPUT` with the failing shell command in `note`, and the next step exit with `0` but won't generate `TEST_OUTPUT` again.

hongweiliu17: My meaning is to add code to `handle_error()` like [this](https://github.com/redhat-appstudio/hacbs-test/pull/158/files#diff-b17e8443b3f0dd1a2f1701af585a3a6222f2631b717c934c6fede3c8b16837ecR173-R178), t hen we still call `trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT` in the steps of tasks, then if one step fails with unexpected error, the failing command will be added to the `note` of `TEST_OUTPUT`, then the followup steps won't generate `TEST_OUTPUT` again, but it prints log `"Task experienced unexpected error in previous step. For details, check Tekton task result and log.` in console and exit with `0`

MartinBasti: To me it looks much cleaner to only return TEST_OUTPUT from steps which are supposed to return it.
We can make it conditional and return TEST_OUTPUT only when `handle_error` has path specified

hongweiliu17: > To me it looks much cleaner to only return TEST_OUTPUT from steps which are supposed to return it. I.e. don't make it conditional, write it explicitly in code where output is returned. We can make it conditional and return TEST_OUTPUT only when `handle_error` has path specified

Ok, let me update https://github.com/redhat-appstudio/hacbs-test/pull/158.

hongweiliu17: https://github.com/redhat-appstudio/hacbs-test/pull/158 is updated

hongweiliu17: Updated to use the new functions in the latest hacbs-test:v1.1.6.

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -28,20 +28,23 @@ spec:
...
 
-        relImgs="$(yq -r '.relatedImages[]?.image' $catalog)"
-        if [ $? -ne 0 ]; then
+        status=0
+        relImgs="$(yq -r '.relatedImages[]?.image' $catalog)" || status=$?
+        if [ $status -ne 0 ]; then
           echo "Processing the catalog with yq failed because catalog.yaml contains data type other than yaml. Attempting to process with jq..."
-          relImgs="$(jq -r '.relatedImages[]?.image' $catalog)"
-          if [ $? -ne 0 ]; then
+          relImgs="$(jq -r '.relatedImages[]?.image' $catalog)" || status=$?
```
MartinBasti, line 46: Status must reset to 0 manually, `|| status=$?` is executed only when previous command failed, thus it would never reset it back to 0 and next condition would fail
```suggestion
          status=0
          relImgs="$(jq -r '.relatedImages[]?.image' $catalog)" || status=$?
```

hongweiliu17: Yes, you are right. Updated. Thanks!

MartinBasti (approved):

dirgim (approved):
LGTM

MartinBasti:
Python repo build test failed, we haven't updated this task, so either repo is broken or clone task
```
STEP-CLONE

+ '[' false = true ']'
+ '[' false = true ']'
+ CHECKOUT_DIR=/workspace/output/source
+ '[' true = true ']'
+ cleandir
+ '[' -d /workspace/output/source ']'
+ test -z ''
+ test -z ''
+ test -z ''
+ /ko-app/git-init -url=https://github.com/redhat-appstudio-qe/devfile-sample-python-basic.git -revision=symlink -refspec= -path=/workspace/output/source -sslVerify=true -submodules=true -depth=1 -sparseCheckoutDirectories=
{"level":"info","ts":1697192832.1426024,"caller":"git/git.go:178","msg":"Successfully cloned https://github.com/redhat-appstudio-qe/devfile-sample-python-basic.git @ cb167402672a0753e2f06bd9cb0ec588051a78fd (grafted, HEAD, origin/symlink) in path /workspace/output/source"}
{"level":"info","ts":1697192832.2145154,"caller":"git/git.go:217","msg":"Successfully initialized and updated submodules in path /workspace/output/source"}
+ cd /workspace/output/source
++ git rev-parse HEAD
+ RESULT_SHA=cb167402672a0753e2f06bd9cb0ec588051a78fd
+ EXIT_CODE=0
+ '[' 0 '!=' 0 ']'
+ printf %s cb167402672a0753e2f06bd9cb0ec588051a78fd
+ printf %s https://github.com/redhat-appstudio-qe/devfile-sample-python-basic.git
STEP-SYMLINK-CHECK

Running symlink check
The cloned repository contains symlink pointing outside of the cloned repository: /workspace/output/source/os-release-symlink
```

MartinBasti:
@mkosiarc can be this issue somehow related to your PR changing workdir?

MartinBasti:
But actually it clones https://github.com/redhat-appstudio-qe/devfile-sample-python-basic/commit/cb167402672a0753e2f06bd9cb0ec588051a78fd  which is branch specific for testing failure  when symlink is out of repo, why is this used in tests?

MartinBasti:
Nevermind what I wrote before, this is the failure
```
  failed to get cosign result for image quay.io/redhat-user-workloads/build-templates-e2e/test-app-570/devfile-sample-python-basic-cwbo:build-a3c96-1697192775@sha256:6fa1eb2ec5ef2411a2e57af23d1baff30c935a6e0e4312f60aed3d27f0ad6796: failed to find cosign results for image quay.io/redhat-user-workloads/build-templates-e2e/test-app-570/devfile-sample-python-basic-cwbo:build-a3c96-1697192775@sha256:6fa1eb2ec5ef2411a2e57af23d1baff30c935a6e0e4312f60aed3d27f0ad6796: error when getting attestation tag: cannot get manifest digest from quay.io/redhat-user-workloads/build-templates-e2e/test-app-570/devfile-sample-python-basic-cwbo:sha256-6fa1eb2ec5ef2411a2e57af23d1baff30c935a6e0e4312f60aed3d27f0ad6796.att image. response body: {"tags": [], "page": 1, "has_additional": false}


  Creating Pipeline "verify-enterprise-contract-run-2jrhv"
  Waiting for pipeline "verify-enterprise-contract-run-2jrhv" to finish
  [FAILED] in [It] - /github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:319 @ 10/13/23 10:29:40.747
• [FAILED] [32.475 seconds]
[build-service-suite Build templates E2E test] HACBS pipelines when the container image for component with Git source URL https://github.com/redhat-appstudio-qe/devfile-sample-python-basic is created and pushed to container registry [It] verify-enterprise-contract check should pass [build, HACBS, pipeline, sbom, slow, build-templates-e2e]
/github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:254

  [FAILED] Expected
      <[]v1beta1.TaskRunResult | len:1, cap:4>: [
          {
              Name: "TEST_OUTPUT",
              Type: "string",
              Value: {
                  Type: "string",
                  StringVal: "{\"timestamp\":\"1697192977\",\"namespace\":\"\",\"successes\":1,\"failures\":0,\"warnings\":1,\"result\":\"WARNING\"}",
                  ArrayVal: nil,
                  ObjectVal: nil,
              },
          },
      ]
  To satisfy at least one of these matchers: [%!s(*matchers.ContainElementsMatcher=&{[0xc001705140] [0xc001705140]}) %!s(*matchers.ContainElementsMatcher=&{[0xc001705180] [0xc001705180]})]
  In [It] at: /github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:319 @ 10/13/23 10:29:40.747
```

MartinBasti:
Root cause is that test expects success
```
ContainElements(tekton.MatchTaskRunResultWithJSONPathValue(constants.TektonTaskTestOutputName, "{$.result}", `["SUCCESS"]`)),
```
but EC returns WARNING
```
STEP-SUMMARY

{
  "timestamp": "1697192977",
  "namespace": "",
  "successes": 1,
  "failures": 0,
  "warnings": 1,
  "result": "WARNING"
}
```
EC detail
```
  warnings:
  - metadata:
      code: slsa_source_correlated.source_code_reference_provided
      collections:
      - minimal
      - slsa1
      - slsa2
      - slsa3
      - redhat
      description: Warn if the expected source code reference is not provided.
      solution: Provide the expected source code reference in inputs.
      title: Source code reference provided
    msg: Expected source code reference was not provided for verification
```

hongweiliu17:
> Root cause is that test expects success
> 
> ```
> ContainElements(tekton.MatchTaskRunResultWithJSONPathValue(constants.TektonTaskTestOutputName, "{$.result}", `["SUCCESS"]`)),
> ```
> 
> but EC returns WARNING
> 
> ```
> STEP-SUMMARY
> 
> {
>   "timestamp": "1697192977",
>   "namespace": "",
>   "successes": 1,
>   "failures": 0,
>   "warnings": 1,
>   "result": "WARNING"
> }
> ```
> 
> EC detail
> 
> ```
>   warnings:
>   - metadata:
>       code: slsa_source_correlated.source_code_reference_provided
>       collections:
>       - minimal
>       - slsa1
>       - slsa2
>       - slsa3
>       - redhat
>       description: Warn if the expected source code reference is not provided.
>       solution: Provide the expected source code reference in inputs.
>       title: Source code reference provided
>     msg: Expected source code reference was not provided for verification
> ```

Thanks very much for debugging it!

PR #576, opened 2023-09-13
STONEBLD-1667 Initial multi platform build tasks

stuartwdouglas:
Note that this includes https://github.com/redhat-appstudio/build-definitions/pull/577 which moves to the multi-arch builder image

### DIFF
--- a/task/buildah-remote/0.1/README.md
+++ b/task/buildah-remote/0.1/README.md
@@ -1,0 +1,37 @@
+# buildah-remote task
+
+This task is programmatically generated from the buildah task to keep it in sync, it should not be manually modified.
+
+Buildah task builds source code into a container image and pushes the image into container registry using buildah tool, however while the standard buildah task is run directly on the cluster, this task is run on a remote host. This must be used in combination with the [Multi Arch Controller](https://github.com/redhat-appstudio/multi-arch-controller) which provides the credentials and host name to use to perform the build.
+
+This task has an additional `PLATFORM` param that is used by the Multi Arch Controller to decide which host should perform the build.
+
+
+
+## Parameters
+| name                | description                                                                                                                                                        |default value|required|
+|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|---|---|
+| PLATFORM            | The target platform.                                                                                                                                               ||true|
+| IMAGE               | Reference of the image buildah will produce.                                                                                                                       ||true|
+| BUILDER_IMAGE       | The location of the buildah builder image.                                                                                                                         |registry.access.redhat.com/ubi9/buildah:9.0.0-19@sha256:c8b1d312815452964885680fc5bc8d99b3bfe9b6961228c71a09c72ca8e915eb|false|
+| DOCKERFILE          | Path to the Dockerfile to build.                                                                                                                                   |./Dockerfile|false|
+| CONTEXT             | Path to the directory to use as context.                                                                                                                           |.|false|
+| TLSVERIFY           | Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)                                                                                      |true|false|
+| DOCKER_AUTH         | unused, should be removed in next task version                                                                                                                     |""|false|
+| HERMETIC            | Determines if build will be executed without network access.                                                                                                       |false|false|
+| PREFETCH_INPUT      | In case it is not empty, the prefetched content should be made available to the build.                                                                             |""|false|
+| IMAGE_EXPIRES_AFTER | Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively. |""|false|
+
+## Results
+|name|description|
+|---|---|
+|IMAGE_DIGEST|Digest of the image just built|
+|IMAGE_URL|Image repository where the built image was pushed|
+|BASE_IMAGES_DIGESTS|Digests of the base images used for build|
+|SBOM_JAVA_COMPONENTS_COUNT|The counting of Java components by publisher in JSON format|
+|JAVA_COMMUNITY_DEPENDENCIES|The Java dependencies that came from community sources such as Maven central.|
+
+## Workspaces
+|name|description|optional|
+|---|---|---|
+|source|Workspace containing the source code to build.|false|
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -1,0 +1,398 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio, hacbs
+  creationTimestamp: null
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: docker
+    build.appstudio.redhat.com/multi-platform-required: "true"
+  name: buildah-remote
+spec:
+  description: |-
+    Buildah task builds source code into a container image and pushes the image into container registry using buildah tool.
+    In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
+    When [Java dependency rebuild](https://redhat-appstudio.github.io/docs.stonesoup.io/Documentation/main/cli/proc_enabled_java_dependencies.html) is enabled it triggers rebuilds of Java artifacts.
+    When prefetch-dependencies task was activated it is using its artifacts to run build in hermetic environment.
+  params:
+  - description: Reference of the image buildah will produce.
+    name: IMAGE
+    type: string
+  - default: quay.io/redhat-appstudio/buildah:v1.31.0@sha256:34f12c7b72ec2c28f1ded0c494b428df4791c909f1f174dd21b8ed6a57cf5ddb
+    description: The location of the buildah builder image.
+    name: BUILDER_IMAGE
+    type: string
+  - default: ./Dockerfile
+    description: Path to the Dockerfile to build.
+    name: DOCKERFILE
+    type: string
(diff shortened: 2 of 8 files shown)


task/buildah-multiplatform/0.1/buildah-multiplatform.yaml
```diff
@@ -0,0 +1,108 @@
...
+        unshare -Uf --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah manifest add $IMAGE $i
+      done
+
+      status=-1
+      max_run=5
+      sleep_sec=10
+      for run in $(seq 1 $max_run); do
+        status=0
+        [ "$run" -gt 1 ] && sleep $sleep_sec
+        echo "Pushing sbom image to registry"
```
lcarva, line 92: Is it an SBOM image? Same question a few lines below.

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -0,0 +1,400 @@
...
+      respectively.
+    name: IMAGE_EXPIRES_AFTER
+    type: string
+  - default: ""
+    description: The image is built from this commit.
+    name: COMMIT_SHA
+    type: string
+  - default: ""
+    description: The architecture being built for a multi arc build
+    name: ARCH
```
lcarva, line 65: Do we need both an ARCH and a PLATFORM parameter? Maybe it's leftover. I don't see it mentioned in the README.md.

task/buildah-remote/0.1/README.md
```diff
@@ -0,0 +1,37 @@
+# buildah-remote task
+
+This task is programmatically generated from the buildah task to keep it in sync, it should not be manually modified.
```
lcarva, line 3: It would be good to include the steps of how someone can regenerate this. Ideally, there would be a step on the CI that checks if it needs to be regenerated.

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: Oops, that was a leftover, I thought I had removed it but it must have got lost somewhere.

task/buildah-multiplatform/0.1/buildah-multiplatform.yaml
stuartwdouglas: copy/paste error, good catch

task/buildah-multiplatform/0.1/buildah-multiplatform.yaml
```diff
@@ -0,0 +1,108 @@
...
+    description: The image is built from this commit.
+    type: string
+    default: ""
+  - name: IMAGES
+    description: List of images that are to be merged into the multi platform image
+    type: array
+  - default: ""
+    description: Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
+    name: IMAGE_EXPIRES_AFTER
+    type: string
```
chmeliik, line 36: BUILDER_IMAGE, COMMIT_SHA and IMAGE_EXPIRES_AFTER look unused

task/buildah-multiplatform/0.1/buildah-multiplatform.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+      # Setting new namespace to run buildah - 2^32-2
+      echo 'root:1:4294967294' | tee -a /etc/subuid >> /etc/subgid
+
+      buildah manifest create $IMAGE
+      for i in $@
+      do
+        unshare -Uf --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $i
+        unshare -Uf --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah manifest add $IMAGE $i
+      done
```
chmeliik, line 84: Can we have a follow-up story to improve this? Pulling the entire image just to add its manifest to an OCI index is unnecessary

Maybe `crane` does it better? https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_index_append.md

And so does OSBS, but in Python

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -0,0 +1,396 @@
...
+      # Save the SBOM produced by Cachi2 so it can be merged into the final SBOM later
+      if [ -n "${PREFETCH_INPUT}" ]; then
+        cp /tmp/cachi2/output/bom.json ./sbom-cachi2.json
+      fi
+
+      buildah push $IMAGE oci:rhtap-final-image
+      REMOTESSHEOF
+      chmod +x scripts/script-build.sh
+      rsync -ra scripts $SSH_HOST:$BUILD_DIR
+      ssh $SSH_ARGS $SSH_HOST $PORT_FORWARD podman  run $PODMAN_PORT_FORWARD -e BUILDAH_FORMAT=oci -e STORAGE_DRIVER=vfs -e HERMETIC=$(params.HERMETIC) -e PREFETCH_INPUT=$(params.PREFETCH_INPUT) -e CONTEXT=$(params.CONTEXT) -e DOCKERFILE=$(params.DOCKERFILE) -e IMAGE=$(params.IMAGE) -e TLSVERIFY=$(params.TLSVERIFY) -e IMAGE_EXPIRES_AFTER=$(params.IMAGE_EXPIRES_AFTER) -e COMMIT_SHA=$(params.COMMIT_SHA)  --rm  -v $BUILD_DIR/workspaces/source:$(workspaces.source.path):Z  -v $BUILD_DIR/scripts:/script:Z --user=0  $(params.BUILDER_IMAGE)  /script/script-build.sh
```
chmeliik, line 224: Unquoted variables can break things, allow shell injection etc. Especially Tekton `$(params.*)` - it's not even possible to quote them correctly. If nothing else, PREFETCH_INPUT may contain whitespace.

Params should be [injected via env](https://github.com/redhat-appstudio/build-definitions/blob/082637759f23a43d694823aa5a7fc1316c0f7c01/task/buildah/0.1/buildah.yaml#L70-L88) and referenced as properly quoted env vars

chmeliik, line 224: How is the performance of `buildah build` inside `podman run`?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -0,0 +1,396 @@
...
+      name: ssh
+      readOnly: true
+    workingDir: $(workspaces.source.path)
+  - image: quay.io/redhat-appstudio/syft:v0.85.0
+    name: sbom-syft-generate
+    resources: {}
+    script: |
+      syft dir:$(workspaces.source.path) --file=$(workspaces.source.path)/sbom-source.json --output=cyclonedx-json
+      find $(cat /workspace/container_path) -xtype l -delete
+      syft dir:$(cat /workspace/container_path) --file=$(workspaces.source.path)/sbom-image.json --output=cyclonedx-json
```
chmeliik, line 249: Does Syft work properly when inspecting images built for arches other than x86_64?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -0,0 +1,396 @@
...
+    workingDir: $(workspaces.source.path)
+  - args:
+    - attach
+    - sbom
+    - --sbom
+    - sbom-cyclonedx.json
+    - --type
+    - cyclonedx
+    - $(params.IMAGE)
+    image: quay.io/redhat-appstudio/cosign:v2.1.1
```
chmeliik, line 383: So each arch-specific image has an SBOM associated with it and the combined index image does not. Is that what we want? And will the [show-sbom](https://github.com/redhat-appstudio/build-definitions/tree/main/task/show-sbom/0.1) task work?

task/buildah-remote/0.1/buildah-remote.yaml
chmeliik: (it's OK to defer this to a follow-up story IMO, let's just make sure we have one if we can't answer this now)

task/buildah-remote/0.1/README.md
stuartwdouglas: I will create a follow up issue for this, I don't think we need it for the initial merge

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: Yes, syft has support for reading from different architectures. I know this because the older version we were originally using had a bug in its handling of big endian architectures, and upgrading to a later version fixed it (while investigating the bug I poked through the code and found that it explicitly has handling for different architectures).

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: Its still an open question. show-sbom will work if you point it at one of the per-arch images, not the manifest image (which is ok for an initial impl in my opinion).

We could make it show each sbom, however that would be a lot of duplication in the output.

task/buildah-remote/0.1/README.md
stuartwdouglas: https://issues.redhat.com/browse/STONEBLD-1757

task/buildah-multiplatform/0.1/buildah-multiplatform.yaml
stuartwdouglas: I have renamed this task to `build-image-manifest`. We will likely replace buildah at some point the name should  not be tied to the implementation.

stuartwdouglas:
I think this should be ready to merge now

tkdchen (approved):
LGMT. A minor thing is if there is nothing to be composed together by kustomize, no need to write a kustomization.yaml.

stuartwdouglas:
@tkdchen I didn't notice your comment until after I hit the button, I filed: https://github.com/redhat-appstudio/build-definitions/pull/598

PR #586, opened 2023-09-20
Clone source code into a dedicated directory

mkosiarc:
The source code will be present in the source directory in the root of
the workspace.

[STONEBLD-1570](https://issues.redhat.com//browse/STONEBLD-1570)

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -143,10 +143,11 @@ spec:
       #!/bin/sh
       set -o verbose
       cd $(workspaces.source.path)
-      if [ -e "$CONTEXT/$DOCKERFILE" ]; then
-        dockerfile_path="$CONTEXT/$DOCKERFILE"
-      elif [ -e "$DOCKERFILE" ]; then
-        dockerfile_path="$DOCKERFILE"
+      SOURCE_CODE_DIR=source
+      if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
+      elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
       elif echo "$DOCKERFILE" | grep -q "^https\?://"; then
         echo "Fetch Dockerfile from $DOCKERFILE"
         dockerfile_path=$(mktemp --suffix=-Dockerfile)
@@ -158,7 +159,7 @@ spec:
         http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path.dockerignore.tmp" "$DOCKERFILE.dockerignore")
         if [ $http_code = 200 ]; then
           echo "Fetched .dockerignore from $DOCKERFILE.dockerignore"
-          mv "$dockerfile_path.dockerignore.tmp" $CONTEXT/.dockerignore
+          mv "$dockerfile_path.dockerignore.tmp" $SOURCE_CODE_DIR/$CONTEXT/.dockerignore
         fi
       else
         echo "Cannot find Dockerfile $DOCKERFILE"
@@ -208,7 +209,7 @@ spec:
         ${LABELS[@]} \
         --tls-verify=$TLSVERIFY --no-cache \
         --ulimit nofile=4096:4096 \
-        -f "$dockerfile_path" -t $IMAGE $CONTEXT
+        -f "$dockerfile_path" -t $IMAGE $SOURCE_CODE_DIR/$CONTEXT
 
       container=$(buildah from --pull-never $IMAGE)
       buildah mount $container | tee /workspace/container_path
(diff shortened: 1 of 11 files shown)


pipelines/template-build/template-build.yaml
Michkov, line 1: temlate-build.yaml should not be updated since it could cause upgrade issue. Renovate bot is updating tasks, but PipelineRun definition stays the same.

task/buildah/0.1/buildah.yaml
```diff
@@ -100,10 +100,10 @@ spec:
     - name: COMMIT_SHA
       value: $(params.COMMIT_SHA)
     script: |
-      if [ -e "$CONTEXT/$DOCKERFILE" ]; then
-        dockerfile_path="$CONTEXT/$DOCKERFILE"
-      elif [ -e "$DOCKERFILE" ]; then
-        dockerfile_path="$DOCKERFILE"
+      if [ -e "source/$CONTEXT/$DOCKERFILE" ]; then
+        dockerfile_path="source/$CONTEXT/$DOCKERFILE"
```
tisutisu, line 104: Seeing multiple places in this file `source` is used. Will it be good to declare a variable with value `source`?

MartinBasti: use something else than `source` which is bash command

MartinBasti: ah nevermind :D

mmorhun (commented):
Do we need to change `buildah-remote`?

mkosiarc:
> Do we need to change `buildah-remote`?

Hmm, I guess so. I was not aware of that previously, as it is a new task. It will have to be updated and tested.

task/buildah/0.1/buildah.yaml
mkosiarc: I am not exactly sure of the benefit here. Most of the time the source dir is changed in workingDir as
```
workingDir: $(workspaces.source.path)/source
```

Where I don't see a good way to declare that as a variable. I could change it just in the buildah tasks, for the 3 occurences, but that is not any big improvement. Should I go ahead and update if for the buildah task?

tisutisu: I think declaring source directory as variable would be sufficient .. not sure about if there is a way to avoid repetitions in different task

tkdchen (approved):
:+1:

A minor question, any consideration not change `workingDir` for the `build` step?

mkosiarc:
> 👍
> 
> A minor question, any consideration not change `workingDir` for the `build` step?

Because in that case I would have to update other paths that aren't supposed to be in the **source** directory.

For example,
```
mv cachi2 /tmp/
```

But it actually might be better solution..

task/buildah/0.1/buildah.yaml
mkosiarc: updated

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -387,7 +394,8 @@ spec:
     resources: {}
     workingDir: $(workspaces.source.path)
   volumes:
-  - emptyDir: {}
+  - emptyDir:
+      medium: Memory
```
tkdchen, line 398: Do we need to keep this change and the above made to `resources`?

mkosiarc: I don't see a reason why we wouldn't. The buildah task itself was changed 3 days ago with this commit - https://github.com/redhat-appstudio/build-definitions/commit/f8c6a108b3be411b54075941b54142fb2b49fa5c

and the buildah-task is generated from the buildah task, so the changes are valid from my POV

tkdchen: Could  you please make these changes in a separate commit? The changes look like a sync update with https://github.com/redhat-appstudio/build-definitions/commit/f8c6a108b3be411b54075941b54142fb2b49fa5c for buildah-remote. But, if it is related to the source directory change, please describe it in the commit message.

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -314,7 +315,7 @@ spec:
     securityContext:
       runAsUser: 0
     workingDir: $(workspaces.source.path)
-  - image: registry.access.redhat.com/ubi9/python-39:1-133.1692772345
+  - image: registry.access.redhat.com/ubi9/python-39:1-143.1695131427
```
tkdchen, line 318: Looks another irrelevant change.

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: I am working on a way to make PRs fail if they are not in sync.

stuartwdouglas: https://github.com/redhat-appstudio/build-definitions/pull/614

stuartwdouglas: That PR is now passing, once it is merged you should be able to rebase to remove the unrelated changes.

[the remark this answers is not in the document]
mkosiarc: updated

mkosiarc: @tkdchen Updated, removed all other changes - Stuart's #614 will take care of it.

mmorhun (approved):
Tested user flow for docker build, hermetic build and s2i pipelines. Everything worked as expected.

dirgim (approved):
LGTM concerning the changes to the `sast-snyk-check` task

PR #587, opened 2023-09-20
fix(RHTAPBUGS-761): return TEST_OUTPUT when task skipped

MartinBasti:
Fixing a bug when task is not returning skip result when snyk token is undefined


jsztuka (approved):
lgtm

dirgim (approved):
LGTM

MartinBasti:
rebased

PR #599, opened 2023-09-29
fix(STONEINTG-603) remove insecure connection from curl

sonam1412:
curl -k runs insecure connection to the server. This is a security issue to skip SSL certificate errors by adding -k or --insecure option in curl commands. 
This PR is to remove the -k option from the curl command of the deprecated-image-check


dirgim (approved):
LGTM

MartinBasti:
We should also fix older versions of this task

jsztuka (approved):
lgtm

MartinBasti (approved):

PR #606, opened 2023-10-03
fix(clamav): inform users that scan started

MartinBasti:
Instead of telling users that extraction is done and let them to look at empty screen, tell them that scan is in progress and may take a while


sonam1412 (approved):
LGTM

Josh-Everett (approved):

dirgim (approved):
LGTM, just needs a rebase

PR #613, opened 2023-10-09
nudge for a new buildah tekton bundle

arewm:


gabemontero (approved):
/lgtm

task/buildah-10gb/0.1/patch.yaml
```diff
@@ -3,7 +3,7 @@
   value: buildah-10gb
 - op: replace
   path: /spec/steps/0/resources/limits/memory
-  value: 10Gi
+  value: 10.1Gi
```
gabemontero, line 6: I did confirm via some internet searching that you can say something like 10.1Gi

Can't remember the last time I had actually done that

PR #614, opened 2023-10-10
STONEBLD-1757 validate buildah-remote task

stuartwdouglas:


mkosiarc (approved):

hack/generate-buildah-remote.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+DIR=`dirname $0`
+docker run -v $DIR/..:/data quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-221979523603b730feca730aeb9d43f24a3c1e67  --buildah-task=/data/task/buildah/0.1/buildah.yaml --remote-task=/data/task/buildah-remote/0.1/buildah-remote.yaml
```
tkdchen, line 4: Need `:Z` here?

tkdchen (approved):

hack/generate-buildah-remote.sh
stuartwdouglas: Why?

chmeliik: For SELinux: https://docs.podman.io/en/latest/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options
(see "Labeling systems like SELinux ..."

It does work for me without the :Z though, odd. Usually I need to add it for volume mounts

chmeliik:
The script doesn't work for me:
```
$ hack/generate-buildah-remote.sh
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: creating named volume "hack/..": running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
```
With the following changes, it does:
```diff
diff --git a/hack/generate-buildah-remote.sh b/hack/generate-buildah-remote.sh
index 107cbcd..49b6571 100755
--- a/hack/generate-buildah-remote.sh
+++ b/hack/generate-buildah-remote.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
-DIR=`dirname $0`
-docker run -v $DIR/..:/data quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-221979523603b730feca730aeb9d43f24a3c1e67  --buildah-task=/data/task/buildah/0.1/buildah.yaml --remote-task=/data/task/buildah-remote/0.1/buildah-remote.yaml
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+docker run -v "$SCRIPTDIR"/..:/data quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-221979523603b730feca730aeb9d43f24a3c1e67  --buildah-task=/data/task/buildah/0.1/buildah.yaml --remote-task=/data/task/buildah-remote/0.1/buildah-remote.yaml
```

(the SCRIPTDIR approach is used by some other hack scripts)

stuartwdouglas:
I have changed it to the SCRIPTDIR approach, although I don't really understand why it was failing for you with the dirname $0 approach.

hack/generate-buildah-remote.sh
tkdchen: oh, odd. It didn't work for me without Z even for read-only operations in the past, so I always add it for volume mounts.

chmeliik:
> I have changed it to the SCRIPTDIR approach, although I don't really understand why it was failing for you with the dirname $0 approach.

podman requires absolute paths for `-v /HOST-DIR:/CONTAINER-DIR`. Maybe that's different for docker

PR #620, opened 2023-10-13
Use the in-cluster ECP resources by default

lcarva:
Ref: https://issues.redhat.com/browse/EC-143


lcarva:
/retest

quay.io was down which is likely the root cause of the failure

simonbaird (approved):

PR #621, opened 2023-10-17
STONEBLD-1605: Add source image build task

tkdchen:
* Task is added to the pipelines and will be available inside fbc, docker, nodejs and java build pipelines. build-container and this task runs in sequence because results.BASE_IMAGES_DIGESTS is required.
* Add a Dockerfile to build a utility image used for the build.
* The utility image is build and pushed to quay.io/redhat-appstudio. Use the `make image push'.
* Sources from base image are included. Multitage build is supported, only the base image used in the last stage is handled.
* Prefetched sources of dependencies are included if hermetic build is enabled.
* Make git ignore any directory with name scripts under task/source-build/ directory, so that the task inline script can be edited separately for convenience.


.gitignore
```diff
@@ -0,0 +1 @@
+task/source-build/**/scripts/
```
rcerven, line 1: why is this needed?

containers/source-build/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# image: build-definitions-source-image-utils:latest
+FROM registry.access.redhat.com/ubi9/ubi:9.2-722
+ARG release_url=https://github.com/containers/BuildSourceImage/archive/refs/tags/v0.2.0.tar.gz
+ARG bin_dir=/opt/BuildSourceImage
+ARG patch0=0001-cleanup-directory-with-all-rpms-which-isn-t-used-any.patch
+RUN dnf update -y && dnf install -y python3.11 git jq skopeo file tar && dnf clean all
+ADD $patch0 /tmp
+RUN mkdir -p /opt/BuildSourceImage/ &&  \
```
rcerven, line 8: use 'bin_dir'

.gitignore
tkdchen: It's hard to edit the inline Python code inside the YAML file, so I edit the code by creating a file under this directory, then copy/paste back to the YAML.

containers/source-build/0001-cleanup-directory-with-all-rpms-which-isn-t-used-any.patch
```diff
@@ -0,0 +1,28 @@
+From 9d66f316158bc84d23a64dfae0a4481cb3a14227 Mon Sep 17 00:00:00 2001
```
tisutisu, line 1: Is this file got added by mistake?

tkdchen: No. If you look at the Dockerfile, you see that this patch is applied to `BuildSourceImage.sh`.

.gitignore
rcerven: so this is  just temporary right? so it shouldn't be in the PR, but only in your repo

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+        import urllib.parse
+        from dataclasses import dataclass, field
+        from subprocess import run, CalledProcessError
+        from typing import TypedDict, NotRequired, Literal, Final
+
+        """
+        Requires: git, skopeo, tar, BuildSourceImage
+        """
+
+        logging.basicConfig(level=logging.DEBUG, format="%(asctime)s:%(name)s:%(levelname)s:%(message)s")
```
tisutisu, line 91: will it be good to change logging level to `INFO` when testing is complete?

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+        - "/var/source-build"
+        - "--source-dir"
+        - "$(workspaces.workspace.path)/source"
+        - "--base-images"
+        - "$(params.BASE_IMAGES)"
+        - "--write-result-to"
+        - "$(results.BUILD_RESULT.path)"
+        - "--cachi2-artifacts-dir"
+        - "$(workspaces.workspace.path)/cachi2"
+      script: |
```
MartinBasti, line 68: Seriously 400 lines long python script inlined task?

Why it's not part of the image?

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+                    dest_dir = create_dir(extra_src_dir, extra_src_archive)
+                    sib_dirs.extra_src_dirs.append(dest_dir)  # save this directory for bsi
+
+                    extract_blob_member(digest, blob_member, str(dest_dir),
+                                        rename_to=extra_src_archive,
+                                        work_dir=base_sources_extraction_dir,
+                                        log=log)
+
+                    # FIXME: perhaps the dependency archive (the Cachi2) might be handled differently
+                    run(["tar", "xvf", extra_src_archive], check=True, cwd=dest_dir)
```
rcerven, line 295: why are you unpacking all individual remote sources ?

in osbs source image there is "extra_src_dir" and each cachito source archive is in individual layer

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+                    os.unlink(f"{dest_dir}/{extra_src_archive}")
+
+                else:
+                    log.warning("No known operation happened on layer %s", digest)
+
+            return len(os.listdir(sib_dirs.rpm_dir)) > 0 or len(sib_dirs.extra_src_dirs) > 0
```
rcerven, line 301: sib_dirs.extra_src_dirs will most likely already have 1 element with sources from make_source_archive

so it should be len(sib_dirs.extra_src_dirs) > 1

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+            log = logging.getLogger("source-build.base-image-sources")
+            parts = urllib.parse.urlparse(f"docker://{image}")
+            if parts.netloc != "registry.access.redhat.com":
+                log.info("Image %s does not come from supported registry. Skip handling the sources from this image.", image)
+                return False
+
+            base_image_sources_dir = create_dir(work_dir, "base_image_sources")
+            base_sources_extraction_dir = create_dir(base_image_sources_dir, "extraction_dir")
+
+            source_image_name = image + "-source"
```
rcerven, line 234: and what will create image with such name? because osbs doesn't name it like that

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+        def gather_prefetched_sources(work_dir: str, cachi2_dir: str, sib_dirs: SourceImageBuildDirectories) -> bool:
+            log = logging.getLogger("source-build.prefetched-sources")
+
+            # Guess if hermetic build is enabled
+            # NOTE: this guess does depend on how cachi2 runs inside prefetch-dependencies task.
+            cachi2_output_dir = f"{cachi2_dir}/output"
+            cachi2_env = f"{cachi2_dir}/cachi2.env"
+
+            has_cachi2_output = os.path.exists(cachi2_output_dir) and os.path.isdir(cachi2_output_dir)
+            has_cachi2_env = os.path.exists(cachi2_env) and os.path.getsize(cachi2_env) > 0
```
rcerven, line 313: cachito env file can be actually empty depending on pkg managers used and their needs

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+            if not (has_cachi2_output and has_cachi2_env):
+                log.info("Cannot discover artifacts generated by cachi2. Seems hermetic build is not enabled.")
+                return False
+
+            def _find_prefetch_source_archives():
```
rcerven, line 318: I don't understand this method at all

cachito output has directory structure

└── deps

    └── gomod

        └── pkg

            └── mod

                └── cache

                    └── download

                        ├── github.com
and in there directories for specific sources

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+                for root, dirs, files in os.walk(cachi2_output_dir):
+                    for filename in files:
+                        is_ext = filename.endswith
+                        # .zip for gomod, .tgz for pip, .tar.gz for npm
+                        if not (is_ext(".zip") or is_ext(".tgz") or is_ext(".tar.gz")):
```
rcerven, line 323: and this will use just files which aren't archives, so what exactly are you using?
but again it is at top level directory so there won't even be anything

task/source-build/0.1/source-build.yaml
rcerven: so what you should do is grab whole outputs/deps and tar.gz it and use that

rcerven: and I'd include also cachito2.env file there

task/source-build/0.1/source-build.yaml
tkdchen: I would keep it in DEBUG so that we can get detailed information in case some issue happens in prod.

task/source-build/0.1/source-build.yaml
tkdchen: `extract_blob_member` extracts the tar archive generated by bsi, like `extra_src_0.tar`, then this line extracts the actual source archive, e.g. a remote source in osbs. Then, bsi builds the source image with each of such sources again. As a result, when a user gets the source archive, for example `extra_src_0.tar`, the user is able to get the actual source by unpacking it just once.

task/source-build/0.1/source-build.yaml
tkdchen: This initial implementation handles images pulled from registry.access.redhat.com firstly, for instance the images under registry.access.redhat.com/ubi9/, which may be ubi, python and nodejs. I don't know which tool creates such name. How does osbs name source image now?

task/source-build/0.1/source-build.yaml
tkdchen: Good catch.

task/source-build/0.1/source-build.yaml
tkdchen: > and I'd include also cachito2.env file there

It is included. The code is below.

tkdchen: Here is an example of the output directory:

```
cachi2-output2/
├── bom.json
└── deps
    ├── gomod
    │   └── pkg
    │       └── mod
    │           └── cache
    │               └── download
    │                   ├── github.com
    │                   │   ├── davecgh
    │                   │   │   └── go-spew
    │                   │   │       └── @v
    │                   │   │           ├── list
    │                   │   │           ├── v1.1.0.mod
    │                   │   │           └── v1.1.1.mod
    │                   │   ├── gogo
    │                   │   │   └── protobuf
    │                   │   │       └── @v
    │                   │   │           ├── list
    │                   │   │           ├── v1.3.2.info
    │                   │   │           ├── v1.3.2.lock
    │                   │   │           ├── v1.3.2.mod
    │                   │   │           ├── v1.3.2.zip
    │                   │   │           └── v1.3.2.ziphash
    ├── npm
    │   ├── accepts-1.3.8.tgz
    │   ├── array-flatten-1.1.1.tgz
    │   └── vary-1.1.2.tgz
    └── pip
        ├── Flask-2.3.1.tar.gz
        └── requests-2.30.0.tar.gz
```

`_find_prefetch_source_archives` gathers each of the archive and remain the path which is relative to the output directory. As a result, when extract sources from the source image, extracting all `extra-src-*.tar` will get the original output directory tree populated by cachi2.

> so what you should do is grab whole outputs/deps and tar.gz it and use that

Do you mean make a tar archive from `outputs/deps` and include all files under the `deps/` directory?

I remember we discussed that we want one source per layer. My understanding is one source means a source archive, a .tgz for npm, a .tar.gz for pip and a .zip for gomod.

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,472 @@
...
+
+            work_dir = create_dir(workspace_dir, "source-build")
+            logger.debug("working directory %s", work_dir)
+
+            sib_dirs = SourceImageBuildDirectories()
+
+            make_source_archive(work_dir, args.source_dir, sib_dirs)
+
+            # Handle base image sources
+            if args.base_images:
```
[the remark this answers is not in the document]
tkdchen: It is just for convenience for now. The input value comes from the `tasks.build-container.results.BASE_IMAGES_DIGESTS` without change, which might include multiple base images. Meanwhile, currently we only handle the last one, but passing `BASE_IMAGES_DIGESTS` directly could be flexible for the future if we need to handle base images in a build stage, or do more even for the last base image.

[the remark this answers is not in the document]
tkdchen: In addition, the implementation does not do any check about the order of input base images. It trusts task build-container sets `BASE_IMAGES_DIGESTS` correctly.

.gitignore
tkdchen: This script should not be included in this pr. I'm using it to edit the script outside the task.

task/source-build/0.1/source-build.yaml
tkdchen: My bad. Fixed in the new commits.

task/source-build/0.1/source-build.yaml
tkdchen: Fixed.

task/source-build/0.1/source-build.yaml
tkdchen: Updated. Now, using a flag variable instead.

.gitignore
mmorhun: I think it should stay in your local repository, because different developers have own habits.

task/source-build/0.1/source-build.yaml
rcerven: I'll say it again, you are extracting extra_src_0.tar remote source from osbs source image with cachito remote sources, and ADDing extracted sources to source image, you should be adding that tar, not extracting it

task/source-build/0.1/source-build.yaml
rcerven: it is just unique name like : registry-proxy-stage.engineering.redhat.com/rh-osbs-stage/osbs-test-distscope-base:osbs-test-rhel-8-containers-x86_64-candidate-66403-20231019000900-x86_64

task/source-build/0.1/source-build.yaml
rcerven: oh I see I forgot that extra_src_0.tar is created by bsi and it contains tar.gz from cachito

task/source-build/0.1/source-build.yaml
rcerven: ok it makes sense now

so you are basically finding archives and putting them in the same path as in that cachito output

ok in that case it looks good now

task/source-build/0.1/source-build.yaml
rcerven: it will work also fro cachito2 sources, because you will extract the archive from cachito2 and get that directory structure with archive, and bsi will again make extra_src.tar from it, so it works the same

task/source-build/0.1/source-build.yaml
arewm: I think it would be better to include this script as part of the image used by the task to limit the amount of in-line script as well as enabling linting, easier modification, tests, etc.

containers/source-build/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.2-722
+ARG bsi_source=https://github.com/containers/BuildSourceImage/archive/refs/tags/v0.2.0.tar.gz
+ARG bin_dir=/opt/BuildSourceImage
+ARG patch0=0001-cleanup-directory-with-all-rpms-which-isn-t-used-any.patch
```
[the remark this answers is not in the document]
tkdchen: Nothing special, probably a natural thought coming from the rpm build :sweat_smile:

> Cannot we update the script directly or add a parameter?

Could be. Current Dockerfile includes all the things to do, and outside the Dockerfile, nothing needs to do but just `make image push`. I think once there will be a new release of BuildSourceImage, we will cleanup this Dockerfile by downloading the new release archive.

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,479 @@
...
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "appstudio, hacbs"
+spec:
+  description: Source image build.
+  params:
+    - name: BINARY_IMAGE
+      description: The corresponding binary image based on which to generate source image.
```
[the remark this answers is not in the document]
tkdchen: Phrase "based on" I used may be ambiguous. The binary image is actually not a base of the source image. I'm going to make this change to simplify:

> Binary image name.

WDYT?

pipelines/template-build/template-build.yaml
```diff
@@ -119,6 +123,27 @@ spec:
       workspaces:
         - name: source
           workspace: workspace
+    - name: build-source-image
```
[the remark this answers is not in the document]
tkdchen: What is the "actual build"? And I think I don't understand your question.

task/source-build/0.1/source-build.yaml
tkdchen: for Go, the other files should be metadata files. For the purpose of shipping sources, not sure how much helpful by including those metadata files.

The above example includes two different kinds of Go package directories, `go-spew` does not have archive inside, include it in the source image as well?

pipelines/template-build/template-build.yaml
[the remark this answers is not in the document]
mmorhun: I meant the resulting binary image of the user's Component.

task/source-build/0.1/source-build.yaml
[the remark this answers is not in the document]
mmorhun: Do you mean image with user's Component binary?

task/source-build/0.1/source-build.yaml
chmeliik: Selecting files base on extension will be error-prone. For example, python sdists historically allowed not only `tar.gz`, but all of these extensions - all of which cachi2 supports. https://github.com/containerbuildsystem/cachi2/blob/84ac71f613e2f51becbc7d5b90ccf4c9c72e2768/cachi2/core/package_managers/pip.py#L56
```python
[".zip", ".tar.gz", ".tar.bz2", ".tar.xz", "tar.Z", ".tar"]
```
Once cachi2 implements Python wheel support, there will also be `*.whl` files (which are zip files but with a different extension). I don't remember what the decision was on those, IIRC we wanted to include them?

With Ruby support, there will be `*.gem` files (which are also zip files I think)

chmeliik: And there was an idea floating around that source containers could contain everything needed for a re-build. In which case we should really dump the whole cachi2/ directory in the container, not filter anything out. @arewm did that idea progress anywhere?

containers/source-build/source_build.py
```diff
@@ -0,0 +1,452 @@
...
+        "name": repo_name,
+        "last_commit_sha": commit_sha,
+    }
+
+
+def parse_cli_args():
+    parser = argparse.ArgumentParser(description="Source image build task")
+    parser.add_argument(
+        "-w",
+        "--workspace",
```
chmeliik, line 84: When trying to test this locally, I found out that `--workspace` only works as an absolute path, with a relative path the `git archive` operation fails

```shell
mkdir source-ws
python containers/source-build/source_build.py \
  --source ../build-service \
  --workspace source-ws \
  --output-binary-image foo
```
```
subprocess.CalledProcessError: Command '['git', 'archive', '--prefix', '-99cebd0a67a6b25b8ccffb76522861f526c762de/', '--output', 'source-ws/source-build/source_archive/-99cebd0a67a6b25b8ccffb76522861f526c762de.tar.gz', 'HEAD']' returned non-zero exit status 128.
```

containers/source-build/source_build.py
```diff
@@ -0,0 +1,452 @@
...
+
+def make_source_archive(
+    work_dir: str, source_dir: str, sib_dirs: SourceImageBuildDirectories
+) -> None:
+    log = logging.getLogger("build-source.source-archive")
+    source_archive_dir = create_dir(work_dir, "source_archive")
+    repo_info = get_repo_info(source_dir)
+    name_sha = f"{repo_info['name']}-{repo_info['last_commit_sha']}"
+    output_archive = f"{source_archive_dir}/{name_sha}.tar.gz"
+    git_cmd = ["git", "archive", "--prefix", name_sha + "/", "--output", output_archive, "HEAD"]
```
chmeliik, line 335: So this takes the clean state at the latest commit and throws away the modifications that we do, right? Like the `sed` commands in the buildah task, `cachi2 inject-files` etc.

I guess that's what we want?

task/source-build/0.1/source-build.yaml
arewm: It is still an idea, but not one that has been written up into a feature and scoped. If we can achieve this or work towards it even, then great! If not, then we can defer until it becomes a priority. Additionally, if this work will be harder for some languages, it might need to be added with incremental language support.

tkdchen: Hi, thank you all for feedback and information. This is the initial implementation for phase 1. I would keep it for now for the phase 1, then improve a little in the phase 2. I'll create a story for the archive type detention for the phase 2 that does not rely on the file extension as much as possible.

pipelines/template-build/template-build.yaml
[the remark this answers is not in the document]
tkdchen: `tasks.build-container.results.BASE_IMAGES_DIGESTS` is required.

task/source-build/0.1/source-build.yaml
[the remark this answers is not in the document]
tkdchen: > I meant the resulting binary image of the user's Component.

The binary image you mentioned above. :)

containers/source-build/source_build.py
chmeliik: Another thing - the archive also excludes the .git/ directory, so the source container doesn't include the git history.

That's probably fine for the first implementation, but could be another issue if we're going to use source containers as the only source for re-builds (https://github.com/redhat-appstudio/build-definitions/pull/621#discussion_r1366959493)

chmeliik: We can change that if/when needed, although there will be a slight issue that some existing source containers will not be usable for re-builds

containers/source-build/source_build.py
```diff
@@ -0,0 +1,452 @@
...
+
+    # push to registry
+    push_cmd = [
+        "skopeo",
+        "copy",
+        f"oci://{image_output_dir}:latest-source",
+        f"docker://{dest_image}",
+    ]
+    log.debug("push source image %r", push_cmd)
+    run(push_cmd, check=True)
```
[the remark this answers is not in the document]
tkdchen: Good point. Will open a story for it.

containers/source-build/source_build.py
tkdchen: fixed

containers/source-build/source_build.py
```diff
@@ -0,0 +1,455 @@
...
+        log.info(
+            "Image %s does not come from supported registry. Skip handling the sources from it.",
+            image,
+        )
+        return False
+
+    base_image_sources_dir = create_dir(work_dir, "base_image_sources")
+    base_sources_extraction_dir = create_dir(base_image_sources_dir, "extraction_dir")
+
+    source_image_name = image + "-source"
```
chmeliik, line 184: I think this is not quite that simple. For example, for image = `registry.access.redhat.com/ubi9/ubi:latest`, there is no `registry.access.redhat.com/ubi9/ubi:latest-source`

```bash
$ skopeo inspect --raw docker://registry.access.redhat.com/ubi9/ubi:latest-source
FATA[0000] Error parsing image name "docker://registry.access.redhat.com/ubi9/ubi:latest-source": reading manifest latest-source in registry.access.redhat.com/ubi9/ubi: manifest unknown
```

But you can find the source image like this:

```bash
$ skopeo inspect --config docker://registry.access.redhat.com/ubi9/ubi:latest | jq '.config.Labels | "\(.version)-\(.release)"' -r
9.2-755.1697625012

$ skopeo inspect --raw docker://registry.access.redhat.com/ubi9/ubi:9.2-755.1697625012-source
...
```

That is: find the unique `$version-$release` tag, append `-source` to it -> that's the source image (I believe that's how Conjur tags them)

containers/source-build/source_build.py
```diff
@@ -0,0 +1,455 @@
...
+        blob_member = ""
+        symlink_member = ""
+        with tarfile.open(f"{base_sources_extraction_dir}/{digest}", "r:gz") as tar:
+            for member in tar:
+                if member.isfile():
+                    blob_member = member.name
+                elif member.issym():
+                    symlink_member = member.name
+
+        if symlink_member.startswith("./rpm_dir/"):
```
chmeliik, line 231: Are the file paths in the manifest guaranteed to start with `./`? This feels like we're relying on an implementation detail

task/source-build/0.1/source-build.yaml
tkdchen: After rethinking of this, I changed the way to detect the archive, that is based on mimetype. Then, when the requirements mentioned above are clear, we can reconsider a better way to dump the whole cachi2/ directory in the container. For the initial implementation, the task is able to cover most of the cases by not enumerating file extensions.

containers/source-build/source_build.py
```diff
@@ -0,0 +1,455 @@
...
+    return value.strip()
+
+
+def get_repo_info(repo_path: str) -> RepoInfo:
+    _run = functools.partial(run, check=True, text=True, capture_output=True, cwd=repo_path)
+    commit_sha = _run(["git", "rev-parse", "HEAD"]).stdout.strip()
+    repo_url = _run(["git", "config", "--get", "remote.origin.url"]).stdout.strip()
+    # The url could look like https://github.com/namespace/app.git
+    # where, app is required.
+    repo_name, _ = os.path.splitext(repo_url.rsplit("/", maxsplit=1)[-1])
```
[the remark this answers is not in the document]
tkdchen: > Should we remove the .git suffix from the repo name?

I'm not sure I get your point. `os.path.splitext` is called.

> Should we handle urls with trailing /? E.g. https://github.com/namespace/app.git/

This is got from the repo cloned by git-clone. Could `remote.origin.url` return such a url?

containers/source-build/source_build.py
```diff
@@ -0,0 +1,455 @@
...
+    image: str, work_dir: str, sib_dirs: SourceImageBuildDirectories
+) -> bool:
+    log = logging.getLogger("source-build.base-image-sources")
+    parts = urllib.parse.urlparse(f"docker://{image}")
+    if parts.netloc != "registry.access.redhat.com":
+        log.info(
+            "Image %s does not come from supported registry. Skip handling the sources from it.",
+            image,
+        )
+        return False
```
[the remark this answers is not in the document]
tkdchen: The investigation document just covers registry.access.redhat.com for the initial implementation. registry.redhat.io requires authentication. AFAIK, I haven't seen builds in the pipelines. Or, maybe I missed something. @arewm do you have any input?

containers/source-build/source_build.py
tkdchen: Yes, relying on bsi's implementation detail :) I did this to avoid unpacking all the SRPMs and possible extra sources with `tar xvf`.

containers/source-build/source_build.py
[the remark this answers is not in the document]
chmeliik: Ah, my bad, I missed the splitext

> This is got from the repo cloned by git-clone. Could remote.origin.url return such a url?

Yeah, it can
```bash
$ git clone git@github.com:containerbuildsystem/cachi2.git/
$ cd cachi2
$ git config --get remote.origin.url
containerbuildsystem/cachi2.git/
```

containers/source-build/source_build.py
tkdchen: Created a story for this. Thanks.

containers/source-build/source_build.py
chmeliik: I was going to suggest something like this to make it more robust:
```python
if Path(symlink_member).parts[0] == "rpm_dir":
    ...
```
But we also rely on the leading `./` when extracting (`# strip 3 components: ./blobs/sha256`)

We could change the extraction to be done by python's tarfile library and then we wouldn't have to rely on the `./`. But for now, relying on the implementation detail is fine

containers/source-build/app/source_build.py
```diff
@@ -0,0 +1,465 @@
...
+    for root, filename in _find_prefetch_source_archives():
+        src_dir = f"src-{next(source_counter)}"
+        copy_dest_dir = f"{prepared_sources_dir}/{src_dir}/{relative_to(root, cachi2_output_dir)}"
+        os.makedirs(copy_dest_dir)
+
+        src = f"{root}/{filename}"
+        dest = f"{copy_dest_dir}/{filename}"
+        log.debug("copy prefetched source %s to %s", src, dest)
+        shutil.copy(src, dest)
+        sib_dirs.extra_src_dirs.append(f"{prepared_sources_dir}/{src_dir}")
```
chmeliik, line 319: So currently, this is one `extra_src_dir` for each archive in the cachi2 output, right? And, if I'm not mistaken, BSI puts each extra_src_dir in a separate layer?

This can easily result in source containers with 1000s of layers. Is that OK?

MartinBasti: Original idea of source containers was that because each source is in separate layer, it will be a mechanism for deduplication of data, as quay can reuse layers in multiple image. I'm not sure how this is working in practice, any byte change would result into different digest and no data deduplication is happening.

Also in OCI each layer can have annotation (and BSI should be filling this), so you should be able to find exact source by checking just index annotations, and the just fetch the source you care about.

chmeliik: Ok, sounds like more layers is good then. The archive fetched by cachi2 for a specific name==version of a dependency should be byte-for-byte identical every time (if the user is verifying checksums)

MartinBasti: > Unless filesystem metadata matters as well (timestamp)
This is what bothers me, but there is no other way than testing it :)

chmeliik: Hmm, but I ran into an issue when testing this with quay

```bash
python containers/source-build/app/source_build.py \
  --bsi /tmp/BSI/BuildSourceImage.sh \
  -s /tmp/test/quay \
  -w source-ws \
  --cachi2-artifacts-dir /tmp/test/cachi2 \
  --output-binary-image foo 1>stdout.log 2>stderr.log
```
```bash
$ grep 'too long' stderr.log
/tmp/BSI/BuildSourceImage.sh: line 1049: /tmp/work/build-definitions/source-ws/source-build/bsi_build/tmp/SrcImg.S4EPU8/driver/extra_src_dir/extra-src-011111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111.tar.json: File name too long
```
Fun fact: line 2775 in stderr.log (the one just before the error, which logs the bsi command) is 243_000 characters long :smile:

chmeliik: It looks like BSI will only be able to handle less than 300 extra sources because of the naming it uses for the files (0, 01, 011, 0111, ...)

chmeliik: And yes, it is just a stupid bash bug in BSI :smile: 

I'm pretty sure that the intention wasn't

```bash
counter=0
counter+=1
# now counter is "01"
```
but rather
```bash
counter=0
((counter+=1))
# now counter is "1"

Comment (MartinBasti) [reply containers/source-build/app/source_build.py:319]: buildah source?

Comment (tkdchen) [reply containers/source-build/app/source_build.py:319]: I thought `((counter+=1))` is the one bsi and we want, but not sure `counter+=1` is intended.

Comment (chmeliik) [reply containers/source-build/app/source_build.py:315]: Maybe we can apply that as another patch in the Dockerfile? Until we drop BSI and move to `buildah source`?

File: task/source-build/0.1/source-build.yaml [hunk 24]
```diff
@@ -0,0 +1,81 @@
...
+        Base images used to build the binary image. Each image per line in the same order of FROM
+        instructions specified in a multistage Dockerfile. Default to an empty string, which means
+        to skip handling a base image.
+      type: string
+      default: ""
+  results:
+    - name: BUILD_RESULT
+      description: Build result.
+    - name: SOURCE_IMAGE_URL
+      description: The source image url.
```
simonbaird, line 28: I think it should include a additional result here named `SOURCE_IMAGE_DIGEST` to match the Tekton Chains hinting conventions. This will allow EC to verify the digest matches the expected value, and also allow chains to sign the source image.

tkdchen: Pushed a commit to output the digest. PTAL.

chmeliik:
Two main things left from my side:
- will fail for any build that uses cachi2 and downloads more than 237 dependencies https://github.com/redhat-appstudio/build-definitions/pull/621#discussion_r1369749446
  - IMO we need to address this now
- errors impossible to debug https://github.com/redhat-appstudio/build-definitions/pull/621#discussion_r1367132589
  - follow-up story is fine by me

containers/source-build/source_build.py
```diff
@@ -0,0 +1,452 @@
...
+def make_source_archive(
+    work_dir: str, source_dir: str, sib_dirs: SourceImageBuildDirectories
+) -> None:
+    log = logging.getLogger("build-source.source-archive")
+    source_archive_dir = create_dir(work_dir, "source_archive")
+    repo_info = get_repo_info(source_dir)
+    name_sha = f"{repo_info['name']}-{repo_info['last_commit_sha']}"
+    output_archive = f"{source_archive_dir}/{name_sha}.tar.gz"
+    git_cmd = ["git", "archive", "--prefix", name_sha + "/", "--output", output_archive, "HEAD"]
+    run(git_cmd, check=True, capture_output=True, cwd=source_dir)
```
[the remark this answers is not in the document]
tkdchen: So far, `capture_output` is set in two methods, for `skopeo inspect` and git operations in method `get_repo_info`. The former does not actually care about the output but just the exit code of the process, and the latter has to capture the output to get commit sha and repository URL.

Traceback is printed for any exception in `main`, which should provide enough information for debug once any of the mentioned commands fails.

`capture_output` has been removed from this line your commented out.

tkdchen:
> * will fail for any build that uses cachi2 and downloads more than 237 dependencies

Added a patch.

task/source-build/0.1/source-build.yaml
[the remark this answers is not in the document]
chmeliik: > Phrase "based on" I used may be ambiguous. The binary image is actually not a base of the source image. I'm going to make this change to simplify:
> 
> > Binary image name.
> 
> WDYT?

How about

"Binary image name. The source container image is pushed to ${BINARY_IMAGE}.src"

(If I understand correctly, that's the only use of the BINARY_IMAGE param)

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - name: SOURCE_IMAGE_DIGEST
+      description: The source image digest.
+  workspaces:
+  - name: workspace
+    description: Workspace where task works inside. Source code is included in this workspace and any artifacts generated by this task are also stored here.
+  volumes:
+    - name: source-build-work-place
+      emptyDir:
+        medium: Memory
```
chmeliik, line 37: Is the description for `workspace` correct? Doesn't the task work inside the source-build-work-place volume, which is a different volume than the workspace?

chmeliik (commented):
LGTM, just some minor nitpicks for the task definition

containers/source-build/app/source_build.py
```diff
@@ -0,0 +1,473 @@
...
+        log.info(
+            "Image %s does not come from supported registry. Skip handling the sources from it.",
+            image,
+        )
+        return False
+
+    base_image_sources_dir = create_dir(work_dir, "base_image_sources")
+    base_sources_extraction_dir = create_dir(base_image_sources_dir, "extraction_dir")
+
+    source_image_name = image + "-source"
```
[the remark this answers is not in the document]
tkdchen: Yes, it is STONEBLD-1831

task/source-build/0.1/source-build.yaml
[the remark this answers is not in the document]
tkdchen: > "Binary image name. The source container image is pushed to ${BINARY_IMAGE}.src"

@chmeliik This includes implementation detail. How about "Binary image name, from which to generate the source image name."

task/source-build/0.1/README.md
```diff
@@ -0,0 +1,20 @@
...
+|name|description|default value|required|
+|---|---|---|---|
+|BINARY_IMAGE|The corresponding binary image based on which to generate source image.||true|
+|BASE_IMAGES|Base images used to build the binary image. Each image per line in the same order of FROM instructions specified in a multistage Dockerfile. Default to an empty string, which means to skip handling a base image.|""|false|
+
+## Results
+|name|description|
+|---|---|
+|BUILD_RESULT|Build result.|
+|SOURCE_IMAGE_URL|The source image url.|
```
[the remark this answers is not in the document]
tkdchen: Done

task/source-build/0.1/source-build.yaml
tkdchen: Updated.

chmeliik (approved):
LGTM, the commits just need a cleanup now :slightly_smiling_face:

.gitignore
tkdchen: Mark this as resolved since the code has been updated and a new story was opened for moving to a separate git repo.

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,86 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: source-build
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "appstudio, hacbs"
```
mmorhun, line 10: Does `hacbs` has special meaning? If no, please remove.

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,86 @@
...
+spec:
+  description: Source image build.
+  params:
+    - name: BINARY_IMAGE
+      description: Binary image name from which to generate the source image name.
+      type: string
+    - name: BASE_IMAGES
+      description: >-
+        Base images used to build the binary image. Each image per line in the same order of FROM
```
mmorhun, line 19: > Base images used to build the binary image.

This could be a bit confusing (especially if read after previous parameter description) . Can we stress that it's Component's or project's binary image or something like this? But up to you.

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,86 @@
...
+        instructions specified in a multistage Dockerfile. Default to an empty string, which means
+        to skip handling a base image.
+      type: string
+      default: ""
+  results:
+    - name: BUILD_RESULT
+      description: Build result.
```
mmorhun, line 26: Could it have better explanation?

task/source-build/0.1/source-build.yaml
```diff
@@ -0,0 +1,86 @@
...
+      description: The source image url.
+    - name: SOURCE_IMAGE_DIGEST
+      description: The source image digest.
+  workspaces:
+    - name: workspace
+      description: The workspace where source code is included.
+  volumes:
+    - name: source-build-work-place
+      emptyDir:
+        medium: Memory
```
mmorhun, line 37: Are you sure about memory volume? What's the average estimated consumption?

mmorhun (approved):

task/source-build/0.1/source-build.yaml
rcerven: in osbs source can have up to 20-40gb

rcerven: but I certainly wouldn't be adding sources in memory and working with that

tkdchen: > Are you sure about memory volume? What's the average estimated consumption?

We don't have real source image build yet. The existing general idea comes from the OSBS that is source image build would consume storage too much as we mentioned multiple times in the meetings.

> in osbs source can use up to 20-40gb (but that was including srpms)

@mmorhun Does it work if we set PVC to request 20Gi for the task for every build pipeline?

task/source-build/0.1/source-build.yaml
tkdchen: Just wrote it as the same tags within other tasks. Removed it from source build task.

tkdchen:
Commits are squashed into a single one. Two minor changes are added: 1) remove hacbs from task tags, 2) increase memory to 8Gi for bsi for this phase 1 implementation.

chmeliik (approved):

PR #627, opened 2023-10-18
Use /source subdir in build-bundles task

lcarva:
A [previous change](https://github.com/redhat-appstudio/build-definitions/commit/0765369ae500ce4240f3a09dd0b9e8b5e20a9450) updated the git-clone task to clone repositories into a subdir which has the default value of `source`. The build-bundles task was missed during that update.

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -66,7 +66,7 @@ spec:
             - name: build-bundles
               image: quay.io/redhat-appstudio/appstudio-utils:{{ revision }}
               workingDir: $(workspaces.source.path)
-              command: ["hack/build-and-push.sh"]
+              command: ["./hack/build-and-push.sh"]
               env:
                 - name: MY_QUAY_USER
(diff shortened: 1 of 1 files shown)


lcarva:
@mkosiarc, can you take a look?

.tekton/push.yaml
```diff
@@ -65,8 +65,8 @@ spec:
           steps:
             - name: build-bundles
               image: quay.io/redhat-appstudio/appstudio-utils:{{ revision }}
-              workingDir: $(workspaces.source.path)
-              command: ["hack/build-and-push.sh"]
+              workingDir: $(workspaces.source.path)/source
+              command: ["./hack/build-and-push.sh"]
```
MartinBasti, line 69: you accidentally reverted previous fix here

mkosiarc (approved):

mkosiarc:
I am so sorry that I missed that, thank you for taking care of it.

PR #630, opened 2023-10-19
Remove the inspect-image task from the default pipeline

arewm:
The default pipeline does not depend on the output generated from the inspect-image task. Therefore, we can remove it from that definition and only include it in the FBC pipeline.

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -141,6 +141,7 @@ spec:
                   --input "oci:${DATA_BUNDLE_REPO}:latest" \
                   --output "oci:${DATA_BUNDLE_REPO}:${TAG}" \
                   --freshen \
+                  --prune \
                   ${BUNDLES_PARAM[@]}
 
                 # To facilitate usage in some contexts, tag the image with the floating "latest" tag.


arewm:
@dirgim @mmorhun, FYI

arewm:
This will need coordination with e2e test changes.

pipelines/tekton-bundle-builder/kustomization.yaml
```diff
@@ -27,6 +27,8 @@ patches:
       path: /spec/tasks/9  # sbom-json-check
     - op: remove
       path: /spec/tasks/5  # deprecated-base-image-check
+    - op: remove
+      path: /spec/tasks/4  # build-source-image
```
arewm, line 31: @tkdchen, I added this removal for the source image tasks which I think was missed in https://github.com/redhat-appstudio/build-definitions/pull/621

arewm:
@mmorhun @dirgim , this is passing e2e tests now. Is it okay to merge?

dirgim (approved):
LGTM

mmorhun (approved):

PR #635, opened 2023-10-20
Fixing inconsistencies in utilization of source directory

arewm:
When cloning source code into a dedicated directory (#586), the ramifications of this change on the generated SBOM data was missed. This PR ensures that we have the proper source SBOM data for merging.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -279,7 +279,7 @@ spec:
       with open("./sbom-image.json") as f:
         image_sbom = json.load(f)
 
-      with open("./sbom-source.json") as f:
+      with open("./source/sbom-source.json") as f:
         source_sbom = json.load(f)
 
       # fetch unique components from available SBOMs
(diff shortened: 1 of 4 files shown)


task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -279,7 +279,7 @@ spec:
       with open("./sbom-image.json") as f:
         image_sbom = json.load(f)
 
-      with open("./sbom-source.json") as f:
+      with open("./source/sbom-source.json") as f:
```
stuartwdouglas, line 282: Are you sure this is right? The syft command above is: `syft dir:$(workspaces.source.path)/source --file=$(workspaces.source.path)/sbom-source.json --output=cyclonedx-json` which means the output file will not be under `./source` I think the current code is correct, as it generates the sbom outside of the users source directory.

stuartwdouglas (changes requested):

arewm:
Thanks for the feedback -- closing the PR. I created this initially when troubleshooting a related problem. You are right, this is not a correct fix.

PR #649, opened 2023-10-27
WIP: DIY Trusted Artifacts

lcarva:
This PR modifies the git-clone and the buildah task to use a pattern that ensures the buildah task is using the unmodified contents of the git repo cloned by the git-clone task.

There is an ongoing effort in the Tekton community to introduce the concept of trusted artifacts, TEP-0130
https://github.com/tektoncd/community/pull/1044

This change is directly influenced by that effort.

The approach taken here is a stop gap until the platform, i.e. Tekton, supports the idea of trusted artifacts natively.

Here's how it works...

I created a couple of little bash scripts to encapsulate producing and consuming an artifact: https://github.com/lcarva/trusted-artifacts These are available in the image: https://quay.io/lucarval/trusted-artifacts-demo This is meant to make it easier to use them in a Tekton task, instead of just stuffing bash scripts all over the place.

The creation of an artifact is pretty simple. It just creates a tarball containing whichever files are deemed part of the artifact. In this case, the artifact is the directory containing a cloned git repo.

The consumption of the artifact is the inverse. It extracts the tarball into a certain location. There are two things that make this step essential.

First, the hash of the archive must match what is expected. Otherwise, the consumption step fails.

Second, the archive is extracted to an emptyDir volume which is only accessible by a single Task. This means once the archive is consumed, no other Tasks can influence its contents. This avoids certain issues with Time of Check vs Time of Usage, (think parallel tasks sharing a workspace).

The SLSA Provenance generated for a PipelineRun using these tasks records the results of the git-clone task, and the parameters for the buildah task. This allows us to create policies that ensure the buildah task used the exact archive the git-clone task produced.

I was able to run this in production and produce an image, e.g. quay.io/redhat-user-workloads/lucarval-tenant/devfile-sample-code-with-quarkus/devfile-sample-code-with-quarkus:3e33c4d702b468aaa46751dc94cf8de5d5233e67

Here's what the SLSA Provenance says for the git-clone task results:

```
$ < /tmp/attestation.json jq '.payload | @base64d | fromjson | .predicate.buildConfig.tasks[] | select(.name == "clone-repository").results[] | select(.name=="source").value.hash'
"6d99e197c4bda4564211943f774561cf85fd0e2ab33226b34eb73d60edaac7aa"
```

And here's what it says for the parameters of the buildah task:

```
$ < /tmp/attestation.json jq '.payload | @base64d | fromjson | .predicate.buildConfig.tasks[] | select(.name == "build-container").invocation.parameters.SOURCE.hash'
"6d99e197c4bda4564211943f774561cf85fd0e2ab33226b34eb73d60edaac7aa"
```

Why bother with this change? This is about ensuring certain segments of the Pipeline are protected. Today, EC expects all the Tasks on the Pipeline to be trusted. This is secure, but it's very rigid. It's not possible for someone to run, for example, a custom Taks to run unit tests between git-clone and buildah without also customizing the organization policy. The proposed changes here would allow EC to be more specific on what it checks instead of just using a general umbrella.

This change is a proof of concept. So it may not fully work as expected. But hopefully it's enough to help us decide on whether or not to move forward with this.

### DIFF
--- a/pipelines/docker-build/patch.yaml
+++ b/pipelines/docker-build/patch.yaml
@@ -30,6 +30,8 @@
     value: "$(params.image-expires-after)"
   - name: COMMIT_SHA
     value: "$(tasks.clone-repository.results.commit)"
+  - name: SOURCE
+    value: "$(tasks.clone-repository.results.source)"
 - op: add
   path: /spec/results/-
   value:
--- a/task/buildah-10gb/0.1/patch.yaml
+++ b/task/buildah-10gb/0.1/patch.yaml
@@ -2,8 +2,8 @@
   path: /metadata/name
   value: buildah-10gb
 - op: replace
-  path: /spec/steps/0/resources/limits/memory
+  path: /spec/steps/2/resources/limits/memory
   value: 10.1Gi
 - op: replace
-  path: /spec/steps/0/resources/requests/memory
+  path: /spec/steps/2/resources/requests/memory
   value: 8Gi
--- a/task/buildah-6gb/0.1/patch.yaml
+++ b/task/buildah-6gb/0.1/patch.yaml
@@ -2,8 +2,8 @@
   path: /metadata/name
   value: buildah-6gb
 - op: replace
-  path: /spec/steps/0/resources/limits/memory
+  path: /spec/steps/2/resources/limits/memory
   value: 6.1Gi
 - op: replace
-  path: /spec/steps/0/resources/requests/memory
+  path: /spec/steps/2/resources/requests/memory
   value: 4Gi
--- a/task/buildah-8gb/0.1/patch.yaml
+++ b/task/buildah-8gb/0.1/patch.yaml
@@ -2,8 +2,8 @@
   path: /metadata/name
   value: buildah-8gb
 - op: replace
-  path: /spec/steps/0/resources/limits/memory
+  path: /spec/steps/2/resources/limits/memory
   value: 8.1Gi
 - op: replace
-  path: /spec/steps/0/resources/requests/memory
+  path: /spec/steps/2/resources/requests/memory
   value: 6Gi
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -54,6 +54,15 @@ spec:
     description: The image is built from this commit.
     type: string
     default: ""
+  - name: SOURCE
+    type: object
+    properties:
+      path:
+        type: string
+      hash:
+        type: string
+      type:
+        type: string
   results:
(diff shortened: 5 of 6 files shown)


brianwcook:
I guess this tarball would be available on pvc so thatit is available universally in the pipeline?
say I want to run some unit tests, can I also use this tarball in my unit test task? or in snyk?

lcarva:
> I guess this tarball would be available on pvc so thatit is available universally in the pipeline? say I want to run some unit tests, can I also use this tarball in my unit test task? or in snyk?

Exactly! Same for multi-arch buildah tasks. They all have a clean slate.

arewm:
This looks pretty cool! We would need to replicate this pattern for the prefetched dependencies as well for the generation and snyk for the consumption.

chmeliik:
If I understand it correctly, this approach won't always work for hermetic builds. The prefetch-dependencies task sometimes needs to modify config files / lockfiles in the source code to enable offline installs.

For npm, cachi2 sometimes modifies package.json and always modifies package-lock.json: https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#example-npm

For pip, cachi2 sometimes modifies requirements.txt: https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#inject-project-files-pip

These modifications currently happen in the prefetch-dependencies task: https://github.com/redhat-appstudio/build-definitions/blob/d8620827fd23d3ba7996c83380959f540f856177/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L40-L41

Some ideas:
- Should the prefetch task somehow be allowed to modify the trusted source?
- Maybe the inject-files command could run in the buildah task, not in the prefetch task
  - But it would probably have to be a separate step so that you can use the cachi2 image to run the step. Would it be possible (and allowed, from a "trusted artifacts" perspective) to persist the changes between task steps?

lcarva:
> If I understand it correctly, this approach won't always work for hermetic builds. The prefetch-dependencies task sometimes needs to modify config files / lockfiles in the source code to enable offline installs.

Yeah. I do think we can make it work as is like this: 

1. git-clone task produces an artifact, A
2. prefetch-dependencies task consumes artifact A, modifies files as needed, then produce a new artifact, B. (If no changes are needed, it can simply "produce" the same artifact it has consumed.)
3. buildah task consumes artifact B.

We can still have the same assurances on the validation side.

@chmeliik, WDYTY?

lcarva:
Closing this out as it has served its purpose as a proof of concept. [EC-20](https://issues.redhat.com/browse/EC-20) is now planned to implement the changes properly.

PR #651, opened 2023-10-31
Don't fail when image sbom does not contain components

mkosiarc:
In previous versions of syft, if the image did not contain any components, the field was empty list:

  "components": []

But in newer versions, the field is omitted completely.


mmorhun (approved):

chmeliik (approved):
LGTM :+1:

PR #654, opened 2023-11-01
Revert "Emit SBOM_BLOB_URL result on build tasks"

lcarva:
This reverts commit a26fa79cde9fd37cf7fa4f58fcf83080b9780e05.

The additional result is causing the max Tekton result size to be reached which causes pipeline failures. Let's revert this change until we come up with an alternative idea.

Ref: RHTAPBUGS-935


chmeliik (approved):
LGTM :+1: 

I tested the affected build, and it does indeed pass if the SBOM_BLOB_URL is removed. But we're bytes away from hitting the limit :fearful:

ralphbean (approved):

PR #657, opened 2023-11-02
STONEBLD-1803 Add rpm-ostree task

stuartwdouglas:
This allows the multi platform controller to create images that are bootable on VMs.

### DIFF
--- a/task/rpm-ostree/0.1/rpm-ostree.yaml
+++ b/task/rpm-ostree/0.1/rpm-ostree.yaml
@@ -1,0 +1,223 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio, hacbs
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: rpm-ostree
+    build.appstudio.redhat.com/multi-platform-required: "true"
+  name: rpm-ostree
+spec:
+  description: RPM Ostree
+  params:
+  - description: Reference of the image rpm-ostree will produce.
+    name: IMAGE
+    type: string
+  - default: quay.io/redhat-user-workloads/project-sagano-tenant/ostree-builder/ostree-builder-fedora-38:d124414a81d17f31b1d734236f55272a241703d7
+    description: The location of the rpm-ostree builder image.
+    name: BUILDER_IMAGE
+    type: string
+  - default: .
+    description: Path to the directory to use as context.
+    name: CONTEXT
+    type: string
+  - description: The file to use to build the image
+    name: IMAGE_FILE
+    type: string
+  - default: "true"
+    description: Verify the TLS on the registry endpoint (for push/pull to a non-TLS
+      registry)
+    name: TLSVERIFY
+    type: string
+  - default: ""
+    description: Delete image tag after specified time. Empty means to keep the image
+      tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks,
+      respectively.
+    name: IMAGE_EXPIRES_AFTER
+    type: string
+  - default: ""
+    description: The image is built from this commit.
+    name: COMMIT_SHA
+    type: string
+  - description: The platform to build on
+    name: PLATFORM
+    type: string
+  results:
+  - description: Digest of the image just built
+    name: IMAGE_DIGEST
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
+  - description: Digests of the base images used for build
+    name: BASE_IMAGES_DIGESTS
+  stepTemplate:
+    env:
+    - name: CONTEXT
+      value: $(params.CONTEXT)
+    - name: IMAGE_FILE
+      value: $(params.IMAGE_FILE)
+    - name: IMAGE
+      value: $(params.IMAGE)
+    - name: TLSVERIFY
+      value: $(params.TLSVERIFY)
+    - name: IMAGE_EXPIRES_AFTER
+      value: $(params.IMAGE_EXPIRES_AFTER)
+    - name: BUILDER_IMAGE
+      value: $(params.BUILDER_IMAGE)
+    name: ""
+    resources: {}
+  steps:
+  - env:
+    - name: COMMIT_SHA
+      value: $(params.COMMIT_SHA)
+    image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f
+    imagePullPolicy: Always
+    name: build
+    resources:
+      limits:
+        cpu: "0.5"
+        memory: 512Mi
+      requests:
+        cpu: 250m
+        memory: 128Mi
+    script: |-
+      set -o verbose
+      set -eu
+      set -o pipefail
+      mkdir -p ~/.ssh
+      if [ -e "/ssh/error" ]; then
+        #no server could be provisioned
+        cat /ssh/error
+        exit 1
+      fi
+      cp /ssh/id_rsa ~/.ssh
+      chmod 0400 ~/.ssh/id_rsa
+      export SSH_HOST=$(cat /ssh/host)
+      export BUILD_DIR=$(cat /ssh/user-dir)
+      export SSH_ARGS="-o StrictHostKeyChecking=no"
+      mkdir -p scripts
+      echo "$BUILD_DIR"
+      ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp"
+
+      rsync -ra $(workspaces.source.path)/ "$SSH_HOST:$BUILD_DIR/workspaces/source/"
+      cat >scripts/script-build.sh <<'REMOTESSHEOF'
+      #!/bin/sh
+      set -o verbose
+      cd $(workspaces.source.path)
+      rpm-ostree compose image --initialize-mode always --format oci "source/$IMAGE_FILE" rhtap-final-image
+
+      REMOTESSHEOF
+      chmod +x scripts/script-build.sh
+      rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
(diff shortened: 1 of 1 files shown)


chmeliik:
Would you be able to explain more in the commit message? I don't really have any context for what this is and why we want it

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+    name: create-purl-sbom
+    resources: {}
+    script: |
+      #!/bin/python3
+      import json
+
+      with open("./sbom-cyclonedx.json") as f:
+        cyclonedx_sbom = json.load(f)
+
+      purls = [{"purl": component["purl"]} for component in cyclonedx_sbom["components"] if "purl" in component]
```
tkdchen, line 159: There was an update to this line for other build tasks: `cyclonedx_sbom.get("components", [])`

stuartwdouglas: Thanks, I have made the same change.

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -104,7 +104,7 @@ spec:
   - env:
     - name: COMMIT_SHA
       value: $(params.COMMIT_SHA)
-    image: quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/hacktask-image-multi-platform-controller:build-dfd73-1698382902@sha256:d733531a07ede2e73500c500a33cdaf78332db146ffdc23d3c1a71b253a734b6
+    image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f
```
chmeliik, line 107: IIUC the commit-hash tag for this image is pinned in the taskgen script, which means the image doesn't self-update (via Renovate). Updating the image would require bumping the tag in https://github.com/redhat-appstudio/multi-platform-controller/blob/main/cmd/taskgen/main.go, and bumping the taskgen image tag in `hack/generate-buildah-remote.sh`

Could be improved in the future?

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+    build.appstudio.redhat.com/build_type: rpm-ostree
+    build.appstudio.redhat.com/multi-platform-required: "true"
+  name: rpm-ostree
+spec:
+  description: RPM Ostree
+  params:
+  - description: Reference of the image rpm-ostree will produce.
+    name: IMAGE
+    type: string
+  - default: quay.io/sdouglas/rpm-ostree:dev
```
chmeliik, line 18: Assuming this is temporary

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+  - env:
+    - name: COMMIT_SHA
+      value: $(params.COMMIT_SHA)
+    image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f
+    imagePullPolicy: Always
+    name: build
+    resources:
+      limits:
+        cpu: "2"
+        memory: 4402341478400m
```
chmeliik, line 86: If I'm reading it right, this is about 4 Gi? Could it be specified in a more readable way?

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+      chmod +x scripts/script-build.sh
+      rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
+      ssh $SSH_ARGS "$SSH_HOST" podman run --mount type=bind,source=$BUILD_DIR/tmp,target=/var/tmp,relabel=shared --privileged -e CONTEXT="$CONTEXT" -e IMAGE_FILE="$IMAGE_FILE" -e IMAGE="$IMAGE" -e IMAGE_EXPIRES_AFTER="$IMAGE_EXPIRES_AFTER" -e COMMIT_SHA="$COMMIT_SHA"  --rm  -v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z"  -v $BUILD_DIR/scripts:/script:Z --user=0  "$BUILDER_IMAGE" /script/script-build.sh
+      rsync -ra "$SSH_HOST:$BUILD_DIR/workspaces/source/" "$(workspaces.source.path)/"
+      cp -r rhtap-final-image /var/lib/containers/rhtap-final-image
+      buildah pull oci:rhtap-final-image
+      buildah images
+      buildah tag localhost/rhtap-final-image "$IMAGE"
+      container=$(buildah from --pull-never "$IMAGE")
+    securityContext:
```
chmeliik, line 126: Is the `container=...` command needed here?

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
+  - description: Digests of the base images used for build
+    name: BASE_IMAGES_DIGESTS
+  - description: The counting of Java components by publisher in JSON format
+    name: SBOM_JAVA_COMPONENTS_COUNT
+    type: string
+  - description: The Java dependencies that came from community sources such as Maven
+      central.
+    name: JAVA_COMMUNITY_DEPENDENCIES
```
chmeliik, line 59: Are these two needed? Is it that all the build tasks must declare the same results?

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -0,0 +1,230 @@
...
+    workingDir: $(workspaces.source.path)
+  - image: quay.io/redhat-appstudio/syft:v0.94.0
+    name: sbom-syft-generate
+    resources:
+      limits:
+        memory: 6Gi
+      requests:
+        memory: 6Gi
+    script: |
+      syft oci-dir:/var/lib/containers/rhtap-final-image --file=$(workspaces.source.path)/sbom-cyclonedx.json --output=cyclonedx-json
```
chmeliik, line 145: `--file` is deprecated for this syft version https://github.com/redhat-appstudio/build-definitions/commit/327db8cc4fc4a28ae1aca08d441cf46b3a6beeb7

chmeliik (commented):
Overall seems reasonable :+1: 

Got a couple details

hack/generate-buildah-remote.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-podman run -v "$SCRIPTDIR"/..:/data quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-221979523603b730feca730aeb9d43f24a3c1e67  --buildah-task=/data/task/buildah/0.1/buildah.yaml --remote-task=/data/task/buildah-remote/0.1/buildah-remote.yaml
+podman run -v "$SCRIPTDIR"/..:/data quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-e008b351d34597c08b6731f21de37b3103662431  --buildah-task=/data/task/buildah/0.1/buildah.yaml --remote-task=/data/task/buildah-remote/0.1/buildah-remote.yaml
```
chmeliik, line 4: This update (and the corresponding buildah-remote update) doesn't seem connected to the rest of this change. Could you split it into a separate commit?

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: Oops, I had been testing with a RHTAP built image in redhat-user-workloads, and missed the default was still personal. I will update it to the redhat-user-workloads one, but I can't just build on GH actions like other images as it needs to be multi platform image>

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: This was copied from the buildah-remote task originally, as I thought I could mostly keep them in sync. They are two different and this does not need 4g anyway, I will update.

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: no, it was a leftover from the copy

hack/generate-buildah-remote.sh
stuartwdouglas: without this the buildah-remote task and this one would end up out of sync with the images they are using.

chmeliik: It doesn't seem like this script touches rpm-ostree though :thinking:

task/rpm-ostree/0.1/rpm-ostree.yaml
chmeliik: Could be built in RHTAP eventually?

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: I have removed it. The original intention was to make it as close to the buildah task as possible, but it was not really possible.

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: fixed

hack/generate-buildah-remote.sh
stuartwdouglas: Yea, there is another PR. It has been a long day.

chmeliik: Ah, ok. We can probably drop the change in the buildah-remote task as well then :slightly_smiling_face:

chmeliik (approved):
LGTM

Just the change in `task/buildah-remote/0.1/buildah-remote.yaml` is not necessary

chmeliik:
> LGTM
> 
> Just the change in `task/buildah-remote/0.1/buildah-remote.yaml` is not necessary

Oh, and your CI check caught that anyway :tada:

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: It definitely can be improved.

cgwalters:
> Would you be able to explain more in the commit message? I don't really have any context for what this is and why we want it

It's used by https://github.com/centos/sagano

See also https://github.com/containers/bootc and https://coreos.github.io/rpm-ostree/container/#creating-base-images

(At the current time the command requires privileges, but that's a bug we will eventually fix)

PR #658, opened 2023-11-03
prefetch: parameterized option to commit changes

lance:
When cachi2 updates the Dockerfile this makes the git tree dirty and may affect scripts or other build processes that use git commit tree information for things like version numbers and other metadata. When the git tree has been modified, the version is indicated as being "dirty".

This change allows projects who use this information to inject a `git commit` in the workflow with the cachi2 changes applied so that and git version information includes a non-dirty hash.

Note that this does not push that commit back to the origin repository. I'm not sure if that's problematic or not. It doesn't seem like it should be.

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -80,6 +80,8 @@ spec:
       value: $(params.HERMETIC)
     - name: PREFETCH_INPUT
       value: $(params.PREFETCH_INPUT)
+    - name: PREFETCH_COMMIT
+      value: $(params.PREFETCH_COMMIT)
     - name: CONTEXT
       value: $(params.CONTEXT)
     - name: DOCKERFILE
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -47,6 +47,10 @@ spec:
     name: PREFETCH_INPUT
     type: string
   - default: ""
+    description: In case it is not empty, the prefetch modifications to the containerfile will be applied to the source repository with a git commit
+    name: PREFETCH_COMMIT
+    type: string
+  - default: ""
     description: Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
     name: IMAGE_EXPIRES_AFTER
     type: string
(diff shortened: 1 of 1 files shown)


arewm (changes requested):
This has the potential to add confusion to the build process as the commit that is _built_ is not the commit that the build is triggered from. This is further confusing because the git reference won't be present in the source as we are only committing to the local repository and not the remote.

If what we are interested in is being able to have the Cachi2 environment file sources without a git commit, can it be the Component owners' responsibility to make a Dockerfile specific for this build that already has the sourcing of the environment file injected into it? This _can_ be hard coded into the git repository, but if you want to abstract it from the git repo, you can potentially have the path as a build argument to buildah?

Another alternative can be solved without adding parameters to the task. If the addition of the utilization of the environment file is contingent on the logic _not_ being present in the git repo already then it can be committed to the repo and the source-modifying behavior will simply be skipped.

lance:
> This has the potential to add confusion to the build process as the commit that is _built_ is not the commit that the build is triggered from. This is further confusing because the git reference won't be present in the source as we are only committing to the local repository and not the remote.

Good point - and a good reason not to take this route.

> Another alternative can be solved without adding parameters to the task. If the addition of the utilization of the environment file is contingent on the logic _not_ being present in the git repo already then it can be committed to the repo and the source-modifying behavior will simply be skipped.

OK I like this. How does the pipeline know that it should skip the source-modifying behavior?

arewm:
> OK I like this. How does the pipeline know that it should skip the source-modifying behavior?

You can run the `sed` only on lines that do not inject the line, i.e. https://stackoverflow.com/a/12178023.

chmeliik:
Instead of modifying the Dockerfile in place, can we generate a modified copy and build from the copy?

By the way, Dockerfile modifications are far from the only modifications that RHTAP may do, but if it helps in your case this would still be worth it IMO

chmeliik:
> Instead of modifying the Dockerfile in place, can we generate a modified copy and build from the copy?
> 
> By the way, Dockerfile modifications are far from the only modifications that RHTAP may do, but if it helps in your case this would still be worth it IMO

(this would require some care, since the buildah task modifies the Dockerfile in more than one place) https://github.com/redhat-appstudio/build-definitions/blob/d0d9953564c2b87e49646d11a35984c8d66402c0/task/buildah/0.1/buildah.yaml#L126

chmeliik:
Looks abandoned, can we close this?

lance:
> Looks abandoned, can we close this?

Yes, I believe so. Closing...

PR #659, opened 2023-11-03
STONEBLD-1758 don't pull the images

stuartwdouglas:
When creating manifest lists don't pull the images. The issue was that the docker image:tag@sha256:digest format requires buildah to pull the image. If we remove the tag section and just use the digest then the manifest list builds without needing to pull.

### DIFF
--- a/task/build-image-manifest/0.1/build-image-manifest.yaml
+++ b/task/build-image-manifest/0.1/build-image-manifest.yaml
@@ -60,21 +60,25 @@ spec:
       value: $(params.COMMIT_SHA)
     args: ["$(params.IMAGES[*])"]
     script: |
-
-
+      #!/bin/bash
       # Fixing group permission on /var/lib/containers
+      set -eu
+      set -o pipefail
       chown root:root /var/lib/containers
 
       sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
 
-      # Setting new namespace to run buildah - 2^32-2
-      echo 'root:1:4294967294' | tee -a /etc/subuid >> /etc/subgid
-
       buildah manifest create "$IMAGE"
       for i in $@
       do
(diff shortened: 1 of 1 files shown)


task/build-image-manifest/0.1/build-image-manifest.yaml
```diff
@@ -46,7 +46,7 @@ spec:
     - name: TLSVERIFY
       value: $(params.TLSVERIFY)
   steps:
-  - image: quay.io/redhat-appstudio/buildah:v1.31.0@sha256:34f12c7b72ec2c28f1ded0c494b428df4791c909f1f174dd21b8ed6a57cf5ddb
+  - image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f
```
chmeliik, line 49: What is the reason for this change?

task/build-image-manifest/0.1/kustomization.yaml
```diff
@@ -1,5 +0,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
-resources:
-- build-image-manifest.yaml
```
chmeliik, line 5: These deleted files seem unrelated to this PR

chmeliik (commented):
Nice! :+1:

tkdchen (commented):
:+1:

task/build-image-manifest/0.1/build-image-manifest.yaml
stuartwdouglas: oops, this was originally because I was going to use podman instead, which is part of that image, but it turned out that the issue was not with buildah, but rather with the use of tag + digest. I will revert.

task/build-image-manifest/0.1/kustomization.yaml
stuartwdouglas: We don't need these files. There has been a PR open to deleted them for weeks.

arewm (approved):

PR #663, opened 2023-11-04
tasks/rpm-ostree: Support older rpm-ostree

cgwalters:
The `--initialize-mode` argument is only in newer releases; this ensures we work with current c9s.


cgwalters:
cc @stuartwdouglas

stuartwdouglas (approved):

PR #667, opened 2023-11-06
Trigger an update of the source-build Task

lcarva:
This commit adds a no-op change to the source-build Task in order to debug an issue with how acceptable bundles are being tracked by EC.

Ref: EC-230


lcarva:
@rcerven , @chmouel , @mmorhun , @tkdchen could you PTAL? 

Unfortunately, merging this change is the only approach I can think of to diagnose the issue reported in the link ticket.

arewm (approved):

lcarva:
@chmouel , I definitely meant to ping @chmeliik :sweat_smile:

PR #684, opened 2023-11-23
feat: add sonar cloud properties file for better control

sawood14012:
Adding Sonar Cloud Properties File for better control 

https://issues.redhat.com/browse/STONEBLD-1940


.sonarcloud.properties
```diff
@@ -0,0 +1,9 @@
+# Path to sources
+sonar.sources=.
+sonar.exclusions=**/*test*, **/vendor/**, **/zz_generated.*, **/*.yaml,
```
tisutisu, line 3: this repo mostly contains pipelines and tasks definitions, there is no golang code here, will it be good to run sonar cloud check only on the `python` or `shell` scripts present in this repo?

chmeliik (approved):

tisutisu (commented):
LGTM, I think we can merge this one, since e2e-tests PR check does not relevant for this change.

mmorhun (approved):
Let's try this configuration and fix if we have any issues

PR #685, opened 2023-11-23
Update image quay.io/mkovarik/ose-cli-git:4.11

prietyc123:
PR https://github.com/openshift-pipelines/pipeline-service/pull/867 needs kustomize cli, so updating the image to get more cli and also its mkovarik personal quay image

### DIFF
--- a/task/update-infra-deployments/0.1/update-infra-deployments.yaml
+++ b/task/update-infra-deployments/0.1/update-infra-deployments.yaml
@@ -25,7 +25,7 @@ spec:
       default: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8:v1.8.2-8@sha256:a538c423e7a11aae6ae582a411fdb090936458075f99af4ce5add038bb6983e8
     - name: SCRIPT_IMAGE
       description: Image reference for SCRIPT execution
-      default: quay.io/mkovarik/ose-cli-git:4.11
+      default: registry.redhat.io/openshift-gitops-1/gitops-rhel8@sha256:475a7933a06b56ac63a7e2a2dd7d426bb13a5f70ba601da8ed1435145b3b8886
     - name: shared-secret
       default: infra-deployments-pr-creator
       description: secret in the namespace which contains private key for the GitHub App


mmorhun (approved):

chmeliik:
I think CI won't pass, e2e tests for build-definitions are currently failing due to a broken secret, which nobody hasbeen able to fix because apparently the build-templates-e2e namespace has no admins

prietyc123:
> I think CI won't pass, e2e tests for build-definitions are currently failing due to a broken secret, which nobody has been able to fix because apparently the build-templates-e2e namespace has no admins

ok but the change in pr https://github.com/openshift-pipelines/pipeline-service/pull/867 will fail until we have this fix. Is there any workaround in this case to merge this pr?

chmeliik:
> > I think CI won't pass, e2e tests for build-definitions are currently failing due to a broken secret, which nobody has been able to fix because apparently the build-templates-e2e namespace has no admins
> 
> ok but the change in pr [openshift-pipelines/pipeline-service#867](https://github.com/openshift-pipelines/pipeline-service/pull/867) will fail until we have this fix. Is there any workaround in this case to merge this pr?

The image is configurable, in this PR you're only changing the default. You can pass the SCRIPT_IMAGE param to the task in your pipeline definition

tkdchen (changes requested):
Please execute `hack/generate-build-team-readme.sh` to update the README.md

gabemontero:
FYI https://issues.redhat.com/browse/RHTAPBUGS-1043 was opened to move off of `quay.io/mkovarik/ose-cli-git:4.11` 

@prietyc123 you should mark whatever Jiras you have that depend on this as blocked by that bug to help @redhat-appstudio/build-team prioritize 

Wit respect to @tisutisu 's comment at https://github.com/redhat-appstudio/build-definitions/pull/685#pullrequestreview-1747883166 is not correct, as the RHTAP override of the default tekton SA has the credentials linked.

In fact, there is already use of registry.redhat.io in this very task, namely the default for `GIT_IMAGE`

@prietyc123 is simply changing the default for `SCRIPT_IMAGE`

In fact, you see use of registry.redhat.io in other both PaC based jobs in this repo, namely

https://github.com/redhat-appstudio/build-definitions/blob/main/.tekton/pull-request.yaml

As well as well used tasks like `init` and `git-clone`

Now, all that good news, said, there is the question of verifying this new image works.

@prietyc123 can you elaborate here on how you've tested, so we have a record here (vs. where you may have reported elsewhere)

prietyc123:
> @prietyc123 can you elaborate here on how you've tested, so we have a record here (vs. where you may have reported elsewhere)

@gabemontero Thanks for your detailed input here. There is another problem with the image is that multiple team using this image. So changing it needs proper analysis. I don't think I have bandwidth for that. So for now maybe I will let someone else pick up this task. I will close this pull request for now.

PR #687, opened 2023-11-28
fix(STONEBUGS-124): settings.xml injection fix

stuartwdouglas:
This will append the '-s settings.xml' directly after the mvn command,
so that commands like 'mvn install && rm target' will still work.
Previously this would have appended the settings to the rm command.


chmeliik:
Could you explain, in the commit message, what this fixes and how?

chmeliik (approved):
LGTM :+1:

PR #673, opened 2023-11-09
chore: Remove CPU limits from tasks

msugakov:
CPU `limits` are known as unhelpful and actually harmful as they cause throttling.
When containers have CPU `requests`, they should always get at least as much as requested. They can also get extra CPU cycles when available unless there are `limits` which prevent that.

This is well described in the following article: https://home.robusta.dev/blog/stop-using-cpu-limits

----

I simply searched for `cpu:` in all files of the repo and removed attributes that were in `limits:`. I feel some CPU `requests` can be tweaked but that's likely a job for separate PRs and requires more observations.

Initially, I created this PR observing that our build task gets throttled at 2 CPUs while it could get a speedup benefit if given more.

Testing: I haven't tested this change, and I hope it should work as expected.

PR #691, opened 2023-12-05
RHTAPBUGS-1025: set CPU requests _and_ limits for buildah task

tommartensen:
In #673 , the CPU limit was removed from this task. 
Because of that, the LimitRanger applied its CPU limit of 350m - which led to PipelineRun timeouts, because the build did not finish in time within the hour. Instead of continuing to fiddle with the LimitRanger and influence the CPU limit for all tasks in the namespace, we suggest to reintroduce the CPU limit.

We increased the CPU request to `1`, so that the over-provisioning is reasonable and help the scheduler.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -101,9 +101,10 @@ spec:
   steps:
   - computeResources:
       limits:
+        cpu: 2
         memory: 4Gi
       requests:
-        cpu: 250m
+        cpu: 1
         memory: 512Mi
     env:
     - name: COMMIT_SHA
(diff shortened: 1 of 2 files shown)


amfred:
100 millicores is a reasonable starting point for a small pod that doesn't have to do much work. Can you increase all of the CPU requests to at least 100 millicores?  10 millicores is too tiny.

jhutar (approved):
/lgtm

tommartensen:
@jhutar @amfred I noticed this PR is still open. I am not part of your org, would you do the necessary to run the tests and merge? Thanks!

tommartensen:
Obsolete

PR #706, opened 2023-12-14
STONEBLD-1779 propagate docker credentials

stuartwdouglas:
This will propagate docker credentials to the remote build agent. We now have tests to verify that these are cleaned up.

### DIFF
--- a/hack/generate-buildah-remote.sh
+++ b/hack/generate-buildah-remote.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-IMG=quay.io/redhat-user-workloads/rhtap-build-tenant/multi-arch-controller/multi-arch-controller:taskgen-57750ec21414607fa20acdef7984f32bbb7730af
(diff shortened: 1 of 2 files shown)


arewm (approved):
We are just copying over the docker credentials here .. making all secrets linked to the ServiceAccount available in the remote environment. Is there a plan to use RemoteSecrets eventually or this will be the only resolution to https://issues.redhat.com/browse/RHTAPBUGS-1060?

tkdchen (approved):
nitpick: breaking the long line properly could make the code easier-to-read. But, it's a personal preference, just FYI. For instance:

```bash
ssh $SSH_ARGS "$SSH_HOST" $PORT_FORWARD podman  run $PODMAN_PORT_FORWARD \
-e BUILDAH_FORMAT="$BUILDAH_FORMAT" \
-e STORAGE_DRIVER="$STORAGE_DRIVER" \
-e HERMETIC="$HERMETIC" \
-e PREFETCH_INPUT="$PREFETCH_INPUT" \
-e CONTEXT="$CONTEXT" \
...
--rm  \
-v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z"  \
-v "$BUILD_DIR/.docker/:/root/.docker:Z"
...
```

stuartwdouglas:
RemoteSecret is a kubernetes thing, while this is a pure VM based solution, so I don't really see how RemoteSecret can be applicable here.

stuartwdouglas:
@tkdchen those lines are auto generated by the task generator, I would need to update the generator to produce split lines. This is easy enough, but it should be done separately.

PR #712, opened 2023-12-19
Introduce trusted artifacts

zregvart:
Adds the use of trusted artifacts[1] to the Tasks and Pipelines. The `prefetch-dependencies` Task is no longer skipped based on the `hermetic` parameter -- it needs to run always to produce the trusted artifacts as the build* tasks depend on these, so if it were skipped the build* tasks would be skipped as well.

This also includes one small bugfix in the `buildah` task, it no longer tries to pull the `scratch` image.

Ref: https://issues.redhat.com/browse/EC-251

[1] https://github.com/redhat-appstudio/build-trusted-artifacts

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -69,6 +69,14 @@ spec:
     description: Target path on the container in which yum repository files should
       be made available
     name: YUM_REPOS_D_TARGET
+  - default: ""
+    description: The source trusted artifact URI
+    name: SOURCE_ARTIFACT
+    type: string
+  - default: ""
+    description: The prefetched dependencies trusted artifact URI
+    name: CACHI2_ARTIFACT
+    type: string
   - description: The platform to build on
     name: PLATFORM
     type: string
@@ -115,6 +123,15 @@ spec:
     - name: BUILDER_IMAGE
       value: $(params.BUILDER_IMAGE)
   steps:
+  - args:
+    - use
+    - --store
+    - $(workspaces.artifacts.path)
+    - $(params.SOURCE_ARTIFACT)=$(workspaces.source.path)/source
+    - $(params.CACHI2_ARTIFACT)=$(workspaces.source.path)/cachi2
+    computeResources: {}
+    image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:dbcf0102d0e9f21c2bcb06dcaa9af680cb9151f1984f5cec1fb397e1356ae771
+    name: use-trusted-artifact
   - computeResources:
       limits:
         memory: 4Gi
@@ -155,6 +172,7 @@ spec:
       fi
 
       rsync -ra $(workspaces.source.path)/ "$SSH_HOST:$BUILD_DIR/workspaces/source/"
+      rsync -ra $(workspaces.artifacts.path)/ "$SSH_HOST:$BUILD_DIR/workspaces/artifacts/"
       rsync -ra "$HOME/.docker/" "$SSH_HOST:$BUILD_DIR/.docker/"
       rsync -ra "/tekton/results/" "$SSH_HOST:$BUILD_DIR/tekton-results/"
       cat >scripts/script-build.sh <<'REMOTESSHEOF'
@@ -276,11 +294,13 @@ spec:
        -e YUM_REPOS_D_TARGET="$YUM_REPOS_D_TARGET" \
        -e COMMIT_SHA="$COMMIT_SHA" \
        -v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z" \
+       -v "$BUILD_DIR/workspaces/artifacts:$(workspaces.artifacts.path):Z" \
        -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
        -v "$BUILD_DIR/tekton-results/:/tekton/results:Z" \
        -v $BUILD_DIR/scripts:/script:Z \
       --user=0  --rm  "$BUILDER_IMAGE" /script/script-build.sh
       rsync -ra "$SSH_HOST:$BUILD_DIR/workspaces/source/" "$(workspaces.source.path)/"
+      rsync -ra "$SSH_HOST:$BUILD_DIR/workspaces/artifacts/" "$(workspaces.artifacts.path)/"
       rsync -ra "$SSH_HOST:$BUILD_DIR/tekton-results/" "/tekton/results/"
       buildah pull oci:rhtap-final-image
       buildah images
@@ -452,3 +472,5 @@ spec:
   workspaces:
   - description: Workspace containing the source code to build.
     name: source
+  - description: The trusted artifact store
+    name: artifacts
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -66,6 +66,14 @@ spec:
   - name: YUM_REPOS_D_TARGET
     description: Target path on the container in which yum repository files should be made available
     default: /etc/yum.repos.d
+  - name: SOURCE_ARTIFACT
+    type: string
+    description: The source trusted artifact URI
+    default: ""
+  - name: CACHI2_ARTIFACT
+    type: string
+    description: The prefetched dependencies trusted artifact URI
+    default: ""
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
@@ -105,6 +113,14 @@ spec:
     - name: YUM_REPOS_D_TARGET
       value: $(params.YUM_REPOS_D_TARGET)
   steps:
+  - name: use-trusted-artifact
+    image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:dbcf0102d0e9f21c2bcb06dcaa9af680cb9151f1984f5cec1fb397e1356ae771
+    args:
+      - use
+      - --store
+      - $(workspaces.artifacts.path)
+      - $(params.SOURCE_ARTIFACT)=$(workspaces.source.path)/source
+      - $(params.CACHI2_ARTIFACT)=$(workspaces.source.path)/cachi2
   - image: $(params.BUILDER_IMAGE)
     # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
     # the cluster will set imagePullPolicy to IfNotPresent; our default param above specifies a digest
@@ -392,3 +408,5 @@ spec:
   workspaces:
   - name: source
     description: Workspace containing the source code to build.
+  - name: artifacts
+    description: The trusted artifact store
--- a/.github/workflows/check-buildah-remote.yaml
+++ b/.github/workflows/check-buildah-remote.yaml
@@ -1,5 +1,5 @@
 name: Validate PR - buildah-remote
-on:
+'on':
   pull_request:
     branches: [main]
 jobs:
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -29,7 +29,9 @@ spec:
         taskRef:
           name: git-clone
         workspaces:
-          - name: output
+          - name: source
+            workspace: workspace
+          - name: artifacts
             workspace: workspace
         params:
           - name: url
@@ -58,7 +60,9 @@ spec:
         taskRef:
           name: sast-snyk-check
         workspaces:
-          - name: workspace
+          - name: source
+            workspace: workspace
+          - name: artifacts
             workspace: workspace
       - name: build-container
         runAfter:
@@ -72,6 +76,8 @@ spec:
           name: buildah
         workspaces:
           - name: source
+            workspace: workspace
+          - name: artifacts
             workspace: workspace
       - name: check-partner-tasks
         runAfter:
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -34,7 +34,9 @@ spec:
         taskRef:
           name: git-clone
         workspaces:
-          - name: output
+          - name: source
+            workspace: workspace
+          - name: artifacts
             workspace: workspace
       - name: build-container
         params:
@@ -48,6 +50,8 @@ spec:
           name: buildah
         workspaces:
           - name: source
+            emptyDir: {}
+          - name: artifacts
             workspace: workspace
       - name: build-bundles
         params:
--- a/pipelines/docker-build-dance/patch.yaml
+++ b/pipelines/docker-build-dance/patch.yaml
@@ -84,7 +84,7 @@
     taskRef:
       name: acs-image-check
     workspaces:
-      - name: workspace
+      - name: source
         workspace: workspace
 - op: add
   path: /spec/tasks/-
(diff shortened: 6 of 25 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -15,350 +16,354 @@ spec:
...
+      type: string
+    - default: ""
+      description: Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
+      name: IMAGE_EXPIRES_AFTER
+      type: string
+    - name: COMMIT_SHA
+      description: The image is built from this commit.
+      type: string
+      default: ""
+    - name: ARTIFACTS
```
lcarva, line 61: We probably want to create a new version of each Task that now task the ARTIFACTS parameter or emits an artifact result.

stuartwdouglas:
It looks like you have accidentally formatted everything? There is a huge amount of non-related formatting changes in the PR.

task/buildah/0.1/buildah.yaml
```diff
@@ -15,350 +16,354 @@ spec:
...
-        http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path" "$DOCKERFILE")
-        if [ $http_code != 200 ]; then
-          echo "No Dockerfile is fetched. Server responds $http_code"
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:4581dc538a0bc1550466e8c7df3ee905e444eef3bc70a8ae765b568ab5a10588
+      args:
+        - use
+        - --store
+        - $(workspaces.source.path)
+        - $(params.ARTIFACTS[0])=$(workspaces.source.path)/source
```
stuartwdouglas, line 103: This kinda feels wrong to me. Wouldn't it make more sense to pass these in as explicit named parameters?

Rather than an array pram just use one parameter/result per artifact.

stuartwdouglas: Also this would remove the need to always run the prefetch step.

stuartwdouglas:
Have you performance tested this? Last I heard we were IO bound on the PVCs, and this has the potential to add significantly more PVC IO, especially for builds with a large number of dependencies.

Is the plan to eventually move away from PVC storage altogether?

task/buildah/0.1/buildah.yaml
zregvart: > This kinda feels wrong to me. Wouldn't it make more sense to pass these in as explicit named parameters?
> 
> Rather than an array pram just use one parameter/result per artifact.

The number of artifacts produced by a task is unbounded, e.g. the git-clone task [create one artifact](https://github.com/zregvart/build-definitions/blob/945441d9219ac4233a022da9ece34aa8ef2601d3/task/git-clone/0.1/git-clone.yaml#L269), whereas prefetch-dependencies [produces two](https://github.com/zregvart/build-definitions/blob/945441d9219ac4233a022da9ece34aa8ef2601d3/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L77-L78).

> Also this would remove the need to always run the prefetch step.

I'm curious, can you elaborate a bit?

zregvart:
> It looks like you have accidentally formatted everything? There is a huge amount of non-related formatting changes in the PR.

This was on purpose (see the [separate commit](https://github.com/redhat-appstudio/build-definitions/pull/712/commits/d612aaa8897845770083e1405250478db9134c95)), I would still like to avoid that and from what it seems the .dotfiles are not included in the archive.

> Have you performance tested this? Last I heard we were IO bound on the PVCs, and this has the potential to add significantly more PVC IO, especially for builds with a large number of dependencies.

No I have not, but I can do some preliminary performance tests. I'm pretty sure that the performance will not improve with this.

> Is the plan to eventually move away from PVC storage altogether?

Not something I can comment on, also not seeing how that would fit with Tekton's notion of [a workspace](https://tekton.dev/docs/pipelines/workspaces/), we would need to not use those to begin with.

task/buildah/0.1/buildah.yaml
zregvart: > We probably want to create a new version of each Task that now task the ARTIFACTS parameter or emits an artifact result.

I'm not sold on this, if ARTIFACTS was made optional the backward compatibility would be maintained.

task/buildah/0.1/buildah.yaml
stuartwdouglas: > > This kinda feels wrong to me. Wouldn't it make more sense to pass these in as explicit named parameters?
> > Rather than an array pram just use one parameter/result per artifact.
> 
> The number of artifacts produced by a task is unbounded, e.g. the git-clone task [create one artifact](https://github.com/zregvart/build-definitions/blob/945441d9219ac4233a022da9ece34aa8ef2601d3/task/git-clone/0.1/git-clone.yaml#L269), whereas prefetch-dependencies [produces two](https://github.com/zregvart/build-definitions/blob/945441d9219ac4233a022da9ece34aa8ef2601d3/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L77-L78).

Each artifact is basically just a digest, so you would just name them, and then in the 'use' section specify the results you were reading from. E.g. git-clone would have a result called CHECKOUT_ARTIFACT, this would be pased in as a GIT_CHECKOUT param, and instead of `$(params.ARTIFACTS[0])=$(workspaces.source.path)/source` it would be `$(params.GIT_CHECKOUT=$(workspaces.source.path)/source`.

> 
> > Also this would remove the need to always run the prefetch step.
> 
> I'm curious, can you elaborate a bit?

I guess it would would still need to run, as I don't think Tekton allows subsequent tasks to run if a task is skipped, but you could abort early and write an empty file to the results, and have the artifact code ignore empty hashes.

zregvart: I think I can make `*_ARTIFACT` work, that will make the code easier to follow. I guess I was considering how we would tie the chain of artifacts easier in EC, having the same result/parameter name would help there.

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -184,7 +201,9 @@ spec:
         BUILDAH_ARGS="--pull=never"
         UNSHARE_ARGS="--net"
         for image in $(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}'); do
-          unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
+          if [ "${image}" != "scratch" ]; then
```
lcarva, line 204: This seems like an unrelated change? Same in `task/buildah/0.1/buildah.yaml`

stuartwdouglas: This is https://issues.redhat.com/browse/RHTAPBUGS-1066, maybe it should be moved to its own PR.

zregvart: Pulled into https://github.com/redhat-appstudio/build-definitions/pull/723

lcarva:
@zregvart, what's preventing this from coming out of draft?

zregvart:
> @zregvart, what's preventing this from coming out of draft?

I think I needed to performance test this, but I'm procrastinating...

zregvart:
I'm looking for reviews here. And a decision if we can leave the performance testing for after. I'm not sure (Luiz seems to have some ideas though), how the potential impact on performance can be minimized. If any performance improvements are to be implemented they will not be implemented here but in the https://github.com/redhat-appstudio/build-trusted-artifacts repository.
With the [policy change](https://github.com/enterprise-contract/ec-policies/pull/868) in draft now I'm eager to get this work merged soon.

task/buildah/0.1/buildah.yaml
```diff
@@ -105,6 +111,14 @@ spec:
...
       value: $(params.YUM_REPOS_D_TARGET)
   steps:
+  - name: use-trusted-artifact
+    image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:7a6a70bb972dba913e3d087305eece1a3e0462fd5adae52cd32322b9997bf213
+    args:
+      - use
+      - --store
+      - $(workspaces.source.path)
+      - $(params.SOURCE_ARTIFACT)=$(workspaces.source.path)/source
+      - $(params.CACHI2_ARTIFACT)=$(workspaces.source.path)/cachi2
```
chmeliik: What happens if a user makes some modifications in the source directory in a custom task but doesn't use the trusted artifact mechanisms? I guess this step just throws away their modifications?

Would it be possible to warn the user or even fail the build when trusted-artifacts discards user modifications?

chmeliik:
> Last I heard we were IO bound on the PVCs, and this has the potential to add significantly more PVC IO, especially for builds with a large number of dependencies.

A potential performance improvement could be to extract the trusted artifacts somewhere other than a PVC? `use --store <not-on-a-PVC>`

task/buildah/0.1/buildah.yaml
lcarva: A couple of things may happen:

1. The user adds their custom task to the chain of tasks where it produces a new trusted artifact which is then wired up to the buildah task. In this case, the build will succeed but EC will catch this as a violation because an untrusted task modified the source.
2. The user adds a custom task but doesn't wire things up properly in the pipeline. Then, yes, the changes will just get thrown out.
3. The user adds a custom task that tries to modify the trusted artifact produced by a trusted task. This will cause a build failure when another task tries to use that archive.

zregvart:
Moved out of draft because I'm anxious and ambitious at the same time.

lcarva:
> > Last I heard we were IO bound on the PVCs, and this has the potential to add significantly more PVC IO, especially for builds with a large number of dependencies.
> 
> A potential performance improvement could be to extract the trusted artifacts somewhere other than a PVC? `use --store <not-on-a-PVC>`
> 
> Although that can be done as a follow-up if the performance does get worse

Yes! We hope to eventually start using OCI archives. But we (mostly me) didn't want to add that additional layer of complexity to this first iteration.

chmeliik:
> Yes! We hope to eventually start using OCI archives. But we (mostly me) didn't want to add that additional layer of complexity to this first iteration.

Would it not be possible to do it without OCI archives? When creating the artifact, use `create --store <yes-on-a-PVC>`, when using it `use --store <not-on-a-PVC>`

For some filesystems, writing lots of small files is much worse than one large file, so the `create` step could be much cheaper than `use`.

_Flashbacks to the time when extracting a 1.0 GiB git repo archive to an aws-efs volume took 30 minutes_

lcarva:
> > Yes! We hope to eventually start using OCI archives. But we (mostly me) didn't want to add that additional layer of complexity to this first iteration.
> 
> Would it not be possible to do it without OCI archives? When creating the artifact, use `create --store <yes-on-a-PVC>`, when using it `use --store <not-on-a-PVC>`
> 
> For some filesystems, writing lots of small files is much worse than one large file, so the `create` step could be much cheaper than `use`.
> 
> _Flashbacks to the time when extracting a 1.0 GiB git repo archive to an aws-efs volume took 30 minutes_

Oh! This reminds me of something. @zregvart, when a trusted artifact is being used, it should get extracted to an emptyDir that only the Task has access to. Otherwise, a Task running in parallel can change the contents after they've been verified.

stuartwdouglas:
> > > Yes! We hope to eventually start using OCI archives. But we (mostly me) didn't want to add that additional layer of complexity to this first iteration.
> > 
> > 
> > Would it not be possible to do it without OCI archives? When creating the artifact, use `create --store <yes-on-a-PVC>`, when using it `use --store <not-on-a-PVC>`
> > For some filesystems, writing lots of small files is much worse than one large file, so the `create` step could be much cheaper than `use`.
> > _Flashbacks to the time when extracting a 1.0 GiB git repo archive to an aws-efs volume took 30 minutes_
> 
> Oh! This reminds me of something. @zregvart, when a trusted artifact is being used, it should get extracted to an emptyDir that only the Task has access to. Otherwise, a Task running in parallel can change the contents after they've been verified.

If we are worried about parallel tasks then shouldn't the 'store' task also do all its processing in an off-PVC location? At the moment it writes the archive to the store directory, the gets the digest and moves it/writes the results. If we are worried about parallel tasks then in theory a parallel task could modify the archive between creation and the original task generating the digest.

zregvart:
> If we are worried about parallel tasks then shouldn't the 'store' task also do all its processing in an off-PVC location? At the moment it writes the archive to the store directory, the gets the digest and moves it/writes the results. If we are worried about parallel tasks then in theory a parallel task could modify the archive between creation and the original task generating the digest.

When restoring the archive from the store the hash of the archive is checked again, for this to be a concern a Task running in parallel would also need to be able to influence the `*_ARTIFACT` result of the Task creating the archive, I think that is possible only if the malicious Task also has permissions to modify the TaskRun.

zregvart:
This will conflict with #771, so let's put this one on hold

stuartwdouglas:
> > If we are worried about parallel tasks then shouldn't the 'store' task also do all its processing in an off-PVC location? At the moment it writes the archive to the store directory, the gets the digest and moves it/writes the results. If we are worried about parallel tasks then in theory a parallel task could modify the archive between creation and the original task generating the digest.
> 
> When restoring the archive from the store the hash of the archive is checked again, for this to be a concern a Task running in parallel would also need to be able to influence the `*_ARTIFACT` result of the Task creating the archive, I think that is possible only if the malicious Task also has permissions to modify the TaskRun.

I mean if you have two task running in parallel, with either `ReadWriteMany` or `ReadWriteOnce` but on the same node, then in this code here there is a potential for a race: https://github.com/redhat-appstudio/build-trusted-artifacts/blob/main/create.sh#L46

Basically at line 46 the archive has been created, but we have not yet calculated the digest. A malicious or erroneous task could in theory modify the archive at this point and we would not be able to detect it, as we have not yet calculated the sha.

zregvart:
> I mean if you have two task running in parallel, with either `ReadWriteMany` or `ReadWriteOnce` but on the same node, then in this code here there is a potential for a race: [redhat-appstudio/build-trusted-artifacts@`main`/create.sh#L46](https://github.com/redhat-appstudio/build-trusted-artifacts/blob/main/create.sh?rgh-link-date=2024-01-31T10%3A08%3A41Z#L46)
> 
> Basically at line 46 the archive has been created, but we have not yet calculated the digest. A malicious or erroneous task could in theory modify the archive at this point and we would not be able to detect it, as we have not yet calculated the sha.

Only if the $TEMP directory is on the same PVC, right? That will not be the case.

stuartwdouglas:
What do you mean by $TEMP? There is only a call to `mktemp -p ${store}`, which will create the temp file on the PVC pointed to by store. I think to avoid this you would need to call plain `mktemp`, create the archive in the /tmp dir, take the sha, then put them both onto the PVC.

zregvart:
> What do you mean by $TEMP? There is only a call to `mktemp -p ${store}`, which will create the temp file on the PVC pointed to by store. I think to avoid this you would need to call plain `mktemp`, create the archive in the /tmp dir, take the sha, then put them both onto the PVC.

Right, thanks, I've created https://github.com/redhat-appstudio/build-trusted-artifacts/pull/14 to fix that. I guess the original intent was to make sure that the archive was created on a filesystem with enough free space. Let's see if that becomes an issue at any point.

zregvart:
Submitted a followup in #791, please have a look. I'll close this one as the discussion here will not be applicable over there in most comments.

PR #724, opened 2024-01-03
chore(RHTAPWATCH-697): support mounting repofiles to buildah build

yftacherzog:
This change allows mounting repofiles inside the image being built so those can be used by the buildah build process.

The repofiles can be stored together with the code, or generated on-the-fly using the generate-odcs-compose task, similarly to what's done here:
https://github.com/redhat-appstudio/tekton-tools/pull/22/files

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -59,6 +59,16 @@ spec:
     description: The image is built from this commit.
     name: COMMIT_SHA
     type: string
+  - default: repos.d
+    description: Path in the git repository in which yum repository files are stored
+    name: YUM_REPOS_D_SRC
+  - default: fetched.repos.d
+    description: Path in source workspace where dynamically-fetched repos are present
+    name: YUM_REPOS_D_FETCHED
+  - default: /etc/yum.repos.d
+    description: Target path on the container in which yum repository files should
+      be made available
+    name: YUM_REPOS_D_TARGET
   - description: The platform to build on
     name: PLATFORM
     type: string
@@ -96,6 +106,12 @@ spec:
       value: $(params.TLSVERIFY)
     - name: IMAGE_EXPIRES_AFTER
       value: $(params.IMAGE_EXPIRES_AFTER)
+    - name: YUM_REPOS_D_SRC
+      value: $(params.YUM_REPOS_D_SRC)
+    - name: YUM_REPOS_D_FETCHED
+      value: $(params.YUM_REPOS_D_FETCHED)
+    - name: YUM_REPOS_D_TARGET
+      value: $(params.YUM_REPOS_D_TARGET)
     - name: BUILDER_IMAGE
       value: $(params.BUILDER_IMAGE)
   steps:
@@ -200,6 +216,19 @@ spec:
         echo "Prefetched content will be made available"
       fi
 
+      # if yum repofiles stored in git, copy them to mount point outside the source dir
+      if [ -d "${SOURCE_CODE_DIR}/${YUM_REPOS_D_SRC}" ]; then
+        mkdir -p ${YUM_REPOS_D_FETCHED}
+        cp -r ${SOURCE_CODE_DIR}/${YUM_REPOS_D_SRC}/* ${YUM_REPOS_D_FETCHED}
+      fi
+
+      # if anything in the repofiles mount point (either fetched or from git), mount it
+      if [ -d "${YUM_REPOS_D_FETCHED}" ]; then
+        chmod -R go+rwX ${YUM_REPOS_D_FETCHED}
+        mount_point=$(realpath ${YUM_REPOS_D_FETCHED})
+        VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume ${mount_point}:${YUM_REPOS_D_TARGET}"
+      fi
+
       LABELS=(
         "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
         "--label" "architecture=$(uname -m)"
@@ -242,6 +271,9 @@ spec:
        -e IMAGE="$IMAGE" \
        -e TLSVERIFY="$TLSVERIFY" \
        -e IMAGE_EXPIRES_AFTER="$IMAGE_EXPIRES_AFTER" \
+       -e YUM_REPOS_D_SRC="$YUM_REPOS_D_SRC" \
+       -e YUM_REPOS_D_FETCHED="$YUM_REPOS_D_FETCHED" \
+       -e YUM_REPOS_D_TARGET="$YUM_REPOS_D_TARGET" \
        -e COMMIT_SHA="$COMMIT_SHA" \
        -v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z" \
        -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
--- a/task/buildah/0.1/README.md
+++ b/task/buildah/0.1/README.md
@@ -17,6 +17,9 @@ When prefetch-dependencies task was activated it is using its artifacts to run b
 |HERMETIC|Determines if build will be executed without network access.|false|false|
 |PREFETCH_INPUT|In case it is not empty, the prefetched content should be made available to the build.|""|false|
 |IMAGE_EXPIRES_AFTER|Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.|""|false|
+|YUM_REPOS_D_SRC|Path in the git repository in which yum repository files are stored|repos.d|false|
+|YUM_REPOS_D_FETCHED|Path in source workspace where dynamically-fetched repos are present|fetched.repos.d|false|
+|YUM_REPOS_D_TARGET|Target path on the container in which yum repository files should be made available|/etc/yum.repos.d|false|
 
 ## Results
 |name|description|
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -57,6 +57,15 @@ spec:
     description: The image is built from this commit.
     type: string
     default: ""
+  - name: YUM_REPOS_D_SRC
+    description: Path in the git repository in which yum repository files are stored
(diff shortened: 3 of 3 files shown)


task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -59,6 +59,13 @@ spec:
     description: The image is built from this commit.
     name: COMMIT_SHA
     type: string
+  - default: repos.d
+    description: Path in the repository in which yum repodir files are stored
+    name: REPOSDIR_SRC
```
tkdchen, line 64: Can the description be more specific instead of using abbrev like `repodir`?

My understanding is this is a directory containing custom YUM repository configuration files.

Should/Must this directory be in a relative path? At least, from the update to the script below, looks it is. 

How is the default value `repos.d` determined? Could a different directory name break something? Any other service depends on it?

Could you please update the description to give more information? Update both tasks.

In addition, how is the content of this directory generated and by what service?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -59,6 +59,13 @@ spec:
...
+  - default: /etc/yum.repos.d
+    description: Target path on the container in which repodir files should be made
+      available
+    name: REPOSDIR_TARGET
```
tkdchen, line 68: My preference is, using `YUM_REPOS_D` is much straightforward than `REPOSDIR`.

Then, `REPOSDIR_SRC` -> `YUM_REPOS_D_SRC`, `REPOSDIR_TARGET` -> `YUM_REPOS_D_TARGET`

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -198,6 +209,12 @@ spec:
         echo "Prefetched content will be made available"
       fi
 
+      reposdir_path=$(realpath $SOURCE_CODE_DIR)/$REPOSDIR_SRC
+      if [ -d "${reposdir_path}" ]; then
+        chmod -R go+rwX ${reposdir_path}
```
tkdchen, line 214: Do we need write permission for `go`?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -59,6 +59,13 @@ spec:
     description: The image is built from this commit.
     name: COMMIT_SHA
     type: string
+  - default: repos.d
+    description: Path in the repository in which yum repodir files are stored
+    name: REPOSDIR_SRC
+  - default: /etc/yum.repos.d
```
tkdchen, line 65: Generally, this is the default standard directory containing YUM repo configuration files. What cases require this customization to use an alternative directory?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -198,6 +209,12 @@ spec:
         echo "Prefetched content will be made available"
       fi
 
+      reposdir_path=$(realpath $SOURCE_CODE_DIR)/$REPOSDIR_SRC
```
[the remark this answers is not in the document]
yftacherzog: The overall idea is to allow the user to define yum repositories to use during the build process. The user can either provide it together with the source code or to have it generated by another task.

Having the directory inside  the `source` dir makes it easier for the user to provide their own yum repositories and having them maintained together with the code. Moving it out will mean that the user will need to have it stored on another repo, as far as I see it.

task/buildah-remote/0.1/buildah-remote.yaml
yftacherzog: > Can the description be more specific instead of using abbrev like repodir?

I'll update the description.

> Should/Must this directory be in a relative path?

Having it as an absolute path means that the user needs to understand more about the workspace definitions inside the task and their pipelinerun. I think it also means more limited portability. With that in mind, I think relative path makes sense. Let me know if you see it differently.

> How is the default value repos.d determined? Could a different directory name break something? Any other service depends on it?

It can be something else.

> In addition, how is the content of this directory generated and by what service?

The service that generates such configs within Red Hat is ODCS, but there could be other services to generate it, and it can even be created manually. We're finalizing a task to have that automated and integrated inside the build pipeline. It's implemented on a different repo, as we need access to the internal network to test it. We're also working on a process for automatically mirroring changes made to that task to build-definitions, so it can be consumed like any other task.

This is the task: https://github.com/redhat-appstudio/tekton-tools/blob/main/tasks/generate-odcs-compose.yaml

task/buildah-remote/0.1/buildah-remote.yaml
yftacherzog: I'll look into that

task/buildah-remote/0.1/buildah-remote.yaml
yftacherzog: We run into multiple permissions issues, and ended up using the same permissions used for the `PREFETCH_INPUT` option. In any case, the permissions are restored to the ones defined for the base image once the build process is done and the volume is unmounted. Let me know if you see an issue here, or any suggested alternative.

task/buildah-remote/0.1/buildah-remote.yaml
yftacherzog: Once we mount the repos to that default location, we hide the content of that directory as stored on the base image. Suppose that the user wants to have access to that configs from the base image, they can define a different target dir, and then have the dockerfile move things around so that both the original content and the custom repos are available.

task/buildah-remote/0.1/buildah-remote.yaml
chmeliik: Would `go+rX` be enough? Does the build ever write into the repos dir?

If the repo files can come from the source directory, it would be ideal if the build did _not_ have the ability to overwrite them (again, for source container accuracy)

chmeliik (commented):
The `buildah-*gb` variants will need corresponding changes as well

task/buildah-remote/0.1/buildah-remote.yaml
yftacherzog: > Does the build ever write into the repos dir?

Upon doing any yum-related call (e.g. even cleaning the cache), a file called redhat.repo is created under yum.repos.d, so if we want to be able to mount the repos directly there so the user won't have to move them around, then we need to allow writing.

task/buildah-remote/0.1/buildah-remote.yaml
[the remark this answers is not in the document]
yftacherzog: I think what we can do is to use the same workspace, but not the `source` dir in it.

We'll mount a directory (`repos.d`) on that workspace to `/etc/yum.repos.d` (or a specified target).
This directory will be at the same level as the `source` dir where the content of the git repo is.

The pre-build task we'll push content to that directory, and the buildah task finds content on `source/$REPOSDIR_SRC`, it will copy this under the `repos.d` directory.

This way, the user won't have to move things around, but we also don't risk making changes inside the directory pulled from git.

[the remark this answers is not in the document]
chmeliik: That sounds reasonable :+1:

task/buildah-remote/0.1/buildah-remote.yaml
chmeliik: Ok, I think we'll have to live with that

task/buildah-remote/0.1/buildah-remote.yaml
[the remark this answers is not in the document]
yftacherzog: @chmeliik can you have a look to see if the changes address your concerns?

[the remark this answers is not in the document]
chmeliik: They do :+1: 

In fact, this even solves https://github.com/redhat-appstudio/build-definitions/pull/724#discussion_r1445846933, since we'll be mounting a copy of the repo files, we won't be mounting them directly from the source dir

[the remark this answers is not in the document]
chmeliik: Though eventually, we could consider moving the copying of `YUM_REPOS_D_SRC` to the task that handles repos - IMO buildah shouldn't really have to know about both the locations (the one to mount from and the one to copy from)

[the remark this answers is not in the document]
yftacherzog: > we could consider moving the copying of YUM_REPOS_D_SRC to the task that handles repos

That task will not be used in case yum repos are only taken from the source code repo itself.
Suppose that we move it outside of the buildah task, we'll need to have another task to prep the mount point in a similar manner to what we're doing here now.

yftacherzog:
> The `buildah-*gb` variants will need corresponding changes as well

Do we need to do any changes within this PR for that?

chmeliik:
> > The `buildah-*gb` variants will need corresponding changes as well
> 
> Do we need to do any changes within this PR for that?

Oh right, those are not copy-pasted. Ignore me 😄

task/buildah-remote/0.1/buildah-remote.yaml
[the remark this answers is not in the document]
yftacherzog: @chmeliik do you see this as a blocker for proceeding with the current PR? We plan on doing a demo tomorrow and it will be super helpful if we can get it through.

chmeliik (approved):
LGTM :+1:

task/buildah-remote/0.1/buildah-remote.yaml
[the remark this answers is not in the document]
chmeliik: Not a blocker, just something to potentially clean up for the RHTAP (former Dance) version of the buildah task

gbenhaim:
There is a single test for FBC that is failing because it fails to find the pipelinerun

```
  Expected success, but got an error:
      <*errors.errorString | 0xc0009ff190>: 
      no pipelinerun found for component fbc-sample-repo-ozeu
      {
          s: "no pipelinerun found for component fbc-sample-repo-ozeu",
      }
  In [It] at
```

I can see in splunk that the pipeline was created and deleted 10 minutes after by tekton results. The pipeline also succeeded. (query `"fbc-sample-repo-ozeu" AND objectRef.resource=pipelineruns AND (verb=delete OR verb=create)`)

PR #729, opened 2024-01-08
New buildah task for RHTAP

brunoapimentel:
This PR proposes a new simplified and improved buildah task to be used for the new RHTAP docker build  pipeline.

Compared to the current Konflux buildah task, this task:
- Removes all the Konflux-specific features that are not needed in RHTAP
- Uses a productized version of the buildah image
- Removes root access from all steps
- Sets the correct Syft catalogers when scanning the built image
- Pins the output CycloneDX version
- Avoids littering the source code workspace with temporary files
- Improves the SBOM merge algorithm (dedup by PURL)
- Add a the SBOM OCI artifact URL as a task result
- Changes the tags

### DIFF
--- a/pipelines/docker-build-dance/patch.yaml
+++ b/pipelines/docker-build-dance/patch.yaml
@@ -23,7 +23,7 @@
 - op: replace
   path: /spec/tasks/3/taskRef
   value:
-    name: buildah
+    name: buildah-rhtap
     version: "0.1"
 - op: add
   path: /spec/tasks/3/params
@@ -34,10 +34,6 @@
     value: $(params.dockerfile)
   - name: CONTEXT
     value: $(params.path-context)
-  - name: HERMETIC
-    value: "$(params.hermetic)"
-  - name: PREFETCH_INPUT
-    value: "$(params.prefetch-input)"
   - name: IMAGE_EXPIRES_AFTER
     value: "$(params.image-expires-after)"
   - name: COMMIT_SHA
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -61,9 +61,6 @@ spec:
         exit 1
       fi
 
-      # Fixing group permission on /var/lib/containers
-      chown root:root /var/lib/containers
-
       buildah build \
         --tls-verify=$TLSVERIFY \
         --ulimit nofile=4096:4096 \
@@ -71,10 +68,10 @@ spec:
 
       container=$(buildah from --pull-never $IMAGE)
       buildah mount $container | tee /workspace/container_path
-      echo $container > /workspace/container_name
     securityContext:
       capabilities:
         add:
+          # this is needed so that buildah can write to the mounted /var/lib/containers directory
           - SETFCAP
     volumeMounts:
     - mountPath: /var/lib/containers
@@ -84,12 +81,16 @@ spec:
   - name: generate-sboms
     image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
     script: |
-      syft dir:$(workspaces.source.path)/source --output cyclonedx-json=$(workspaces.source.path)/sbom-source.json
+      syft dir:$(workspaces.source.path)/source --output cyclonedx-json=/tmp/files/sbom-source.json
+
       find $(cat /workspace/container_path) -xtype l -delete
-      syft dir:$(cat /workspace/container_path) --output cyclonedx-json=$(workspaces.source.path)/sbom-image.json
+
+      syft dir:$(cat /workspace/container_path) --output cyclonedx-json=/tmp/files/sbom-image.json
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+    - mountPath: /tmp/files
+      name: tmpfiles
 
   - name: merge-sboms
     image: registry.access.redhat.com/ubi9/python-39:1-158@sha256:967000729b17efdea309e297f4b1961c38b902a1ef18f6d886b8086c2a12f01f
@@ -122,9 +123,10 @@ spec:
       # write the CycloneDX unified SBOM
       with open("./sbom-cyclonedx.json", "w") as f:
         json.dump(image_sbom, f, indent=4)
-    workingDir: $(workspaces.source.path)
-    securityContext:
-      runAsUser: 0
+    volumeMounts:
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: /tmp/files
 
   - name: push-image
     image: registry.access.redhat.com/ubi9/buildah@sha256:04fde77ea72c25b56efb3f71db809c5d7b09938130df2da9175a3c888b94043d
@@ -152,7 +154,6 @@ spec:
       cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
       echo -n "$IMAGE" | tee $(results.IMAGE_URL.path)
     securityContext:
-      runAsUser: 0
       capabilities:
         add:
           - SETFCAP
@@ -171,11 +172,16 @@ spec:
       - --type
       - cyclonedx
       - $(params.IMAGE)
-    workingDir: $(workspaces.source.path)
+    volumeMounts:
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: /tmp/files
 
   volumes:
   - emptyDir: {}
     name: varlibcontainers
+  - emptyDir: {}
+    name: tmpfiles
   workspaces:
   - name: source
     description: Workspace containing the source code to build.
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -6,22 +6,17 @@ metadata:
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
-    tekton.dev/tags: "image-build, appstudio, hacbs"
-  name: buildah
+    tekton.dev/tags: "containers, rhtap"
+  name: buildah-rhtap
 spec:
   description: |-
     Buildah task builds source code into a container image and pushes the image into container registry using buildah tool.
     In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
-    When [Java dependency rebuild](https://redhat-appstudio.github.io/docs.stonesoup.io/Documentation/main/cli/proc_enabled_java_dependencies.html) is enabled it triggers rebuilds of Java artifacts.
-    When prefetch-dependencies task was activated it is using its artifacts to run build in hermetic environment.
   params:
   - description: Reference of the image buildah will produce.
     name: IMAGE
     type: string
   - default: quay.io/redhat-appstudio/buildah:v1.31.0@sha256:34f12c7b72ec2c28f1ded0c494b428df4791c909f1f174dd21b8ed6a57cf5ddb
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
     description: The location of the buildah builder image.
     name: BUILDER_IMAGE
     type: string
@@ -37,26 +32,6 @@ spec:
     description: Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)
     name: TLSVERIFY
     type: string
-  - description: unused, should be removed in next task version
-    name: DOCKER_AUTH
-    type: string
-    default: ""
-  - default: "false"
-    description: Determines if build will be executed without network access.
-    name: HERMETIC
-    type: string
-  - default: ""
-    description: In case it is not empty, the prefetched content should be made available to the build.
-    name: PREFETCH_INPUT
-    type: string
-  - default: ""
-    description: Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
-    name: IMAGE_EXPIRES_AFTER
-    type: string
-  - name: COMMIT_SHA
-    description: The image is built from this commit.
-    type: string
-    default: ""
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
@@ -64,21 +39,10 @@ spec:
     name: IMAGE_URL
   - description: Digests of the base images used for build
     name: BASE_IMAGES_DIGESTS
-  - name: SBOM_JAVA_COMPONENTS_COUNT
-    description: The counting of Java components by publisher in JSON format
-    type: string
-  - name: JAVA_COMMUNITY_DEPENDENCIES
-    description: The Java dependencies that came from community sources such as Maven central.
   stepTemplate:
     env:
-    - name: BUILDAH_FORMAT
-      value: oci
     - name: STORAGE_DRIVER
       value: vfs
-    - name: HERMETIC
-      value: $(params.HERMETIC)
-    - name: PREFETCH_INPUT
-      value: $(params.PREFETCH_INPUT)
     - name: CONTEXT
       value: $(params.CONTEXT)
     - name: DOCKERFILE
@@ -87,101 +51,31 @@ spec:
       value: $(params.IMAGE)
     - name: TLSVERIFY
       value: $(params.TLSVERIFY)
-    - name: IMAGE_EXPIRES_AFTER
-      value: $(params.IMAGE_EXPIRES_AFTER)
   steps:
-  - image: $(params.BUILDER_IMAGE)
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent; our default param above specifies a digest
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
-    name: build
-    computeResources:
-      limits:
-        memory: 4Gi
-      requests:
-        memory: 512Mi
-        cpu: 250m
-    env:
-    - name: COMMIT_SHA
-      value: $(params.COMMIT_SHA)
+  - name: build
+    image: $(params.BUILDER_IMAGE)
     script: |
       SOURCE_CODE_DIR=source
       if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
         dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
       elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
         dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
-      elif echo "$DOCKERFILE" | grep -q "^https\?://"; then
-        echo "Fetch Dockerfile from $DOCKERFILE"
-        dockerfile_path=$(mktemp --suffix=-Dockerfile)
-        http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path" "$DOCKERFILE")
-        if [ $http_code != 200 ]; then
-          echo "No Dockerfile is fetched. Server responds $http_code"
-          exit 1
-        fi
-        http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path.dockerignore.tmp" "$DOCKERFILE.dockerignore")
-        if [ $http_code = 200 ]; then
-          echo "Fetched .dockerignore from $DOCKERFILE.dockerignore"
-          mv "$dockerfile_path.dockerignore.tmp" $SOURCE_CODE_DIR/$CONTEXT/.dockerignore
-        fi
       else
         echo "Cannot find Dockerfile $DOCKERFILE"
         exit 1
-      fi
-      if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] && grep -q '^\s*RUN \(./\)\?mvn' "$dockerfile_path"; then
-        sed -i -e "s|^\s*RUN \(\(./\)\?mvn\)\(.*\)|RUN echo \"<settings><mirrors><mirror><id>mirror.default</id><url>http://$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR/v1/cache/default/0/</url><mirrorOf>*</mirrorOf></mirror></mirrors></settings>\" > /tmp/settings.yaml; \1 -s /tmp/settings.yaml \3|g" "$dockerfile_path"
-        touch /var/lib/containers/java
       fi
 
       # Fixing group permission on /var/lib/containers
       chown root:root /var/lib/containers
 
-      sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
-
-      # Setting new namespace to run buildah - 2^32-2
-      echo 'root:1:4294967294' | tee -a /etc/subuid >> /etc/subgid
-
-      if [ "${HERMETIC}" == "true" ]; then
-        BUILDAH_ARGS="--pull=never"
-        UNSHARE_ARGS="--net"
-        for image in $(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}'); do
-          unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
-        done
-        echo "Build will be executed with network isolation"
-      fi
-
-      if [ -n "${PREFETCH_INPUT}" ]; then
-        cp -r cachi2 /tmp/
-        chmod -R go+rwX /tmp/cachi2
-        VOLUME_MOUNTS="--volume /tmp/cachi2:/cachi2"
-        sed -i 's|^\s*run |RUN . /cachi2/cachi2.env \&\& \\\n    |i' "$dockerfile_path"
-        echo "Prefetched content will be made available"
-      fi
-
-      LABELS=(
-        "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
-        "--label" "architecture=$(uname -m)"
-        "--label" "vcs-type=git"
-      )
-      [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
-      [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
-
-      unshare -Uf $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah build \
-        $VOLUME_MOUNTS \
-        $BUILDAH_ARGS \
-        ${LABELS[@]} \
-        --tls-verify=$TLSVERIFY --no-cache \
+      buildah build \
+        --tls-verify=$TLSVERIFY \
         --ulimit nofile=4096:4096 \
         -f "$dockerfile_path" -t $IMAGE $SOURCE_CODE_DIR/$CONTEXT
 
       container=$(buildah from --pull-never $IMAGE)
       buildah mount $container | tee /workspace/container_path
       echo $container > /workspace/container_name
-
-      # Save the SBOM produced by Cachi2 so it can be merged into the final SBOM later
-      if [ -n "${PREFETCH_INPUT}" ]; then
-        cp /tmp/cachi2/output/bom.json ./sbom-cachi2.json
-      fi
-
     securityContext:
       capabilities:
         add:
@@ -191,11 +85,8 @@ spec:
       name: varlibcontainers
     workingDir: $(workspaces.source.path)
 
-  - name: sbom-syft-generate
+  - name: generate-sboms
     image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
     script: |
       syft dir:$(workspaces.source.path)/source --output cyclonedx-json=$(workspaces.source.path)/sbom-source.json
       find $(cat /workspace/container_path) -xtype l -delete
@@ -203,29 +94,9 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
-  - name: analyse-dependencies-java-sbom
-    image: quay.io/redhat-appstudio/hacbs-jvm-build-request-processor:1d417e6f1f3e68c6c537333b5759796eddae0afc
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
-    script: |
-      if [ -f /var/lib/containers/java ]; then
-        /opt/jboss/container/java/run/run-java.sh analyse-dependencies path $(cat /workspace/container_path) -s $(workspaces.source.path)/sbom-image.json --task-run-name $(context.taskRun.name) --publishers $(results.SBOM_JAVA_COMPONENTS_COUNT.path)
-        sed -i 's/^/ /' $(results.SBOM_JAVA_COMPONENTS_COUNT.path) # Workaround for SRVKP-2875
-      else
-        touch $(results.JAVA_COMMUNITY_DEPENDENCIES.path)
-      fi
-    volumeMounts:
-    - mountPath: /var/lib/containers
-      name: varlibcontainers
-    securityContext:
-      runAsUser: 0
 
-  - name: merge-syft-sboms
+  - name: merge-sboms
     image: registry.access.redhat.com/ubi9/python-39:1-158@sha256:967000729b17efdea309e297f4b1961c38b902a1ef18f6d886b8086c2a12f01f
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
     script: |
       #!/bin/python3
       import json
@@ -259,57 +130,11 @@ spec:
     securityContext:
       runAsUser: 0
 
-  - name: merge-cachi2-sbom
-    image: quay.io/redhat-appstudio/cachi2:0.3.0@sha256:46097f22b57e4d48a3fce96d931e08ccfe3a3e6421362d5f9353961279078eef
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
-    script: |
-      if [ -n "${PREFETCH_INPUT}" ]; then
-        echo "Merging contents of sbom-cachi2.json into sbom-cyclonedx.json"
-        /src/utils/merge_syft_sbom.py sbom-cachi2.json sbom-cyclonedx.json > sbom-temp.json
-        mv sbom-temp.json sbom-cyclonedx.json
-      else
-        echo "Skipping step since no Cachi2 SBOM was produced"
-      fi
-    workingDir: $(workspaces.source.path)
-    securityContext:
-      runAsUser: 0
-
-  - name: create-purl-sbom
-    image: registry.access.redhat.com/ubi9/python-39:1-158@sha256:967000729b17efdea309e297f4b1961c38b902a1ef18f6d886b8086c2a12f01f
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
-    script: |
-      #!/bin/python3
-      import json
-
-      with open("./sbom-cyclonedx.json") as f:
-        cyclonedx_sbom = json.load(f)
-
-      purls = [{"purl": component["purl"]} for component in cyclonedx_sbom.get("components", []) if "purl" in component]
-      purl_content = {"image_contents": {"dependencies": purls}}
-
-      with open("sbom-purl.json", "w") as output_file:
-        json.dump(purl_content, output_file, indent=4)
-    workingDir: $(workspaces.source.path)
-    securityContext:
-      runAsUser: 0
-
-  - name: inject-sbom-and-push
+  - name: push-image
     image: $(params.BUILDER_IMAGE)
-    computeResources: {}
     script: |
       # Expose base image digests
       buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' | grep -v $IMAGE > $(results.BASE_IMAGES_DIGESTS.path)
-
-      base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $IMAGE | cut -f1 -d'@')
-      base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $IMAGE)
-      container=$(buildah from --pull-never $IMAGE)
-      buildah copy $container sbom-cyclonedx.json sbom-purl.json /root/buildinfo/content_manifests/
-      buildah config -a org.opencontainers.image.base.name=${base_image_name} -a org.opencontainers.image.base.digest=${base_image_digest} $container
-      buildah commit $container $IMAGE
 
       status=-1
       max_run=5
@@ -317,20 +142,19 @@ spec:
       for run in $(seq 1 $max_run); do
         status=0
         [ "$run" -gt 1 ] && sleep $sleep_sec
-        echo "Pushing sbom image to registry"
+        echo "Pushing image to registry"
         buildah push \
           --tls-verify=$TLSVERIFY \
           --digestfile $(workspaces.source.path)/image-digest $IMAGE \
           docker://$IMAGE && break || status=$?
       done
       if [ "$status" -ne 0 ]; then
-          echo "Failed to push sbom image to registry after ${max_run} tries"
+          echo "Failed to push image to registry after ${max_run} tries"
           exit 1
       fi
 
       cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
       echo -n "$IMAGE" | tee $(results.IMAGE_URL.path)
-
     securityContext:
       runAsUser: 0
       capabilities:
@@ -343,9 +167,6 @@ spec:
 
   - name: upload-sbom
     image: quay.io/redhat-appstudio/cosign:v2.1.1@sha256:c883d6f8d39148f2cea71bff4622d196d89df3e510f36c140c097b932f0dd5d5
-    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
-    # the cluster will set imagePullPolicy to IfNotPresent
-    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
     args:
       - attach
       - sbom
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -72,12 +72,12 @@ spec:
       buildah push \
         --tls-verify=$TLSVERIFY \
         --retry=5 \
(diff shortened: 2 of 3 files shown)


task/buildah-dance/0.1/buildah-dance.yaml
```diff
@@ -0,0 +1,253 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: "docker"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "image-build, appstudio, hacbs"
+  name: buildah-dance
```
arewm, line 10:
```suggestion
  name: buildah-rhtap
```

and we can change the directory/filename too.

task/buildah-dance/0.1/buildah-dance.yaml
```diff
@@ -0,0 +1,253 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: "docker"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "image-build, appstudio, hacbs"
```
arewm, line 9: Do we need to continue pulling the `appstudio` and `hacbs` tags?

task/buildah-dance/0.1/buildah-dance.yaml
```diff
@@ -0,0 +1,253 @@
...
+      runAsUser: 0
+      capabilities:
+        add:
+          - SETFCAP
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+    workingDir: $(workspaces.source.path)
+
+  - name: upload-sbom
```
arewm, line 233: @lcarva, do you have thoughts on whether we should change the process for uploading SBOMs to the registry for customer utilization? Is `cosign attach sbom` still the best option?

task/buildah-dance/0.1/buildah-dance.yaml
```diff
@@ -0,0 +1,253 @@
...
+      image_sbom_components.sort(key=lambda c: get_identifier(c))
+
+      # write the CycloneDX unified SBOM
+      with open("./sbom-cyclonedx.json", "w") as f:
+        json.dump(image_sbom, f, indent=4)
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
+
+  - name: inject-sbom-and-push
```
arewm, line 190: You indicated that you removed this:

> Copying of the SBOM into the built container image

So it it intentional that it is here still or was the comment inaccurate?

task/buildah-dance/0.1/buildah-dance.yaml
```diff
@@ -0,0 +1,253 @@
...
+    Buildah task builds source code into a container image and pushes the image into container registry using buildah tool.
+    In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
+  params:
+  - description: Reference of the image buildah will produce.
+    name: IMAGE
+    type: string
+  - default: registry.access.redhat.com/ubi9/buildah@sha256:04fde77ea72c25b56efb3f71db809c5d7b09938130df2da9175a3c888b94043d
+    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+    # the cluster will set imagePullPolicy to IfNotPresent
+    # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
```
arewm, line 22: I don't think we need to include these comments in the pipeline definition.

task/buildah-dance/0.1/buildah-dance.yaml
arewm, line 233: Are we properly generating the results/annotations to enable the SBOM to be visualized in the UI? I don't see that present anywhere but maybe I am missing it.

task/buildah-dance/OWNERS
```diff
@@ -0,0 +1 @@
+Stonesoup Build Team
```
arewm, line 1: Let's change this too.

arewm (changes requested):

task/buildah-dance/0.1/buildah-dance.yaml
lcarva: Funny thing is that `cosign attach sbom` actually uploads the sbom to the registry as a blob which is great. what we do need is to record the digest of the sbom so we can ensure it has not changed post-build. We had this at some point but had to revert because that was the tipping point for exceeding the size of the results in the buildah task.

In konflux, we can get around this because the sbom is always embedded in the image itself, so EC can just use that.

brunoapimentel: We are removing the embedded SBOM for RHTAP in this PR, so we won't be able to rely on that in case we need to. :grimacing: 
(no changes for Konflux though)

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: Changed it.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: I'm really not sure what these tags do nowadays. Maybe @mmorhun can help.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: I did remove the one [line](https://github.com/redhat-appstudio/build-definitions/pull/729/commits/d32f8491341dd368d020c5b47ee324026c9e6b8c#diff-e27d0664a577cb8c73878188508eccd2fe7d26a67e8c4a1687df758f7aae03a6L310) that was embedding the SBOM. The name of the step is misleading, so I changed it.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: Agreed. Removed all of them.

task/buildah-dance/OWNERS
brunoapimentel: Done.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: What are the exact results is the UI expecting regarding SBOMs?  The buildah task currently emits three results, but they're only related to image digests.

Maybe you were looking for this task [here](https://github.com/redhat-appstudio/build-definitions/blob/main/task/show-sbom/0.1/show-sbom.yaml) here?

lcarva: Oh... Can we introduce the [SBOM_BLOB_URL](https://github.com/redhat-appstudio/build-definitions/pull/654/files) for this task, then?

I'm happy to do so as a follow up PR.

arewm: Whatever is needed for https://docs.google.com/document/d/1_1YXFx0ymzjl4b9M_LDjmmGrEYey5mDrfTdNn_56hpM/edit#heading=h.y4al2k1ggzhu

@rohitkrai03, would you expect that here or in `show-sbom`?

brunoapimentel: I found this is already a requirement for RHTAP (related [comment](https://github.com/redhat-appstudio/build-definitions/pull/729#discussion_r1445274263)). It is supposed to be renamed to `LINK_TO_SBOM`, but I believe it has the same purpose.

Although, I'm confused on how to implement this now. I entered the [deprecate cosign attachment](https://github.com/sigstore/cosign/issues/2755) rabbit hole once I saw the deprecation message in the CLI, and the task currently still is dependent on attach.

brunoapimentel: I was not aware of this at all. Are these annotations supposed to be added to the task itself? And is saving the SBOM URL in a tekton result enough for it to be shown in the UI?

As for the `show-sbom` task, it seems that its purpose is to download the SBOM and show it within the task logs... apparently the design is supposing that both scenarios could exist (link and logs)?

lcarva: My suggestion is to keep using the `cosign attach` command as it is done today. It is deprecated, but right now we don't really have a better alternative. The main thing here is that that command uploads the SBOM to the registry as a "blob". So just record the sh256 digest of the SBOM and we should be able to find it in the registry, e.g. `quay.io/foo/bar@sha256:abc..`.

I can help you with this. Feel free to ping me.

task/buildah-dance/0.1/buildah-dance.yaml
chmeliik: Probably nothing? It might be nice to have a common tag for all RHTAP (former Dance) tasks though.

task/buildah-rhtap/OWNERS
```diff
@@ -0,0 +1 @@
+RHTAP Build Team
```
chmeliik, line 1: I think this should be a github team: https://github.com/orgs/redhat-appstudio/teams

chmeliik (commented):
Oh no, taking a good look at our buildah task did not end well. Congratulations @brunoapimentel, you get a review for the whole task :grin: 

Some of the comments are definitely not blockers, we can probably just file new storiesor say it's acceptable

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+      value: $(params.IMAGE_EXPIRES_AFTER)
+  steps:
+  - image: $(params.BUILDER_IMAGE)
+    name: build
+    computeResources:
+      limits:
+        memory: 4Gi
+      requests:
+        memory: 512Mi
+        cpu: 250m
```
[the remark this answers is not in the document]
chmeliik: Probably not true, clamav-scan also [defines computeResources](https://github.com/redhat-appstudio/build-definitions/blob/54ecd81ce2791a83da05e289bd7630b87e201d55/task/clamav-scan/0.1/clamav-scan.yaml#L43) and I believe users do have a way to override them (right @MartinBasti?)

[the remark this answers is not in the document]
MartinBasti: Yeah, I cannot override it and steps overrides are only alpha

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+        fi
+        http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path.dockerignore.tmp" "$DOCKERFILE.dockerignore")
+        if [ $http_code = 200 ]; then
+          echo "Fetched .dockerignore from $DOCKERFILE.dockerignore"
+          mv "$dockerfile_path.dockerignore.tmp" $SOURCE_CODE_DIR/$CONTEXT/.dockerignore
+        fi
+      else
+        echo "Cannot find Dockerfile $DOCKERFILE"
+        exit 1
+      fi
```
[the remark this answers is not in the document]
arewm: I don't think it breaks L3, no, but it would probably be better if the references were pinned and we enabled the hash verification of fetched resources. Do we take advantage of this for enabling "s2i" builds for various languages?

pipelines/docker-build-dance/patch.yaml
```diff
@@ -23,7 +23,7 @@
 - op: replace
   path: /spec/tasks/3/taskRef
   value:
-    name: buildah
+    name: buildah-dance
```
lkolacek, line 26: Should we stick everywhere to rhtap name?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+  description: |-
+    Buildah task builds source code into a container image and pushes the image into container registry using buildah tool.
+    In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
+  params:
+  - description: Reference of the image buildah will produce.
+    name: IMAGE
+    type: string
+  - default: registry.access.redhat.com/ubi9/buildah@sha256:04fde77ea72c25b56efb3f71db809c5d7b09938130df2da9175a3c888b94043d
+    description: The location of the buildah builder image.
+    name: BUILDER_IMAGE
```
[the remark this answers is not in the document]
brunoapimentel: I removed this parameter. I don't think we need it to be configurable, and the default buildah task in the catalog does not allow this to be set either.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: I removed the resources definition from the task.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: From the discussions we had, this custom logic was added to enable devfile builds. Since there's no plans to support them in the near future, I dropped it.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: "docker"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "image-build, appstudio, hacbs"
+  name: buildah-rhtap
```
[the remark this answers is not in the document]
brunoapimentel: Sure, if you want the long version, here's my [attempt](https://github.com/redhat-appstudio/build-definitions/commit/1e548f9709015d4a0cd0ea2ffe83b9feadfddbb0) to port the task into build-definitions as it is, and a [version](https://github.com/redhat-appstudio/build-definitions/commit/4466dc6f3fb93fa1d18407ca8a3500c7a6b609af) with the SBOM generation added to it. 

But in summary, the default buidah task:
- validates input params, existence of the Dockerfile and build context
- builds the image
    - can use a dockerconfig secret (optional) 
    - sets the --no-cache option
    - allows for extra args build args to be added
- pushes image 
    - allows for extra push args to be added
    - can be skipped
- emits the results
    - digest sum
    - image name and tag

It is much more straightforward, it doesn't use things such as root access or a mounted `/var/lib/containers` volume, as it does build an push in the same step.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+      chown root:root /var/lib/containers
+
+      sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
+
+      LABELS=(
+        "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
+        "--label" "architecture=$(uname -m)"
+        "--label" "vcs-type=git"
+      )
+      [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
```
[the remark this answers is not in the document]
brunoapimentel: For now, I just removed this piece, as it seemed not necessary for our initial implementation.

brunoapimentel:
Newer pushes are addressing some of the comments, but there are still quite a few comments to address.

The idea of using the default buildah task from the catalog [proved](https://github.com/redhat-appstudio/build-definitions/compare/main...brunoapimentel:build-definitions:buildah-rhtap?expand=1) not very promising, at least not for the current phase. I think the way the catalog handles tasks (by having a separate [repository](https://github.com/openshift-pipelines/task-containers/) for scripts and templates, and generating the actual tasks with the scripts files added as base64 encoded [strings](https://github.com/openshift-pipelines/tektoncd-catalog/blob/p/tasks/task-buildah/0.2.0/task-buildah.yaml#L130-L132)) is very different from what we currently do, and just adding the raw scripts code proves to be very verbose, not to mention we lose the ability to reuse scripts such as `common.sh`.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
chmeliik: I don't know if that's the solution - unless we can ensure that every user will set the requests and limits in their own pipelineRun (do the limits get applied if they're set in the pipelineRun)?

If they do get applied, maybe the solution would be:
* document that the user must set requests and limits, suggest defaults
* in the pipelineRun template that we ship, set default requests and limits

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
chmeliik: > can use a dockerconfig secret (optional)

I think ours does too, but implicitly (tekton automatically injects some secrets to predefined directories)

> allows for extra args build args to be added

As in `--build-arg`? Or any argument at all? That would give the default buildah task way more flexibility. Proably too much flexibility to be EC-verifiable, since the user could make buildah do just about anything with the extra args

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
chmeliik: Just FYI: EC already blocks Dockerfiles downloaded from urls https://enterprisecontract.dev/docs/ec-policies/release_policy.html#buildah_build_task__buildah_uses_local_dockerfile

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: Any argument at all... this is what it looks like:
```
buildah \
      --storage-driver="${PARAMS_STORAGE_DRIVER}" \
      --tls-verify="${PARAMS_TLS_VERIFY}" \ 
      bud ${PARAMS_BUILD_EXTRA_ARGS} \
            --no-cache \
            --file="${DOCKERFILE_FULL}" \
            --tag="${PARAMS_IMAGE}" \
            ${PARAMS_CONTEXT}
 ```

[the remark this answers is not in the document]
Comment (chmeliik) [reply task/buildah-rhtap/0.1/buildah-rhtap.yaml:10]: Yeah, this would allow them to mount any volumes and secrets and who knows what else. Which we may need to support in the future, but probably in a more controlled way

File: task/buildah-rhtap/0.1/buildah-rhtap.yaml [hunk 14]
```diff
@@ -0,0 +1,237 @@
...
+          existing_components.append(get_identifier(component))
+
+      image_sbom_components.sort(key=lambda c: get_identifier(c))
+
+      # write the CycloneDX unified SBOM
+      with open("./sbom-cyclonedx.json", "w") as f:
+        json.dump(image_sbom, f, indent=4)
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
```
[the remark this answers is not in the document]
brunoapimentel: I don't think so. I changed the steps so they wouldn't need this anymore.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+          - SETFCAP
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+    workingDir: $(workspaces.source.path)
+
+  - name: sbom-syft-generate
+    image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
+    script: |
+      syft dir:$(workspaces.source.path)/source --output cyclonedx-json=$(workspaces.source.path)/sbom-source.json
```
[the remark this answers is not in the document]
brunoapimentel: Added.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+      find $(cat /workspace/container_path) -xtype l -delete
+      syft dir:$(cat /workspace/container_path) --output cyclonedx-json=$(workspaces.source.path)/sbom-image.json
```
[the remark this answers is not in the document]
brunoapimentel: Ok, this took me some time to figure out. I think we pontentially have two approaches to solve this:
- Try to use `oci-dir` instead of `dir`. I couldn't figure out a way to get/mount the image as an oci-dir though
- Manually set the catalogers. I added a patch that does that (in a very inelegant way)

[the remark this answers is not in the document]
brunoapimentel: For the latter approach, since we still use0.98, we can't use tags to set catalogers (introduced in [0.101](https://github.com/anchore/syft/tree/v0.101.0)), such as:
```
syft <some container image> --select-catalogers "image"
```

So in the commit I added, I needed to manually specify the catalogers as describe in the [0.98 docs](https://github.com/anchore/syft/tree/v0.98.0#image-scanning). Dumb me is still struggling with generating a config yaml file from within the script block from within the task yaml file :laughing:

Any hints are appreciated, but for now, I added a very ugly single-liner that solves the problem (apparently, syft also does not like when we break the args in multiple lines using \, I'm still not sure why).

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+    computeResources: {}
+    script: |
+      # Expose base image digests
+      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' | grep -v $IMAGE > $(results.BASE_IMAGES_DIGESTS.path)
+
+      base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $IMAGE | cut -f1 -d'@')
+      base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $IMAGE)
+      container=$(buildah from --pull-never $IMAGE)
+      buildah config -a org.opencontainers.image.base.name=${base_image_name} -a org.opencontainers.image.base.digest=${base_image_digest} $container
+      buildah commit $container $IMAGE
```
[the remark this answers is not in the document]
brunoapimentel: Done :feelsgood:

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+        "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
+        "--label" "architecture=$(uname -m)"
+        "--label" "vcs-type=git"
+      )
+      [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
+      [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+
+      buildah build \
+        ${LABELS[@]} \
+        --tls-verify=$TLSVERIFY --no-cache \
```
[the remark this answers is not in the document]
chmeliik: If the official buildah task does it, we can do it as well. I do believe builds will be isolated even without --pull=always

https://github.com/redhat-appstudio/build-definitions/pull/729#discussion_r1454319886

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
chmeliik: > Try to use oci-dir instead of dir. I couldn't figure out a way to get/mount the image as an oci-dir though

Would this work?

```shell
buildah push "$image" oci-archive:image.tar
syft oci-archive:image.tar
```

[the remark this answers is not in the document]
brunoapimentel: `oci-archive:` does not do it, but `oci-dir:` works. For reference, these were the catalogers executed this time:
- alpm-db-cataloger
- apk-db-cataloger
- binary-cataloger
- conan-info-cataloger
- dpkg-db-cataloger
- dotnet-portable-executable-cataloger
- go-module-binary-cataloger
- java-archive-cataloger
- graalvm-native-image-cataloger
- javascript-package-cataloger
- nix-store-cataloger
- php-composer-installed-cataloger
- portage-cataloger
- python-installed-package-cataloger
- r-package-cataloger
- rpm-db-cataloger
- ruby-installed-gemspec-cataloger
- sbom-cataloger

pipelines/docker-build-dance/patch.yaml
brunoapimentel: Fixed it.

task/buildah-rhtap/OWNERS
brunoapimentel: Do you think that adding the full url to the github team is enough?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+      status=-1
+      max_run=5
+      sleep_sec=10
+      for run in $(seq 1 $max_run); do
+        status=0
+        [ "$run" -gt 1 ] && sleep $sleep_sec
+        echo "Pushing sbom image to registry"
+        buildah push \
+          --tls-verify=$TLSVERIFY \
+          --digestfile $(workspaces.source.path)/image-digest $IMAGE \
```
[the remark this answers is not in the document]
brunoapimentel: I agree. I changed this so it is only written in the tmpfiles volume.

task/buildah-rhtap/OWNERS
chmeliik: Just the name probably? I thought these were the Kubernetes OWNERS files, but looking at the docs our syntax is not correct so IDK https://www.kubernetes.dev/docs/guide/owners/

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+      # load SBOMs
+      with open("./sbom-image.json") as f:
+        image_sbom = json.load(f)
+
+      with open("./sbom-source.json") as f:
+        source_sbom = json.load(f)
+
+      # fetch unique components from available SBOMs
+      def get_identifier(component):
+        return component["name"] + '@' + component.get("version", "")
```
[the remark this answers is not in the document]
brunoapimentel: I changed the algorithm so it looks at the `purl` now. Thing is, there is one case in which the purl is always absent: operating-system components. That should be fine, as they're only reported once in the image scan, but it led me to allow any component that lacks a purl to be present in the final SBOM (at risk of duplication).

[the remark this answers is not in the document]
brunoapimentel: I don't think this inline Python script looks good, though, so I think we should improve Cachi2's merge [script](https://github.com/containerbuildsystem/cachi2/blob/main/utils/merge_syft_sbom.py) to be more generalistic, able to handle any CycloneDX SBOM regardless of the generation tool.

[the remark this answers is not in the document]
chmeliik: Trying to do general SBOM de-duplication will be rough, because purls => packages is an N:1 relation, what with all the optional qualifiers etc. Some duplicates will always remain.

And if the SBOM generation tool is not careful or doesn't care, purls => packages can also be 1:M. For example, you can't be sure if `pkg:golang/sigs.k8s.io/cluster-api-provider-aws/v2` really means the original module or the [`github.com/openshift/cluster-api-provider-aws`](https://github.com/openshift/cluster-api-provider-aws/blob/4c2d9b8e1626337cdbe8cdd5a0ad43c2d520ecd1/go.mod#L1C8-L1C47) fork (with cachi2 you probably can, with Syft you can't). You also can't be sure if the tool neglected to include a `?repository` qualifier for a package that comes from a non-default repository, `?vcs_url` for a package that comes straight from git, etc. In this case, by trusting the purl you end up discarding non-duplicate packages, which may or may not be fine.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: I'm still not sure what's the best approach here. Should we remove these and add a new `rhtap` tag? The tekton catalog only has `containers` as [tag](https://github.com/openshift-pipelines/tektoncd-catalog/blob/p/tasks/task-buildah/0.2.0/task-buildah.yaml#L12).

Also, the min version seems off too.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: So, @lcarva, I took the same [approach](https://github.com/redhat-appstudio/build-definitions/pull/729/commits/9912beaddf1ffd58022a90a3069d20aac66d36a2) to generate the digest as in the PR you linked. Thing is, taking the `sha256sum sbom-cyclonedx.json` will give me the digest of the layer, not of the SBOM image that is pushed to the registry. 

You can see the test I did in https://quay.io/repository/bpimente/test-images:
- the SBOM image digest is `ed244f35bd7959a1149836997cd568361dbae6e496ffa560821e0d29c46c94fc`
- the only layer existing has a digest of `57a8c881a24e554cecd66dce96645b55b331138c9613586c3507482b09f0e3c4` 

I can compute the latter if I
```
cosign download sbom quay.io/bpimente/test-images:test-rhtap > sbom.json
sha256sum sbom.json
```

Maybe I'm doing something wrong?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: Yeah, excelent point. So I think we'll need to keep doing a tool-based dedup, since we can better predict its behavior then. Maybe we could move this script to the rh-syft repo in the future? I don't think it belongs in Cachi2.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,237 @@
...
+    description: The image is built from this commit.
+    type: string
+    default: ""
+  results:
+  - description: Digest of the image just built
+    name: IMAGE_DIGEST
+  - description: Image repository where the built image was pushed
+    name: IMAGE_URL
+  - description: Digests of the base images used for build
+    name: BASE_IMAGES_DIGESTS
```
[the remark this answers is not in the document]
brunoapimentel: As far as I could see, this is only used in the [deprecated-base-image-check](https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/template-build/template-build.yaml#L137-L147), which is [removed](https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/docker-build-dance/patch.yaml#L67-L68) from the RHTAP pipeline.

Any chance this is used in any other place?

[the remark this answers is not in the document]
chmeliik: Yeah, it's used for an EC check https://enterprisecontract.dev/docs/ec-policies/release_policy.html#base_image_registries_package

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
chmeliik: > Maybe we could move this script to the rh-syft repo in the future?

The fewer things we productize (and have to support) the better. I wanted the `syft` executable to be the only thing in the image, but I may end up including a ubi-micro base. But probably nothing past that.

If we do need a script long-term, it should probably be tightly coupled to the task, versioned together with the task etc. Though it shouldn't be inlined in the yaml and it should be tested.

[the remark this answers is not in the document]
chmeliik: Btw, one of the cataloggers that `syft` includes is `sbom`. Maybe we don't need a script for syft X syft deduplication at all

task/buildah-dance/0.1/buildah-dance.yaml
mmorhun: I'd remove outdated stuff. To me only `rhtap` makes sense. We might also need that `containers` for consistency.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: So I think we'll need to keep this for now, and probably move STONEBLD-2042 into the SBOM epic.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: Ok, I changed the tags to `containers, rhtap`.

task/buildah-rhtap/OWNERS
brunoapimentel: Well, looking at the other tasks, they just use an internal team name. I don't think we need to set a new standard in this PR. I'll leave a note for this to be improved later.

`Konflux Build Team` seems to be what it is for now, so that's what I picked, but maybe we can stick to `Stonesoup Build Team` and change it the same time the other tasks do. :shrug:

mmorhun: This is out of scope of this PR, let's just have a value for now and align all later.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      with open("./sbom-cyclonedx.json", "w") as f:
+        json.dump(image_sbom, f, indent=4)
+
+
+      ### write the SBOM digest result ###
+
+      with open("./sbom-cyclonedx.json", "rb") as f:
+        sbom_digest = hashlib.file_digest(f, "sha256").hexdigest()
+
+      image_repo = os.getenv("IMAGE").split(":")[0]
```
mmorhun, line 161: It will fail if registry has port, e.g. `registry.com:1234/user/image:tag`

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      component_list = image_sbom.get("components", [])
+      existing_purls = list(filter(lambda c: "purl" in c, component_list))
+
+      for component in source_sbom.get("components", []):
+        if "purl" in component:
+          if component["purl"] not in existing_purls:
+            component_list.append(component)
+            existing_purls.append(component["purl"])
+        else:
+          # We won't try to dedup components that lack a purl.
```
mmorhun, line 141:
```suggestion
          # We won't try to deduplicate components that lack a purl.
```

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: I'll file a story so we can investigate this further.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: I did some early testing and the sobm cataloger seems promising. I will file a story to investigate it further.

task/buildah-dance/0.1/buildah-dance.yaml
brunoapimentel: I'll move this to a separate PR so we can wrap up the work on this one here.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
+      elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
+      else
+        echo "Cannot find Dockerfile $DOCKERFILE"
+        exit 1
+      fi
+
+      sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
```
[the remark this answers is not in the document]
brunoapimentel: I copied this from the current buildah task without giving it much thought. Thing is, apparently this enables short-name resolution (see related [commit](https://github.com/redhat-appstudio/build-definitions/commit/018608552f790f4d1bd529f366173be13ee7bd51)). 

I don't think we want to keep this for the RHTAP task, as it is less secure than requiring full names. Wdyt?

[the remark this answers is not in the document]
chmeliik: Oh, good catch, I thought this line was there to _disable_ short name search :exploding_head: 

I think we need "enforcing" https://man.archlinux.org/man/containers-registries.conf.5.en#Short-Name_Aliasing:_Modes, which is the default in the redhat buildah image

```
$ podman run --rm -ti registry.access.redhat.com/ubi9/buildah@sha256:04fde77ea72c25b56efb3f71db809c5d7b09938130df2da9175a3c888b94043d cat /etc/containers/registries.conf

...
short-name-mode = "enforcing"
```

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      ### write the SBOM digest result ###
+
+      with open("./sbom-cyclonedx.json", "rb") as f:
+        sbom_digest = hashlib.file_digest(f, "sha256").hexdigest()
+
+      image_repo = os.getenv("IMAGE").split(":")[0]
+      sbom_blob_url = f"{image_repo}@sha256:{sbom_digest}"
+
+      print(image_repo)
+      print(sbom_blob_url)
```
chmeliik, line 165: Are the prints intentional?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      name: varlibcontainers
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: $(workspaces.source.path)
+
+  - name: generate-sboms
+    image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
+    script: |
+      syft dir:$(workspaces.source.path)/source --output cyclonedx-json@1.4=/tmp/files/sbom-source.json
+      syft oci-dir:/tmp/files/image --output cyclonedx-json@1.4=/tmp/files/sbom-image.json
```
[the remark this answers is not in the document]
brunoapimentel: I already changed this locally to 1.5, will push the fix later.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: Nope, they were accidentally commited in some rebase :face_exhaling:

mmorhun (approved):
I think we can start from current state of the new Buildah task and fix found problems in follow up PRs.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: Since the deafult is enforcing, I removed this line here.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' | grep -v $IMAGE > $(results.BASE_IMAGES_DIGESTS.path)
+      cat /tmp/files/image-digest | tee $(results.IMAGE_DIGEST.path)
+      echo -n "$IMAGE" | tee $(results.IMAGE_URL.path)
+
+      # Save the image so it can be used in the generate-sbom step
+      buildah push "$IMAGE" oci:/tmp/files/image
+    securityContext:
+      capabilities:
+        add:
+          - SETFCAP
```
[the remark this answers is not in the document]
brunoapimentel: Added.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+
+
+      ### attempt to deduplicate components ###
+
+      component_list = image_sbom.get("components", [])
+      existing_purls = list(filter(lambda c: "purl" in c, component_list))
+
+      for component in source_sbom.get("components", []):
+        if "purl" in component:
+          if component["purl"] not in existing_purls:
```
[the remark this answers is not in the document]
brunoapimentel: You're absolutely right, I stopped halfway through that filter thing. It should be right now (this part at least... deduplication will need major rework)

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: I dropped this commit, let's handle the SBOM_BLOB_URL in a follow up PR.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -68,25 +68,23 @@ spec:
...
   - name: generate-sboms
     image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
-    script:
+    script: |
       syft dir:$(workspaces.source.path)/source --output cyclonedx-json=/tmp/files/sbom-source.json
-
-      find $(cat /workspace/container_path) -xtype l -delete
-
-      syft dir:$(cat /workspace/container_path) --output cyclonedx-json=/tmp/files/sbom-image.json
+      syft oci-dir:/tmp/sboms/image --output cyclonedx-json=/tmp/files/sbom-image.json
```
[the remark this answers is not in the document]
brunoapimentel: I caught one instance of a `/tmp/sboms` leftover and fixed it. Not sure if I overlooked more commit-specific errors.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      buildah build \
+        --tls-verify=$TLSVERIFY --no-cache \
+        --ulimit nofile=4096:4096 \
+        -f "$dockerfile_path" -t $IMAGE $SOURCE_CODE_DIR/$CONTEXT
+
+      # Push the image
+      buildah push \
+        --tls-verify=$TLSVERIFY \
+        --digestfile /tmp/files/image-digest $IMAGE \
+        docker://$IMAGE
```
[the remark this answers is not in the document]
brunoapimentel: I added the --retry option, and kept it to 5 as it was the number set before.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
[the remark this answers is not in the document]
brunoapimentel: Fixed.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: This commit was entirely dropped.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+      value: $(params.CONTEXT)
+    - name: DOCKERFILE
+      value: $(params.DOCKERFILE)
+    - name: IMAGE
+      value: $(params.IMAGE)
+    - name: TLSVERIFY
+      value: $(params.TLSVERIFY)
+  steps:
+  - image: registry.access.redhat.com/ubi9/buildah@sha256:04fde77ea72c25b56efb3f71db809c5d7b09938130df2da9175a3c888b94043d
+    name: build
```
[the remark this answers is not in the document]
brunoapimentel: Fixed.

chmeliik (approved):

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,197 @@
...
+    script: |
+      # Check if the Dockerfile exists
+      SOURCE_CODE_DIR=source
```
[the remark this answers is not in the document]
brunoapimentel: I will add this to the list of follow up stories.

task/buildah-dance/0.1/buildah-dance.yaml
lcarva: Bruno and I chatted offline, summarizing it here for posterity/transparency. 

We want to record the digest of the layer (which should match the digest of the SBOM file). This is what EC will fetch. An SBOM image could, in theory, contain multiple SBOMs. Adding an SBOM changes the image digest, but not the digest of the other SBOMs (as layers).

brunoapimentel:
The newer push is just rebasing the branch. We still need to create the repository for the new task's image, since the CI is failing because of it.

chmeliik:
@arewm it looks like your change request is blocking the merge button

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,174 @@
...
+          - SETFCAP
+    volumeMounts:
+    - mountPath: /var/lib/containers
+      name: varlibcontainers
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: $(workspaces.source.path)
+
+  - name: generate-sboms
+    image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
```
arewm, line 98: @brunoapimentel , is modifying this a follow up story as well?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -0,0 +1,174 @@
...
+
+      with open("./sbom-cyclonedx.json", "w") as f:
+        json.dump(image_sbom, f, indent=4)
+    volumeMounts:
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: /tmp/files
+
+  - name: upload-sbom
+    image: quay.io/redhat-appstudio/cosign:v2.1.1@sha256:c883d6f8d39148f2cea71bff4622d196d89df3e510f36c140c097b932f0dd5d5
```
arewm, line 153: I think we should be able to modify this to the RHTAS images

arewm (approved):

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: Do you mean changing this for the productized version Adam is working on? If so, yes, it should be a follow up.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: Do you have a link for them?

arewm: https://catalog.redhat.com/software/containers/rhtas-tech-preview/cosign-rhel9/652971cd584d77e9d922883b?architecture=amd64&image=65c5506e19b6942d637e5603

task/buildah-rhtap/0.1/buildah-rhtap.yaml
arewm: yes, great!

PR #730, opened 2024-01-10
chore: sync tasks from tekton tools

yftacherzog:
This PR copying two new tasks coming from repo
https://github.com/redhat-appstudio/tekton-tools
In the future, this sort of sync will happen automatically upon changes to the tasks, but we're still working on the automation.

### DIFF
--- a/task/generate-odcs-compose/0.1/generate-odcs-compose.yaml
+++ b/task/generate-odcs-compose/0.1/generate-odcs-compose.yaml
@@ -1,0 +1,58 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: generate-odcs-compose
+spec:
+  params:
+    - name: IMAGE
+      type: string
+      description: The image to use for this task
+    - name: KT_PATH
+      type: string
+      description: path to mount keytab
+      default: /tmp/kt
+    - name: KRB_CACHE_PATH
+      description: path to krb cache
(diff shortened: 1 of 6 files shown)


avi-biton (approved):
LGTM

task/generate-odcs-compose/0.1/generate-odcs-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: generate-odcs-compose
+spec:
+  params:
+    - name: IMAGE
+      type: string
+      description: The image to use for this task
+    - name: KT_PATH
```
MartinBasti, line 11: From where does it take krb5 config?

Or maybe I just miss little bit of more context how krb5 client should be configured.

MartinBasti: I'm getting lost here.

So you have keytab and you can use keytab to acquire krb ticket, so why do you need cronjob?

MartinBasti: okay, I'm worried to see krb in such workloads, but we will see

yftacherzog: Yes. Unfortunately, only kerberos is supported for authentication with ODCS at the moment.

PR #737, opened 2024-01-11
Remove unused parameters and result

mkosiarc:
The parameters skip-optional, pipelinerun-name and pipelinerun-uid are
no longer used in init task and neither is the container-registry-secret
result.
New version for init task was created together with the migration doc
and the reference was updated in the pipeline

[STONEBLD-2000](https://issues.redhat.com//browse/STONEBLD-2000)

### DIFF
--- a/pipelines/enterprise-contract-everything.yaml
+++ b/pipelines/enterprise-contract-everything.yaml
@@ -65,7 +65,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
-            value: quay.io/enterprise-contract/ec-task-bundle:snapshot@sha256:962693658bc1ab9e2efd726d34704ffd9ec0951d500f6be78a7bb1d7b3f88c81
+            value: quay.io/enterprise-contract/ec-task-bundle:snapshot@sha256:778bd7a74a2909e608cd28f15f81078534e86ff99561a0b8fd2fdf01465de153
           - name: name
             value: verify-enterprise-contract
           - name: kind
--- a/pipelines/enterprise-contract-redhat-no-hermetic.yaml
+++ b/pipelines/enterprise-contract-redhat-no-hermetic.yaml
@@ -65,7 +65,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
-            value: quay.io/enterprise-contract/ec-task-bundle:snapshot@sha256:962693658bc1ab9e2efd726d34704ffd9ec0951d500f6be78a7bb1d7b3f88c81
+            value: quay.io/enterprise-contract/ec-task-bundle:snapshot@sha256:778bd7a74a2909e608cd28f15f81078534e86ff99561a0b8fd2fdf01465de153
           - name: name
             value: verify-enterprise-contract
           - name: kind
--- a/pipelines/enterprise-contract-redhat.yaml
+++ b/pipelines/enterprise-contract-redhat.yaml
@@ -65,7 +65,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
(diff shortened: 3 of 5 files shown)


mmorhun (approved):
LGTM

task/init/0.2/README.md
```diff
@@ -0,0 +1,19 @@
...
+|rebuild|Rebuild the image if exists|false|false|
+|skip-checks|Skip checks against built image|false|false|
+|pipelinerun-name|unused, should be removed in next task version|""|false|
+|pipelinerun-uid|unused, should be removed in next task version|""|false|
+
+## Results
+|name|description|
+|---|---|
+|build|Defines if the image in param image-url should be built|
+|container-registry-secret|unused, should be removed in next task version|
```
chmeliik, line 18: > container-registry-secret|unused, should be removed in next task version|

Maybe we could drop this one too while we're at it?

mkosiarc: Seems like we could, good catch

mkosiarc: there is also

```
|pipelinerun-name|unused, should be removed in next task version|""|false|
|pipelinerun-uid|unused, should be removed in next task version|""|false|
```
I guess it's safe to remove those as well, right @mmorhun ?

task/init/0.2/MIGRATION.md
```diff
@@ -0,0 +1,10 @@
+# Migration from 0.1 to 0.2
+
+The parameter `skip-optional` used by `init` task was removed
+
+## Action from users
+
+Update files in Pull-Request created by RHTAP bot:
+- Search for the task named `init`
+- Remove the `skip-optional` parameter from the `specs.params` and from the `specs.steps.env` sections in the task
```
tkdchen, line 9: User should not need to update the task directly. `skip-optional` param can be removed from generated PipelineRuns under `.tekton/`. The task is referenced via bundle resolver.

task/init/0.2/README.md
mmorhun: Yes, it should be safe, but will require more migration. Let's remove unused stuff.

task/init/0.2/MIGRATION.md
mkosiarc: I don't have any experience with creating new versions of the tasks/migration. I was basing my approach on the way things were handled in `deprecated-image-check` e.g https://github.com/redhat-appstudio/build-definitions/blob/main/task/deprecated-image-check/0.2/MIGRATION.md

@mmorhun Could you please share your opinion here?

chmeliik: The users have [this](https://github.com/redhat-appstudio/image-controller/blob/889c0b1d674feb10123817a4270d0b4cfe39636e/.tekton/image-controller-pull-request.yaml#L142-L156) in their repo, so they should only need to remove deprecated params from the `params` section

mkosiarc: I see

mkosiarc: Removed all parameteres and the result and updated the migration doc

chmeliik (approved):
LGTM, tiny nitpick in the migration doc

task/init/0.2/MIGRATION.md
```diff
@@ -0,0 +1,10 @@
+# Migration from 0.1 to 0.2
+
+The parameters `skip-optional`, `pipelinerun-name` and ` pipelinerun-uid` used by `init` task were removed.
+
+## Action from users
+
+Update files in Pull-Request created by RHTAP bot:
+- Search for the task named `init`
+- Remove the `skip-optional`, `pipelinerun-name` and ` pipelinerun-uid` parameters from the params section
+`
```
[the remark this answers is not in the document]
mkosiarc: that's not tiny, that's huge :D Removed

mmorhun (approved):

PR #754, opened 2024-01-19
Pin Syft's output format to CycloneDX 1.4

brunoapimentel:
Newer versions of Syft may default to newer versions of CycloneDX (as observed in the recent [0.99.0](https://github.com/anchore/syft/tree/v0.99.0) version). The problem is that the schema is bound to change between CycloneDX versions, which can bring breaking changes to the pipelines, specially to the scripts that merge different SBOMS.

This patch unblocks the update of Syft to `>=0.99` (#722).

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -303,9 +303,9 @@ spec:
     image: quay.io/redhat-appstudio/syft:v0.98.0@sha256:4d3856e6a2622700b9a9d5d74d9aaf5d8a55671653f80bf6c636677658680ede
     name: sbom-syft-generate
     script: |
-      syft dir:$(workspaces.source.path)/source --output cyclonedx-json=$(workspaces.source.path)/sbom-source.json
+      syft dir:$(workspaces.source.path)/source --output cyclonedx-json@1.4=$(workspaces.source.path)/sbom-source.json
       find $(cat /workspace/container_path) -xtype l -delete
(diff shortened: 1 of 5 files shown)


lcarva:
I'm pretty sure it's been producing with 1.5 format for a while. So much so that we [hard-coded 1.5 as the expected schema](https://github.com/enterprise-contract/ec-policies/blob/main/policy/release/sbom_cyclonedx_schema.rego)  version in EC.

brunoapimentel:
> I'm pretty sure it's been producing with 1.5 format for a while. So much so that we [hard-coded 1.5 as the expected schema](https://github.com/enterprise-contract/ec-policies/blob/main/policy/release/sbom_cyclonedx_schema.rego) version in EC.
> 
> I think we need to update EC to accept either. Can you confirm?

While this is very odd, since apparently the 1.5 format was only adopted with this [patch](https://github.com/anchore/syft/pull/2413) (which made it's way into [0.99](https://github.com/anchore/syft/compare/v0.98.0...v0.99.0)), I can confirm Syft 0.98 already outputs CycloneDX 1.5 :exploding_head: 

The initial reason for this proposed change is that the CI is failing when upgrading to Syft >=0.99. But the real reason for it is that Syft changed the format to specify `.metadata.tools` ([docs](https://cyclonedx.org/docs/1.5/json/#metadata_tools)), not the actual CycloneDX version. Version 1.5 still supports the legacy format from <=1.4.

Well, I think there's no point in reverting the CycloneDX version if updating Syft is not urgent right now. I'll close this PR, and the problem will be handled in https://github.com/containerbuildsystem/cachi2/pull/452.

PR #755, opened 2024-01-22
Disable PR EC checks

lcarva:
As part of https://issues.redhat.com/browse/EC-356, a check was added to ensure Task steps use images from reputable registries. However, some existing Tasks do not meet this criteria. This will be addressed in https://issues.redhat.com/browse/KONFLUX-2034. This effort is being tracked by the Epic https://issues.redhat.com/browse/EC-332.

Also, the existing check uses `ec validate definition` which doesn't fully allow customization of which policies are executed. It just runs all of the ones from the "task" namespace.

Finally, the previous rules basically checks that a Task has the kind attribute set to "Task", which is not very valuable.

For now, let's disable the EC checks altogether. They will be properly re-enabled as part of the epic EC-332.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -201,14 +201,15 @@ spec:
         workspaces:
           - name: source
             workspace: workspace
-      - name: ec-task-checks
-        runAfter:
-          - fetch-repository
-        taskRef:
-          name: ec-checks
-        workspaces:
-          - name: source
-            workspace: workspace
+      # This will be re-enabled as part of https://issues.redhat.com/browse/EC-332
+      # - name: ec-task-checks
+      #   runAfter:
+      #     - fetch-repository
+      #   taskRef:
+      #     name: ec-checks
+      #   workspaces:
+      #     - name: source
+      #       workspace: workspace
       - name: check-task-migration-md
         runAfter:
           - fetch-repository


chmeliik:
> Also, the existing check uses ec validate definition which doesn't fully allow customization of which policies are executed. It just runs all of the ones from the "task" namespace.

I guess this means there's no way to exclude just the one rule?

chmeliik (approved):
LGTM, hoping we can re-enable soon and not introduce more violations in the meantime

lcarva:
> I guess this means there's no way to exclude just the one rule?

That's right. We need to switch the check to use `ec validate input` instead of `ec validate definition`. That will allow us to use a policy config where we can pick and choose what to check.

lcarva:
Same e2e failure...

```
[FAILED] in [It] - /github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:145 @ 01/22/24 16:14:50.74
  << Timeline

  [FAILED] Timed out after 300.003s.
  timed out when waiting for the PipelineRun to start for the Component build-templates-e2e/nodejs-no-dockerfile-7n6w
  Expected success, but got an error:
      <*errors.errorString | 0xc001a523e0>: 
      no pipelinerun found for component nodejs-no-dockerfile-7n6w
      {
          s: "no pipelinerun found for component nodejs-no-dockerfile-7n6w",
      }
  In [It] at: /github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:145 @ 01/22/24 16:14:50.74
```

PR #761, opened 2024-01-29
Re-enable EC Task checks

lcarva:
Ref: https://issues.redhat.com/browse/EC-359

### DIFF
--- a/.tekton/tasks/ec-checks.yaml
+++ b/.tekton/tasks/ec-checks.yaml
@@ -25,34 +25,30 @@ spec:
   - name: validate-all-tasks
     workingDir: "$(workspaces.source.path)/source"
     image: quay.io/enterprise-contract/ec-cli:snapshot
-    command: [ec]
-    args:
-      - validate
-      - definition
-      - "--file"
-      - "./all_tasks-ec"
-      - "--policy"
-      - "git::https://github.com/enterprise-contract/ec-policies//policy/task"
-      - "--policy"
-      - "git::https://github.com/enterprise-contract/ec-policies//policy/lib"
-      - "--data"
-      - "git::https://github.com/release-engineering/rhtap-ec-policy//data"
-      - "--strict"
+    script: |
+      set -euo pipefail
+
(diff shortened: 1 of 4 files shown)


lcarva:
This has a hard-requirement on the changes from https://github.com/enterprise-contract/config/pull/78 propagating to infra-deployments [here](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/enterprise-contract/ecp.yaml#L115-L133).

I expect the Konflux build pipeline to fail without it. (This will be a good indicator that the EC check is indeed blocking as expected.)

zregvart:
https://github.com/redhat-appstudio/infra-deployments/pull/3176 for the EC update

lcarva:
Looks like an unrelated yamllint failure. Fixing it in https://github.com/redhat-appstudio/build-definitions/pull/770

.tekton/tasks/ec-checks.yaml
```diff
@@ -25,34 +25,28 @@ spec:
...
+    script: |
+      set -euo pipefail
+
+      # Generate list of file parameters, e.g. --file=foo.yaml --file=bar.yaml
+      files=$(ls all_tasks-ec/*.yaml | awk '{printf " --file=" $0}')
+      echo "[DEBUG] Files parameter: ${files}"
+
+      policy='enterprise-contract-service/redhat-trusted-tasks'
+
+      ec validate input --policy "${policy}" --output yaml --strict=true ${files}
```
zregvart, line 37: This can be done without shelling out to external commands:
```suggestion
      files=(all_tasks-ec/*.yaml)
      args=${files[*]/#/--file }
      echo "[DEBUG] Files parameter: ${args[*]}"

      policy=enterprise-contract-service/redhat-trusted-tasks

      ec validate input --policy "${policy}" --output yaml --strict=true ${args[*]}
```

lcarva: That is fantastic. I'll incorporate that approach.

lcarva: Done.

zregvart (approved):

PR #763, opened 2024-01-30
Enforce a standard YAML format

stuartwdouglas:
This PR requires all files to be formatted with the google YAML formatter before submitting. This can be done with the following commands:

go install github.com/google/yamlfmt/cmd/yamlfmt@v0.10.0 yamlfmt .

This will enforce a standard style that should remove any issues with formatting of PRs.


mmorhun (approved):

prietyc123:
@stuartwdouglas Is there any specific reason you are using `v0.10.0` yamlfmt. Can't be do it based on Red Hat YAML lint? 
One of the ex: With the current yaml format I see this approves the checks even if the yamls not started with 
```
---
```

.github/workflows/check-yaml-format.yaml
```diff
@@ -0,0 +1,28 @@
+name: Validate PR - YAML Format
```
prietyc123, line 1: Isn't It would be best if the file started with ---

prietyc123 (changes requested):

stuartwdouglas:
Doesn't yaml lint just check formatting? Or can it automatically fix the formatting as well?

The idea of this is that there is a single standard format, and you can just run the formatter and not have to think about formatting ever again. It means you won't get PRs with formatting changes, and all the files will be consistent.

.github/workflows/check-yaml-format.yaml
stuartwdouglas: Why?

prietyc123:
> Doesn't yaml lint just check formatting? Or can it automatically fix the formatting as well?

yes yamllint just check formatting but my point is the formatting that yamlfmt is doing might be lacking few of the standard like file should start with `---` or if some values started with block but end with scalar then it should wrap it inside the quotes https://github.com/spack/spack/issues/38408 etc. 
How would you fix those? Manually?

.github/workflows/check-yaml-format.yaml
prietyc123: I think `---` makes file structure more explicit, clearer and readable, especially when dealing with complex yaml documents that contains multiple documents concatenated together. I don't think yamlfmt handles this. Correct me if I am wrong.

prietyc123:
ok maybe for now I will remove the `document-start: enable` from the sample pipeline repo yamllint check and this fix should help the syncing linter issue. So lets merge this if everything else is good

/lgtm

chmeliik:
Looks abandoned, can we close this?

PR #770, opened 2024-01-30
Fix yamllint

lcarva:
```
Ignore files without spec in yamllint

This commit modifies the `ensure-params-not-in-script` step of the
yaml-lint task to not emit an error when processing YAML files that do
not contain a `.spec` attribute.

The functionality is unchanged. The only change is that `yq` quits early
instead of failing.
```

```
Fix yaml lint violation on tkn-bundle task

The task was merged at the same time as the yaml lint checks were
enabled.
```

### DIFF
--- a/task/tkn-bundle/0.1/tkn-bundle.yaml
+++ b/task/tkn-bundle/0.1/tkn-bundle.yaml
@@ -44,6 +44,8 @@ spec:
     env:
     - name: CONTEXT
       value: $(params.CONTEXT)
+    - name: STEPS_IMAGE
+      value: $(params.STEPS_IMAGE)
     script: |
       #!/bin/env bash
       set -o errexit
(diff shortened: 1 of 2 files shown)


task/tkn-bundle/0.1/tkn-bundle.yaml
```diff
@@ -100,9 +100,9 @@ spec:
         fi
       done
 
-      if [[ -n "$(params.STEPS_IMAGE)" ]]; then
+      if [[ -n "${STEPS_IMAGE}" ]]; then
```
joejstuart, line 103: Why change it from a param to an env var?

task/tkn-bundle/0.1/tkn-bundle.yaml
```diff
@@ -115,6 +115,8 @@ spec:
       value: $(params.CONTEXT)
     - name: IMAGE
       value: $(params.IMAGE)
+    - name: STEPS_IMAGE
```
joejstuart, line 118: I think this should be defined in the step before this?

task/tkn-bundle/0.1/tkn-bundle.yaml
lcarva: I should've added to the commit message: https://tekton.dev/docs/pipelines/tasks/#substituting-in-script-blocks

The yaml-lint task was actually raising those errors.

task/tkn-bundle/0.1/tkn-bundle.yaml
lcarva: Thanks for catching this! Fixed.

task/tkn-bundle/0.1/tkn-bundle.yaml
joejstuart: Ah, I didn't know that. thanks!

joejstuart (approved):

PR #772, opened 2024-01-31
Fix tkn-bundle task and add test

zregvart:
The `registry.access.redhat.com/ubi9/toolbox` image does not contain the `yq` binary, this changes the `modify-task-files` step's image to use `quay.io/redhat-appstudio/hacbs-test:latest` image instead.

The added test makes sure that the modification occurred by inspecting the image of the Tasks in the built bundle.

Resolves: https://issues.redhat.com/browse/EC-379


lcarva (approved):
Thanks!

PR #774, opened 2024-02-01
fix: update args max of htmlnormalize scriptnormalize  from clamac-scan task

hongweiliu17:
Update max-htmlnormalize max-scriptnormalize from clamav-scan task because they have not worked well.

They cause the issue in https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1706699932388599 and https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1704986484024359

I test scanning users's image quay.io/redhat-user-workloads/rh-acs-tenant/acs/scanner:on-pr-2554e4fa14648ce3dadfa75c7f15d105bf214f7f mentioned [here](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1704986484024359) but it times out after 1 hour, and then it works if I remove these two args.
So let's remove these two args for now

### DIFF
--- a/task/clamav-scan/0.1/clamav-scan.yaml
+++ b/task/clamav-scan/0.1/clamav-scan.yaml
@@ -74,7 +74,7 @@ spec:
         echo "Scanning image. This operation may take a while."
         clamscan -ri --max-scansize=4095M --max-filesize=4095M \
           --max-scantime=0 --max-files=0 --max-recursion=1000 --max-dir-recursion=20000 --max-embeddedpe=4095M \
-          --max-htmlnormalize=4095M --max-htmlnotags=4095M --max-scriptnormalize=4095M --max-ziptypercg=4095M \
+          --max-htmlnormalize=10M --max-htmlnotags=4095M --max-scriptnormalize=5M --max-ziptypercg=4095M \
           --max-partitions=50000 --max-iconspe=100000 --max-rechwp3=20000 --pcre-match-limit=100000000 --pcre-recmatch-limit=2000000 \
           --pcre-max-filesize=4095M --alert-exceeds-max=yes \
           --alert-encrypted=yes --alert-encrypted-archive=yes --alert-encrypted-doc=yes --alert-macros=yes \


task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -74,7 +74,7 @@ spec:
         echo "Scanning image. This operation may take a while."
         clamscan -ri --max-scansize=4095M --max-filesize=4095M \
           --max-scantime=0 --max-files=0 --max-recursion=1000 --max-dir-recursion=20000 --max-embeddedpe=4095M \
-          --max-htmlnormalize=4095M --max-htmlnotags=4095M --max-scriptnormalize=4095M --max-ziptypercg=4095M \
```
MartinBasti, line 77: by removing it, we are basically getting the default which is 

```
       --max-htmlnormalize=#n
              Maximum size of HTML file to normalize. You may pass the value in kilobytes in format xK or xk, or megabytes in format xM or xm, where x is a number (default: 10 MB, max: <4 GB).
```

Do we want to rather change it to 10M instead of removing?

jsztuka: I would say we could try cutting it in half and not have the upper limit set as we do now those 4 gigs.

jsztuka:
There is also mentioned influence of `--max-scantime=0`
`Also, I observed a significant time difference.

--max-htmlnormalize=1898M (which causes no errors) runs for 1:08,08 total time (1m8s).
--max-htmlnormalize=1899M --max-scantime=0 (which causes errors) executes for 7:47,49 total (7m47s).`

task/clamav-scan/0.1/clamav-scan.yaml
msugakov: Unfortunately, 2 gigs will cause issues for the component I looked at. 10M should be safe.

1 gig as well, for this file I looked at. Not sure if there will be other inputs where we'll see problems with 1 gig.

MartinBasti (approved):

jsztuka (approved):
lgtm

msugakov (approved):
Not sure if I am supposed to approve, but this looks good to me as a user of this Konflux task.

sonam1412 (approved):

PR #775, opened 2024-02-01
Add ACS tasks

mmorhun:
Adds ACS tasks:
- acs-image-check
- acs-image-scan
- acs-deploy-check

Task logs:

<details open>
<summary>ACS image check</summary>

```
Using rox central endpoint central-rhacs-operator.apps.cluster.lab.upshift.com:443
Download roxctl cli
roxctl image check
Policy check results for image: quay.io/mmorhun-org/test-builds@sha256:646f38919bfcdb760e1b26ac23da00d6ddb3f1767d5732591813426bdfc4b9db
(TOTAL: 2, LOW: 1, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

+--------------------------------+----------+--------------+--------------------------------+--------------------------------+--------------------------------+
|             POLICY             | SEVERITY | BREAKS BUILD |          DESCRIPTION           |           VIOLATION            |          REMEDIATION           |
+--------------------------------+----------+--------------+--------------------------------+--------------------------------+--------------------------------+
|   Fixable Severity at least    |   HIGH   |      X       |   Alert on deployments with    |    - Fixable RHSA-2023:6745    |  Use your package manager to   |
|           Important            |          |              |  fixable vulnerabilities with  |      (CVSS 8.1) (severity      |  update to a fixed version in  |
|                                |          |              |   a Severity Rating at least   | Important) found in component  |  future builds or speak with   |
|                                |          |              |           Important            |    'curl-minimal' (version     | your security team to mitigate |
|                                |          |              |                                |   7.76.1-23.el9_2.4.x86_64),   |      the vulnerabilities.      |
|                                |          |              |                                |      resolved by version       |                                |
|                                |          |              |                                |      0:7.76.1-26.el9_3.2       |                                |
|                                |          |              |                                |                                |                                |
|                                |          |              |                                |    - Fixable RHSA-2023:6745    |                                |
|                                |          |              |                                |      (CVSS 8.1) (severity      |                                |
|                                |          |              |                                | Important) found in component  |                                |
|                                |          |              |                                |    'libcurl-devel' (version    |                                |
|                                |          |              |                                |   7.76.1-23.el9_2.4.x86_64),   |                                |
|                                |          |              |                                |      resolved by version       |                                |
|                                |          |              |                                |      0:7.76.1-26.el9_3.2       |                                |
|                                |          |              |                                |                                |                                |
|                                |          |              |                                |    - Fixable RHSA-2023:6745    |                                |
|                                |          |              |                                |      (CVSS 8.1) (severity      |                                |
|                                |          |              |                                | Important) found in component  |                                |
|                                |          |              |                                |   'libcurl-minimal' (version   |                                |
|                                |          |              |                                |   7.76.1-23.el9_2.4.x86_64),   |                                |
|                                |          |              |                                |      resolved by version       |                                |
|                                |          |              |                                |      0:7.76.1-26.el9_3.2       |                                |
|                                |          |              |                                |                                |                                |
|                                |          |              |                                |    - Fixable RHSA-2023:6746    |                                |
|                                |          |              |                                |      (CVSS 7.5) (severity      |                                |
|                                |          |              |                                | Important) found in component  |                                |
|                                |          |              |                                |     'libnghttp2' (version      |                                |
|                                |          |              |                                |   1.43.0-5.el9_2.1.x86_64),    |                                |
|                                |          |              |                                |      resolved by version       |                                |
|                                |          |              |                                |       0:1.43.0-5.el9_3.1       |                                |
+--------------------------------+----------+--------------+--------------------------------+--------------------------------+--------------------------------+
|   Red Hat Package Manager in   |   LOW    |      -       |   Alert on deployments with    |        - Image includes        | Run `rpm -e --nodeps $(rpm -qa |
|             Image              |          |              |     components of the Red      |    component 'dnf' (version    |  '*rpm*' '*dnf*' '*libsolv*'   |
|                                |          |              |   Hat/Fedora/CentOS package    |     4.14.0-5.el9_2.noarch)     |   '*hawkey*' 'yum*')` in the   |
|                                |          |              |       management system.       |                                |   image build for production   |
|                                |          |              |                                |        - Image includes        |          containers.           |
|                                |          |              |                                |    component 'rpm' (version    |                                |
|                                |          |              |                                |    4.16.1.3-22.el9.x86_64)     |                                |
|                                |          |              |                                |                                |                                |
|                                |          |              |                                |        - Image includes        |                                |
|                                |          |              |                                |    component 'yum' (version    |                                |
|                                |          |              |                                |     4.14.0-5.el9_2.noarch)     |                                |
+--------------------------------+----------+--------------+--------------------------------+--------------------------------+--------------------------------+
WARN:	A total of 2 policies have been violated
ERROR:	failed policies found: 1 policies violated that are failing the check
ERROR:	Policy "Fixable Severity at least Important" - Possible remediation: "Use your package manager to update to a fixed version in future builds or speak with your security team to mitigate the vulnerabilities."
ERROR:	checking image failed after 3 retries: failed policies found: 1 policies violated that are failing the check
```

</details>


task/acs-image-check/0.1/acs-image-check.yaml
```diff
@@ -0,0 +1,77 @@
...
+        echo "rox API token is not set, demo will exit with success"
+        echo "TODO: configure the pipeline to have access to ROXCTL"
+        exit 0
+      fi
+
+      echo "Using rox central endpoint $(params.rox-central-endpoint)"
+
+      echo "Download roxctl cli"
+      IMAGE=$(params.image)@$(params.image-digest)
+      curl -s -k -L -H "Authorization: Bearer $ROX_API_TOKEN" \
```
zregvart, line 65: If we need to pull the CLI from a remote location, could we not use `-k` (`--insecure`)?
Could we pull from mirror.openshift.com [as documented](https://docs.openshift.com/acs/4.3/cli/installing-the-roxctl-cli.html) instead? Ideally an image with the `roxctl` would be the best option here, is there an image we could use instead?

mmorhun: Downloading is done mainly due to the following reasons:
 - The image containing `roxctl` is stored in `registry.redhat.com` which requires subscription to pull.
 - `roxctl` version in the image might be different from ACS version used by consumer.

Regarding insecure, I agree, we have to have it controller. Probably by `insecure-skip-tls-verify` parameter.

task/acs-deploy-check/0.1/acs-deploy-check.yaml
```diff
@@ -0,0 +1,82 @@
...
+          exit 0
+        fi
+
+        echo "Using rox central endpoint $(params.rox-central-endpoint)"
+
+        echo "Configuration"
+        echo "Directory = $(pwd)"
+        ls -al
+
+        curl -s -k -L -H "Authorization: Bearer $ROX_API_TOKEN" \
```
lcarva, line 67: Is this the best way to get the CLI? Maybe use an image that already has the roxctl? There is no guarantee the version being fetched is the version expected.

task/acs-deploy-check/0.1/acs-deploy-check.yaml
```diff
@@ -0,0 +1,82 @@
...
+          "https://$(params.rox-central-endpoint)/api/cli/download/roxctl-linux" \
```
lcarva, line 68: NACK on this. This effectively allows users to execute an arbitrary binary in this task. This is effectively introduces a variation of the issue https://issues.redhat.com//browse/STONEBLD-2109.

task/acs-image-check/0.1/acs-image-check.yaml
lcarva: > The image containing roxctl is stored in registry.redhat.com which requires subscription to pull.

There are ways to solve this, e.g. document that you need a subscription... This is not a reason not to use that registry IMO.

> roxctl version in the image might be different from ACS version used by consumer.

Then provide different version of the Task for different ACS versions?

zregvart: I see there is quay.io/repository/stackrox-io/roxctl, could that be used? In principal if the version matches the ACS server version? Perhaps using that from the start and then giving a reasonable window of supported CLI vs server version could be an option? 

Another idea, we could have many versions of the Task here matching the (major?) versions of ACS/stackrox using the compatible CLI.

I would not like to see any insecure fetching, that could allow for man in the middle attacks.

mmorhun: > There are ways to solve this, e.g. document that you need a subscription... This is not a reason not to use that registry IMO.

If we use images that require subscription to pull in the RHTAP pipeline, it would mean, that whole RHTAP pipeline wouldn't run without subscription. I'm not sure that it's a desired "feature" for the RHTAP pipeline.
cc @jduimovich @arewm

mmorhun: > Another idea, we could have many versions of the Task 

I'd really like to avoid this.
Otherwise, we only pile up a lot of tasks we'll have to maintain.

task/acs-deploy-check/0.1/acs-deploy-check.yaml
jduimovich: The "bring your own ACS" which is off-cluster is a use case we need to support.

task/acs-image-check/0.1/acs-image-check.yaml
jduimovich: FWIW - this same code, with insecure download -k code is found in the "official" task https://hub.tekton.dev/tekton/task/stackrox-image-scan. 
We need to work with the stackrox team to resolve the one correct way to get a binary for roxctl. Not sure why they control access to the CLI, as you can only use it with an installed ACS server.

jduimovich (approved):
Despite the issues concerning insecure download approved so we can continue to test

task/acs-deploy-check/0.1/acs-deploy-check.yaml
lcarva: I understand the requirement, but this can't be the only possible way to achieve it.

Is ACS on the same cluster? If so, you can assume a certain URL, no?

As a stop gap, we need to, at a bare minimum, add an EC policy that verifies the value of the `rox-central-endopint` parameter is one from a list of pre-authorized values (which can be customized.)

lcarva:
> Despite the issues concerning insecure download approved so we can continue to test

My understanding is that this is not what merging a PR in this repo means. As soon as it is merged we are giving the go-ahead that users can start using the Task. There are security concern that we should address. We may need to have special EC rules to ensure the Task is being used as expected, for example.

If you want to test, it's quite trivial to load the Task from a fork. Why not use that?

task/acs-image-check/0.1/acs-image-check.yaml
arewm: > If we use images that require subscription to pull in the RHTAP pipeline, it would mean, that whole RHTAP pipeline wouldn't run without subscription
Can we just reference the image in the task itself such that we can protect against the lack of valid credentials?

Users should have credentials for registry.redhat.io to use, but would this require _each_ user to specify the credentials (i.e. not the PEs)?

task/acs-deploy-check/0.1/acs-deploy-check.yaml
arewm: Would it be preferable to have that validation in EC or within the task itself? In-task validation seems more desirable for this use case than some others that have cropped up.

lcarva: Unclear what the best approach is here. It comes down to do we trust the task or do we only trust the task when used in a certain way.

task/acs-image-check/0.1/acs-image-check.yaml
mmorhun: @arewm this would require pull secret in the pipeline namespace (user's namespace).
Sharing the same token is not secure, so each user must provide own secret.

PR #777, opened 2024-02-01
Skip the pipeline-bundles when ec track bundle

simonbaird:
Tracking the pipeline bundles causes 845 queries to quay, which (we believe) is contributing to timeout problems causing the acceptable bundles image update to fail, which then causes bogus EC failures saying that tasks are using an unacceptable digest.

This is a quick minimal fix aimed at getting the "Red Hat Trusted App Pipeline / build-definitions-bundle-push" CI working again.

Ref: [EC-236](https://issues.redhat.com/browse/EC-236)
Ref: [EC-351](https://issues.redhat.com/browse/EC-351)

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -134,10 +134,17 @@ spec:
                 #!/usr/bin/env bash
                 set -euo pipefail
 
+                # Quick fix for timeout causing the acceptable bundles push to
+                # fail, see https://issues.redhat.com/browse/EC-351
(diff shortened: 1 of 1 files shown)


simonbaird:
@lcarva says this won't have the desired effect because `ec track bundle ... --freshen` is where all the existing pipeline bundles are pull in from the existing image.

simonbaird:
Replaced by #778 some time ago.

PR #778, opened 2024-02-01
Longer timeout for ec track bundle

simonbaird:
Lately the "Red Hat Trusted App Pipeline /
build-definitions-bundle-push" has been timing out which causes the acceptable bundles data to go stale, which causes bogus EC failures out unacceptable tasks.

This is a quick fix to get the CI working again. It's likely the root causes still need to be addressed. See the following Jiras for details:

Ref: https://issues.redhat.com/browse/EC-236
Ref: https://issues.redhat.com/browse/EC-351

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -154,6 +154,7 @@ spec:
                 ec track bundle --debug \
                   --input "oci:${DATA_BUNDLE_REPO}:latest" \
                   --output "oci:${DATA_BUNDLE_REPO}:${TAG}" \
+                  --timeout "15m0s" \
                   --freshen \
                   --prune \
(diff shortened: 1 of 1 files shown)


lcarva (approved):

.tekton/push.yaml
```diff
@@ -154,6 +154,7 @@ spec:
                 ec track bundle --debug \
                   --input "oci:${DATA_BUNDLE_REPO}:latest" \
                   --output "oci:${DATA_BUNDLE_REPO}:${TAG}" \
+                  --timeout "15m0s" \
```
simonbaird, line 157: Default is "5m0s" if you're wondering.

PR #788, opened 2024-02-06
Introduce `TARGET_STAGE` build parameter

msugakov:
In RHACS, we have multiple "slim" containers that are the same as their non-slim counterparts with just few things absent. These are `collector-slim`-`collector`, `scanner-slim`-`scanner`, `scanner-db-slim`-`scanner-db`.

It is natural for these cases to have a single Dockerfile for both "slim" and non-slim containers done in the following way

```Dockerfile
FROM some-base AS slim

# common build for slim and non-slim

FROM slim AS full

# extra stuff specific to non-slim
```

Both `docker build` and `buildah build` support specifying target stage to build. This way, building something-slim would be
```bash
$ buildah build --target=slim ...
```
Building something non-slim would go simply without the `--target` argument because the Dockerfile is built entirely by default
```bash
$ buildah build ...
```

Not having an option to select target forces us to have two dockerfiles for slim&non-slim pair with mostly identical content. This is inconvenient and creates maintenance overhead.

This change would allow us reuse code when building in Konflux.

I don't know what's the process to test this change for regressions and hope you can advise.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -69,6 +69,11 @@ spec:
     description: Target path on the container in which yum repository files should
       be made available
     name: YUM_REPOS_D_TARGET
+  - default: ""
+    description: Target stage in Dockerfile to build. If not specified, the Dockerfile
+      is processed entirely to (and including) its last stage.
+    name: TARGET_STAGE
+    type: string
   - description: The platform to build on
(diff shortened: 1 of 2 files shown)


arewm (approved):

msugakov:
It looks like `e2e-tests` failed but I don't know how to troubleshoot that.

chmeliik (approved):

PR #792, opened 2024-02-07
Do not allow image customization for task steps

rcerven:
[STONEBLD-2109](https://issues.redhat.com//browse/STONEBLD-2109)


stuartwdouglas (approved):

chmeliik (commented):
Could you explain more about why it's done this way? If I understand correctly, none of the params were removed, they're just silently ignored now.

Are we sure this doesn't break users?

mmorhun:
@chmeliik Robert has done investigation if / where the parameters are set. It turned out, that we have only internal usages of the parameter (in ci namespace). All users stick to the default value, so it's not a breaking change for them.

mmorhun:
@rcerven I think we should fail or at least warn users if they try to use deprecated image parameter.

mmorhun (approved):
LGTM

rcerven:
added warning when deprecated parameters are provided

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -169,6 +169,10 @@ spec:
       set -o verbose
       set -e
       cd $(workspaces.source.path)
+      if [ -n "$(params.BUILDER_IMAGE)" ]; then
```
chmeliik, line 172: I think our CI checks are broken - this should have caused a failure :confused: 

https://github.com/redhat-appstudio/build-definitions/blob/bc5eeaa574694a9ea449f38c69039edadc6127a4/.tekton/tasks/yaml-lint.yaml#L35

zregvart (approved):
For what it's worth +1 from me

task/buildah-remote/0.1/buildah-remote.yaml
zregvart: Yeah, I'm getting:

```
Tasks contains params in script section (https://tekton.dev/docs/pipelines/tasks/#substituting-in-script-blocks)
task/acs-deploy-check/0.1/acs-deploy-check.yaml
task/acs-image-check/0.1/acs-image-check.yaml
task/acs-image-scan/0.1/acs-image-scan.yaml
task/buildah/0.1/buildah.yaml
task/buildah-remote/0.1/buildah-remote.yaml
task/s2i-java/0.1/s2i-java.yaml
task/s2i-nodejs/0.1/s2i-nodejs.yaml
task/update-infra-deployments/0.1/update-infra-deployments.yaml
```

With the extracted script from https://github.com/redhat-appstudio/build-definitions/blob/main/.tekton/tasks/yaml-lint.yaml#L41C1-L51C11

chmeliik: Filed a bug for this

PR #794, opened 2024-02-07
Remove all labels inherited from last parent image

rcerven:
[STONEBLD-1767](https://issues.redhat.com//browse/STONEBLD-1767)

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -111,6 +111,44 @@ spec:
     - name: TARGET_STAGE
       value: $(params.TARGET_STAGE)
   steps:
+  - name: remove-labels-from-parent-image
+    image: quay.io/redhat-appstudio/hacbs-test:v1.2.0@sha256:73ff1801ceb0bb4c09f182f73d721e8f4aca31644c35e1f75c91550009ee5cf8
+    securityContext:
+      runAsUser: 0
+      capabilities:
+        add:
+          - SETFCAP
+    workingDir: $(workspaces.unset_labels_options.path)
+    script: |
+      #!/usr/bin/env bash
+      set -euo pipefail
+
+      SOURCE_CODE_DIR=source
+      if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
+      elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
+        dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
+      elif echo "$DOCKERFILE" | grep -q "^https\?://"; then
+        echo "Fetch Dockerfile from $DOCKERFILE"
+        dockerfile_path=$(mktemp --suffix=-Dockerfile)
+        http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path" "$DOCKERFILE")
+        if [ $http_code != 200 ]; then
+          echo "No Dockerfile is fetched. Server responds $http_code"
+          exit 1
+        fi
+      else
+        echo "Cannot find Dockerfile $DOCKERFILE"
+        exit 1
+      fi
+
+      UNSET_LABELS_OPTIONS_FILE="$(workspaces.unset_labels_options.path)/unset_labels_options"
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Would `buildah inspect` work?

```shell
buildah inspect \
  --format '{{ range $key, $value := .OCIv1.Config.Labels }}{{ printf "%s\n" $key }}{{ end }}' \
  "$image_that_must_be_pulled_locally"
```

task/buildah/0.1/buildah.yaml
```diff
@@ -111,6 +111,43 @@ spec:
...
+      else
+        echo "Cannot find Dockerfile $DOCKERFILE"
+        exit 1
+      fi
+
+      UNSET_LABELS_OPTIONS_FILE="$(workspaces.unset_labels_options.path)/unset_labels_options"
+      LAST_PARENT=$(grep -i '^\s*FROM' "$dockerfile_path" | tail -n 1 | sed 's/--platform=\S*//' | awk '{print $2}')
+      if [ "${LAST_PARENT}" != "scratch" ]; then
+        for I in $(skopeo inspect --no-tags docker://"${LAST_PARENT}" |jq -e ".Labels"|grep ':'|sed -e "s/^\s*\"\([^:]*\)\":.*$/\1/"); do
+          echo -n " --unsetlabel \"$I\"" >> $UNSET_LABELS_OPTIONS_FILE
```
arewm, line 148: While this is what we talked about doing ultimately/eventually, the scope of the issue associated with this PR ([STONEBLD-1767](https://issues.redhat.com//browse/STONEBLD-1767)) only indicates the removal of a subset of these labels. 

If so, the implementation would be much simpler as you would just have a static set of labels to unset.

chmeliik: IMO a static set of labels doesn't make sense generically (i.e. it's not something that should ever be part of the RHTAP buildah task). If anything, the set of labels would have to be configurable. At that point, I think the question is whether we want an "allowlist" (inherit only the specified labels) or a "denylist" (inherit all labels except those specified)

arewm: That's a valid point. The intention is certainly to come back and do a more generic approach later, but that is tied up in [KONFLUX-1329](https://issues.redhat.com/browse/KONFLUX-1329). Our current set of label requirements are ultimately derived from https://github.com/projectatomic/ContainerApplicationGenericLabels which is out of date in that it does not really conform to the OCI spec.

Since this broader effort will take more time and will require coordination with additional teams, I wanted to help improve the experience _today_ to remove specific labels that are added internally (i.e. according to the former spec) that we do not or cannot add -- like a `release`. This would prevent inheriting from parent images like ubi-minimal to therefore reduce the potential confusion of consumers that expect labels to be consistent with a static NVR tag.

rcerven:
> Would `buildah inspect` work?
> 
> ```shell
> buildah inspect \
>   --format '{{ range $key, $value := .OCIv1.Config.Labels }}{{ printf "%s\n" $key }}{{ end }}' \
>   "$image_that_must_be_pulled_locally"
> ```

I wanted to avoid pulling the image

chmeliik:
> > Would `buildah inspect` work?
> > ```shell
> > buildah inspect \
> >   --format '{{ range $key, $value := .OCIv1.Config.Labels }}{{ printf "%s\n" $key }}{{ end }}' \
> >   "$image_that_must_be_pulled_locally"
> > ```
> 
> I wanted to avoid pulling the image

So instead we add another step, copy-paste the Dockerfile-handling code, depend on skopeo and have to use the integration team's image? :smile: 

We already pull all base images when hermetic=true, let's just do it unconditionally?

task/buildah/0.1/buildah.yaml
chmeliik: Ok, I guess we can start with a "static denylist" approach and evaluate what makes sense before we include any similar functionality in RHTAP

lcarva: Adding a list of labels feels a lot like encoding a policy in this Task. I agree that it's ok for it to be a starting point, but we need to move away from that - fast. Can we make that clear in the code? A comment explaining that it's temporary and a link to the Jira issue where it'll be removed?

arewm: @chmeliik, this functionality seems scoped specifically to Konflux and not to RHTAP. If we want to enable similar functionality in RHTAP, I think that it would make more sense to have a model where users explicitly request the labels to _not_ inherit from parent images.

rcerven:
closing this PR as it removal of all labels will be part of KONFLUX-1329

PR #799, opened 2024-02-11
task/rpm-ostree: Use --entrypoint bash

cgwalters:
Required for https://github.com/CentOS/centos-bootc/pull/277

I am planning to potentially change to use bootc-image-builder for this, but it has a non-`bash` entrypoint.

### DIFF
--- a/task/rpm-ostree/0.1/rpm-ostree.yaml
+++ b/task/rpm-ostree/0.1/rpm-ostree.yaml
@@ -129,7 +129,7 @@ spec:
       REMOTESSHEOF
       chmod +x scripts/script-build.sh
       rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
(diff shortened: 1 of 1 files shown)


lmilbaum (approved):

arewm (approved):

arewm:
@cgwalters , you say that you are planning to change the builder image but that has a different entry point. At that time, would you plan to submit a PR to the build definitions to modify the _default_ builder image reference (reducing the need for it to be passed as a parameter) and also modifying the entrypoint? Or would you be more interested at that time to make the entrypoint configurable as well?

cgwalters:
We could try to just change the default builder image here, yes.  I'm ok with that, but it seems like a larger change.

I don't think it really makes sense to make the entrypoint configurable because this pipeline *relies* on it being bash.

arewm:
We can get this merged and address the change in the path with the default builder pipeline if needed in the future.

PR #805, opened 2024-02-13
buildah-rhtap: validate sbom generated from buildah-rhtap

srivickynesh:
buildah-rhtap: validate sbom generated from buildah-rhtap


mmorhun (commented):
Please add a description with the purpose of this task and when / for what it's supposed to be used.

task/sbom-verify-rhtap/0.1/validate-sbom-rhtap.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - $(params.IMAGE)
+      - --output-file
+      - /tmp/files/sbom-cyclonedx.json
+    volumeMounts:
+    - mountPath: /tmp/files
+      name: tmpfiles
+    workingDir: /tmp/files
+
+  - name: validate-sboms
+    image: quay.io/redhat-appstudio/test-image@sha256:a93cf4d0ce4ec82859a0836629dbc3fd3bf3589eb0cb7e64e8721740ec8d3773
```
chmeliik, line 35: How is this image built? Images used by the RHTAP tasks will need to be properly "productized" (still not fully clear what that means, but I think they will at least have to be built in Konflux)

task/sbom-verify-rhtap/0.1/validate-sbom-rhtap.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+      result=$(echo -n $(sbom-utility validate --input-file /tmp/files/sbom-cyclonedx.json))
+      if [[ ! $result =~ "SBOM valid against JSON schema: `true`" ]]
+      then
+        echo "Failure: SBOM Validation Failed" > $(results.validation-result.path)
+        echo "$result"
+        exit 1
+      else
+        echo "$result"
+        echo "Success: SBOM Validation Succeeded > $(results.validation-result.path)
```
tisutisu, line 51: missing closing `"` here

task/sbom-verify-rhtap/0.1/validate-sbom-rhtap.yaml
chmeliik: TBH I don't think that validating the SBOM schema is worth building an `sbom-utility` image ourselves

tisutisu, line 35: how this image is built.. where is the dockerfile?

srivickynesh: is it okay to reuse the image build in konflux, here?

chmeliik: Do you mean reuse this image? https://github.com/redhat-appstudio/build-definitions/blob/261d1c8ace528d3ebae7ebf03871d5fb5c812f9d/task/sbom-json-check/0.1/sbom-json-check.yaml#L20-L21

I don't think so, it's not built in Konflux and the Dockerfile looks like it wouldn't pass policies https://github.com/redhat-appstudio/hacbs-test

srivickynesh: okay, then we don't have any other options here other than building our images. But if that is not worth it then I can probably drop this PR? WDYT?

chmeliik: It depends what exactly is the motivation for including this task, what we want to achieve with it etc. IMO there's not that much value in verifying that Syft produced a valid CycloneDX document. That would be a serious bug in Syft.

Maybe we can have a weaker check (e.g. just check that the SBOM exists and is valid JSON)

Or we can have this check in CI tests, not in the pipeline itself

srivickynesh: Agree. I will include this in e2e tests when the CI is ready.

PR #806, opened 2024-02-13
RHTAPSRE-395: Add more cpus to buildah 24gb

gbenhaim:
The RHOAI team needs 10-20 CPUs to build their
component.

It also makes sense that if a build requires 24GB of RAM, it will also require a large amount of CPUs.

### DIFF
--- a/task/buildah-24gb/0.1/patch.yaml
+++ b/task/buildah-24gb/0.1/patch.yaml
@@ -7,3 +7,9 @@
 - op: replace
   path: /spec/steps/0/computeResources/requests/memory
   value: 20Gi
+- op: replace
+  path: /spec/steps/0/computeResources/limits/cpu
+  value: "20"
+- op: replace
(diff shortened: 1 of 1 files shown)


task/buildah-24gb/0.1/patch.yaml
```diff
@@ -7,3 +7,9 @@
 - op: replace
   path: /spec/steps/0/computeResources/requests/memory
   value: 20Gi
+- op: replace
+  path: /spec/steps/0/computeResources/limits/cpu
+  value: "10"
+- op: replace
+  path: /spec/steps/0/computeResources/requests/cpu
+  value: "20"
```
hugares, line 15: I think you reversed the numbers, should be requests 10 and limits 20

gbenhaim: fixed

hugares (approved):

gbenhaim:
@chmeliik we added dedicated node pool to handle this task.

PR #810, opened 2024-02-14
Use image digest instead of commit tag in order to have renovate updates

mmorhun:


.tekton/tasks/e2e-test.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       type: string
   steps:
     - name: e2e-test
-      image: quay.io/redhat-appstudio/e2e-tests:42f4d88c5decbb098dbea2937aef111f48585252
+      image: quay.io/redhat-appstudio/e2e-tests@sha256:90d6e83c272d1a75827f9789e1670c2de635a2f3850386c35f8fae28db1f5ccc
```
chmeliik, line 21: We somehow do get updates for these :thinking: #809

lcarva:
> Use image digest instead of commit tag in order to have renovate updates

To be clear, removing the tag means renovate will assume the tag `latest`. Not sure if those repos do use that tag.

stuartwdouglas (approved):

stuartwdouglas (changes requested):
I didn't see the earlier comment, and we use latest for JVM build service. I think think we currently have a way to produce this tag without adding a new pipeline run.

PR #814, opened 2024-02-15
Enable step_image_registries check for Tasks

lcarva:
This PR includes two commits:

1. Enable step_image_registries check for Tasks
    > This commit effectively enables the step_image_registries[1] policy rules on the Task definitions in this repo.
    > Since some Tasks do not yet comply with some of those policy rules, exceptions are added for those cases. 
    > Enabling this check now should help prevent new violations from being introduced.
    >
    > Since the policy now includes a very specific exclusion list for this
    > repo, the policy itself is now defined here.
    > 
    > Ref: https://issues.redhat.com/browse/EC-374
    > 
    > [1] https://enterprisecontract.dev/docs/ec-policies/task_policy.html#step_image_registries_package
2. Disable gitInitImage param from git-clone Task
    > This was missed in [STONEBLD-2109](https://issues.redhat.com/browse/STONEBLD-2109).

### DIFF
--- a/.tekton/tasks/ec-checks.yaml
+++ b/.tekton/tasks/ec-checks.yaml
@@ -33,7 +33,7 @@ spec:
       args=${files[*]/#/--file }
       echo "[DEBUG] Files parameter: ${args[*]}"
 
-      policy='enterprise-contract-service/redhat-trusted-tasks'
+      policy='./policies/all-tasks.yaml'
 
       ec validate input --policy "${policy}" --output yaml --strict=true ${args[*]}
   - name: validate-build-tasks
--- a/policies/all-tasks.yaml
+++ b/policies/all-tasks.yaml
@@ -1,0 +1,19 @@
+---
+# These policies are meant to be applied to all of the Tasks in this repo.
+sources:
+  - policy:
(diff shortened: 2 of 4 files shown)


lcarva:
cc @chmeliik , @rcerven

lcarva:
yaml lint is failing (unrelated to the changes in this PR). https://github.com/redhat-appstudio/build-definitions/pull/815 fixes it.

chmeliik (approved):
LGTM :+1:

chmeliik:
Related to the step_image_registries check, I'll mention that the rpm-ostree task gets away with having a [configurable image](https://github.com/redhat-appstudio/build-definitions/blob/101e6883ad546fdf4a55662394450a16a1738885/task/rpm-ostree/0.1/rpm-ostree.yaml#L18-L23), because it uses it in a much more [obfuscated way](https://github.com/redhat-appstudio/build-definitions/blob/101e6883ad546fdf4a55662394450a16a1738885/task/rpm-ostree/0.1/rpm-ostree.yaml#L132). Per discussions in Slack, the use case is valid but may need some special treatment

lcarva:
Ack. We have https://issues.redhat.com/browse/EC-415 for handling that particular task.

PR #816, opened 2024-02-15
buildah-rhtap: emit the SBOM_BLOB_URL result

brunoapimentel:
This Tekton result will contain the URL to the layer containing the SBOM, so it can be later be accessed by the EC. 

For future reference, the SBOM can be retrieved from a layer by using:
```
oras blob fetch quay.io/bpimente/test-images@sha256:509e7e73b03bb9c02f6e14a80acc81296396f538834dff09013f134f35a98591 --output -
```

### DIFF
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -35,6 +35,8 @@ spec:
     name: IMAGE_URL
   - description: Digests of the base images used for build
     name: BASE_IMAGES_DIGESTS
+  - description: Link to the SBOM layer pushed to the registry as part of an OCI artifact.
+    name: SBOM_BLOB_URL
   stepTemplate:
     env:
     - name: STORAGE_DRIVER
@@ -106,10 +108,15 @@ spec:
       name: tmpfiles
 
   - name: merge-sboms
-    image: registry.access.redhat.com/ubi9/python-39:1-165@sha256:4da8ddb12096a31d8d50e58ea479ba2fe2f252f215fbaf5bf90923a1827463ba
+    image: registry.access.redhat.com/ubi8/python-311@sha256:8ded4b6d8087706b6819ddda5d31f22b80e5aa4efa772e94d750699ccfbf98eb
+    env:
+      - name: RESULT_PATH
+        value: $(results.SBOM_BLOB_URL.path)
     script: |
       #!/bin/python3
+      import hashlib
       import json
+      import os
 
       ### load SBOMs ###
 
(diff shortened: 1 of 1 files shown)


brunoapimentel:
I wrote the algorithm to calculate the digest of the layer using Python, since it's currently part of the merge-sboms step, which is also writen in Python. Maybe it's best to do this in a separate script and write in shell?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -35,6 +35,8 @@ spec:
     name: IMAGE_URL
   - description: Digests of the base images used for build
     name: BASE_IMAGES_DIGESTS
+  - description: Link to the SBOM layer pushed to the registry as part of an OCI artifact.
+    name: SBOM_BLOB_URL
```
chmeliik, line 39: If the buildah task has a README, it would be good to document how this url is meant to be used. How does one even download the SBOM from it and verify its  checksum?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -106,10 +108,15 @@ spec:
       name: tmpfiles
 
   - name: merge-sboms
-    image: registry.access.redhat.com/ubi9/python-39:1-165@sha256:4da8ddb12096a31d8d50e58ea479ba2fe2f252f215fbaf5bf90923a1827463ba
+    image: registry.access.redhat.com/ubi8/python-311@sha256:8ded4b6d8087706b6819ddda5d31f22b80e5aa4efa772e94d750699ccfbf98eb
+    env:
+      - name: SBOM_BLOB_URL
+        value: $(results.SBOM_BLOB_URL.path)
```
chmeliik, line 114: nitpick: the name should make it clear that this is the result path, not a url

pipelines/docker-build-rhtap/patch.yaml
```diff
@@ -119,3 +119,8 @@
     taskRef:
       kind: Task
       name: acs-deploy-check
+- op: add
+  path: /spec/results/-
+  value:
+    name: SBOM_BLOB_URL
+    value: "$(tasks.build-container.results.SBOM_BLOB_URL)"
```
chmeliik, line 126: > needs to be properly added to the pipeline as well.

Why?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
chmeliik: As well as document why it's the layer url and not the url of the whole artifact

chmeliik:
> I wrote the algorithm to calculate the digest of the layer using Python, since it's currently part of the merge-sboms step, which is also writen in Python. Maybe it's best to do this in a separate step and write in shell?

Makes sense to me to keep it in the same step. Adding steps can also be counter-productive, since it reduces the overall result size limit :disappointed:

pipelines/docker-build-rhtap/patch.yaml
brunoapimentel: From my testing, if the result is not added to the pipeline, it does not show in the Openshift UI. I didn't try to retrieve it via CLI, so I just assumed it was not available at all.

chmeliik: It will be available on the TaskRun yaml, which EC does have access to. The result doesn't really seem like a user-facing one (unless it can serve as the LINK_TO_SBOM as well?)

brunoapimentel: Good point, so I'll drop this commit.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: I still need to add a README to it. For reference, we can fetch the layer using:
```
oras blob fetch quay.io/bpimente/test-images@sha256:509e7e73b03bb9c02f6e14a80acc81296396f538834dff09013f134f35a98591 --output -
```

Is it ok if the README is added as a follow up PR?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
brunoapimentel: Changed it. SBOM_BLOB_URL_RESULT_PATH sounded to Javaish to me, so I left it only as RESULT_PATH since we only have a single one for this step.

chmeliik (approved):
LGTM :+1:

lcarva (approved):

PR #817, opened 2024-02-16
Add update-deployment task

mmorhun:
Add task to update gitops repository.


task/update-deployment/0.1/update-deployment.yaml
jduimovich, line 1: This ok for now  but the component-id shoud become a parameter so when we have multiple components in a gitops repo it can update the correct one. 
That will require changes in the caller pipeline to separate multiple src-repos (src-1,src2) and both can update components in a single unrelated name common-gitops

rcerven (approved):

PR #825, opened 2024-02-20
buildah-rhtap: use productized Syft image

chmeliik:
RHTAP tasks must use productized images. Replace the previous Syft image with one that was recently released to registry.redhat.io.

### DIFF
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -97,7 +97,7 @@ spec:
     workingDir: $(workspaces.source.path)
 
   - name: generate-sboms
-    image: quay.io/redhat-appstudio/syft:v0.105.0@sha256:32a9d2007f2b042ceec4ef32fa1d90b8d28141822e7d9748f240da9d55c56601
+    image: registry.redhat.io/rh-syft-tech-preview/syft-rhel9:0.105.0@sha256:8d34c03188cf294a77339b2a733b1f6811263a369b309e6b170d9b489abc0334
     script: |
       syft dir:$(workspaces.source.path)/source --output cyclonedx-json@1.5=/tmp/files/sbom-source.json
       syft oci-dir:/tmp/files/image --output cyclonedx-json@1.5=/tmp/files/sbom-image.json


task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -97,7 +97,7 @@ spec:
     workingDir: $(workspaces.source.path)
 
   - name: generate-sboms
-    image: quay.io/redhat-appstudio/syft:v0.105.0@sha256:32a9d2007f2b042ceec4ef32fa1d90b8d28141822e7d9748f240da9d55c56601
+    image: registry.redhat.io/rh-syft-tech-preview/syft-rhel9:0.105.0@sha256:8d34c03188cf294a77339b2a733b1f6811263a369b309e6b170d9b489abc0334
```
chmeliik, line 100: Now I'm thinking - do all RHTAP users have to have access to registry.redhat.io? If not, this breaks the buildah task.

Do we have to move rh-syft to registry.access.redhat.com? CC @jduimovich

jduimovich: Yes, I think the images would be best in an open access registry in  the longer term.  I would say in the context of establishing an upstream konflux community which runs on non-openshift all tasks need open access images. 
OpenShift installs will have acccess to registry.redhat.io images 
We currently have some images in registry.redhad.io as per this table of  [Images and tasks](https://github.com/jduimovich/build-definitions/blob/create-owners-table/owners.md)

chmeliik: > Yes, I think the images would be best in an open access registry in the longer term. I would say in the context of establishing an upstream konflux community which runs on non-openshift all tasks need open access images.

Agreed, I'll try to figure out how to make the image publicly accessible

> We currently have some images in registry.redhad.io as per this table of [Images and tasks](https://github.com/jduimovich/build-definitions/blob/create-owners-table/owners.md)

I was going to say that the only immediately relevant task out of those is git-clone, which would be solved by switching to the community task from tektoncd-catalog

But then I looked at the task and it uses registry.redhat.io as well :confused: https://github.com/openshift-pipelines/tektoncd-catalog/blob/29276021fc0c301a9e2781f7a0cc549a2f9eac81/tasks/task-git/0.2.0/task-git.yaml#L209

PR #826, opened 2024-02-20
Add handling of entitlement secret

mkosiarc:
Users can add the entitlement secret "etc-pki-entitlement" to their namespace in the cluster. This will be then available for installing rpms during container build.

[STONEBLD-1589](https://issues.redhat.com//browse/STONEBLD-1589)

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -67,6 +67,10 @@ spec:
     description: Target stage in Dockerfile to build. If not specified, the Dockerfile is processed entirely to (and including) its last stage.
     type: string
     default: ""
+  - name: ENTITLEMENT_SECRET
+    description: Name of secret which contains the entitlement certificates
+    type: string
+    default: "etc-pki-entitlement"
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
@@ -109,6 +113,8 @@ spec:
       value: $(params.TARGET_STAGE)
     - name: PARAM_BUILDER_IMAGE
       value: $(params.BUILDER_IMAGE)
+    - name: ENTITLEMENT_SECRET
+      value: $(params.ENTITLEMENT_SECRET)
   steps:
   - image: quay.io/redhat-appstudio/buildah:v1.31.0@sha256:34f12c7b72ec2c28f1ded0c494b428df4791c909f1f174dd21b8ed6a57cf5ddb
     name: build
@@ -207,6 +213,13 @@ spec:
       [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
       [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
 
+      ENTITLEMENT_PATH="/entitlement"
+      if [ -d "$ENTITLEMENT_PATH" ]; then
+        cp -r --preserve=mode "$ENTITLEMENT_PATH" /tmp/entitlement
+        VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume /tmp/entitlement:/etc/pki/entitlement"
+        echo "Adding the entitlement to the build"
+      fi
+
       unshare -Uf $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah build \
         $VOLUME_MOUNTS \
         ${BUILDAH_ARGS[@]} \
@@ -234,6 +247,8 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+    - mountPath: "/entitlement"
+      name: etc-pki-entitlement
     workingDir: $(workspaces.source.path)
 
   - name: sbom-syft-generate
@@ -403,8 +418,12 @@ spec:
     workingDir: $(workspaces.source.path)
 
   volumes:
-  - emptyDir: {}
-    name: varlibcontainers
+  - name: varlibcontainers
+    emptyDir: {}
+  - name: etc-pki-entitlement
+    secret:
+      secretName: $(params.ENTITLEMENT_SECRET)
+      optional: true
   workspaces:
   - name: source
     description: Workspace containing the source code to build.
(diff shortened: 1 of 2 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -207,6 +213,13 @@ spec:
       [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
       [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
 
+      ENTITLEMENT_PATH="/entitlement"
+      if [ -n "${ENTITLEMENT_SECRET}" ] && [ -d "$ENTITLEMENT_PATH" ]; then
+        cp -r /entitlement /tmp/entitlement
```
mkosiarc, line 218: So this might seem puzzling, but I could not find any other solution so that these kind of warnings weren't shown:
```
time="2024-02-19T16:14:14Z" level=warning msg="pkg/bind: error detaching \"/var/tmp/buildah969021357/mnt/buildah-bind-target-11\": permission denied"
time="2024-02-19T16:14:14Z" level=warning msg="pkg/bind: error removing \"/var/tmp/buildah969021357/mnt/buildah-bind-target-11\": device or resource busy"
time="2024-02-19T16:14:14Z" level=warning msg="pkg/bind: error removing \"/var/tmp/buildah969021357/mnt\": directory not empty"
```

They must be related to the permissions of the volume. I tried changing the permissions of the volume, mounting it to different places and nothing worked expected making a copy of the volume and mounting that copy to the buildah. Any ideas?

mkosiarc:
This is how to test this/reproduce (you must have a redhat portal account with the rhel subscriptions)

1. run registry.access.redhat.com/ubi8/ubi:latest locally
2. run subscription-manager register command
3. copy the generated -key.pem and .pem files from /etc/pki/entitlement to your local machine
4. oc create secret generic etc-pki-entitlement --from-file <>-key.pem --from-file <>.pem
That is based on https://developers.redhat.com/articles/2022/01/24/create-entitled-builds-red-hat-subscriptions-openshift

And here is a testing build that I used, https://github.com/mkosiarc-rhtap/my-test/blob/main/Dockerfile
It won't install the libvirt-libs package successfully without working entitlement.

task/buildah/0.1/buildah.yaml
```diff
@@ -207,6 +213,13 @@ spec:
...
+        VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume /tmp/entitlement:/etc/pki/entitlement"
+        echo "Adding the entitlement to the build"
```
tisutisu, line 220: nitpick: when printing the log, it already added the `entitlement` .. will `Added the entitlement to the build` sound good here?

mkosiarc: Hmm, as I understand it at this point nothing was added, only variable was changed. The mounting to the buildah build command happens below

```
unshare -Uf $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah build \
        $VOLUME_MOUNTS \
```

task/buildah/0.1/buildah.yaml
```diff
@@ -207,6 +213,13 @@ spec:
       [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
       [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
 
+      ENTITLEMENT_PATH="/entitlement"
+      if [ -n "${ENTITLEMENT_SECRET}" ] && [ -d "$ENTITLEMENT_PATH" ]; then
```
chmeliik, line 217: I guess `[ -n "${ENTITLEMENT_SECRET}" ]` is there so that you can disable the entitlement handling even if an `etc-pki-entitlement` secret exists in your namespace? (By setting the param to an empty string).

Does that not cause issues for
```yaml
  - name: etc-pki-entitlement
    secret:
      secretName: ""  # when $(params.ENTITLEMENT_SECRET) is an empty string
      optional: true
```

task/buildah/0.1/buildah.yaml
chmeliik, line 218:
```suggestion
        cp -r "$ENTITLEMENT_PATH" /tmp/entitlement
```
just for consistency

chmeliik: Never seen that :confused: 

I guess copying the secret is fine, but maybe we could copy with `--preserve=mode` to preserve permissions?

mkosiarc: updated

task/buildah/0.1/buildah.yaml
mkosiarc: You are correct, in the case of "" it is a problem. I removed the first check, as I guess it's not really needed. either the entitlement secret is in the namespace or it isn't, so checking that should be enough.

task/buildah/0.1/buildah.yaml
mkosiarc: Updated to use the preserve option. Interestingly, it works without any warnings. This makes the issue of the buildah warnings even more puzzling...

chmeliik (commented):
LGTM, but you'll need to re-generate the buildah-remote task (the failing CI check should have the instructions)

And the Sonarcloud failure should be resolved by rebasing on main

task/buildah/0.1/buildah.yaml
```diff
@@ -67,6 +67,10 @@ spec:
     description: Target stage in Dockerfile to build. If not specified, the Dockerfile is processed entirely to (and including) its last stage.
     type: string
     default: ""
+  - name: ENTITLEMENT_SECRET
```
tisutisu, line 70: the changes is for adding support to download rpms during builds. Do we need to add a e2e-tests to cover this scenario?

mkosiarc: We do, part of the acceptance criteria is "Create a follow up story for e2e tests and link it with this story".

mkosiarc:
Should some docs be updated as well, so the users are aware of this ? @mmorhun

task/buildah/0.1/buildah.yaml
tisutisu: oh I see.. thanks for pointing it now. changes looks good to me.

chmeliik (approved):
LGTM :+1:

PR #832, opened 2024-02-21
Respect user configuration for Syft

chmeliik:
[STONEBLD-2095](https://issues.redhat.com//browse/STONEBLD-2095)

Syft makes many things configurable:
https://github.com/anchore/syft#configuration

For example, users can take advantage of this to get rid of false
positives. This will be useful for the Syft build itself:
https://github.com/redhat-appstudio/rh-syft/pull/21

Currently, our SBOM generation does not respect the user configuration.
Syft reads the config from the current working directory, not from the
target directory (https://github.com/anchore/syft/issues/2465).

Set the working directory to the root of the user's repository to ensure
we respect the configuration.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -351,6 +351,7 @@ spec:
     volumeMounts:
     - mountPath: /var/lib/containers
       name: varlibcontainers
+    workingDir: $(workspaces.source.path)/source
   - computeResources: {}
     image: quay.io/redhat-appstudio/hacbs-jvm-build-request-processor:127ee0c223a2b56a9bd20a6f2eaeed3bd6015f77
     name: analyse-dependencies-java-sbom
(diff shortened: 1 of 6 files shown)


chmeliik:
Testing in https://github.com/redhat-appstudio/rh-syft/pull/22

lcarva:
@chmeliik, are there any concerns that a user config may generate a less accurate SBOM (intentionally or not)?

chmeliik:
Added a topic to the Feb 27 architecture call

brunoapimentel:
Upon first review, LGTM. I think we should sync up with the docs team so that this gets properly documented in Konflux (otherwise, users might not even know this is possible).

I will run a few tests here so I can give you my final ack.

chmeliik:
just a rebase

chmeliik:
I believe tests are failing for the same reason as https://github.com/redhat-appstudio/build-definitions/pull/827#issuecomment-1977185422

The attestation was created 24 minutes after the image was pushed to quay.io

https://quay.io/repository/redhat-user-workloads/build-templates-e2e/test-app-832/devfile-sample-python-basic-qcxh?tab=tags&tag=latest

image

PR #834, opened 2024-02-21
buildah-rhtap: Fix SBOM_BLOB_RESULT for images that lack a tag

brunoapimentel:
The algorithm that is being used to strip the tag from the image will return an empty string in case the image already lacks the tag in the first place (i.e. quay.io/user/my-image).

This results in a SBOM_BLOB_RESULT that lacks the image name, and only contains the digest.

### DIFF
--- a/task/buildah-rhtap/0.1/buildah-rhtap.yaml
+++ b/task/buildah-rhtap/0.1/buildah-rhtap.yaml
@@ -117,6 +117,7 @@ spec:
       import hashlib
       import json
       import os
+      import re
 
       ### load SBOMs ###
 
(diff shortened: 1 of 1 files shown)


brunoapimentel:
I'm not sure if we want to allow RHTAP to build an image that lacks a tag (it will default to latest). But I stumbled upon this error in one of my tests, and we need to address it either way.

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -159,9 +159,13 @@ spec:
...
       image_name_parts = os.getenv("IMAGE").split(":")
-      # discard only the last part since the image can have a port number
-      # (i.e. registry:443/image/path:tag)
-      image_name_without_tag = ":".join(image_name_parts[0:-1])
+
+      # image name contains a port number (i.e. registry:443/image/path:tag)
+      if len(image_name_parts) == 3:
+        image_name_without_tag = ":".join(image_name_parts[0:-1])
+      else:
+        image_name_without_tag = image_name_parts[0]
```
chmeliik, line 167: What about `registry:443/image/path` (no tag)

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -158,10 +159,13 @@ spec:
...
-      image_name_parts = os.getenv("IMAGE").split(":")
-      # discard only the last part since the image can have a port number
-      # (i.e. registry:443/image/path:tag)
-      image_name_without_tag = ":".join(image_name_parts[0:-1])
+      # https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L160
+      tag_regex = "[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}"
+
+      # the tag must be after a colon, but always at the end of the string
+      # this avoids conflict with port numbers
+      image_without_tag = re.sub(f":{tag_regex}$", "", c)
```
chmeliik, line 167: I don't think `c` is the correct variable

brunoapimentel: :facepalm: Fixed it

PR #843, opened 2024-02-27
Consume ACS API endpoint from the secret together with the token

mmorhun:
After this change the secret with the ACS configuration should be created the following way:
```
oc create secret generic rox-api-token --from-literal=rox-api-token="${ROX_TOKEN}" --from-literal=rox-api-endpoint="${ROX_ENDPOINT}"
```


jduimovich (approved):
LGTM

PR #845, opened 2024-02-28
RHTAPSRE-405: Custom CA bundle for git-clone

gbenhaim:
Support custom CA bundle in the git-clone task.
Required when cloning using HTTPs from an SCM that has self-signed certificate.


MartinBasti (approved):

gbenhaim:
@chmeliik I didn't find any other way to do it.

gbenhaim:
@chmeliik here is the openshift documentation I followed https://docs.openshift.com/container-platform/4.14/networking/configuring-a-custom-pki.html#nw-proxy-configure-object_configuring-a-custom-pki

chmeliik (approved):

chmeliik:
> @chmeliik I didn't find any other way to do it.

Ok. Just beware that other tasks will need corresponding changes too.

MartinBasti:
Technically users could define pod templates in pipeline definition and mount CA there for every pod, but that's not something we should have at task level though.

PR #853, opened 2024-03-04
Show source image URL in summary

tkdchen:
[STONEBLD-1608](https://issues.redhat.com//browse/STONEBLD-1608)

This is a rework for the reverted version. In this implementation, the source-build task writes the build result to a file inside workspace so that any other task can read it. Task summary reads source image URL from this file and if the result file does not exist yet, just skip silently.

The workspace mount is optional for task summary. It works for users who update task summary bundle but not PipelineRun YAML.


tkdchen:
Tested with the following cases:

## Case 1
- source build is enabled

Result: task summary runs and source image URL is shown in summary.

## Case 2
- source build is not enabled

Result: task summary runs and source image URL is not shown in the summary logs.

## Case 3
- source build is enabled
- pipeline is not updated by specifying workspace to summary task

Result: task summary runs and no source image URL is shown in the summary logs.

task/summary/0.1/summary.yaml
```diff
@@ -45,5 +51,18 @@ spec:
...
+          for f in ${fields[@]}; do
+            if [ -n "$found" ]; then
+              echo "Generated Source Image is in : $(echo $f | tr -d '",')"
+              break
+            fi
+            if [ "$f" == '"image_url":' ]; then
+              found=yes
+            fi
+          done
+        fi
```
chmeliik, line 66: Instead of parsing JSON in Bash, can the source-build task write something more easily parse-able to the workspace file? Maybe just the image url?

tkdchen: I was intended to make the result file be general for sharing with other tasks, not only for summary even though only summary uses for now.

tkdchen: Oh, you remind me that we can use https://github.com/redhat-appstudio/build-tasks-dockerfiles/tree/main/task-toolset @mmorhun

chmeliik: > make the result file be general for sharing with other tasks

IMO that's generally a bad idea. Files in the workspace can't really be trusted, since the user can always add a custom task that modifies the workspace content. That's also why the EC team is doing the Trusted Artifacts work #791

chmeliik: (although for the show-summary task this is fine, it's purpose is just to inform the user)

tkdchen: @chmeliik PTAL https://github.com/redhat-appstudio/build-definitions/pull/857

mmorhun (approved):

tkdchen:
Rebased. No update to the changes.

PR #854, opened 2024-03-04
Store the base image digests in the correct order

mkosiarc:
Previously the base image digest were not guaranteed to be in the correct order because the output was generated just from the output of "buildah images". The output of that command is sorted according to the creation date of the image, which is not what we want.

The new approach outputs the digests in the order in which they were found in the dockerfile.

[STONEBLD-2203](https://issues.redhat.com//browse/STONEBLD-2203)

The "FROM scratch" problem is not yet resolved here. It will be dealt with in another story/PR as that might be solved on the source-build task side.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -218,10 +218,11 @@ spec:
 
       BUILDAH_ARGS=()
 
+      BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}')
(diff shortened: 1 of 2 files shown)


mmorhun (approved):

tkdchen (approved):

chmeliik (commented):
Could this also be relevant for the s2i build tasks?

mkosiarc:
> Could this also be relevant for the s2i build tasks?

From what I understand, these tasks will contain only one base image, the one that is specified as parameter. So I think there should not be a problem with the order there.

chmeliik (approved):

PR #857, opened 2024-03-05
Show source image URL in summary

tkdchen:
[STONEBLD-1608](https://issues.redhat.com//browse/STONEBLD-1608)

source-build task now always runs but checks the input param
BUILD_SOURCE_IMAGE to determine whether to start the build. Task summary
accepts a new param of source image URL.

Both task source-build and summary have change to the parameters, then
new versions are made and migration documents are prepared for updating
the existing PipelineRun YAML files.

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -118,9 +118,6 @@ spec:
         - input: $(tasks.init.results.build)
           operator: in
           values: ["true"]
-        - input: $(params.build-source-image)
-          operator: in
-          values: ["true"]
       runAfter:
         - build-container
       taskRef:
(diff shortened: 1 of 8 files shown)


chmeliik (commented):
I'm wondering if these changes are worth it. Making the users do migration changes just to show the source image url in the summary doesn't seem like good value

pipelines/template-build/template-build.yaml
```diff
@@ -118,9 +118,6 @@ spec:
         - input: $(tasks.init.results.build)
           operator: in
           values: ["true"]
-        - input: $(params.build-source-image)
```
mmorhun, line 121: Does this change makes building source image mandatory for all users?

PR #863, opened 2024-03-06
Update quay.io/opdev/preflight:stable docker digest to add156…

skattoju:
Updates preflight image to the latest release and adjusts the path in which results are stored due to multi-arch support.

### DIFF
--- a/task/ecosystem-cert-preflight-checks/0.1/ecosystem-cert-preflight-checks.yaml
+++ b/task/ecosystem-cert-preflight-checks/0.1/ecosystem-cert-preflight-checks.yaml
@@ -13,7 +13,7 @@ spec:
       description: Preflight pass or fail outcome.
   steps:
     - name: check-container
-      image: quay.io/opdev/preflight:stable@sha256:d990cbe8fb0db4dff0f7609403992f273f07d55f89dc7d2f046eb8f9c3e48110
+      image: quay.io/opdev/preflight:stable@sha256:add15669e17a86d807be05671f3c9834161d7af6f41bf4a50969be2da0487fbc
       args: ["check", "container", "$(params.image-url)"]
       volumeMounts:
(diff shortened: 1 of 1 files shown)


task/ecosystem-cert-preflight-checks/0.1/ecosystem-cert-preflight-checks.yaml
```diff
@@ -25,7 +25,7 @@ spec:
           mountPath: /artifacts
       script: |
         # Expected results directory
-        RESULT_JSON_PATH=artifacts/results.json
+        RESULT_JSON_PATH=artifacts/amd64/results.json
```
arewm, line 28: Under what conditions will this architecture change?

skattoju: amd64 is the default architecture so this should not change. Additional folders will be created in case preflight is provided a multi-arch image ex arm64 but my understanding is that there will always be a default amd64 folder.

arewm: Instead of hard coding this, would it be possible to inspect the image to see what architectures are present? It would be preferable if we don't need to modify the task to increase the supported architectures. We already support building amd64, arm64, ppc64le, and s390x architectures. While it might be a reasonable assumption that amd64 is always present, that isn't necessarily a requirement.

skattoju: agreed 👍  Will change it to process results from all architectures if present.

MartinBasti:
I think that this is the root cause, it looks like the intermittent error where secret is prematurely removed (or wrong secret is used?)
image

MartinBasti (approved):

PR #867, opened 2024-03-09
[CVP-4034] support scanning Image Index in clair-scan task

yashvardhannanavati:

### DIFF
--- a/task/clair-scan/0.1/clair-scan.yaml
+++ b/task/clair-scan/0.1/clair-scan.yaml
@@ -24,7 +24,38 @@ spec:
       description: Tekton task test output.
     - name: CLAIR_SCAN_RESULT
       description: Clair scan result.
+    - name: IMAGES_PROCESSED
+      description: Images processed in the task.
   steps:
+    - name: get-image-manifests
+      image: quay.io/redhat-appstudio/hacbs-test:v1.3.4@sha256:b909fe6111e04169742e23a5e515aa549aa8c09ce0348f4560ae83d3c174a15a
+      # the clair-in-ci image neither has skopeo or jq installed. Hence, we create an extra step to get the image manifest digests
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+        - name: IMAGE_DIGEST
+          value: $(params.image-digest)
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
+      script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+        . /utils.sh
+
+        imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
+        # strip new-line escape symbol from parameter and save it to variable
+        imageanddigest=$(echo $imagewithouttag@$IMAGE_DIGEST)
+        echo "Inspecting raw image manifest $imageanddigest."
+
+        # Get the arch and image manifests by inspecting the image. This is mainly for identifying image indexes
+        image_manifests=$(get_image_manifests -i ${imageanddigest})
+        if [ -n "$image_manifests" ]; then
+          echo "$image_manifests" | jq -r 'to_entries[] | "\(.key) \(.value)"' | while read -r arch arch_sha; do
+            echo "$arch_sha" > /tekton/home/image-manifest-$arch.sha
+          done
+        fi
     - name: get-vulnerabilities
       image: quay.io/redhat-appstudio/clair-in-ci:v1  # explicit floating tag, daily updates, per arch call this is exempt for now for use of image digest
       imagePullPolicy: Always
@@ -36,11 +67,29 @@ spec:
       script: |
         #!/usr/bin/env bash
 
-        imagewithouttag=$(echo $IMAGE_URL | sed "s/\(.*\):.*/\1/" | tr -d '\n')
-        # strip new-line escape symbol from parameter and save it to variable
-        imageanddigest=$(echo $imagewithouttag@$IMAGE_DIGEST)
+        imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
+        images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
+        digests_processed=()
 
-        clair-action report --image-ref=$imageanddigest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result.json || true
+        for sha_file in /tekton/home/image-manifest-*.sha; do
+          if [ -e "$sha_file" ]; then
+            arch_sha=$(cat "$sha_file")
+            arch=$(basename "$sha_file" | sed 's/image-manifest-//;s/.sha//')
+            arch_specific_digest="$imagewithouttag@$arch_sha"
+
+            echo "Running clair-action on $arch image manifest."
+            # run the scan for each image manifest in the image index
+            clair-action report --image-ref=$arch_specific_digest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result-$arch.json || true
+
+            digests_processed+=("\"$arch_sha\"")
+          fi
+        done
+
+        digests_processed_string=$(IFS=,; echo "${digests_processed[*]}")
+
+        # add the image_index to the processed digests list and store the result in a file
+        images_processed=$(echo "${images_processed_template/\[%s]/[$digests_processed_string]}")
+        echo "$images_processed" > /tekton/home/images-processed.json
     - name: conftest-vulnerabilities
       image: quay.io/redhat-appstudio/hacbs-test:v1.3.5@sha256:975c09b7a97a57563695cf0d51b3f2af0fcc91c468e11f06586c70f85401723f
       # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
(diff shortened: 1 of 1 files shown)


yashvardhannanavati:
Image 3-11-24 at 7 19 PM

Tested on a multi-arch image. The result size is `955 bytes`

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -24,7 +24,38 @@ spec:
...
+
+        imagewithouttag=$(echo $IMAGE_URL | sed "s/\(.*\):.*/\1/" | tr -d '\n')
+        # strip new-line escape symbol from parameter and save it to variable
+        imageanddigest=$(echo $imagewithouttag@$IMAGE_DIGEST)
+        echo "Inspecting raw image manifest $imageanddigest."
+
+        # check if the image is a manifest list
+        skopeo_inspect_raw=$(skopeo inspect --no-tags --raw docker://$imageanddigest)
+        if echo "$skopeo_inspect_raw" | jq .mediaType | grep -q "oci.image.index"; then
+          for arch in $(echo "$skopeo_inspect_raw" | jq -r '.manifests[].platform.architecture'); do
```
MartinBasti, line 52: I did something similar just by using JQ

https://github.com/redhat-appstudio/build-definitions/blob/acacbe1a5b84031157069a4df91b55029543f0d6/task/inspect-image/0.1/inspect-image.yaml#L98

given that multiple tasks will need the same code for detecting manifest images, I'd like to have a reusable function in hacbs-test to be used in all tasks, instead writing ad-hoc code per task

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -39,8 +70,30 @@ spec:
...
+            # run the scan for each image manifest in the image index
+            clair-action report --image-ref=$arch_specific_digest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result-$arch.json || true
 
-        clair-action report --image-ref=$imageanddigest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result.json || true
+            digests_processed+="\"$arch_sha\", "
+          fi
+        done
+
+        # run the scan for the Image Index itself
+        clair-action report --image-ref=$imageanddigest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result-image_index.json || true
```
MartinBasti, line 91: I'm just afraid this will produce duplicated results, as index image would resolve into the x86_64 architecture, and that image would be tested twice.

We should check "index image" only when it's not an index image

MartinBasti:
Please improve commit message and describe also changes in results

task/clair-scan/0.1/clair-scan.yaml
yashvardhannanavati: Added a function to do `skopeo inspect` on the images. Please let me know what you think: https://github.com/redhat-appstudio/hacbs-test/pull/227 

The reason I did not add function to filter the `sha` and` arch` is because each function might parse the output differently.

MartinBasti: I meant function to extract image manifests from index image, not replacement of skopeo inspect call

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -56,30 +109,53 @@ spec:
...
-              high: (.[] | .warnings? // [] | map(select(.metadata.details.name=="clair_high_vulnerabilities").metadata."vulnerabilities_number" // 0)| add // 0),
-              medium: (.[] | .warnings? // [] | map(select(.metadata.details.name=="clair_medium_vulnerabilities").metadata."vulnerabilities_number" // 0)| add // 0),
-              low: (.[] | .warnings? // [] | map(select(.metadata.details.name=="clair_low_vulnerabilities").metadata."vulnerabilities_number" // 0)| add // 0),
-              unknown: (.[] | .warnings? // [] | map(select(.metadata.details.name=="clair_unknown_vulnerabilities").metadata."vulnerabilities_number" // 0)| add // 0)
-            }}' /tekton/home/clair-vulnerabilities.json | tee $(results.CLAIR_SCAN_RESULT.path)
+        clair_vulnerabilities_files=$(ls /tekton/home/clair-vulnerabilities-*.json)
+        results=""
+        for vulnerabilities_file in $clair_vulnerabilities_files; do
+          file_suffix=$(basename $vulnerabilities_file| sed 's/clair-vulnerabilities-//;s/.json//')
+          results+="$(jq -rce --arg TYPE "$file_suffix" \
```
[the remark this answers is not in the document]
yashvardhannanavati: Added the discussed format to the ADR: https://github.com/redhat-appstudio/architecture/pull/174

yashvardhannanavati:
single-arch image
clair-scan-image-manifest

multi-arch image index
clair-scan-multi-arch

task/clair-scan/0.1/clair-scan.yaml
yashvardhannanavati: makes sense. done

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -24,7 +24,39 @@ spec:
...
+        imagewithouttag=$(echo $IMAGE_URL | sed "s/\(.*\):.*/\1/" | tr -d '\n')
+        # strip new-line escape symbol from parameter and save it to variable
+        imageanddigest=$(echo $imagewithouttag@$IMAGE_DIGEST)
+        echo "Inspecting raw image manifest $imageanddigest."
+
+        # check if the image is a manifest list
+        image_index_manifests=$(get_image_index_manifests -i ${imageanddigest})
+        if [ -n "$image_index_manifests" ]; then
+          echo "$image_index_manifests" | jq -r 'to_entries[] | "\(.key) \(.value)"' | while read -r arch arch_sha; do
+            echo "$arch_sha" > /tekton/home/image-manifest-$arch.sha
```
MartinBasti, line 55: What prevents us from merging these steps and usie JSON directly? Every extra step causes extra load, because tekton runs all step containers in paralel.

yashvardhannanavati: @MartinBasti So the original file has two steps, one using the clair image and the other using the hacbs-test image. The issue is that the clair image doesn't have `jq` or `skopeo` so I cannot use that step and we also needed to inspect the image before the clair step. Hence, I had to create a third step.

yashvardhannanavati:
Made the changes that that will be required after https://github.com/redhat-appstudio/hacbs-test/pull/230 is merged. 

Once the hacbs-test PR is merged, I will change the digest for the hacbs-test image in this task.

task/clair-scan/0.1/clair-scan.yaml
MartinBasti: okay, it may deserve comment in code, so I won't find in the hard way during refactoring in future

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -37,12 +67,45 @@ spec:
...
+          if [[ "$IMAGE_DIGEST" == "$digest" ]]; then
+            digest_found=1
+            break
+          fi
+        done
+
+        if [[ $digest_found -eq 0 ]]; then
+          digests_processed+=("$IMAGE_DIGEST")
+        fi
+        # digests_processed_string=$(jq -nc '$ARGS.positional | join(", ")' --args -- "${digests_processed[@]}")
```
MartinBasti, line 100: this looks like forgotten line

yashvardhannanavati: Ah, yes! fixed it :)

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -37,12 +67,44 @@ spec:
...
+        digest_found=0
+        for digest in "${digests_processed[@]}"; do
+          if [[ "$IMAGE_DIGEST" == "$digest" ]]; then
+            digest_found=1
+            break
+          fi
+        done
+
+        if [[ $digest_found -eq 0 ]]; then
+          digests_processed+=("$IMAGE_DIGEST")
```
MartinBasti, line 98: this is like a grey area, I'm not sure if we want to include index image digests in output

@lcarva Does EC checks also index images?

MartinBasti:
except one comment where I'm not sure, it LGTM :rocket:

task/clair-scan/0.1/clair-scan.yaml
lcarva: Today if an Image Index is provided then EC is just going to check the first architecture. There's planned work to make that better. Eventually EC is going to check each Image Manifest.

I can see an argument here that clair-scan doesn't actually scan the Image Index. It scans each Image Manifest individually and that's what should be reported. At the moment, I don't have a good reason to add the Image Index to the list of images checked. I would suggest not adding it. (We may have to revisit at some point though.)

@arewm, wdyt?

arewm: I would prefer to leave the Image Index. It isn't scanned because there isn't any content to it other than references, therefore the scan of the reference itself is a no-op. I see it being beneficial to include because this is the image digest that many people will primarily see.

lcarva: My concern with leaving it in is that this creates a special case on the verification side, but it's probably not a big deal. Let's leave it in then.

arewm: What is the special case on the verification side?

lcarva: One of the values is the digest of the Image Index, not an Image Manifest. Not exactly sure how that will come into play though.

yashvardhannanavati: I sense most of us are leaning towards leaving it out. I will remove the image index digest from the IMAGES_PROCESSED result. Please let me know if you feel otherwise strongly.

task/clair-scan/0.1/clair-scan.yaml
yashvardhannanavati: Done :)

task/clair-scan/0.1/clair-scan.yaml
arewm: As @lcarva told me, it is always easier to add something later if needed than it is to take something out.

arewm (approved):

MartinBasti (approved):

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -24,7 +24,38 @@ spec:
...
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
+      script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+        . /utils.sh
+
+        imagewithouttag=$(echo $IMAGE_URL | sed "s/\(.*\):.*/\1/" | tr -d '\n')
```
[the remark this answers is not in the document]
yashvardhannanavati: yes, it would. Changed it. thanks! :)

dirgim (approved):
LGTM

task/clair-scan/0.1/clair-scan.yaml
```diff
@@ -24,7 +24,38 @@ spec:
...
+      # the clair-in-ci image neither has skopeo or jq installed. Hence, we create an extra step to get the image manifest digests
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+        - name: IMAGE_DIGEST
+          value: $(params.image-digest)
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
```
jsztuka, line 41: Why are we including SETFCAP here? I don't see any `oc image extract`

jsztuka (approved):
lgtm

yashvardhannanavati:
Screenshot 2024-04-09 at 4 50 37 PM

multi-arch test with resolved merge conflicts

Screenshot 2024-04-09 at 6 35 41 PM

single-arch test with resolved merge conflicts

PR #871, opened 2024-03-12
fix(KFLUXBUGS-1152): add support for multi-arch

scoheb:
- if IMAGE_URL is a multi-arch one, print out sbom
      for each arch present

### DIFF
--- a/task/show-sbom/0.1/README.md
+++ b/task/show-sbom/0.1/README.md
@@ -2,8 +2,13 @@
 
 Shows the Software Bill of Materials (SBOM) generated for the built image in CyloneDX JSON format.
 
+The parameter named PLATFORM can be used to specify the arch to display the sbom for in the case of a multi-arch image. 
+In the case of a single arch image, the parameter is ignored. 
+If PLATFORM is empty and the image is multi-arch, the task defaults to 'linux/amd64'
+
 ## Parameters
-|name|description|default value|required|
-|---|---|---|---|
-|IMAGE_URL|Fully qualified image name to show SBOM for.||true|
+| name      | description                                                                                                                                                                                               | default value | required  |
+|-----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|-----------|
+| IMAGE_URL | Fully qualified image name to show SBOM for.                                                                                                                                                              |               | true      |
+| PLATFORM  | Specific architecture to display the SBOM for. An example arch would be "linux/amd64". If IMAGE_URL refers to a multi-arch image and this parameter is empty, the task will default to use "linux/amd64". | linux/amd64   | false     |
 
--- a/task/show-sbom/0.1/show-sbom.yaml
+++ b/task/show-sbom/0.1/show-sbom.yaml
@@ -15,30 +15,57 @@ spec:
     - name: IMAGE_URL
       description: Fully qualified image name to show SBOM for.
       type: string
+    - name: PLATFORM
+      description: Specific architecture to display the SBOM for. An example arch would be "linux/amd64". If IMAGE_URL
+        refers to a multi-arch image and this parameter is empty, the task will default to use "linux/amd64".
+      type: string
+      default: "linux/amd64"
   steps:
   - name: show-sbom
-    image: quay.io/redhat-appstudio/cosign:v2.1.1@sha256:c883d6f8d39148f2cea71bff4622d196d89df3e510f36c140c097b932f0dd5d5
+    image: quay.io/redhat-appstudio/appstudio-utils:3e548a38b3ad183262a25bc2a4eb6b5367b83fb5
     # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
     # the cluster will set imagePullPolicy to IfNotPresent
     # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
     env:
     - name: IMAGE_URL
       value: $(params.IMAGE_URL)
+    - name: PLATFORM
+      value: $(params.PLATFORM)
     script: |
-      #!/busybox/sh
-      status=-1
-      max_try=5
-      wait_sec=2
-      for run in $(seq 1 $max_try); do
-        status=0
-        cosign download sbom $IMAGE_URL 2>>err
-        status=$?
-        if [ "$status" -eq 0 ]; then
-          break
-        fi
-        sleep $wait_sec
-      done
-      if [ "$status" -ne 0 ]; then
+      #!/bin/bash
+
+      download_sbom_with_retry() {
+        status=-1
+        max_try=5
+        wait_sec=2
+
+        PLATFORM_ARG="$1"
+        for run in $(seq 1 $max_try); do
+          status=0
+          cosign download sbom $PLATFORM_ARG $IMAGE_URL 2>>err
+          status=$?
+          if [ "$status" -eq 0 ]; then
+            break
+          fi
+          sleep $wait_sec
+        done
+        if [ "$status" -ne 0 ]; then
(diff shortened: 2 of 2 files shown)


arewm (approved):
LGTM, thanks.

chmeliik:
Similar: #873

ralphbean:
Note, the other architectures are actually important. For the bootc images, for instance, we expect the sbom content to differ between the different arches - and those differences will be important for some users to know.

scoheb:
updated PR to use different task image that includes `skopeo` and included looping over all the arches present.

task/show-sbom/0.1/show-sbom.yaml
```diff
@@ -10,35 +10,65 @@ metadata:
...
+        echo "Single arch image"
+        echo ""
+        download_sbom_with_retry ""
+      else
+        for arch in $ARCHES; do
+          echo ""
+          echo "Arch: $arch"
+          echo ""
+          download_sbom_with_retry " --platform=linux/$arch "
+        done
```
arewm, line 73: Since this is just a task to show the summary in the console, I don't know if users would get any value iterating through all of the arches and showing _all_ SBOMs. I wouldn't expect this to be a primary means to see the information. Instead, this might decrease the utility as it just increases the size of the logs.

scoheb:
> Note, the other architectures are actually important. For the bootc images, for instance, we expect the sbom content to differ between the different arches - and those differences will be important for some users to know.
> 
> For example, try one of the centos-bootc images: quay.io/redhat-user-workloads/centos-bootc-tenant/centos-bootc/centos-bootc. Download the sbom twice, once with --platform linux/amd64 and once with --platform linux/arm64. Note the grub2-pc package in the diff between those SBOMs.

task/show-sbom/0.1/show-sbom.yaml
scoheb: See @ralphbean 's comment:

> https://github.com/redhat-appstudio/build-definitions/pull/871#issuecomment-1991902775

arewm: Yes, the SBOM content is important, but I don't know if it is important in _this_ context. This Tekton task is just an "quick method" to show users what is in the builds. I posit that if users want to see detailed information then they will likely try to either access it locally where they can target specific architectures or they will upload the SBOMs to some tool like guac.sh.

scoheb:
so we remove the task from the pipeline? I am not familiar with the history of why this task was needed in the first place.

scoheb:
That's what I was thinking too. Although we still have to detect if it is a multi-arch image because you cannot specify a platform for a single Arch image with cosign.

scoheb:
@arewm @ralphbean please review

task/show-sbom/0.1/show-sbom.yaml
```diff
@@ -15,30 +15,60 @@ spec:
...
+          fi
+          sleep $wait_sec
+        done
+        if [ "$status" -ne 0 ]; then
           echo "Failed to get SBOM after ${max_try} tries" >&2
           cat err >&2
+        fi
+      }
+      echo "Inspecting image ${IMAGE_URL}"
+
```
ralphbean, line 57: Does this need to be dropped so as not to break the e2e-tests and other sensitive parsers who are expecting only JSON here?

task/show-sbom/0.1/show-sbom.yaml
```diff
@@ -15,30 +15,60 @@ spec:
...
+      fi
+
+      if [ -z "${ARCHES}" ] ; then
+        # single arch image
+        download_sbom_with_retry ""
+      else
+        if [ -z "${PLATFORM}" ]; then
+          PLATFORM="amd64"
+        fi
+        download_sbom_with_retry " --platform=linux/$PLATFORM "
```
ralphbean, line 73: Let's not hardcode the "linux/" prefix here to make the task more generic. Let PLATFORM be exactly whatever the user specifies and modify the documentation to tell users to set it to "linux/amd64" if they want amd64.

Note, there are actually non-linux containers like `mcr.microsoft.com/windows/servercore:ltsc2022`.

task/show-sbom/0.1/README.md
```diff
@@ -2,8 +2,13 @@
...
+If PLATFORM is empty and the image is multi-arch, the task defaults to 'amd64'
+
 ## Parameters
-|name|description|default value|required|
-|---|---|---|---|
-|IMAGE_URL|Fully qualified image name to show SBOM for.||true|
+| name      | description                                                                                                                                                                                   | default value | required  |
+|-----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|-----------|
+| IMAGE_URL | Fully qualified image name to show SBOM for.                                                                                                                                                  |               | true      |
+| PLATFORM  | Specific architecture to display the SBOM for. An example arch would be "amd64". If IMAGE_URL refers to a multi-arch image and this parameter is empty, the task will default to use "amd64". |               | false     |
```
ralphbean, line 13: See comment below - I think this should be updated to expect a string like "linux/amd64" so that we don't box ourselves in.

ralphbean (changes requested):
See inline comments above. Consider changing both the PLATFORM string and the echo statement.

task/show-sbom/0.1/show-sbom.yaml
scoheb: true! dropped.

task/show-sbom/0.1/show-sbom.yaml
scoheb: updated to be more flexible.

task/show-sbom/0.1/README.md
scoheb: updated

task/show-sbom/0.1/show-sbom.yaml
```diff
@@ -15,30 +15,59 @@ spec:
...
+        ARCHES=$(jq -r '.manifests[].platform.architecture' <<< $RAW_OUTPUT)
+      fi
+
+      if [ -z "${ARCHES}" ] ; then
+        # single arch image
+        download_sbom_with_retry ""
+      else
+        if [ -z "${PLATFORM}" ]; then
+          PLATFORM="linux/amd64"
+        fi
```
arewm, line 71: Would it be simpler if we just provided a default for the parameter instead of this?

scoheb: yes sir! updated.

ralphbean (approved):

task/show-sbom/0.1/README.md
```diff
@@ -2,8 +2,13 @@
...
+
 ## Parameters
-|name|description|default value|required|
-|---|---|---|---|
-|IMAGE_URL|Fully qualified image name to show SBOM for.||true|
+| name      | description                                                                                                                                                                                               | default value | required  |
+|-----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|-----------|
+| IMAGE_URL | Fully qualified image name to show SBOM for.                                                                                                                                                              |               | true      |
+| PLATFORM  | Specific architecture to display the SBOM for. An example arch would be "linux/amd64". If IMAGE_URL refers to a multi-arch image and this parameter is empty, the task will default to use "linux/amd64". |               | false     |
```
arewm, line 13: Nit: The default value is not included in the table column

arewm (approved):

task/show-sbom/0.1/README.md
scoheb: fixed.

arewm (approved):
Thanks!

tkdchen (commented):
LGTM.

A new parameter is added, we need to make a new version with MIGRATION.md for this task.

scoheb:
> A new parameter is added, we need to make a new version with MIGRATION.md for this task.

README mentions this: **Adding a new parameter with a default value does not require the task version increase**

so we should be ok as-is

PR #872, opened 2024-03-12
Add platform specificity to the show-sbom task

ralphbean:
I noticed in the logs of the show-sbom task for a multiarch image, the following error:

```
[show-sbom] This multiarch image does not have an SBOM attached at the index level.
[show-sbom] Try using --platform with one of the following architectures:
[show-sbom] linux/amd64, linux/arm64
[show-sbom] 
[show-sbom] Error: no SBOM found attached to image index
[show-sbom] main.go:74: error during command execution: no SBOM found attached to image index
```

This should make the show-sbom task default to linux/amd64, but give the user a way to show other architecture by setting params in their pipeline or adding additional show-sbom tasks, one per architecture, if that is important to them.

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -224,6 +224,8 @@ spec:
       params:
       - name: IMAGE_URL
         value: $(tasks.build-container.results.IMAGE_URL)
+      - name: PLATFORM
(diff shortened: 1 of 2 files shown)


ralphbean:
Dropping this. Need to rebase.

PR #873, opened 2024-03-12
Add platform specificity to the show-sbom task

ralphbean:
I noticed in the logs of the show-sbom task for a multiarch image, the following error:

```
[show-sbom] This multiarch image does not have an SBOM attached at the index level.
[show-sbom] Try using --platform with one of the following architectures:
[show-sbom] linux/amd64, linux/arm64
[show-sbom] 
[show-sbom] Error: no SBOM found attached to image index
[show-sbom] main.go:74: error during command execution: no SBOM found attached to image index
```

This should let users specify which architecture they want to see the SBOM for, by parameterizing their pipeline.

### DIFF
--- a/task/show-sbom/0.1/show-sbom.yaml
+++ b/task/show-sbom/0.1/show-sbom.yaml
@@ -15,6 +15,9 @@ spec:
     - name: IMAGE_URL
       description: Fully qualified image name to show SBOM for.
       type: string
+    - name: PLATFORM
+      description: Specific architecture to display the SBOM for.
+      type: string
   steps:
   - name: show-sbom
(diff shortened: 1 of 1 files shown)


arewm (changes requested):
I don't think that we need to accept this as a parameter. Instead, if we just detect like @scoheb did in https://github.com/redhat-appstudio/build-definitions/pull/871. If we only generate a single SBOM then we will always be defaulting to building on `linux/amd64` if present in an Image Index.

ralphbean:
Nice! Great minds! @scoheb we started working on this at the same time, it seems, spontaneously.

ralphbean:
Let's move discussion to https://github.com/redhat-appstudio/build-definitions/pull/871 for now.

We can re-open this if we determine that it is too cumbersome or awkward to include skopeo in the task image for whatever reason.

PR #876, opened 2024-03-12
docker-build-rhtap: create a new show-sbom task compatible with RHDH

brunoapimentel:
Previous title: emit the LINK_TO_SBOM result

~This result contains the URL to the OCI artifact that stores the SBOM. It is meant primarily to be consumed by [RHDH](https://github.com/janus-idp/backstage-plugins/issues/988) and [ODC](https://github.com/openshift/console/pull/13314). The URl follows the format:~

~registry.io/user/image:sha256-hash.sbom~

~Note that this URL is not meant to be used with cosign, as it automatically resolves it when performing a "cosign download sbom registry.io/user/image:actual-tag".~

This PR was changed based on the comments on this [thread](https://github.com/redhat-appstudio/build-definitions/pull/876#discussion_r1522576564). We need to annotate the `show-sbom` task instead of `buildah-rhtap`. But since Pipelines as Code will clear the annotations, they need to be re-added using a dedicated step in the task.

To avoid breaking the Konflux pipeline, we're temporarily splitting the `show-sbom` into a new `show-sbom-rhdh` so that the latter can have all the necessary changes for the RHDH integration.

### DIFF
--- a/task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
+++ b/task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
@@ -1,0 +1,63 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: show-sbom-rhdh
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: "docker"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "containers, rhtap"
+    task.results.format: application/text
+    task.results.key: LINK_TO_SBOM
+    task.output.location: results
+spec:
+  description: >-
+    Shows the Software Bill of Materials (SBOM) generated for the built image in CyloneDX JSON format.
+    The 'task.*' annotations are processed by Red Hat Developer Hub (RHDH) so that the log content can be rendered in its UI.
+  params:
+    - name: IMAGE_URL
+      description: Fully qualified image name to show SBOM for.
+      type: string
+  results:
+  - description: Placeholder result meant to make RHDH identify this task as the producer of the SBOM logs.
+    name: LINK_TO_SBOM
+  steps:
+  - name: annotate-task
+    image: registry.redhat.io/openshift4/ose-cli:4.13@sha256:73df37794ffff7de1101016c23dc623e4990810390ebdabcbbfa065214352c7c
+    script: |
+      #!/usr/bin/env bash
+
+      # When this task is used in a pipelineRun triggered by Pipelines as Code, the annotations will be cleared,
+      # so we're re-adding them here
+      oc annotate taskrun $(context.taskRun.name) task.results.format=application/text
+      oc annotate taskrun $(context.taskRun.name) task.results.key=LINK_TO_SBOM
+      oc annotate taskrun $(context.taskRun.name) task.output.location=results
+  - name: show-sbom
+    image: registry.redhat.io/rhtas-tech-preview/cosign-rhel9@sha256:151f4a1e721b644bafe47bf5bfb8844ff27b95ca098cc37f3f6cbedcda79a897
+    env:
+    - name: IMAGE_URL
+      value: $(params.IMAGE_URL)
+    script: |
+      #!/busybox/sh
+      status=-1
+      max_try=5
+      wait_sec=2
+      for run in $(seq 1 $max_try); do
+        status=0
+        cosign download sbom $IMAGE_URL 2>>err
+        status=$?
+        if [ "$status" -eq 0 ]; then
+          break
+        fi
+        sleep $wait_sec
+      done
+      if [ "$status" -ne 0 ]; then
+          echo "Failed to get SBOM after ${max_try} tries" >&2
+          cat err >&2
+      fi
+
+      # This result will be ignored by RHDH, but having it set is actually necessary for the task to be properly
+      # identified. For now, we're adding the image URL to the result so it won't be empty.
+      echo -n "$IMAGE_URL" > $(results.LINK_TO_SBOM.path)
--- a/pipelines/docker-build-rhtap/patch.yaml
+++ b/pipelines/docker-build-rhtap/patch.yaml
@@ -135,3 +135,8 @@
     taskRef:
       kind: Task
       name: update-deployment
+- op: replace
+  path: /spec/finally/0/taskRef
+  value:
+    name: show-sbom-rhdh
+    version: "0.1"
--- a/task/show-sbom-rhdh/0.1/README.md
+++ b/task/show-sbom-rhdh/0.1/README.md
@@ -1,0 +1,10 @@
+# show-sbom-rhdh task
+
+Shows the Software Bill of Materials (SBOM) generated for the built image in CyloneDX JSON format.
+The 'task.*' annotations are processed by Red Hat Developer Hub (RHDH) so that the log content can be rendered in its UI.
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|IMAGE_URL|Fully qualified image name to show SBOM for.||true|
+
--- a/task/show-sbom-rhdh/OWNERS
+++ b/task/show-sbom-rhdh/OWNERS
@@ -1,0 +1,1 @@
+Stonesoup Build Team


task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -7,6 +7,10 @@ metadata:
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "containers, rhtap"
+    task.results.format: application/text
+    task.results.type: external-link
```
karthikjeeyar, line 11: Since we don't have an external sbom viewer mechanism, we can skip this annotation so that UI will open the sbom task logs modal to show the output instead of redirecting to external page. Once we have an external sbom viewer then we can add this annotation back.

example sbom logs modal:

image



```suggestion
 
```

chmeliik: So what does the `LINK_TO_SBOM` do? Does RHDH just automatically link to the logs of the task that emits a `LINK_TO_SBOM` result (when the `results.type` is not `external-link`)?

Does that mean the LINK_TO_SBOM result needs to be added to the task that prints the SBOM and the value of the result doesn't matter?

brunoapimentel: But the build task ([buildah-rhtap](https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah-rhtap/0.1/buildah-rhtap.yaml) in this case) does not print the SBOM in its logs, so if I understood it correctly, RHDH would open am modal and show the logs, but the SBOM wouldn't be printed there, just the usual build logs.

karthikjeeyar: Right, [show-sbom](https://github.com/redhat-appstudio/build-definitions/blob/main/task/show-sbom/0.1/show-sbom.yaml) task needs to be annotated.

brunoapimentel: @karthikjeeyar I completely reworked the PR. See if the new `show-sbom-rhdh` task has everything we need.

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
```diff
@@ -0,0 +1,60 @@
...
+      #!/usr/bin/env bash
+      echo "show-sbom-rhdh $(context.taskRun.name)"
+
+      # When this task is used in a pipelineRun triggered by Pipelines as Code, the annotations will be cleared,
+      # so we're re-adding them here
+      oc annotate taskrun $(context.taskRun.name) task.results.format=application/text
+      oc annotate taskrun $(context.taskRun.name) task.results.key=LINK_TO_SBOM
+      oc annotate taskrun $(context.taskRun.name) task.output.location=logs
+  - name: show-sbom
+    image: quay.io/redhat-appstudio/cosign:v2.1.1@sha256:c883d6f8d39148f2cea71bff4622d196d89df3e510f36c140c097b932f0dd5d5
```
chmeliik, line 39: We should use the productized cosign (https://catalog.redhat.com/software/containers/rhtas-tech-preview/cosign-rhel9/652971cd584d77e9d922883b?architecture=amd64&image=65c5506e19b6942d637e5603)

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
```diff
@@ -0,0 +1,60 @@
...
+        status=$?
+        if [ "$status" -eq 0 ]; then
+          break
+        fi
+        sleep $wait_sec
+      done
+      if [ "$status" -ne 0 ]; then
+          echo "Failed to get SBOM after ${max_try} tries" >&2
+          cat err >&2
+      fi
```
chmeliik, line 60: We don't write the LINK_TO_SBOM result, will this work? I think you need `touch $(results.LINK_TO_SBOM.path)`, otherwise the task won't have any results and I'm not sure what RHDH will think.

chmeliik (commented):
> the step that re-adds the annotations is forbidden in Konflux by the Enterprise Contract.

For my info, how is that forbidden? What's the error message and where did you see it?

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
brunoapimentel: Yeah, but this image needs credentials to be pulled. I wasn' t able to test this in a full RHTAP install, so would this still work on it?

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
brunoapimentel: I was really not sure about this, I kept the result in the Task definition because apparently it's what RHDH requires. But I see your point, maybe `touch` is the way to go.

brunoapimentel:
> > the step that re-adds the annotations is forbidden in Konflux by the Enterprise Contract.
> 
> For my info, how is that forbidden? What's the error message and where did you see it?

@mmorhun was the one that told me the EC doesn't allow the tasksRuns to modify the pipeline in any way. Do you have more info on it, Mykola?

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
chmeliik: It would, the [git-clone task works as well](https://github.com/redhat-appstudio/build-definitions/blob/3e548a38b3ad183262a25bc2a4eb6b5367b83fb5/task/git-clone/0.1/git-clone.yaml#L146). Tekton somehow magically has permissions to pull the images used to execute tasks from registry.redhat.io (maybe that's OpenShift magic)

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
chmeliik: We definitely need the result in the Task spec, but if you don't write to it in the script the result will not appear in the TaskRun status. Not sure which one RHDH cares about

karthikjeeyar: In this case, since we do not have a external-link annotation set,  RHTAP UI will just ignore the value of LINK_TO_SBOM result and make use of `task.results.key: LINK_TO_SBOM` annotation to identify and choose this sbom task logs.

Having said that, I agree it is a good idea to have result value set here (maybe we can set to ${params.IMAGE_URL} atm? ) and  later when we have a valid URL we must replace the value of LINK_TO_SBOM result.

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
```diff
@@ -0,0 +1,60 @@
...
+    image: registry.redhat.io/openshift4/ose-cli:4.13@sha256:73df37794ffff7de1101016c23dc623e4990810390ebdabcbbfa065214352c7c
+    script: |
+      #!/usr/bin/env bash
+      echo "show-sbom-rhdh $(context.taskRun.name)"
+
+      # When this task is used in a pipelineRun triggered by Pipelines as Code, the annotations will be cleared,
+      # so we're re-adding them here
+      oc annotate taskrun $(context.taskRun.name) task.results.format=application/text
+      oc annotate taskrun $(context.taskRun.name) task.results.key=LINK_TO_SBOM
+      oc annotate taskrun $(context.taskRun.name) task.output.location=logs
```
karthikjeeyar, line 37:
```suggestion
      oc annotate taskrun $(context.taskRun.name) task.output.location=results
```

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
brunoapimentel: We could also keep using the actual link to the SBOM OCI artifact (the link that this PR was previously handling). Let me know if you think it'd be better this way.

karthikjeeyar: Sure, you can add that but again the value will be ignored thats why I suggested IMAGE_URL since doesn't require much changes.

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
brunoapimentel: Fixed this.

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
brunoapimentel: I agree. I'm adding IMAGE_URL for now.

karthikjeeyar (approved):
Thank you!   LGTM

chmeliik (approved):
LGTM with a tiny nitpick

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
```diff
@@ -0,0 +1,63 @@
...
+        sleep $wait_sec
+      done
+      if [ "$status" -ne 0 ]; then
+          echo "Failed to get SBOM after ${max_try} tries" >&2
+          cat err >&2
+      fi
+
+      # This result will be ignored by RHDH, but having it set is actually necessary for the task to be properly
+      # identified. For now, we're adding the image URL to the result so it won't be empty.
+      echo "$IMAGE_URL" > $(results.LINK_TO_SBOM.path)
```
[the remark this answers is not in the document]
brunoapimentel: The :smiling_imp: is in the details. Fixed that.

mmorhun (approved):

mmorhun:
e2e tests fail on known nissue.

PR #878, opened 2024-03-14
tasks/acs*: don't fail when ACS is not configured

chmeliik:
* The report step in all tasks now cats the file only if it exists
* The image-scan task writes an empty SCAN_OUTPUT when skipped
  * Otherwise the pipeline fails because the ACS_SCAN_OUTPUT result references a non-existent task result

### DIFF
--- a/task/acs-deploy-check/0.1/acs-deploy-check.yaml
+++ b/task/acs-deploy-check/0.1/acs-deploy-check.yaml
@@ -133,4 +133,6 @@ spec:
           mountPath: /workspace/repository
       script: |
         #!/usr/bin/env bash
-        cat /workspace/repository/acs-deploy-check.json
+        if [ -f /workspace/repository/acs-deploy-check.json ]; then
+          cat /workspace/repository/acs-deploy-check.json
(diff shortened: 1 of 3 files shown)


jduimovich:
I assume the empty ACS_SCAN_OUTPUT  causes the 0 values in the vulnerabilties  ?

mmorhun (approved):

chmeliik:
> I assume the empty ACS_SCAN_OUTPUT causes the 0 values in the vulnerabilties ?

Probably yes. I don't think see an easy way to avoid that

chmeliik:
From a discussion with John: the 0s in the vulnerabilities report are a problem, we'd like a better solution

> maybe the solution is to have a pre-check task the extracts the secrets, validates access and if successful, the ACS tasks can run

> The other solution which could work is not annotate if not successful. Of course when they fix pac, the annotations won't be lost

> The other behaviour could to use the onError: continue so you can error but the pipeline succeeds overall and when you read the error message the "You have not configured your rox secret etc" can be printed

PR #881, opened 2024-03-15
Do not run update-deployment task on pull request

mmorhun:
Prevents running update deployment task if pipeline was trigerred by pull_request event.
Also, fixes some formatting.
https://github.com/redhat-appstudio/tssc-sample-pipelines/pull/24 adds `event-type` parameter to the RHTAP pipeline run.


chmeliik (commented):
What is going to set the param to "pull_request"? Could you explain that in the description?

PR #890, opened 2024-03-19
gitops-pull-request: check only changed images

chmeliik:
The pipeline will only check images that changed in the pull request

### DIFF
--- a/pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
+++ b/pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
@@ -11,6 +11,10 @@ spec:
       name: revision
       type: string
       default: ""
+    - description: The target branch for the pull request
+      name: target-branch
+      type: string
+      default: "main"
     - description: Enterprise Contract policy to validate against
       name: ec-policy-configuration
       type: string
@@ -38,6 +42,10 @@ spec:
           value: $(params.git-url)
         - name: revision
           value: $(params.revision)
+        # We're more interested in remote branches than tags, but this fetches both.
+        # Branch refs are needed for the TARGET_BRANCH feature of gather-deploy-images.
+        - name: fetchTags
+          value: "true"
       taskRef:
         name: git-clone
         version: "0.1"
@@ -47,6 +55,9 @@ spec:
         - name: basic-auth
           workspace: git-auth
     - name: gather-deploy-images
+      params:
+        - name: TARGET_BRANCH
+          value: $(params.target-branch)
       runAfter:
         - clone-repository
       taskRef:
--- a/pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
+++ b/pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
@@ -82,6 +82,10 @@ spec:
           value: $(params.ec-tuf-mirror)
       runAfter:
         - gather-deploy-images
+      when:
+        - input: $(tasks.gather-deploy-images.results.IMAGES_TO_VERIFY)
+          operator: notin
+          values: [""]
       taskRef:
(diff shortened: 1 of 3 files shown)


chmeliik:
Tested in https://github.com/acmiel-rhtap/rhtap-foo-gitops/pull/2

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
```diff
@@ -38,6 +42,10 @@ spec:
           value: $(params.git-url)
         - name: revision
           value: $(params.revision)
+        # We're more interested in remote branches than tags, but this fetches both.
+        # Branch refs are needed for the TARGET_BRANCH feature of gather-deploy-images.
+        - name: fetchTags
```
mmorhun, line 47: Is it used anywhere?

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
```diff
@@ -71,6 +82,10 @@ spec:
           value: $(params.ec-tuf-mirror)
       runAfter:
         - gather-deploy-images
+      when:
+        - input: $(tasks.gather-deploy-images.results.NOTHING_TO_VERIFY)
+          operator: notin
+          values: ["true"]
```
mmorhun, line 88: Is it possible to have something like
```
 - input: $(tasks.gather-deploy-images.results.IMAGES_TO_VERIFY)
   operator: notin
   values: [""]
```
instead of a new result?

chmeliik: This would probably work (after fixing some trailing newlines)
```yaml
 - input: $(tasks.gather-deploy-images.results.IMAGES_TO_VERIFY)
   operator: notin
   values: ['{"components":[]}']
```
That feels a bit fragile, like we're exposing too much of the implementation details. If the output wasn't in `jq -c` format, the JSON could be formatted a bit differently

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
mmorhun: If it fails without fetching them, we do we need a parameter?

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
mmorhun: The above is definitely fragile, but we can check `/tmp/all-images.txt` before invoking `jq` and exit with empty result earlier.

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
chmeliik: In this section we're not adding a parameter, we're just passing fetchTags: true to the clone task (the default is false)

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
chmeliik: good point, that should work

chmeliik:
re-testing, don't merge yet

chmeliik:
Yup, still works :tada: https://github.com/acmiel-rhtap/rhtap-foo-gitops/pull/1

First the EC check was skipped with no image changes, then it ran as expected (and failed, but for unrelated reasons)

pipelines/gitops-pull-request-rhtap/gitops-pull-request.yaml
chmeliik: Adjusted as suggested :+1:

mmorhun (approved):

jduimovich (approved):

PR #892, opened 2024-03-20
Change the shebang in the show-sbom-rhdh task

brunoapimentel:
The current shebang (/busybox/sh) does not exist in the RHTAS image, and will cause the taskRun to fail.


mmorhun (approved):

task/show-sbom-rhdh/0.1/show-sbom-rhdh.yaml
```diff
@@ -40,7 +40,7 @@ spec:
     - name: IMAGE_URL
       value: $(params.IMAGE_URL)
     script: |
-      #!/busybox/sh
+      #!/bin/bash
```
eskultety, line 43: Just my 2c: `/bin/sh`  might be a better choice in general, but I'd double check with `shellcheck`.

PR #898, opened 2024-03-21
Annotate EC task to align with RHDH requirement

karthikjeeyar:
Fixes:
https://issues.redhat.com/browse/RHTAPBUGS-1179

Description:

Add required annotations to verify-enterprise-contract task so RHDH shows the EC report in the output tab.

### DIFF
--- a/task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -7,6 +7,10 @@ metadata:
     tekton.dev/displayName: Verify Enterprise Contract
     tekton.dev/pipelines.minVersion: "0.19"
     tekton.dev/tags: ec, chains, signature, conftest
+    task.results.format: application/json
+    task.results.type: ec
+    task.output.location: logs
+    task.results.container: step-report-json
   labels:
     app.kubernetes.io/version: "0.1"
 spec:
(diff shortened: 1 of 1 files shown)


karthikjeeyar:
cc: @jduimovich

task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -103,6 +107,15 @@ spec:
         value: "$(params.HOMEDIR)"
 
   steps:
+    - name: annotate-task
```
zregvart, line 110: Why is this required? The annotations from the definition are present on the TaskRun. We do not want to allow TaskRuns to modify their own metadata this could lead to all sort of security issues.

zregvart (changes requested):
The task in here is a copy of https://github.com/enterprise-contract/ec-cli/blob/main/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml and any changes should be made there first.

task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
karthikjeeyar: PAC is removing these custom annotations which is required for RHDH UI to identify the ec task to show the reports, so this seems like a workaround to overcome this issue. Similar fix is done in ACS and SBOM tasks ref- https://github.com/redhat-appstudio/build-definitions/pull/859 . Tagging @jduimovich who can add more details to this.

karthikjeeyar:
@zregvart After talking to John, I realised we dont want to put the RHTAP specific annotations on here and in enterprise-contract repo and moreover this is a temporary workaround until PAC fixes this issue. so closing this PR in favor of https://github.com/redhat-appstudio/tssc-sample-pipelines/pull/32

PR #900, opened 2024-03-21
Remove usage of task-toolset image

mmorhun:
This PR removes last usage of `quay.io/redhat-appstudio/task-toolset` image by replacing it with `uay.io/redhat-appstudio/appstudio-utils` image that has needed tools installed and already used in other tasks, so we reduce number of used images by 1.


chmeliik (approved):

mkosiarc (approved):

tkdchen (approved):
BTW, usege, a typo probably?

PR #912, opened 2024-03-29
Revert "Removing unneeded tasks from FBC pipeline (#865)"

ralphbean:
This reverts commit 117850230cc180562e2fd6116b2c7858c8f20bab.

The original change is essentially correct, but for unknown reasons I think that https://github.com/redhat-appstudio/e2e-tests/pull/1065 is not effective in the way we thought it should be. My theory is that 117850230cc180562e2fd6116b2c7858c8f20bab is currently causing e2e tests to fail in the build-definitions repo for all other PRs.

### DIFF
--- a/pipelines/fbc-builder/patch.yaml
+++ b/pipelines/fbc-builder/patch.yaml
@@ -28,35 +28,6 @@
     value: "$(params.image-expires-after)"
   - name: COMMIT_SHA
     value: "$(tasks.clone-repository.results.commit)"
-# Remove tasks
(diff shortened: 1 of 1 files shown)


tkdchen (approved):

mmorhun:
We are trying to fix tests, so I'd like to postpone merging.

mmorhun:
The fix is confirmed to work and merged into tests repo. Waiting on https://github.com/redhat-appstudio/build-definitions/pull/919 to apply new tests image

mmorhun:
Closing since the fix is applied and tested in https://github.com/redhat-appstudio/build-definitions/pull/918

PR #913, opened 2024-03-29
Trusted Artifacts with OCI storage

lcarva:
This PR extends https://github.com/redhat-appstudio/build-definitions/pull/791 to use an OCI registry as the storage for Trusted Artifacts.

Pros:

- PVCs are no longer required in the Pipeline! Small data is shared between Tasks via the usual Tekton results while large data is shared via an OCI registry.
- Trusted artifacts outlive the life-cycle of the PipelineRun. They can be inspected and used later on.
- As with previous DIY Trusted Artifacts implementations, artifacts are addressed by digest creating the assurance that data is not modified between Tasks.

Cons:

- Involved migration for users. Although there are no new Pipeline parameters, existing Pipeline parameters must be passed to certain Tasks. Manual migration steps are required for users that have the expanded version of the Pipeline in their repos which is the majority, if not all, of users.
- There is still overhead in creating and consuming artifacts. Fundamentally, there's additional work required to perform these operations. The hope is that any overhead added by the Trusted Artifacts approach outweighs the limitations of requiring PVCs.
- Users will notice the additional `<image-ref>-clone` and `<image-ref>-prefetch` tags in their OCI repositories. Some users may be adverse to such "clutter".
- This effectively sets the same access level for the source code to the OCI repository. This may be an issue for cases where the source code is not open. A simple mitigation is to set the OCI repository to private in such cases.

Below is a list of potential future work that could leverage Trusted Artifacts to improve existing processes. (Mentally say "additional work required" at the end of each bullet point :smiley_cat:)

- Cache could be implemented in terms of OCI Trusted Artifacts. For example, the `git-clone` Task could be modified to instead of cloning a git repository, simply use an existing Trusted Artifact. The changes in this PR by no means implement this, but I believe it paves the way.
- It should be trivial to extend Tasks like `clair-scan` to upload the complete CVE report to the OCI registry. This would increase the level of granularity for EC policies, for example. It may also provide an easier way for users to view the report without relying on logs. (This could also help while we transition away from `cosign attach sbom`.)
- Creating source container builds could be streamlined to leverage the Trusted Artifacts already in the OCI registry. In such cases, there's no need for the `source-build` Task to download the Trusted Artifacts. It can simply use them by reference.
- The `buildah-remote` Task could be modified so the Trusted Artifacts are fetched directly from the remote host, instead of downloading it from the Task and pushing to the remote host via ssh.
- Trusted Artifacts could be "deeply" inspected to ensure modifications are scoped to the expectations. For example, the `prefetch-dependencies` Task may alter the contents of the clone repository, mainly regarding lockfiles. With this approach, it's possible to verify what changes exist between the artifact produced by the `git-clone` Task and the `prefetch-dependencies` Task. Likely too heavy of a check for EC, but nonetheless it's at least possible. This may be useful for debugging and ad-hoc auditing purposes.


ralphbean:
What are the performance characteristics of this change?

> Users will notice the additional <image-ref>-clone and <image-ref>-prefetch tags in their OCI repositories. Some users may be adverse to such "clutter".

How can you know the "<image-ref>" at the time the clone task runs? The image-ref is not known until after build-container, right?

> This effectively sets the same access level for the source code to the OCI repository. This may be an issue for cases where the source code is not open. A simple mitigation is to set the OCI repository to private in such cases.

This is okay. image-controller already [changes the visibility](https://github.com/redhat-appstudio/image-controller/blob/main/controllers/imagerepository_controller.go#L560) based on the visibility of the source repo.

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -0,0 +1,519 @@
+apiVersion: tekton.dev/v1
```
arewm, line 1: In order for these to pass CI, we will need to update the verification process

https://github.com/redhat-appstudio/multi-platform-controller/blob/e9bab24a0a21c6e0a10b285dd7df385d178c1d61/cmd/taskgen/main.go
https://github.com/redhat-appstudio/build-definitions/blob/main/hack/generate-buildah-remote.sh

arewm (commented):
I only looked at the one commit that @lcarva added in this brief review.

There are a lot of breaking changes here. @rcerven wanted to propose many additional breaking changes in a previous PR which we ended up not including due to the impact. In order to make _all_ of these changes once, we shouldn't merge this directly to `main` but instead use a temporary branch to stage all of the breaking changes so that we can create a single version bump for all changes.

arewm:
> Users will notice the additional <image-ref>-clone and <image-ref>-prefetch tags in their OCI repositories. Some users may be adverse to such "clutter".

We already have much clutter present. This will certainly increase the clutter ... but calling this out is valid as we will be pushing a lot more that users will have to sift through. Is there some way that we can change the image tags for PRs to more clearly separate PR clutter from push clutter?

@zregvart, what are the race conditions that you were concerned about?

> Involved migration for users. Although there are no new Pipeline parameters, existing Pipeline parameters must be passed to certain Tasks. Manual migration steps are required for users that have the expanded version of the Pipeline in their repos which is the majority, if not all, of users.

While the migration may be "painful", it would likely be simpler for users to convert to the default build and then back to the custom one to get a refreshed pipeline definition. The benefit of removing a PVC limit here _might_ be a big enough improvement for people to be more willing to perform the action than they otherwise might be. That being said, if there are any additional breaking changes that we might introduce with the follow-on work (especially nearer term follow on work), we should stage these breaking changes and introduce them once. I am thinking about you, caching... 

> Creating source container builds could be streamlined to leverage the Trusted Artifacts already in the OCI registry. In such cases, there's no need for the source-build Task to download the Trusted Artifacts. It can simply use them by reference.

@lcarva, What do you mean by this? Are you trying to suggest that the source container might just reference the cache objects instead of trying to recreate the sources? Would we be able to improve the source generation by performing _within_ the prefetch step so that we can push dependencies to the registry at the same time, benefitting from layer caching? This might get more complex once we support prefetching binary content as the source container content will be different.

> The buildah-remote Task could be modified so the Trusted Artifacts are fetched directly from the remote host, instead of downloading it from the Task and pushing to the remote host via ssh.

I think that this should be a high priority before these changes are "GA" or at least widely adopted (I reviewed similar on the PR before reading the summary text).

> This may be useful for debugging and ad-hoc auditing purposes.

The presence of these sources outside of the PLR might be useful for auditing/troubleshooting regardless of whether we have the deep inspection capabilities.

> One issue with TA based on PVCs is that if prefetch-dependencies is a no-op, the archives with source and fetched dependencies are created regardless, this implementation helps as it de-duplicates automatically (assuming TA images are hermetic)

With the OCI artifact approach, will we require that the prefetch step always runs, even if a no-op such that the OCI artifact is always populated? Or does the implementation of the OCI-based artifact allow a no-op if the pullspec is invalid?

> What are the performance characteristics of this change?

My suspicion is that the performance characteristics of this change will likely not be that good. A larger benefit will come from follow-on work (i.e. caching strategies).

zregvart:
> @zregvart, what are the race conditions that you were concerned about?

[This](https://github.com/redhat-appstudio/build-definitions/pull/913/commits/f4b51549e18fd6594354453e3d1ea8edf6ef48b8#diff-073dcc2fd850949d01892f5daa50e15d7ba496f9603e783cdd356ae6027e4622R80-R81) (I think) can be same for more than one CI run

lcarva:
> What are the performance characteristics of this change?

(Copying response from slack here for posterity and transparency.) At the moment it is unknown. KONFLUX-2547 will measure performance.

> >It should be trivial to extend Tasks like clair-scan to upload the complete CVE report to the OCI registry. This would increase the level of granularity for EC policies, for example. It may also provide an easier way for users to view the report without relying on logs.
> This could be better suited for attestations, not necessarily a TA concern

I agree. I mentioned it here because this is a valid stop-gap for environments where this is currently not possible, e.g. Konflux.

> Is there a chance for race conditions?
> [This](https://github.com/redhat-appstudio/build-definitions/pull/913/commits/f4b51549e18fd6594354453e3d1ea8edf6ef48b8#diff-073dcc2fd850949d01892f5daa50e15d7ba496f9603e783cdd356ae6027e4622R80-R81) (I think) can be same for more than one CI run

Indeed! It's important to note that the ref for a trusted artifact is always to a blob/layer, not an Image Manifest. A race condition may cause a certain layer to become "dangling" (not part of an Image Manifest) and thus garbage collected, but it should never cause it to change. It's definitely something we will need to have a closer look if/when this POC gets the go-ahead to be properly implemented.

> While the migration may be "painful", it would likely be simpler for users to convert to the default build and then back to the custom one to get a refreshed pipeline definition. The benefit of removing a PVC limit here might be a big enough improvement for people to be more willing to perform the action than they otherwise might be. That being said, if there are any additional breaking changes that we might introduce with the follow-on work (especially nearer term follow on work), we should stage these breaking changes and introduce them once. I am thinking about you, caching...

Good point. KONFLUX-2500 also looks promising going forward. We floated the idea of having some sort of `devel` or `next` version that we can make multiple commits before promoting it to a proper version. This would avoid renovate bot upgrading users prematurely.

> > Creating source container builds could be streamlined to leverage the Trusted Artifacts already in the OCI registry. In such cases, there's no need for the source-build Task to download the Trusted Artifacts. It can simply use them by reference.
> @lcarva, What do you mean by this? Are you trying to suggest that the source container might just reference the cache objects instead of trying to recreate the sources? Would we be able to improve the source generation by performing within the prefetch step so that we can push dependencies to the registry at the same time, benefitting from layer caching? This might get more complex once we support prefetching binary content as the source container content will be different.

Yes. Further work is required here, of course. Recall that this is under the _potential_ future work.

> > The buildah-remote Task could be modified so the Trusted Artifacts are fetched directly from the remote host, instead of downloading it from the Task and pushing to the remote host via ssh.
> I think that this should be a high priority before these changes are "GA" or at least widely adopted (I reviewed similar on the PR before reading the summary text).

We don't really the concept of GA or anything remote to it in this repo. (We probably should!) But regardless, if pursued this change should be isolated to the implementation of the buildah-remote task. I don't think it's reasonable to halt adoption until this is sorted out. IMO this is a perfect candidate for an incremental improvement.

> With the OCI artifact approach, will we require that the prefetch step always runs, even if a no-op such that the OCI artifact is always populated? Or does the implementation of the OCI-based artifact allow a no-op if the pullspec is invalid?

We should improve the prefetch-dependencies Task to better take advantage of Trusted Artifacts. This is something we should probably address before making the switch, but it could also, arguably, be an incremental improvement after the fact.

jhutar:
Hello @lcarva ! I can not find tag `docker-build-914a773e173427ec045ee607be32ccb5138a932b` in https://quay.io/repository/redhat-appstudio-tekton-catalog/pull-request-builds?tab=tags. Am I looking for the correct tag? If so, I assume it expired. Could you please regenerate it?

lcarva:
@jhutar, apologies for the late reply. This should be available now: 
```
quay.io/redhat-appstudio-tekton-catalog/pull-request-builds:docker-build-101a1defaa6bcf2bb77ff6683b4ad4b034d4b743
```

I've tweaked the CI for this PR, so if that gets garbage collected again, we can simply comment `/retest` here to regenerate. (New ref will be in logs)

arewm:
If you want to use that image without retriggering the PR, you can always do something like this (untested)

```
$ cat Containerfile
FROM quay.io/redhat-appstudio-tekton-catalog/pull-request-builds:docker-build-101a1defaa6bcf2bb77ff6683b4ad4b034d4b743
$ buildah build --unsetlabel "quay.expires-after" Containerfile```

Or change the label to a larger value.

**Comment by jhutar** [comment]
Thank you! `docker-build-101a1defaa6bcf2bb77ff6683b4ad4b034d4b743` is there.

**Comment by vdemeester** [comment]
> Also, folk need to migrate away from workspaces

@zregvart that's not necessary. workspaces can be bound with emptydir. The main "benefit" of usage task's workspace is that it provides a shared "folder" (a volume) accross all containers in the Task's pod — be it bound by pvc or emptydir.

File: task/git-clone/0.2/git-clone.yaml [hunk 2]
```diff
@@ -0,0 +1,282 @@
...
+  results:
+  - description: The precise commit SHA that was fetched by this Task.
+    name: commit
+  - description: The precise URL that was fetched by this Task.
+    name: url
+  - description: Produced trusted source artifact
+    name: SOURCE_ARTIFACT
+    type: string
+  stepTemplate:
+    volumeMounts:
```
vdemeester, line 98: Could still use workspace for this though (and bind it with emptydir…)

zregvart:
> > Also, folk need to migrate away from workspaces
> 
> @zregvart that's not necessary. workspaces can be bound with emptydir. The main "benefit" of usage task's workspace is that it provides a shared "folder" (a volume) accross all containers in the Task's pod — be it bound by pvc or emptydir.

@vdemeester what I meant is that there is an additional migration step for the users: the expanded PipelineRun definitons today reference a workspace, these Tasks do not, so simply upgrading the version of the Task is not the only migration step needed

task/git-clone/0.2/git-clone.yaml
zregvart: What would be the benefit of using a workspace instead of a volume here?

vdemeester: A bit less verbose (no need for volume+volumemounts) and a possiblity, for the task, to be used with a PVC or something else (if used outside of konflux, …) for example.

lcarva:
@jhutar, yeah... Let's use this one instead:

```yaml
  pipelineRef:
    resolver: bundles
    params:
      - name: bundle
        value: quay.io/lucarval/pipeline-docker-build:2024-04-18-141354@sha256:203cb2d5c91e0e8af0ed3c5815c3ab33e1d17a912190a6ebf99327ad78c92ca2
      - name: name
        value: docker-build
      - name: kind
        value: pipeline
```

That should stay around.

jhutar:
Thank you! Mine rerun for yesterday evening failed with same error which is strange as I can see that image tag in Quay. Anyway, updated config to use your image and re-running again.

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -141,15 +141,11 @@ spec:
         name: workdir
   steps:
   - name: use-trusted-artifact
-    image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:2741aaaf0c06ab784dcab99545be615696d05a578bd4ae5a1b2d6e17e5c569c4
+    image: quay.io/lucarval/build-trusted-artifacts:latest@sha256:8f65579cdc92e0d68f261997d8c61040fe3313d0493da4c0fe64ba4252d582ae
```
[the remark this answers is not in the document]
stuartwdouglas: buildah-remote should be generated from buildah, I will need to make some changes to the generation code to handle both use cases.

task/buildah-remote/0.2/buildah-remote.yaml
stuartwdouglas: This needs some significant updates, I will have a look. 

Note that I don't think this will affect CI yet, as the verification won't be checking this new task yet.

PR #922, opened 2024-04-04
Save base_images_from_dockerfile to /workspaces/source

mkosiarc:
The reason is that this was failing in buildah remote task because the line

echo "$BASE_IMAGES" > /workspace/base_images_from_dockerfile

was executed on a VM via SSH and it was not restored afterwards. However, /workspaces/source is actually restored via

rsync -ra "$SSH_HOST:$BUILD_DIR/workspaces/source/" "$(workspaces.source.path)/" 
https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah-remote/0.1/buildah-remote.yaml#L326

So storing it in /workspaces/source should make it available in the end.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -295,7 +295,7 @@ spec:
       done
 
       # Needed to generate base images SBOM
-      echo "$BASE_IMAGES" > /workspace/base_images_from_dockerfile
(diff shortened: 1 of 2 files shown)


mkosiarc:
failure happened in infra-deployments e2e tests in https://github.com/redhat-appstudio/infra-deployments/pull/3515

rcerven (approved):

psturc:
this PR is supposed to fix [this issue](https://issues.redhat.com/browse/KFLUXBUGS-1189) that is currently affecting infra-deployments CI

PR #923, opened 2024-04-04
Examine merge commits on bundle updates

joejstuart:
Update git diff command to examine merge commits, git diff-tree only examined non merge commits.

Also, remove revision from the task format. EC tracker does not support it.

This change is dependent on https://github.com/enterprise-contract/ec-cli/pull/1494

### DIFF
--- a/.tekton/scripts/build-acceptable-bundles.sh
+++ b/.tekton/scripts/build-acceptable-bundles.sh
@@ -8,20 +8,21 @@ BUNDLES=${BUNDLES:-()}
 # store a list of changed task files
 task_records=()
 # loop over all changed files
-for path in $(git diff-tree -c --name-only --no-commit-id -r ${REVISION}); do
+for path in $(git log -m -1 --name-only --pretty="format:" ${REVISION}); do
     # check that the file modified is the task file
     if [[ "${path}" == task/*/*/*.yaml ]]; then
-    IFS='/' read -r -a path_array <<< "${path}"
-    dir_name_after_task="${path_array[1]}"
-    file_name=$(basename "${path_array[-1]}" ".yaml")
+        IFS='/' read -r -a path_array <<< "${path}"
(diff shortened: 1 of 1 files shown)


.tekton/scripts/build-acceptable-bundles.sh
```diff
@@ -8,20 +8,21 @@ BUNDLES=${BUNDLES:-()}
...
-        # GIT_URL is the repo_url from PAC (https://hostname/org/repo)
-        task_records+=("git+${GIT_URL}.git//${path}@${REVISION}")
-    fi
+        IFS='/' read -r -a path_array <<< "${path}"
+        dir_name_after_task="${path_array[1]}"
+        file_name=$(basename "${path_array[-1]}" ".yaml")
+
+        if [[ "${dir_name_after_task}" == "${file_name}" ]]; then
+            # GIT_URL is the repo_url from PAC (https://hostname/org/repo)
+            task_records+=("git+${GIT_URL}.git//${path}")
```
zregvart, line 20: Needs revision here. See [this](https://github.com/enterprise-contract/ec-cli/blob/01b6c39ad710b0c6d8e3a70885557c36343e0db5/internal/tracker/tracker.go#L259), the `--freshen` helps but adds an additional git lookup which is not necessary here as we already know the revision we want to add.

joejstuart: I wonder if `--freshen` is needed for the bundle updates though. I'll have to test it out.

zregvart: IIRC we added `--freshen` to pickup any bundle digests missed when CI on main fails, e.g. quay flakiness results in the acceptable bundles not being pushed

zregvart (approved):

robnester-rh (approved):
LGTM

PR #928, opened 2024-04-09
rpm-ostree: use prefetched RPMs when available

chmeliik:
STONEBLD-2281

If the prefetch-dependencies task has prefetched RPMs, make the
rpm-ostree build use them. Do this by removing all the original repo
files from the source directory and replacing them with the generated
cachi2.repo.

---

Note that the rpm-ostree task determines whether there are prefetched
RPMs by checking the existence of the top-level deps/rpm directory, not
the platform-specific subdirectory. This means that if the prefetch task
has prefetched RPMs for at least one platform, the build tasks for *all*
platforms will attempt to use them (and fail, if the RPMs have not been
prefetched for their platform).

This is intentional, to reduce the chance of accidentally prefetching
RPMs only for one platform and having the other builds download RPMs
from the internet.

There is a caveat though. The user could structure their pipeline in
such a way that the prefetch task runs before the build for platform A,
but not before the build for platform B. In this case, platform B would
still download RPMs from the internet.

The issues would best be solved by a 'hermetic' flag to cut off network
connection - if such a thing can be achieved with rpm-ostree.

---

Also expose the `--dev-package-managers` flag for the prefetch task.


chmeliik:
Tested in https://github.com/chmeliik/centos-bootc-tmp/pull/7, https://github.com/chmeliik/centos-bootc-tmp/pull/8 and https://github.com/chmeliik/centos-bootc-tmp/pull/9

Prefetch + hermetic works, hermetic without prefetch fails as expected, non-hermetic without prefetch still works.

Some notes.
- There are caveats around referencing unused repos in the manifest file. See https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/dd6e1dda389349e9a7eacce8c812e1701c1079fb and https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/53204d50251aee536c2667beeddd26792bd467a5
- For some reason, when using the original `quay.io/centos-bootc/bootc-image-builder` image, I was getting GPG verification failures. I had to modify this image by moving the RPM-GPG-KEY-CentOS-Official file into /etc/pki/rpm-gpg https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/b1b8f7c7a3303f212d7b2425c34569ebaf425a3d
- For the CentOS bootc build, the total size of the prefetched RPMs and SRPMs is about 5.3 Gi. I bumped the size of the source workspace to 7Gi because it also has to fit the oci-archive of the built image. https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/78f2964de18c51807962deffc04c6985a7e83d04
- Generating the lockfile is an ordeal. I have a set of hacky scripts that make it almost bearable. https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/15bc3925432670e30d1097d2c62cc29c27111ad2
  - Note that the set of repo files must also include the -sources repos, otherwise these scripts don't find any SRPMs https://github.com/chmeliik/centos-bootc-tmp/pull/7/commits/9a6f27e1f7c8415abac0b7f1b98f7ca18634657b

chmeliik:
Also tested how this works with the source build task (see final pipeline in https://github.com/chmeliik/centos-bootc-tmp/blob/hermetically/.tekton/ostree-build.yaml). There is a bug in the source build task which causes the source container to be almost empty. Will fix.

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -126,6 +126,25 @@ spec:
...
       fi
+
+      prefetched_rpms=./cachi2/output/deps/rpm/$(uname -m)
+      if [ -d "$prefetched_rpms" ]; then
+        # move all repo files out of the source repository to avoid conflicts with the cachi2.repo
+        mkdir /tmp/original-repos
+        find ./source -maxdepth 1 -name '*.repo' -exec mv {} /tmp/original-repos \;
+
+        # copy the platform-specific cachi2.repo into the source repository
+        cp "$prefetched_rpms/repos.d/cachi2.repo" ./source
```
eskultety, line 143: Just an FYI: This one will become `$prefetched_rpms/<arch>/repos.d/cachi2.repo`

chmeliik: prefetched_rpms is the arch-specific directory - I will rename the variable

chmeliik:
> the total size of the prefetched RPMs and SRPMs is about 5.3 Gi

Also, if I'm not wrong, 2.4 Gi of that are duplicates

```bash
$ fdupes -r cachi2-output > dupes.txt
$ cat dupes.txt | xargs du -b | awk '{ s += $1 } END { print s }' | numfmt --to=iec-i 
4.8Gi
```

Of those, roughly 400Mi are noarch duplicates (these are needed - each noarch RPM has to be in every arch-specific subdirectory, otherwise only one arch will have access to them). The remaining 2 Gi are duplicate SRPMs (these are not needed).

```bash
$ grep 'noarch.rpm$' dupes.txt | xargs du -b | awk '{ s += $1 } END { print s }' | numfmt --to=iec-i 
792Mi

$ grep 'src.rpm$' dupes.txt | xargs du -b | awk '{ s += $1 } END { print s }' | numfmt --to=iec-i 
4.0Gi
```

TLDR, the lockfile generation needs a lot of work. My scripts will happily include the same SRPM 3 times per arch when it's found in more than one -sources repo

chmeliik:
> Also tested how this works with the source build task (see final pipeline in https://github.com/chmeliik/centos-bootc-tmp/blob/hermetically/.tekton/ostree-build.yaml). There is a bug in the source build task which causes the source container to not include any SRPMs. Will fix.

The fix: https://github.com/redhat-appstudio/build-tasks-dockerfiles/pull/63

chmeliik:
Just screenshotting the passing CI for future reference :sweat_smile: 

image

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -126,12 +132,54 @@ spec:
       else
         CONFIG_FILE_ARG="  --image-config=source/$CONFIG_FILE "
       fi
-      rpm-ostree compose image --initialize --format oci $CONFIG_FILE_ARG "source/$IMAGE_FILE" rhtap-final-image
 
+      prefetched_rpms_for_my_arch=./cachi2/output/deps/rpm/$(uname -m)
```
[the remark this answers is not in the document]
chmeliik: added :+1:

chmeliik:
/retest

e2e-tests failed with 
```
error when sending request to Github API: GET https://api.github.com/repos/redhat-appstudio-appdata/test-app-928-Dyf2k-demand-manage: 404 Not Found []
```
first time seeing that

chmeliik:
Another EC pipeline timeout
```
[build-service-suite Build templates E2E test] HACBS pipelines when the container image for component with Git source URL https://github.com/redhat-appstudio-qe/source-build-parent-image-with-digest-only is created and pushed to container registry [It] verify-enterprise-contract check should pass [build, build-templates, HACBS, pipeline, sbom, slow, build-templates-e2e]
/github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:391

  Timeline >>
  Creating Pipeline "verify-enterprise-contract-run-jvm6j"
  Waiting for pipeline "verify-enterprise-contract-run-jvm6j" to finish
  [FAILED] in [It] - /github.com/redhat-appstudio/e2e-tests/tests/build/build_templates.go:459 @ 04/10/24 13:47:04.455
  << Timeline

  [FAILED] Expected success, but got an error:
      <context.deadlineExceededError>: 
      context deadline exceeded
```

chmeliik:
`PipelineRun source-build-parent-image-with-digest-only-gjxw-9b6hq reason: CouldntGetTask`

interesting

/retest

chmeliik:
```
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```
While pushing bundles. Thanks Quay.

PR #935, opened 2024-04-11
new build-args-file parameter which allows to pass build arguments to

rcerven:
buildah task

[KONFLUX-268](https://issues.redhat.com//browse/KONFLUX-268)

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -78,6 +78,11 @@ spec:
     description: Name of secret which contains the entitlement certificates
     name: ENTITLEMENT_SECRET
     type: string
+  - default: ""
+    description: Path to a file with build arguments which will be passed to podman
+      during build
+    name: BUILD_ARGS_FILE
+    type: string
   - description: The platform to build on
     name: PLATFORM
     type: string
@@ -127,6 +132,8 @@ spec:
       value: $(params.BUILDER_IMAGE)
     - name: ENTITLEMENT_SECRET
       value: $(params.ENTITLEMENT_SECRET)
+    - name: BUILD_ARGS_FILE
+      value: $(params.BUILD_ARGS_FILE)
     - name: BUILDER_IMAGE
       value: quay.io/redhat-appstudio/buildah:v1.31.0@sha256:34f12c7b72ec2c28f1ded0c494b428df4791c909f1f174dd21b8ed6a57cf5ddb
   steps:
@@ -234,6 +241,10 @@ spec:
         BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
       fi
 
+      if [ -n "${BUILD_ARGS_FILE}" ]; then
+        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}")
+      fi
+
       if [ -n "${PREFETCH_INPUT}" ]; then
         cp -r cachi2 /tmp/
         chmod -R go+rwX /tmp/cachi2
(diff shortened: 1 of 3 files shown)


task/buildah/0.1/buildah.yaml
```diff
@@ -185,6 +193,10 @@ spec:
         BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
       fi
 
+      if [ -n "${BUILD_ARGS_FILE}" ]; then
+        BUILDAH_ARGS+=("--build-arg-file=${BUILD_ARGS_FILE}")
```
arewm, line 197: Does the path need to be local to the context, i.e. similar to the Dockerfile?
```
$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE
```

rcerven: yep,

fixed

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -234,6 +241,10 @@ spec:
         BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
       fi
 
+      if [ -n "${BUILD_ARGS_FILE}" ]; then
+        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${CONTEXT}/${BUILD_ARGS_FILE}")
```
arewm, line 245: How will precedence be determined between the buildah args that we set and those that are set via this file? Should we do any inspection of the args file to warn/fail if there is collision between the arguments?

rcerven: these are use as --build-arg argument=value   to be exposed as env variables during build

those aren't regular arguments for buildah build

and since we don't pass any --build-arg, there isn't any conflict

arewm (approved):

task/buildah/0.1/buildah.yaml
```diff
@@ -185,6 +193,10 @@ spec:
         BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
       fi
 
+      if [ -n "${BUILD_ARGS_FILE}" ]; then
+        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${CONTEXT}/${BUILD_ARGS_FILE}")
```
chmeliik, line 197: IMO we shouldn't assume the BUILD_ARGS_FILE is relative to the context dir

```suggestion
        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}")
```

Buildah only does this for the Containerfile, not for the build-arg-file

```bash
$ mkdir context
$ echo 'FROM scratch' > context/Containerfile
$ buildah build -f Containerfile context/
# works
$ echo foo=bar > context/args.conf
$ buildah build -f Containerfile --build-arg-file args.conf context /
Error: open args.conf: no such file or directory
```

rcerven: but we won't be passing just "args.conf" but full path to the context dir

build args are for component, so using context dir makes sense to me

chmeliik: > build args are for component, so using context dir makes sense to me

I don't think we have any basis for this. The user may want to pass the same args to all their components or different args to each component.

I would say that making the path relative to the context dir makes things unnecessarily more confusing

```yaml
- name: path-context
  value: context
- name: build-arg-file
  value: context/args.conf

# =>
# Error: open source/context/context/args.conf: no such file or directory

# Huh? Why is Konflux saying the file doesn't exist? It exists in my repo.
```

Especially considering that buildah doesn't behave like that (doesn't consider the --build-arg-file as relative to the context dir)

chmeliik (approved):
LGTM :+1:

rcerven:
I've changed it to "        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}")

because that is what I wanted initially do

and it is up to users, if they will provide builargs from the root of the repo "./args.conf", or from context "./context/args.conf"

anyway it is "full" path within repo which will be provided in the --build-arg-file

mmorhun (approved):

PR #937, opened 2024-04-12
export path to docker config json for preflight 

skattoju:
- **add support for private container repos by passing credentials to preflight via exported docker config json**


MartinBasti (approved):

jpopelka:
I can confirm that it works now, thanks.

PR #938, opened 2024-04-14
Configurable public key path for EC pipelines

gbenhaim:
Allow to mention the path to the public key in the enterprise-contract pipelines. This is needed when running on non OCP k8s cluster.


lcarva:
Looks reasonable to me besides what Zoran commented.

gbenhaim:
@lcarva @zregvart done

lcarva (approved):
/lgtm

zregvart (approved):

PR #945, opened 2024-04-17
fix(KFLUXBUGS-1215): fix use of PREFETCH_INPUT

scoheb:
- quotes within variable are not handled very well
      by ssh command and it impacts the buildah-remote task
      (which is based on this task)
      we only need to know if prefetch was used, not the actual
      input. Therefore, we inspect whether cachi2 has been called.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -108,8 +108,6 @@ spec:
       value: vfs
     - name: HERMETIC
       value: $(params.HERMETIC)
-    - name: PREFETCH_INPUT
-      value: $(params.PREFETCH_INPUT)
     - name: CONTEXT
       value: $(params.CONTEXT)
     - name: DOCKERFILE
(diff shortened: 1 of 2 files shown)


task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -258,6 +258,16 @@ spec:
...
         fi
+
+        # quotes within variable are not handled very well
+        # by ssh command and it impacts the buildah-remote task (which is based on this task)
+        # we only need to know if prefetch was used, not the actual
+        # input. Therefore, we simplify the variable.
+        #
+        if [ -n "${PREFETCH_INPUT}" ] ; then
+          PREFETCH_INPUT="prefetched"
+        fi
```
chmeliik, line 269: What specifically is the problem? I don't see how this would help. Here we're manipulating the variable in the script executed in the podman container on the VM. If SSH mangles the variable value, it's already been mangled before this point.

chmeliik: If there really is a problem, it might be better to stop using PREFETCH_INPUT and instead rely on checking `if [ -d "$(workspaces.source.path)/cachi2" ]`

scoheb: This is what I choose to do in the end. please review

task/buildah/0.1/buildah.yaml
```diff
@@ -96,8 +96,6 @@ spec:
       value: vfs
     - name: HERMETIC
       value: $(params.HERMETIC)
-    - name: PREFETCH_INPUT
-      value: $(params.PREFETCH_INPUT)
```
chmeliik, line 100: Huh, today I learned that removing a param is not a breaking change, Tekton will just ignore unknown params :open_mouth: 

cc @lcarva, I think I saw that in your migration table somewhere

@scoheb could you also remove the param from the `template-build.yaml`?

scoheb: I did not remove the param. only the env var.

chmeliik: Ohh, I see. But I actually tried passing something completely nonexistent to the buildah task, and it worked

chmeliik (approved):
LGTM :+1:

PR #962, opened 2024-04-23
Allow gather sources from images pulled from quay.io

tkdchen:
[STONEBLD-1831](https://issues.redhat.com//browse/STONEBLD-1831)

### DIFF
--- a/task/source-build/0.1/source-build.yaml
+++ b/task/source-build/0.1/source-build.yaml
@@ -80,6 +80,7 @@ spec:
         registry_allowlist="
         registry.access.redhat.com
         registry.redhat.io
+        quay.io/redhat-user-workloads
         "
 
         ## This is needed for the builds performed by the rpm-ostree task


task/source-build/0.1/source-build.yaml
```diff
@@ -80,6 +80,7 @@ spec:
         registry_allowlist="
         registry.access.redhat.com
         registry.redhat.io
+        quay.io
```
mmorhun, line 83: Does it make sense to limit it to specific org(s) ?

chmeliik: I think users can technically configure the build pipeline to push to any Quay organization they want, no?

tkdchen: I opened https://github.com/redhat-appstudio/build-tasks-dockerfiles/pull/70, which may be helpful for configuring this list in a more flexible way.

PR #965, opened 2024-04-24
Support prefetch-dependencies without hermetic

simonbaird:
IIUC there is some value in doing a prefetch-dependencies even if you're not quite ready to enable a hermetic build. This change makes it easier and more intuitive for people who want to do that.

For a pipeline with a non-empty value for prefetch-input, let's assume that the intention is to run the prefetch-dependencies task.

(This is a fast patch for an idea that occurred to me while working on https://issues.redhat.com/browse/EC-585 . Not sure how/where to test it. If the idea seems reasonable I'd be happy to create a fresh Jira for it and take some advice on what kind of testing it should have.)

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -87,9 +87,7 @@ spec:
           workspace: git-auth
     - name: prefetch-dependencies
       when:
-      - input: $(params.hermetic)
-        operator: in
-        values: ["true"]
+      - cel: "'$(params.prefetch-input) != '' || '$(params.hermetic)' == 'true'"
       params:
         - name: input
           value: $(params.prefetch-input)


simonbaird:
Actually it doesn't work - all the conditions must match for the task to run and I was expecting that it would run if any one of them matched.

It would be possible to write a cel expression to do it, but it makes the change more complex. For now I think I'll abandon this.

chmeliik:
> Actually it doesn't work - all the conditions must match for the task to run and I was expecting that it would run if any one of them matched.
> 
> It would be possible to write a cel expression to do it, but it makes the change more complex. For now I think I'll abandon this.
> 
> Update: Actually I did push a version with the cel expression, however I have not tested this cel expression.

Maybe we can go with your original attempt but drop the hermetic condition? There's no real reason to run the prefetch task when `hermetic=true` and `prefetch-input` is empty, the task is just a no-op in that case anyway.

```yaml
when:
- input: $(params.prefetch-input)
  operator: notin
  values: [""]
```

I don't actually remember why we had the hermetic condition :thinking: Maybe @brunoapimentel does?

brunoapimentel:
> I don't actually remember why we had the hermetic condition 🤔 Maybe @brunoapimentel does?

I only remember we were discussing cases in which the build could be hermetic even without prefetching. Maybe vendored deps for a non-supported package manager? It could also be needed in case of Java, since the prefetch is not and will not be done by Cachi2.

IMO, I think we should keep these two features (hermetic and prefetch) independent in the pipeline, in a way that you could do either, none or both by setting the params. We can accomodate more use cases like this.

simonbaird:
One thing I'm not sure about, is there any scenario where `params.prefetch-input` is empty and you still want to run the `prefetch-dependencies` task?

If so then the test for non-empty `params.prefetch-input` is not sufficient.

simonbaird:
Looking at https://github.com/redhat-appstudio/build-definitions/blob/ffc53980a69641def06e90e188bceeb754b2d55d/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L55-L59 I think I know the answer.

If we skip the whole task when the input param is blank no one is every going to see that "Build will be executed with network isolation..." message because the whole task would be skipped.

I would guess that's okay though.

simonbaird:
I might do a fresh PR.

brunoapimentel:
> One thing I'm not sure about, is there any scenario where `params.prefetch-input` is empty and you still want to run the `prefetch-dependencies` task?
> 
> If so then the test for non-empty `params.prefetch-input` is not sufficient.

I don't think so, since the `prefetch-input` is required by Cachi2, as it carries the [input package](https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#pre-fetch-dependencies) definitions. Not having it would just make the Cachi2 process fail.

> I think I know the answer.
> 
> If we skip the whole task when the input param is blank no one is every going to see that "Build will be executed with network isolation..." message because the whole task would be skipped.

Right, because we had a different idea in mind for these parameters. The way they're set up now, it you set `hermetic` to `true`, then `prefetch-dependencies` will still be executed, if only to print that message and `exit 0`.

I think that if we want truly independent params, we need to make `prefetch-dependencies` depend only on `prefetch-input`, and the network isolation in `buildah` depend only on `hermetic`.

We can drop that warning message then.

simonbaird:
See if you like #970, I think it's doing what you suggested above.

One possibility of skipping the prefetch-dependencies task is that folks doing hermetic builds that don't set prefetch-input might start to fail EC unless since the prefetch-dependencies task is required by default in the standard EC policy.

They would need something like this to tell EC it's okay to skip it:

```
...
exclude:
- tasks.required_tasks_found:prefetch-dependencies
...
```

simonbaird:
Replaced by #970.

PR #984, opened 2024-05-02
Revert "Dummy change to get the new Task bundle built"

zregvart:
Reverts redhat-appstudio/build-definitions#980


zregvart:
We need this so that the

PR #987, opened 2024-05-02
Dummy change to the deprecated-image-check 0.4

zregvart:
This is to get the the rebuild of the Task bundle so that the acceptable bundles are updated with this and previous versions.


zregvart:
@mmorhun can we get this merged?

mmorhun (approved):

PR #989, opened 2024-05-03
STONEBLD-2419: disabled source build test with parent using latest image

tisutisu:

### DIFF
--- a/.tekton/tasks/e2e-test.yaml
+++ b/.tekton/tasks/e2e-test.yaml
@@ -38,7 +38,8 @@ spec:
       - name: APP_SUFFIX
         value: "$(params.app_suffix)"
       - name: COMPONENT_REPO_URLS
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

lcarva:
Consider adding an explanation in the commit message regarding why this change is being made.

mmorhun:
@lcarva the problem is that the test is using [latest](https://github.com/redhat-appstudio-qe/source-build-use-latest-parent-image/blob/main/Dockerfile#L1) image and a new version has no signature on the parent source image. It looks like a bug on ubi side, but it affects us. Ideally we should downgrade the tag, but to avoid try/fail and to quickly unblock tests that always failing now, we just disable the test.

mmorhun:
It interesting, but the Konflux pipeline result was not reported back to this PR. The check above shows it "in progress" while actual pipeline `build-definitions-pull-request-94s8f` has already finished with success (logs are from tekton results):
```
[e2e-tests : e2e-test] Ran 82 of 676 Specs in 2012.338 seconds
[e2e-tests : e2e-test] SUCCESS! -- 82 Passed | 0 Failed | 39 Pending | 555 Skipped
[e2e-tests : e2e-test] PASS
```

mmorhun:
Merging this PR to unblock tests for all other PRs.

PR #992, opened 2024-05-04
feat(KONFLUX-2742): upload test results to quay as OCI artifacts

ralphbean:
For https://issues.redhat.com/browse/KONFLUX-2742

Will post example output

### DIFF
--- a/task/clamav-scan/0.1/clamav-scan.yaml
+++ b/task/clamav-scan/0.1/clamav-scan.yaml
@@ -15,6 +15,8 @@ spec:
       description: Tekton task test output.
     - name: IMAGES_PROCESSED
       description: Images processed in the task.
+    - name: RESULT_URI
+      description: OCI pullspec where the full clamav output can be found.
   params:
     - name: image-digest
       description: Image digest to scan.
@@ -67,6 +69,7 @@ spec:
 
         images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
         digests_processed=()
+        mkdir logs
         mkdir content
         cd content
         echo "Extracting image(s)."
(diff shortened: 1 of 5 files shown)


ralphbean:
I see a similarity in this and #913. We should consider:

* Should the result name convention be aligned with #913? i.e., should the results here be named SARIF_ARTIFACT and CLAMAV_ARTIFACT?
* Should the image used to push the artifact into quay be aligned?

zregvart:
Was there a consideration of using in-toto [vulnerabilities predicate](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md)?

ralphbean:
@zregvart no, it wasn't considered at all - not until you raised it just now.

task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -15,6 +15,8 @@ spec:
       description: Tekton task test output.
     - name: IMAGES_PROCESSED
       description: Images processed in the task.
+    - name: RESULT_URI
```
yashvardhannanavati, line 18: Do we need to update the ADR with this new result? 

https://github.com/konflux-ci/architecture/blob/main/ADR/0030-tekton-results-naming-convention.md

ralphbean: No. Good idea, but let's go in the opposite direction and rename this to be something _non_standard, like `CLAMAV_RESULT_URI`. Something just for consumers who know _this_ task.

ralphbean: Done in ef92eeb0c595fafba4b480fedeef861e239c10ef.

ralphbean:
For the moment, this is blocked on https://issues.redhat.com/browse/KONFLUX-2838

task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -141,6 +144,38 @@ spec:
...
+        - name: IMAGE_URL
+          value: $(params.image-url)
+        - name: IMAGE_DIGEST
+          value: $(params.image-digest)
+      workingDir: /work
+      script: |
+        #!/usr/bin/env bash
+
+        IFS=':' read -ra REPO <<< "$IMAGE_URL"
+        REPO="${REPO[0]}"
```
gbenhaim, line 159: This won't work if the IMAGE_URL has a port for the registry defined.

task/clamav-scan/0.1/clamav-scan.yaml
```diff
@@ -141,6 +144,38 @@ spec:
...
+
+        for UPLOAD_FILE in $(find . -name "clamscan-result*.log"); do
+          MEDIA_TYPE=text/vnd.clamav
+          args+=("${UPLOAD_FILE}:${MEDIA_TYPE}")
+        done
+        for UPLOAD_FILE in $(find . -name "clamscan-ec-test*.json"); do
+          MEDIA_TYPE=application/vnd.konflux.test_output+json
+          args+=("${UPLOAD_FILE}:${MEDIA_TYPE}")
+        done
+        retry oras push --no-tty --registry-config "$HOME/.docker/config.json" "${CLAMAV_RESULT_URI}" "${args[@]}"
```
gbenhaim, line 174: shouldn't we use `oras attach` for creating an artifacts graph?

task/sast-snyk-check/0.1/sast-snyk-check.yaml
```diff
@@ -84,5 +96,33 @@ spec:
           ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
         fi
         echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee $(results.TEST_OUTPUT.path)
+    - name: upload
```
gbenhaim, line 99: Same comments as above.

task/clamav-scan/0.1/clamav-scan.yaml
gbenhaim: Do we need a new result? if the scan results are `attached` to the container image they can be discovered using `oras discover ...`

gbenhaim: I do see a value in introducing a new result if it's going to be displayed in the UI (like we show the pull reference to the image). With that being said I still think we should use `oras attach`.

ralphbean: Ack, let's do both.

ralphbean: Tested at https://github.com/ralphbean/python-app/pull/51

```
❯ oras discover --distribution-spec v1.1-referrers-tag quay.io/redhat-user-workloads/ralphjbean-tenant/demo/python-app@sha256:5e3aff51d8342e2dbbfbabea4b647e580a022b471b2fc428f69931e7bb811017
quay.io/redhat-user-workloads/ralphjbean-tenant/demo/python-app@sha256:5e3aff51d8342e2dbbfbabea4b647e580a022b471b2fc428f69931e7bb811017
├── application/sarif+json
│   └── sha256:e55bc2e1bfc7be5339e95c332f6dc4aeefe8dd24d4009d252fa003bcb0dd23ad
└── application/vnd.clamav
    └── sha256:73ef70da4321875976f49e31f9a82607d8baef6cc4542dee7639390d4b98d726
```

task/clamav-scan/0.1/clamav-scan.yaml
ralphbean: Dropped this.

task/clamav-scan/0.1/clamav-scan.yaml
ralphbean: Done.

task/sast-snyk-check/0.1/sast-snyk-check.yaml
ralphbean: Done.

dirgim (approved):
LGTM

PR #1000, opened 2024-05-08
export the quay token to push bundles during e2e tests

joejstuart:
The tkn-bundle tests need to push a bundle to quay. This change export the token from a secret.

https://issues.redhat.com/browse/EC-624

### DIFF
--- a/.tekton/tasks/e2e-test.yaml
+++ b/.tekton/tasks/e2e-test.yaml
@@ -48,6 +48,11 @@ spec:
           secretKeyRef:
             name: github
             key: token
+      - name: QUAY_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: quay-push-secret
+            key: .dockerconfigjson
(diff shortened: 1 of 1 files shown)


.tekton/tasks/e2e-test.yaml
```diff
@@ -48,6 +48,11 @@ spec:
           secretKeyRef:
             name: github
             key: token
+      - name: QUAY_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: quay-repository
```
tisutisu, line 54: in the `tekton-ci` namespace of production cluster, I think the secret is named as [quay-push-secret](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/tekton-ci/production/quay-push-secret.yaml)

joejstuart: Thanks, just pushed the update.

zregvart (approved):

PR #1003, opened 2024-05-09
Script to manually update trusted tasks

zregvart:
This script can be run if there is a CI failure and trusted tasks bundle has not been updated with the newest revisions (digest bumps) or versions (new versions of Tasks).

By default it will update both git and oci references, resolving the Tasks from the local git repository and from the remote quay.io repository (quay.io/redhat-appstudio-tekton-catalog by default).

Since there is no resolver that supports `kustomize`, the Task definitions that rely on it will not be included.

Fixes: https://issues.redhat.com/browse/EC-545

### DIFF
--- a/hack/trusted-tasks-update.sh
+++ b/hack/trusted-tasks-update.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# Resolves all tasks from the `/task` directory of this repository and any Task
+# bundles in the `$QUAY_NAMESPACE` quay.io namespace (defaults to
+# `redhat-appstudio-tekton-catalog`) and runs `ec track bundle` with those git
+# and image references in order to catalogue the latest versions in the trusted
+# tasks list in `$INPUT_IMAGE` written to `$OUTPUT_IMAGE` (both default to
+# `quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest`).
+#
+# By default both git and image references are collected, what is collected can
+# be controlled by `$COLLECT` which can be set to `git` or `oci`
+#
+# Parameters via environment variables:
+# COLLECT        - can be set to `oci` or `git` (defaults to both), what to
+#                  resolve
+# INPUT_IMAGE    - Conftest OCI bundle containing the trusted tasks list, defaults
+#                  to:
+#                  `quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest`
+# OUTPUT_IMAGE   - Conftest OCI bundle to write the trusted task list to, defaults
+#                  to `$INPUT_IMAGE`
+# QUAY_NAMESPACE - Quay namespace to query for Task bundles (defaults to
+#                  `redhat-appstudio-tekton-catalog`)
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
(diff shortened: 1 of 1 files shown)


hack/trusted-tasks-update.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+# Resolves all tasks from the `/task` directory of this repository and any Task
+# bundles in the `$QUAY_NAMESPACE` quay.io namespace (defaults to
+# `redhat-appstudio-tekton-catalog`) and runs `ec track bundle` with those git
+# and image references in order to catalogue the latest versions in the trusted
+# tasks list in `$INPUT_IMAGE` written to `$OUTPUT_IMAGE` (both default to
+# `quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest`).
+#
+# By default both git and image refrences are collected, what is collected can
```
simonbaird, line 9:
```suggestion
# By default both git and image references are collected, what is collected can
```

hack/trusted-tasks-update.sh
```diff
@@ -0,0 +1,86 @@
...
+#                  to `$INPUT_IMAGE`
+# QUAY_NAMESPACE - Quay namespace to query for Task bundles (defaults to
+#                  `redhat-appstudio-tekton-catalog`)
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+COLLECT=${COLLECT:-git oci}
+mapfile -td ' ' < <(echo "${COLLECT}")
```
simonbaird, line 28: I can't grok this line. What var is being set? Should it be this?
```suggestion
mapfile COLLECT -td ' ' < <(echo "${COLLECT}")
```

simonbaird: Also I can't make the space delimiter work like that..

```
$ mapfile BAZ -td ' ' < <(echo "foo bar")
$ echo ${BAZ[0]}
foo bar

$ mapfile BAZ < <(printf "foo\nbar")
$ echo ${BAZ[0]}
foo
```

zregvart: I figured it out, the variable name needs to be after the parameters it seems:

```
bash -c 'mapfile -td " " BAZ < <(echo -n "foo bar"); echo "BAZ[0]: ${BAZ[0]}"; echo "BAZ[1]: ${BAZ[1]}"'
```

zregvart:
I like this version with the `case` a bit better, there was one echo that bugged me and now it is no longer there...

simonbaird (approved):

simonbaird:
Lgtm. A few ideas (not blockers):
* A way to do a dry run, i.e. spit out the collection output, but skip the ec track bundle command.
* A way to use (or list out in the case of a dry run) only the git or oci params that are not already present in the input image.

zregvart:
> Lgtm. A few ideas (not blockers):
> 
>     * A way to do a dry run, i.e. spit out the collection output, but skip the ec track bundle command.
> 
>     * A way to use (or list out in the case of a dry run) only the git or oci params that are not already present in the input image.

I think these should not be part of this PR, but the way to achieve them with the state as is on this PR would be to set the `OUTPUT_IMAGE` to a different image ref than the `INPUT_IMAGE` and then compare those two images.

PR #1004, opened 2024-05-10
Add new task for tagging images

mkosiarc:
This new task "apply-tags" allows specifying additional-tags that will be added to the specified image.

[STONEBLD-2418](https://issues.redhat.com//browse/STONEBLD-2418)

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -224,6 +224,15 @@ spec:
         value: $(tasks.build-container.results.IMAGE_URL)
       - name: IMAGE_DIGEST
         value: $(tasks.build-container.results.IMAGE_DIGEST)
+    - name: apply-tags
+      runAfter:
+        - build-container
+      taskRef:
+        name: apply-tags
+        version: "0.1"
+      params:
+      - name: IMAGE
+        value: $(tasks.build-container.results.IMAGE_URL)
 
   finally:
     - name: show-sbom
--- a/task/apply-tags/0.1/README.md
+++ b/task/apply-tags/0.1/README.md
@@ -1,0 +1,16 @@
+# apply-tags task
+
+Apply-tags task will apply additional tags to the specified IMAGE. These additional tags can be provided via the ADDITIONAL_TAGS array parameter or they can also be provided in the image label "konflux.additional-tags". If you specify more than one additional tag in the label, they must be separated by a comma or a blank space, e.g:
+
+```
+LABEL konflux.additional-tags="tag1, tag2"
+```
+```
+LABEL konflux.additional-tags="tag tag2"
+```
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|IMAGE|Reference of image that was pushed to registry in the buildah task.||true|
+|ADDITIONAL_TAGS|Additional tags that will be applied to the image in the registry.|[]|false|
--- a/task/apply-tags/0.1/apply-tags.yaml
+++ b/task/apply-tags/0.1/apply-tags.yaml
@@ -1,0 +1,63 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "appstudio, hacbs"
+  name: apply-tags
+spec:
+  description: >-
+    Applies additional tags to the built image.
+  params:
+  - name: IMAGE
+    description: Reference of image that was pushed to registry in the buildah task.
+    type: string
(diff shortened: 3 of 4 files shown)


task/apply-tags/0.1/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+```
+LABEL konflux.additional-tags="tag1, tag2"
+```
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|IMAGE|Reference of image that was pushed to registry in the buildah task.||true|
+|ADDITIONAL_TAGS|Additional tags that will be applied to the image in the registry.|[]|false|
```
mmorhun, line 13: What about ability to push into another registry?

mmorhun (approved):

task/apply-tags/0.1/README.md
mkosiarc: We agreed not supporting pushes to another registry in the design phase

mkosiarc: so for now, it would always push the new tags to the place where the original image is.

task/apply-tags/0.1/apply-tags.yaml
```diff
@@ -0,0 +1,63 @@
...
+    type: string
+  - name: ADDITIONAL_TAGS
+    description: Additional tags that will be applied to the image in the registry.
+    type: array
+    default: []
+  steps:
+    - name: apply-additional-tags-from-parameter
+      image: registry.access.redhat.com/ubi9/skopeo:9.4-6@sha256:c4d70dec3eb0a0c831490192145ea25431fe04d1cf307f8d61e2d87adb41e7e3
+      args:
+        - $(params.ADDITIONAL_TAGS)
```
chmeliik, line 25:
```suggestion
        - $(params.ADDITIONAL_TAGS[*])
```

I thought array params were supposed to be used like this https://tekton.dev/docs/pipelines/tasks/#substituting-array-parameters. Does it work the same way without the `[*]`?

task/apply-tags/0.1/README.md
```diff
@@ -0,0 +1,13 @@
+# apply-tags task
+
+Apply-tags task will apply additional tags to the specified IMAGE. These additional tags can be provided via the ADDITIONAL_TAGS array parameter or they can also be provided in the image label "konflux.additional-tags". If you specify more than one additional tag in the label, they must be separated by a comma, e.g:
```
chmeliik, line 3: > they must be separated by a comma, e.g:

Technically, it also works if they are space-separated 

```bash
$ IFS=', ' read -r -a array <<< 'v1 v1.0 latest'
$ printf "%s\n" "${array[@]}"
v1
v1.0
latest
```

That's probably a good thing

pipelines/template-build/template-build.yaml
```diff
@@ -224,6 +224,15 @@ spec:
...
         value: $(tasks.build-container.results.IMAGE_DIGEST)
+    - name: apply-tags
+      runAfter:
+        - build-container
+      taskRef:
+        name: apply-tags
+        version: "0.1"
+      params:
+      - name: IMAGE
+        value: $(tasks.build-container.results.IMAGE_URL)
```
chmeliik, line 235: This might be a bit safer (in case IMAGE is not a unique tag)

```suggestion
        value: $(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)
```

But would require an update to the IMAGE_WITHOUT_TAG logic

task/apply-tags/0.1/apply-tags.yaml
mkosiarc: I tested it again and the output is identical whether I use [*] or not. I noticed that yamlint task does not use it
https://github.com/mkosiarc/build-definitions/blob/main/.tekton/tasks/yaml-lint.yaml#L31

However, I'll update it so it matches the documentation

task/apply-tags/0.1/README.md
mkosiarc: Good point. I updated the README docs instead of changing the behavior

pipelines/template-build/template-build.yaml
mkosiarc: Even if the tag wouldn't be unique, IMAGE_URL and IMAGE_DIGEST will point to the same image at build time (after buildah task finishes and apply-tags starts), so the tags will be added to the correct image. If the original-non-unique tag changes afterwards, this should not affect the already applied additional-tags right?

chmeliik: True. The only case where this would matter is if you have two builds running at the same time - then pipeline A could accidentally apply the additional tags to the image built by pipeline B. But that's not a big deal. Maybe it's even the desired behavior, since the image built by pipeline B would be the newer one.

chmeliik (approved):
LGTM :+1:

PR #1009, opened 2024-05-13
chore: drop comments explaining the tag+digest combination

ralphbean:
This comment is getting copied everywhere everytime we add a new step or task.

It's true I asked for this to be put in place, but it takes on a life of its own when it lives as a comment in the task definitions like this.


.tekton/pull-request.yaml
```diff
@@ -81,7 +81,6 @@ spec:
               image: quay.io/redhat-appstudio/pull-request-builds:build-definitions-utils-{{revision}}
               # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
```
lcarva, line 82: Maybe this is also out of place? Not sure if there's any sense in copying this everywhere.

lcarva (approved):

PR #1015, opened 2024-05-15
buildah: always use a context of .

owtaylor:
In certain cases, having a context set could change the meaning of a Containerfile. See:

https://github.com/containers/buildah/issues/5529

For details of a situation affecting Flatpaks and bootc images. This applies to both to any context from the CONTEXT parameter, and the implicit context from SOURCE_CODE_DIR. To avoid any possible problems, just change to the context directory and use a context argument of .

[edited to fix a wrong link]

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -188,9 +188,9 @@ spec:
 
       SOURCE_CODE_DIR=source
       if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
-        dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
+        dockerfile_path="$(pwd)/$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
       elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
-        dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
+        dockerfile_path="$(pwd)/$SOURCE_CODE_DIR/$DOCKERFILE"
       elif echo "$DOCKERFILE" | grep -q "^https\?://"; then
         echo "Fetch Dockerfile from $DOCKERFILE"
         dockerfile_path=$(mktemp --suffix=-Dockerfile)
@@ -240,7 +240,7 @@ spec:
       fi
 
       if [ -n "${BUILD_ARGS_FILE}" ]; then
-        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}")
+        BUILDAH_ARGS+=("--build-arg-file=$(pwd)/$SOURCE_CODE_DIR/${BUILD_ARGS_FILE}")
       fi
 
       if [ -d "$(workspaces.source.path)/cachi2" ]; then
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
LGTM :+1:

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -188,9 +188,9 @@ spec:
 
       SOURCE_CODE_DIR=source
       if [ -e "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE" ]; then
-        dockerfile_path="$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
+        dockerfile_path="$(pwd)/$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"
       elif [ -e "$SOURCE_CODE_DIR/$DOCKERFILE" ]; then
-        dockerfile_path="$SOURCE_CODE_DIR/$DOCKERFILE"
+        dockerfile_path="$(pwd)/$SOURCE_CODE_DIR/$DOCKERFILE"
```
zregvart, line 193: This looks odd to me, if we're testing for existence of `"$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"` and `"$SOURCE_CODE_DIR/$DOCKERFILE"` why prepend the `$(pwd)` to their path? Is it to get the full path to the Dockerfile?

For that I think `realpath` should be better alternative, i.e. something like:

```
      if [ "${DOCKERFILE}" =~ ^http(s)?://.* ]]; then
        echo "Fetch Dockerfile from $DOCKERFILE"
        // ... rest of the fetching code
      else
        dockerfile_path="$(realpath "$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE")"
        [ ! -f "${dockerfile_path}" ] && dockerfile_path="$(realpath "$SOURCE_CODE_DIR/$DOCKERFILE")"
        [ ! -f "${dockerfile_path}" ] && dockerfile_path="$(pwd)/$SOURCE_CODE_DIR/$DOCKERFILE"
        if [ ! -f "${dockerfile_path}" ]; then
          echo "Cannot find Dockerfile $DOCKERFILE"
          exit 1
        fi
      fi
```

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -240,7 +240,7 @@ spec:
       fi
 
       if [ -n "${BUILD_ARGS_FILE}" ]; then
-        BUILDAH_ARGS+=("--build-arg-file=${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}")
+        BUILDAH_ARGS+=("--build-arg-file=$(pwd)/$SOURCE_CODE_DIR/${BUILD_ARGS_FILE}")
```
zregvart, line 243: Prepending `$(pwd)` here is very odd, see the comment above

task/buildah-remote/0.1/buildah-remote.yaml
owtaylor: > This looks odd to me, if we're testing for existence of `"$SOURCE_CODE_DIR/$CONTEXT/$DOCKERFILE"` and `"$SOURCE_CODE_DIR/$DOCKERFILE"` why prepend the `$(pwd)` to their path? Is it to get the full path to the Dockerfile?

The reason for this is that we are going to change the working directory before actually running `buildah build`; dockerfile_path is referenced in various places so an absolute path is most convenient. I agree that using `realpath` would be a fine alternative (*), but since we know SOURCE_CODE_DIR is relative (in fact it's always `source`), the $(pwd)/ technique should be reliable as well.

If it's desired, I can certainly do a new PR that uses realpath and/or adds comments about why an absolute path.

(*) I might argue that realpath might turn `/workspace` into /mnt/kubernetes/<uuid>/ or something like that - resolving symlinks isn't always better. But in this case we don't even log the buildah command line, so it doesn't much matter either way.

PR #1016, opened 2024-05-16
Buildah: Add SKIP_UNUSED_STAGES parameter

owtaylor:
Expose the --skip-unused-stages argument to Buildah as a parameter. This is needed to handle certain Containerfiles where stages have hidden side-effects like writing things to the context directory.

See https://github.com/containers/buildah/issues/5529 for details.


chmeliik (approved):
LGTM :+1:

owtaylor:
I had thought I had a workaround that avoided the need for this, but it apparently doesn't work in all scenarios, so we still need this.

Issue #1014, opened 2024-05-15
RFE: way to set up a "install root" within a buildah task


owtaylor:
In various contexts, we want to install packages into a sub directory of the root directory, and then use them as the root directory of a subsequent stage. Examples include:

 * Creating a standard base image from a Containerfile
 * [Creating a bootc base image using rpm-ostree](https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/Containerfile)
 * [Creating a Flatpak runtime](https://github.com/owtaylor/fedora-flatpak-runtime/-/blob/main/Containerfile)

In all of these cases, we really don't want to install into a bare directory - this will result in weirdness like redirections to /dev/null going into a actual file. The directory at least needs a skeleton /dev populated and would ideally have /proc and /sys as well.

There are two basic ways that this could be enabled:

 * Allow running `buildah build` with `--cap-add=all --security-opt=label=disable` ; this is sufficient to allow the Containerfile to create its own mount namespace and populate the root directory itself.
 * Instead specify the install directory as a task parameter, and have the buildah task add something like:

   ```
          --device=/dev/null:"$INSTALL_ROOT"/dev/null
          --device=/dev/random:"$INSTALL_ROOT"/dev/random
          --device=/dev/urandom:"$INSTALL_ROOT"/dev/urandom
          --device=/dev/zero:"$INSTALL_ROOT"/dev/zero
   ``` 

The advantage of this is that it is more flexible and satisfies some other needs for nested sandboxing[^1]. It may the only way that actually works for the bootc case [@cgwalters - is that right?]. On the other hand, the second approach is more obviously secure. [^2]

[^1]: To delve some into this, if we were using buildah with user namespaces, then --cap-add=all would not be necessary since the Containerfile could create a nested user namespace where it had the SYS_ADMIN capability, and then create a mount namespace and set things up there. But with `--isolation=chroot` the run commands are being run in an active chroot, and the Linux kernel disables user namespace creation when a chroot is active, so we have to actually give the SYS_ADMIN capability to the subprocess so it can create a mount namespace

[^2]: `--cap-add=all` will not compromise overall system security, since we're counting on the pod running the task for that, and `buildah --isolation=chroot` is already not that strong. And of course, an actual malicious component could specify their own pipeline with its own build steps already. The main weakness it introduces is that it could make it easier for malicious content pulled into the Containerfile during the build process to mess with the build artifacts and build metadata.

chmeliik:
[...]

PR #1017, opened 2024-05-16
buildah: Add an ADD_CAPABILITIES parameter

owtaylor:
This corresponds to 'buildah build --cap-add=<value>'

One use case is when we want to do nested sandboxing within the build, and need to set up a new mount namespace - in this case we need ADD_CAPABILITIES=SYS_ADMIN. (Since we are running buildah using --isolation=chroot.)

Related: #1014

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -86,6 +86,11 @@ spec:
     description: Path to a file with build arguments, see https://www.mankier.com/1/buildah-build#--build-arg-file
     name: BUILD_ARGS_FILE
     type: string
+  - default: ""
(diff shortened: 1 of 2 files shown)


chmeliik (commented):
@lcarva @zregvart @simonbaird 

For this feature, we would need a policy checks that disallows the param by default but lets the policy author allowlist some specific capabilities. Is that feasible?

ralphbean:
I pushed a rebase of this at https://github.com/konflux-ci/build-definitions/pull/1084 (with an additional call to `./hack/generate-ta-tasks.sh` included).

ralphbean:
Merging this in https://github.com/konflux-ci/build-definitions/pull/1084.

PR #1020, opened 2024-05-16
Move repo to konflux-ci

chmeliik:
[STONEBLD-2339](https://issues.redhat.com//browse/STONEBLD-2339)

### DIFF
--- a/hack/trusted-tasks-update.sh
+++ b/hack/trusted-tasks-update.sh
@@ -27,7 +27,7 @@ set -o pipefail
 mapfile -td ' ' COLLECT < <(echo -n "${COLLECT:-git oci}")
 INPUT_IMAGE=${INPUT_IMAGE:-quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest}
 OUTPUT_IMAGE=${OUTPUT_IMAGE:-$INPUT_IMAGE}
-GIT_REPOSITORY=git+https://github.com/redhat-appstudio/build-definitions.git
+GIT_REPOSITORY=git+https://github.com/konflux-ci/build-definitions.git
 QUAY_NAMESPACE=${QUAY_NAMESPACE:-redhat-appstudio-tekton-catalog}
 
 HACK_DIR="$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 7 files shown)


hack/trusted-tasks-update.sh
```diff
@@ -27,7 +27,7 @@ set -o pipefail
 mapfile -td ' ' COLLECT < <(echo -n "${COLLECT:-git oci}")
 INPUT_IMAGE=${INPUT_IMAGE:-quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest}
 OUTPUT_IMAGE=${OUTPUT_IMAGE:-$INPUT_IMAGE}
-GIT_REPOSITORY=git+https://github.com/redhat-appstudio/build-definitions.git
+GIT_REPOSITORY=git+https://github.com/konflux-ci/build-definitions.git
```
chmeliik, line 30: @zregvart how much will moving this repo to github.com/konflux-ci break the trusted task data?

I suppose we'll need to allow both github.com/konflux-ci and github.com/redhat-appstudio for all the tasks? (If anyone is using the tasks with the git resolver today, they're using github.com/redhat-appstudio and they can keep using it thanks to the automatic redirects)

chmeliik: Updated the script to add both urls as trusted, LMK what you think

chmeliik:
Updated the build-acceptable-bundles script as well. Tested with:
* make an arbitrary change to some task yaml, commit the change
* `REVISION=$(git rev-parse HEAD) .tekton/scripts/build-acceptable-bundles.sh /dev/null`

```
Tasks to be added:
git+https://github.com/redhat-appstudio/build-definitions.git//task/prefetch-dependencies/0.1/prefetch-dependencies.yaml@763afc9a8c72576ee7034e90dcb1879d913ad31e
git+https://github.com/konflux-ci/build-definitions.git//task/prefetch-dependencies/0.1/prefetch-dependencies.yaml@763afc9a8c72576ee7034e90dcb1879d913ad31e
Bundles to be added:

```

zregvart (approved):

chmeliik:
After offline discussion, we decided that tracking the old redhat-appstudio url for trusted tasks was not necessary (there is no evidence that anyone would be referencing our tasks via git resolver today).

After the repo is moved and this PR is merged, the git urls in trusted task data will only contain the konflux-ci references.

PR #1022, opened 2024-05-16
Use consistent naming for -oci-ta Tasks

zregvart:
We should be consistent in our naming, makes for easier manipulation via tools.


lcarva:
I think the naming of parameters and results should be consistent within the Task. Given that different Tasks in this repo are managed by different teams, I don't see the value in keeping such naming consistent between them. In fact, the parameters/results just look odd in the context of each Task.

We could pursue an approach where we are consistent across all the Tasks, but I'd rather tackle that as its own larger effort.

zregvart:
I agree, it is odd that we do not follow any single naming convention. For Trusted Artifacts to be usable in a policy, we do however need some naming conventions (e.g. the results need to be named `*_ARTIFACT`, so at least we can have some semblance of a naming convention there.

lcarva:
Ack. From the perspective of being able to generalize the policy rules, that makes sense.

PR #1024, opened 2024-05-16
Add docker-build-oci-ta Pipeline

lcarva:
This Pipeline, based on the docker-build Pipeline, uses Trusted Artifacts stored in an OCI registry to share data between Tasks. No PVC is needed.

Ref: [EC-555](https://issues.redhat.com//browse/EC-555)

### DIFF
--- a/pipelines/docker-build-oci-ta/patch.yaml
+++ b/pipelines/docker-build-oci-ta/patch.yaml
@@ -1,0 +1,126 @@
+---
+- op: replace
+  path: /metadata/name
+  value: docker-build-oci-ta
+- op: replace
+  path: /metadata/labels
+  value:
+    "pipelines.openshift.io/used-by": "build-cloud"
+    "pipelines.openshift.io/runtime": "generic"
+    "pipelines.openshift.io/strategy": "docker"
+- op: remove
+  path: /spec/workspaces/0
+# Order of Tasks from the base docker-build Pipeline:
+# $ kustomize build pipelines/docker-build | yq .spec.tasks.[].name | nl -v 0
+#  0  init
+#  1  clone-repository
+#  2  prefetch-dependencies
+#  3  build-container
+#  4  build-source-image
+#  5  deprecated-base-image-check
+#  6  clair-scan
+#  7  ecosystem-cert-preflight-checks
+#  8  sast-snyk-check
+#  9  clamav-scan
+# 10  sbom-json-check
+# 11  apply-tags
+
+# clone-repository Task
+- op: replace
+  path: /spec/tasks/1/taskRef/name
+  value: git-clone-oci-ta
+- op: add
+  path: /spec/tasks/1/params/-
+  value:
+    name: ociStorage
+    value: $(params.output-image).git
+- op: add
+  path: /spec/tasks/1/params/-
+  value:
+    name: ociArtifactExpiresAfter
+    value: $(params.image-expires-after)
+- op: remove
+  path: /spec/tasks/1/workspaces/0
+
+# prefetch-dependencies Task
+- op: replace
+  path: /spec/tasks/2/taskRef/name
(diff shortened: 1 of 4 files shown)


lcarva:
I'll likely have to adjust this once https://github.com/redhat-appstudio/build-definitions/pull/1022 is merged.

lcarva:
Moving this to draft because there are two gaps here:
1. We need to also introduce a version of the show-sbom Task that understand trusted artifacts :cry: 
2. The patch.yaml here needs to actually remove the workspace from the docker-build-oci-ta Pipeline definition

lcarva:
Ready for review. We decided that the summary Task doesn't provide a whole lot of value. We will exclude it from the `docker-build-oci-ta` Pipeline. See the comments [here](https://issues.redhat.com/browse/EC-643) for additional context.

(Also, if you do think we should include the summary Task in this new Pipeline, let me know!)

zregvart (approved):

pipelines/docker-build-oci-ta/patch.yaml
```diff
@@ -0,0 +1,122 @@
...
+  value:
+    name: SOURCE_ARTIFACT
+    value: $(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)
+- op: add
+  path: /spec/tasks/4/params/-
+  value:
+    name: CACHI2_ARTIFACT
+    value: $(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)
+- op: remove
+  path: /spec/tasks/4/workspaces/0
```
simonbaird, line 99: Did you consider removing the whole thing when you expect it to be empty afterwards?
```suggestion
  path: /spec/tasks/2/workspaces
```

Just an idea for tidier yaml - I don't think it would make any functional difference.

simonbaird:
I did this to compare the old/new:

```
vimdiff <(kustomize build pipelines/docker-build) <(kustomize build pipelines/docker-build-oci-ta)
```

Looks nice and tidy except I guess for the `show-summary` removal.

pipelines/docker-build-oci-ta/patch.yaml
```diff
@@ -0,0 +1,122 @@
...
+- op: remove
+  path: /spec/tasks/8/workspaces/0
+
+# Order of finally Tasks from the base docker-build Pipeline:
+# $ kustomize build pipelines/docker-build | yq .spec.finally.[].name | nl -v 0
+#  0  show-sbom
+#  1  show-summary
+
+# show-summary
+- op: remove
```
simonbaird, line 121: Maybe add a comment to explain why we want to nuke this one.

simonbaird (approved):

pipelines/docker-build-oci-ta/patch.yaml
lcarva: I'm gonna leave it as is for consistency. The empty array is not a big deal.

pipelines/docker-build-oci-ta/patch.yaml
lcarva: Good call. Thanks.

zregvart (approved):

lcarva:
CI is failing here with an odd error:

```
Trying to pull registry.access.redhat.com/ubi8/nodejs-18-minimal:latest...
Error: creating build container: copying system image from manifest list: Source image rejected: None of the signatures were accepted, reasons: Signature for identity registry.access.redhat.com/ubi8/nodejs-18-minimal:1-115.1716502126 is not accepted; Signature for identity registry.redhat.io/ubi8/nodejs-18-minimal:1-115.1716502126 is not accepted
```

I can reproduce the issue locally with `podman pull` on Fedora 39. Pulling by the `latest` tag shows the same error. Pulling by the NVR tag works.

PR #1025, opened 2024-05-16
Custom CA bundle for build-task

ashwindasr:
Replicates https://github.com/redhat-appstudio/build-definitions/pull/942, but for build task

Slack [convo](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1715889582174049?thread_ts=1715795662.148729&cid=C04PZ7H0VA8)

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -86,6 +86,15 @@ spec:
     description: Path to a file with build arguments, see https://www.mankier.com/1/buildah-build#--build-arg-file
     name: BUILD_ARGS_FILE
     type: string
+  - default: trusted-ca
+    description: The name of the ConfigMap to read CA bundle data from.
+    name: caTrustConfigMapName
+    type: string
+  - default: ca-bundle.crt
+    description: The name of the key in the ConfigMap that contains the CA bundle
+      data.
+    name: caTrustConfigMapKey
+    type: string
   - description: The platform to build on
     name: PLATFORM
     type: string
@@ -189,6 +198,13 @@ spec:
       cd $(workspaces.source.path)
       if [ -n "${PARAM_BUILDER_IMAGE}" ]; then
         echo "WARNING: provided deprecated BUILDER_IMAGE parameter has no effect."
+      fi
+
+      ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+      if [ -f "$ca_bundle" ]; then
+        echo "INFO: Using mounted CA bundle: $ca_bundle"
+        cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+        update-ca-trust
       fi
 
       SOURCE_CODE_DIR=source
@@ -365,6 +381,9 @@ spec:
       name: varlibcontainers
     - mountPath: /entitlement
       name: etc-pki-entitlement
+    - mountPath: /mnt/trusted-ca
+      name: trusted-ca
+      readOnly: true
     - mountPath: /ssh
       name: ssh
       readOnly: true
@@ -537,6 +556,13 @@ spec:
     secret:
       optional: true
       secretName: $(params.ENTITLEMENT_SECRET)
+  - configMap:
+      items:
+      - key: $(params.caTrustConfigMapKey)
+        path: ca-bundle.crt
+      name: $(params.caTrustConfigMapName)
+      optional: true
+    name: trusted-ca
   - name: ssh
     secret:
       optional: false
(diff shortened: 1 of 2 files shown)


gbenhaim:
looks good to me, but I would like someone from the build team to approve @mmorhun @chmeliik

task/buildah/0.1/buildah.yaml
```diff
@@ -75,6 +75,14 @@ spec:
     description: Path to a file with build arguments which will be passed to podman during build
     type: string
     default: ""
+  - name: caTrustConfigMapName
+    type: string
+    description: The name of the ConfigMap to read CA bundle data from.
+    default: trusted-ca
+  - name: caTrustConfigMapKey
```
chmeliik, line 82: I'm conflicted about the consistency of these param names.

On the one hand, the buildah task uses UPPERCASE_NAMES, so this is inconsistent. On the other hand, we already dropped the ball in #942

(And in the context of the whole pipeline, the params should be consistent across all the tasks.)

For this PR, I don't really care what you choose. We really need to make the param casing across all tasks consistent. Breaking changes on that scale will be fun :smiling_face_with_tear:

chmeliik (commented):
LGTM, could you just improve the commit message? `Custom CA bundle for prefetch-dependencies for build-task` doesn't really make sense :)

yftacherzog:
I think we'd need to do something with the `upload-sbom` step as well so that it works with self-signed certificate.
I'm not sure whether cosign has such option though. It might be that we need to use `--allow-insecure-registry` there.

gbenhaim:
@yftacherzog maybe it can be done in a followup pr?

task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -186,6 +195,13 @@ spec:
         echo "WARNING: provided deprecated BUILDER_IMAGE parameter has no effect."
       fi
 
+      ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+      if [ -f "$ca_bundle" ]; then
+        echo "INFO: Using mounted CA bundle: $ca_bundle"
+        cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
```
ifireball, line 201: Perhaps we can symlink instead of copy to avoid creating another copy of the entire bundle in RAM.

For now we end up with 3 copies at least:
1. The config map
2. The file in `/etc/pki/ca-trust/source/anchors`
3. The resulting bundle file after running `update-ca-trust`

Maybe we can even mount the config map directly into `/etc/pki/ca-trust/source/anchors` ?

yftacherzog:
> @yftacherzog maybe it can be done in a followup pr?

I'm ok with that, but does this change provide added functionality without the upload-sbom task?
If I understand correctly, with this change, running the task will fail on `upload-sbom` instead of on `build`.

chmeliik:
> I think we'd need to do something with the `upload-sbom` step as well so that it works with self-signed certificate. I'm not sure whether cosign has such option though. It might be that we need to use `--allow-insecure-registry` there.

Good catch :+1: 

> I'm ok with that, but does this change provide added functionality without the upload-sbom task?
If I understand correctly, with this change, running the task will fail on upload-sbom instead of on build.

Now that you mention it, @ashwindasr @gbenhaim what exactly is the purpose of this change? Is to let the buildah task pull images from / push images to registries with self-signed certs?

Or is to let any instructions in the Containerfile access any servers with self-signed certs? If it's the latter, I'm pretty sure this PR doesn't achieve it

task/buildah-remote/0.1/buildah-remote.yaml
gbenhaim: maybe, but we now that this approach works and I don't think the memory overhead is too big.
I think we should proceed with this PR for unblocking @ashwindasr

gbenhaim:
@yftacherzog, what @ashwindasr needs it to pull images from a registry with self signed certificates.

task/buildah-remote/0.1/buildah-remote.yaml
ifireball: How hare is it to test using `ln -s` instead of `cp` ?

ashwindasr:
Thanks @yftacherzog 

> needs it to pull images from a registry with self signed certificates

So do you think this PR would help us fix this issue, or do you think additional changes are required?

yftacherzog:
> Thanks @yftacherzog
> 
> > needs it to pull images from a registry with self signed certificates
> 
> So do you think this PR would help us fix this issue, or do you think additional changes are required?

As far as I understand, it should.

ashwindasr:
@gbenhaim looks like you're not an approver?

gbenhaim:
@chmeliik can you please approve the pr?

chmeliik (approved):

zregvart:
The corresponding update to the buildah-oci-ta Task is in #1039

PR #1028, opened 2024-05-17
buildah: support passing build args as Tekton parameter list

msugakov:
## Description

For building ACS containers, we need to pass the same parameter to multiple tasks, one of them is `buildah`.

While there is `BUILD_ARGS_FILE` parameter (added in https://github.com/redhat-appstudio/build-definitions/pull/935), it requires us _writing_ a file in a workspace which, well, requires creating a custom task which will be a pain to get rid of in order to conform with EC.

Here I introduce an additional parameter of Tekton's type `list` - `BUILD_ARGS_LIST`.

## Notes on testing

I developed it in a toy task and also verified how it runs in a pipeline: <https://github.com/stackrox/scanner/pull/1512>. ([Example runs](https://console.redhat.com/preview/application-pipeline/workspaces/rh-acs/applications/acs/commit/2b810bde1582d0bbf51268b38b277c16193b3825), if you have access to the ACS tenant.)

A thing I'm concerned about is that now `--build-arg-file` option will point to a file that's outside of `buildah` Context directory (`/build-args/build-args-unified`). I think it should be fine but I was not able to test whether `buildah` will object (the latest version I get on my machine is 1.28 and it does not have that command-line option).

More broadly, I don't know how to test this change on real pipelines before it's merged so I'll appreciate some guidance. Also, leaving the template text :point_down: as a reminder.

### DIFF
--- a/task/buildah-remote/0.1/buildah-remote.yaml
+++ b/task/buildah-remote/0.1/buildah-remote.yaml
@@ -79,10 +79,17 @@ spec:
     name: ENTITLEMENT_SECRET
     type: string
   - default: ""
-    description: Path to a file with build arguments which will be passed to podman
-      during build
+    description: |
+      Path to a file with build arguments which will be passed to buildah during build.
+      The path is relative to the repository root.
     name: BUILD_ARGS_FILE
     type: string
+  - default: []
+    description: |
+      List of arguments that will be passed to buildah during build. Each item in the list should be in the format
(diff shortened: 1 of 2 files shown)


chmeliik:
@msugakov could you check #1023? Do you see some advantages in your approach over it?

msugakov:
Closing in favor of https://github.com/redhat-appstudio/build-definitions/pull/1023

PR #1032, opened 2024-05-22
fix: add golang in the image

tisutisu:
Will be using this image in prepare-e2e-tasks task in integration tests of build-task-dockerfiles repo
Part of the story: https://issues.redhat.com/browse/STONEBLD-2353

### DIFF
--- a/appstudio-utils/Dockerfile
+++ b/appstudio-utils/Dockerfile
@@ -13,6 +13,7 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     git \
     skopeo \
     make \
+    golang \
     && dnf clean all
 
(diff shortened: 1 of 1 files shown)


appstudio-utils/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     git \
     skopeo \
     make \
+    go-toolset \
```
chmeliik, line 16: Can we install just `golang`? It looks like `go-toolset` only adds the `delve` debugger and nothing else

chmeliik (approved):
LGTM :+1:

tkdchen (approved):

PR #1033, opened 2024-05-22
 Community Docker build pipeline

gbenhaim:
Author: gbenhaim <gbenhaim@redhat.com>
Date:   Wed May 22 11:18:42 2024 +0300

    Add docker-build community pipeline
    
    It has the same logic has the docker-build pipeline,
    but it uses only tasks that use public images.
    
    Signed-off-by: gbenhaim <gbenhaim@redhat.com>

commit 04d9148e2cda4a4d8eb0a48f0fab652920e61f04
Author: gbenhaim <gbenhaim@redhat.com>
Date:   Wed May 22 11:15:34 2024 +0300

    Add git-clone community task
    
    It has the same logic such as the git-clone task, but it's
    using an image which is publicly available.
    
    Signed-off-by: gbenhaim <gbenhaim@redhat.com>

commit ff790de061e19df89505236522a86f8f4e9a9384
Author: gbenhaim <gbenhaim@redhat.com>
Date:   Wed May 22 11:11:47 2024 +0300

    Add init community task
    
    It has the same logic such as the init task, but it's
    using an image which is publicly available.
    
    Signed-off-by: gbenhaim <gbenhaim@redhat.com>

### DIFF
--- a/task/git-clone-community/0.1/patch.yaml
+++ b/task/git-clone-community/0.1/patch.yaml
@@ -1,0 +1,8 @@
+---
+- op: replace
+  path: /metadata/name
+  value: git-clone-community
+
+- op: replace
+  path: /spec/steps/0/image
+  value: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.40.2
--- a/task/init/0.2/init.yaml
+++ b/task/init/0.2/init.yaml
@@ -25,7 +25,7 @@ spec:
 
   steps:
     - name: init
-      image: registry.redhat.io/openshift4/ose-cli:4.13@sha256:73df37794ffff7de1101016c23dc623e4990810390ebdabcbbfa065214352c7c
+      image: quay.io/openshift/origin-cli@sha256:086156148c5ef531223b959c3636f0cdb556272eb5a181a3a2fe43d822f41b4f
       env:
         - name: IMAGE_URL
           value: $(params.image-url)
--- a/pipelines/docker-build-community/README.md
+++ b/pipelines/docker-build-community/README.md
@@ -1,0 +1,226 @@
+# "docker-build pipeline"
+## Parameters
+|name|description|default value|used in (taskname:taskrefversion:taskparam)|
+|---|---|---|---|
+|build-args| Array of --build-arg values ("arg=value" strings) for buildah| []| build-container:0.1:BUILD_ARGS|
+|build-args-file| Path to a file with build arguments for buildah, see https://www.mankier.com/1/buildah-build#--build-arg-file| | build-container:0.1:BUILD_ARGS_FILE|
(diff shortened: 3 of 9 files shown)


task/init/0.2/init.yaml
```diff
@@ -25,7 +25,7 @@ spec:
 
   steps:
     - name: init
-      image: registry.redhat.io/openshift4/ose-cli:4.13@sha256:73df37794ffff7de1101016c23dc623e4990810390ebdabcbbfa065214352c7c
+      image: quay.io/openshift/origin-cli@sha256:086156148c5ef531223b959c3636f0cdb556272eb5a181a3a2fe43d822f41b4f
```
chmeliik, line 28: We could instead rework the `init` task to use `skopeo` rather than `oc image`, and then we could use the registry.access.redhat.com skopeo image

task/git-clone-community/0.1/patch.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- op: replace
+  path: /metadata/name
+  value: git-clone-community
+
+- op: replace
+  path: /spec/steps/0/image
+  value: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.40.2
```
chmeliik, line 8: We would need a mechanism to keep this image up to date as well as a mechanism to keep `task/git-clone-community/0.1/git-clone.yaml` in sync with with the "non-community" version

But:
* What rule prohibits using community images for the redhat deployment of Konflux? Can we have just one git-clone task?
* If not, can the community pipeline use the [community git-clone task](https://hub.tekton.dev/tekton/task/git-clone)?

task/init/0.2/init.yaml
gbenhaim: great idea! do you have the full reference for this skopeo image?

chmeliik: In the apply-tags task we use `registry.access.redhat.com/ubi9/skopeo:9.4-6@sha256:c4d70dec3eb0a0c831490192145ea25431fe04d1cf307f8d61e2d87adb41e7e3`

task/git-clone-community/0.1/patch.yaml
chmeliik: > I assumed that we need the tasks to be provided from our bundle to for passing enterprise-contract.

Hmm, I think that will soon put the build team in an impossible situation, having to maintain all the tasks for the konflux-ci community

chmeliik: How hard will it be to get rid of the git-clone-community task if we do create it?

I would want to get rid of it ASAP and get the openshift-pipelines team to release the git-init image to registry.access.redhat.com. But if someone starts using community pipelines by then, it could be really difficult.

PR #1037, opened 2024-05-23
prefetch-dependencies: Make the task fail if the input is empty

brunoapimentel:
This change aims to make it clear that a successful execution of the prefetch-dependencies task means that Cachi2 was executed successfully, as opposed to the task being executed as a no-op.

Accepting an empty input was initially allowed because, previously, the task execution was tied to hermetic builds, and we wanted to allow use cases in which there could be a hermetic build but no prefetch. Since this behavior is now changed, allowing the task to be a no-op is no longer necessary.

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
@@ -80,13 +80,6 @@ spec:
         mountPath: /mnt/trusted-ca
         readOnly: true
     script: |
-      if [ -z "${INPUT}" ]
-      then
-        # Confirm input was provided though it's likely the whole task would be skipped if it wasn't
-        echo "No prefetch will be performed because no input was provided for cachi2 fetch-deps"
-        exit 0
-      fi
-
       if [ "$DEV_PACKAGE_MANAGERS" = "true" ]; then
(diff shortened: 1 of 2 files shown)


chmeliik:
> Accepting an empty input was initially allowed because, previously, the task execution was tied to hermetic builds, and we wanted to allow use cases in which there could be a hermetic build but no prefetch. Since this behavior is now changed, allowing the task to be a no-op is no longer necessary.

Are you basing this on https://github.com/konflux-ci/build-definitions/pull/970? Changes to the pipeline definition do not propagated to existing users, so your change here would probably break everyone who is using prefetch-less hermetic builds.

You would have to create prefetch-dependencies 0.2

brunoapimentel:
> Are you basing this on #970? Changes to the pipeline definition do not propagate to existing users, so your change here would probably break everyone who is using prefetch-less hermetic builds.
> 
> You would have to create prefetch-dependencies 0.2

Yeah, I completely overlooked this detail. I'd be ok with doing a 0.2, if it weren't for the fact that we need to copy the whole task code in a separate yaml. I approached this as an easy fix, but I think I'll just close it for now, and we'll revisit this when we investigate how to enforce the prefetch to be executed for repos that use supported package managers.

brunoapimentel:
Closing as fixing this would require versioning the task. We will handle this at another opportunity.

PR #1039, opened 2024-05-23
Include tash in the Task bundle build

zregvart:
This includes the tool to generate the Trusted Artifacts Task variants in the `hack/build-and-push.sh` script, now if the task has a `recipe.yaml` file within its directory the tool will be used to regenerate the Trusted Artifact Task from the base Task.

Fixes https://issues.redhat.com/browse/EC-650

### DIFF
--- a/hack/generate-ta-tasks.sh
+++ b/hack/generate-ta-tasks.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o posix
+
+shopt -s globstar nullglob
+
+HACK_DIR="$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
+TASK_DIR="$(realpath "${HACK_DIR}/../task")"
+
+if ! command -v tash &> /dev/null; then
+  echo INFO: tash command is not available will download and use the latest version
+  tash_dir="$(mktemp -d)"
+  trap 'rm -rf ${tash_dir}' EXIT
+  tash_url=https://github.com/enterprise-contract/hacks/releases/download/latest/tash
+  echo INFO: downloading from ${tash_url} to "${tash_dir}"
+  curl --no-progress-meter --location --output "${tash_dir}/tash" "${tash_url}"
+  echo INFO: SHA256: "$(sha256sum "${tash_dir}/tash")"
+  chmod +x "${tash_dir}/tash"
+  tash() {
+    "${tash_dir}/tash" "$@"
+  }
+fi
+
+declare -i changes=0
+emit() {
+  if [ "${GITHUB_ACTIONS:-false}" == "true" ]; then
+    printf "::error file=%s,line=1,col=0::%s\n" "$1" "$2"
+  else
+    printf "INFO: \033[1m%s\033[0m %s\n" "$1" "$2"
+  fi
+  changes=$((changes + 1))
+}
+
+
+cd "${TASK_DIR}"
+for recipe_path in **/recipe.yaml; do
+    task_path="${recipe_path%/recipe.yaml}/$(basename "${recipe_path%/*/*}").yaml"
+    tash "${recipe_path}" > "${task_path}"
+    readme_path="${recipe_path%/recipe.yaml}/README.md"
+    "${HACK_DIR}/generate-readme.sh" "${task_path}" > "${readme_path}"
+    if ! git diff --quiet HEAD "${task_path}"; then
+        emit "task/${task_path}" "file is out of date and has been updated"
+    fi
+    if ! git diff --quiet HEAD "${readme_path}"; then
+        emit "task/${readme_path}" "file is out of date and has been updated"
+    fi
+done
+
+if [[ ${changes} -gt 0 ]]; then
+  if [ "${GITHUB_ACTIONS:-false}" == "true" ]; then
+    exit 1
+  else
+    printf "INFO: \033[1mMake sure to include the regenerated files in your changeset\033[0m\n"
+  fi
+fi
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -2,13 +2,13 @@
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
+  name: buildah-oci-ta
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio, hacbs
   labels:
     app.kubernetes.io/version: "0.1"
-    build.appstudio.redhat.com/build_type: "docker"
-  annotations:
-    tekton.dev/pipelines.minVersion: "0.12.1"
-    tekton.dev/tags: "image-build, appstudio, hacbs"
-  name: buildah-oci-ta
+    build.appstudio.redhat.com/build_type: docker
 spec:
   description: |-
     Buildah task builds source code into a container image and pushes the image into container registry using buildah tool.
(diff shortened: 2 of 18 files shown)


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -16,87 +16,96 @@ spec:
...
+  - description: The counting of Java components by publisher in JSON format
+    name: SBOM_JAVA_COMPONENTS_COUNT
     type: string
-  - name: JAVA_COMMUNITY_DEPENDENCIES
-    description: The Java dependencies that came from community sources such as Maven central.
+  - description: The Java dependencies that came from community sources such as Maven
+      central.
+    name: JAVA_COMMUNITY_DEPENDENCIES
   stepTemplate:
+    computeResources: {}
```
lcarva, line 108: NIT: It would be nice to just omit this attribute.

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -1,13 +1,13 @@
----
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio, hacbs
+  creationTimestamp: null
```
lcarva, line 7: NIT: It would be nice to omit this.

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -183,7 +193,7 @@ spec:
       sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
 
       # Setting new namespace to run buildah - 2^32-2
-      echo 'root:1:4294967294' | tee -a /etc/subuid >> /etc/subgid
+      echo 'root:1:4294967294' | tee -a /etc/subuid >>/etc/subgid
```
lcarva, line 196: NIT: I do prefer the whitespace, but it's not a big deal if it's much of an effort to keep those in.

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -274,29 +284,25 @@ spec:
...
       name: varlibcontainers
-    - mountPath: "/entitlement"
+    - mountPath: /entitlement
       name: etc-pki-entitlement
     workingDir: /var/workdir
-
-  - name: sbom-syft-generate
+  - computeResources: {}
     image: quay.io/redhat-appstudio/syft:v0.105.1@sha256:1910b829997650c696881e5fc2fc654ddf3184c27edb1b2024e9cb2ba51ac431
-    # Respect Syft configuration if the user has it in the root of their repository
```
lcarva, line 297: NIT: Any chance we can keep the comment?

lcarva (approved):
This is great!

appstudio-utils/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN curl -L https://github.com/tektoncd/cli/releases/download/v0.32.2/tkn_0.32.2
 RUN curl -L https://github.com/sigstore/rekor/releases/download/v0.5.0/rekor-cli-linux-amd64 -o /usr/bin/rekor-cli && chmod +x /usr/bin/rekor-cli
 RUN curl -L https://github.com/open-policy-agent/conftest/releases/download/v0.32.0/conftest_0.32.0_Linux_x86_64.tar.gz | tar -xz --no-same-owner -C /usr/bin
 RUN curl -L https://github.com/enterprise-contract/ec-cli/releases/download/snapshot/ec_linux_amd64 -o /usr/bin/ec && chmod +x /usr/bin/ec && ec version
+ADD --chmod=755 --chown=0:0 https://github.com/enterprise-contract/hacks/releases/download/latest/tash /usr/local/bin
```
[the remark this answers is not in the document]
zregvart: Separate repository? Easily done, though I was also hoping the tool will not be that long lived...

hack/build-and-push.sh
```diff
@@ -7,6 +7,20 @@ QUAY_ORG=redhat-appstudio-tekton-catalog
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 WORKDIR=$(mktemp -d --suffix "-$(basename "${BASH_SOURCE[0]}" .sh)")
 
+if ! command -v tash &> /dev/null; then
+  echo INFO: tash command is not available will download and use the latest version
```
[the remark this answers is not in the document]
zregvart: I agree, I did not want to end up troubleshooting and assisting folk in getting that latest version including the fix. And I'm sure there will be fixes as we progress here. I also considered moving the code here and using `go run` but I also use go 1.22 to build so that might be an issue for some folk (not sure?). I could expand this to prompt if the user wants the tool to be downloaded/installed and include a digest or similar.
This will be run only locally, the CI has the latest version included in the appstudio-util image...

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -1,13 +1,13 @@
----
```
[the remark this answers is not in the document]
zregvart: I should upload my wonderful tkn fmt tool, I was making some refinements and I think I broke something. With it this would be `diff -u <(tkn fmt <(curl -sL $MAIN)) <(tkn fmt <(curl -sL $PR))` or something along those lines...

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
zregvart: Yeah I was a bit puzzled why it is there, but I didn't really want to address it, I guess it might be missing `omitempty`? I'll look into it...

zregvart:
Thanks for the suggestions I'll look into addressing them, other reviews welcome...

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
zregvart: The issue here is that the YAML is unmarshalled to a golang struct and comments are lost, I missed this because the test code also does this, this might take some tinkering

hack/build-and-push.sh
```diff
@@ -134,7 +162,12 @@ do
         yq e "$sub_expr_1" -i "${filename}"
         yq e "$sub_expr_2" -i "${filename}"
     done
-done
+done < <(find task/*/*/ -maxdepth 0 -type d | awk -F '/' '{ print $0, $2, $3 }')
+
+if [[ ${errors} -gt 0 ]]; then
+  echo "ERROR: Found differences between base and OCI-TA Task variants, run hack/build-and-push.sh locally and include the regenerated files in your changeset"
```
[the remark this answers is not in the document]
chmeliik: (this is conceptually the same as the `.github/workflows/check-buildah-remote.yaml` workflow)

.tekton/scripts/check-task-pipeline-bundle-repos.sh
```diff
@@ -22,7 +22,7 @@ echo "Checking existence of task and pipeline bundle repositories ..."
 
 # tasks
 for task_dir in $(find task/*/*/ -maxdepth 0 -type d); do
-    if [ ! -f $task_dir/kustomization.yaml ]; then
+    if [[ ! -f $task_dir/kustomization.yaml || ! -f $task_dir/recipe.yaml ]]; then
       task_name=$(oc apply --dry-run=client -f "$task_dir/*.yaml" -o jsonpath='{.metadata.name}')
     else
       task_name=$(oc apply --dry-run=client -k "$task_dir" -o jsonpath='{.metadata.name}')
```
chmeliik, line 28: Not sure I understand the logic here

```
if the task directory has neither kustomization.yaml nor recipe.yaml:
    read task name directly from task yaml
else
    run kustomize and read task name from the result
```
I think the decision should still be based only on the existence of kustomization.yaml, no?

chmeliik (commented):
Nice, no more manual syncing of multiple versions of the same task :tada:

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
zregvart: The issue with timestamps is that they're using [zero value](https://pkg.go.dev/time#Time.IsZero) of `time.Time`, and the custom Marshaller implementation converts that value to `null`, there is no option to omit `null` values in golang's json. New version of tkn-fmt handles this

appstudio-utils/Dockerfile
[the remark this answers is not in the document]
zregvart: Decided to be lazy and work with hacks

hack/build-and-push.sh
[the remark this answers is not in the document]
zregvart: One small affordance I did is to do a sha256sum over the downloaded file...

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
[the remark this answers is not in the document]
zregvart: Dropping a link to https://github.com/zregvart/tkn-fmt

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
zregvart: Done by the shell parser/formatter and the https://pkg.go.dev/mvdan.cc/sh/v3@v3.8.0/syntax#SpaceRedirects doesn't help it seems

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
zregvart: I had a try of restoring comments, the issue is that comments that are part of sequences loose their position as the tool adds or removes elements from the sequence, i.e. a comment on `spec.steps[n]` can't be guaranteed to be placed on the same place, and the option of tracking the comments by each item's identifier (i.e. step name) seemed like a lot of work. The attempt I had added comments to wrong places and for whatever reason (read: implementation bug on my side) did not restore this comment in particular. So the lazy option is to loose the comments, at least for now.

hack/build-and-push.sh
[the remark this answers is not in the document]
zregvart: Yeah, reworked this to run similar to the buildah-remote

.tekton/scripts/check-task-pipeline-bundle-repos.sh
zregvart: Dropped this logic altogether, not sure if it made sense before, I think it might be leftover from a version I did when the generated files were not versioned...

zregvart:
I think I addressed all the comments, please do have another look. Notable differences from the previous version:
 * TA variant generation is no longer part of `hack/build-and-push.sh` but done via `hack/generate-ta-tasks.sh`, there is now a new workflow to check if the files need to be regenerated
 * the Task YAML files are now tkn-fmt-ed, so diffing base tasks is best done by installing https://github.com/zregvart/tkn-fmt and doing something like `diff <(tkn fmt task.yaml) task-oci-ta.yaml`
 * YAML file comments are difficult to maintain, so the lazy option of ignoring them has been implemented (though this was the same in the previous version)

chmeliik (commented):
buildah-remote-oci-ta seems to be missing the recipe.yaml, otherwise LGTM :taco:

chmeliik:
> buildah-remote-oci-ta seems to be missing the recipe.yaml, otherwise LGTM 🌮

Oh wait, that one is generated from buildah-oci-ta, isn't it

chmeliik (approved):

hack/generate-ta-tasks.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o posix
```
chmeliik, line 6: Just out of curiosity, why do you enable the posix option?

zregvart:
> buildah-remote-oci-ta seems to be missing the recipe.yaml, otherwise LGTM 🌮

It is generated using `hack/generate-buildah-remote.sh` so that's why it is not there.

hack/generate-ta-tasks.sh
zregvart: I think it was because of the exit code from subshells (or something akin to that), might be required in the previous version when this was included in `hack/build-and-push.sh`.

There is a list of changes in behavior when in the posix mode here: https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html

zregvart:
Looks like one of the e2e tests timed out:

```
Summarizing 1 Failure:
  [INTERRUPTED] [build-service-suite Build templates E2E test] HACBS pipelines [It] should eventually finish successfully for component with Git source URL https://github.com/cachito-testing/pip-e2e-test [build, build-templates, HACBS, pipeline, build-templates-e2e, source-build-e2e]
  /github.com/konflux-ci/e2e-tests/tests/build/build_templates.go:207

Ran 24 of 607 Specs in 2327.058 seconds
FAIL! - Interrupted by User -- 23 Passed | 1 Failed | 36 Pending | 547 Skipped
--- FAIL: TestE2E (2327.07s)
FAIL
```

/retest

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -16,43 +16,45 @@ spec:
     When [Java dependency rebuild](https://redhat-appstudio.github.io/docs.stonesoup.io/Documentation/main/cli/proc_enabled_java_dependencies.html) is enabled it triggers rebuilds of Java artifacts.
```
stuartwdouglas, line 16: This code is now generated from two different places, as the remote task generation is expecting to generate this from the buildah-oci-ta.

zregvart: From what I understand the `hack/generate-buildah-remote.sh` script will generate `task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml`, there is no `recipe.yaml` in `task/buildah-remote-oci-ta/0.1` so tash and the `hack/generate-ta-tasks.sh` script will not try to generate this file.

lcarva:
e2e tests keep failing with:

```
[e2e-tests : e2e-test] • [FAILED] [660.100 seconds]
[e2e-tests : e2e-test] [build-service-suite Build templates E2E test] HACBS pipelines [BeforeAll] triggers PipelineRun for symlink component with source URL https://github.com/redhat-appstudio-qe/devfile-sample-python-basic.git [build, build-templates, HACBS, pipeline, build-templates-e2e, source-build-e2e]
[e2e-tests : e2e-test]   [BeforeAll] /github.com/konflux-ci/e2e-tests/tests/build/build_templates.go:96
[e2e-tests : e2e-test]   [It] /github.com/konflux-ci/e2e-tests/tests/build/build_templates.go:184
[e2e-tests : e2e-test] 
[e2e-tests : e2e-test]   Timeline >>
[e2e-tests : e2e-test]   [FAILED] in [BeforeAll] - /github.com/konflux-ci/e2e-tests/tests/build/build_templates.go:135 @ 06/11/24 15:45:01.939
```

lcarva:
I wonder if this needs a rebase now that HAS has been disabled?

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
stuartwdouglas: ok, I missed that it skipped this one, looks good to me.

I have a PR open at https://github.com/konflux-ci/build-definitions/pull/1068 that moves this generator to the build-definitions repo.

zregvart: Yeah I saw that, I think we'll move this from ec/hacks here as well.

PR #1040, opened 2024-05-23
Add commit timestamp

SiddharthRajaraman:


zregvart (approved):
LGTM

chmeliik (commented):
Copying from Slack:

> acmiel
> Not really a fan of adding that result to the clone task. If we let the clone task be responsible for returning git metadata, soon it will be returning dozens of results to satisfy everyone's use case. (Unless it hits the Tekton result size limit before then.)
>
> I'd prefer a generic git metadata task that lets users specify exactly what data they want, or that everyone make their own custom task

PR #1042, opened 2024-05-28
Pass --all when building manifests

stuartwdouglas:
This means the task can also take manifest lists as an input, which is useful when dealing with OCI artifacts.

### DIFF
--- a/task/build-image-manifest/0.1/build-image-manifest.yaml
+++ b/task/build-image-manifest/0.1/build-image-manifest.yaml
@@ -45,6 +45,8 @@ spec:
       value: $(params.IMAGE)
     - name: TLSVERIFY
       value: $(params.TLSVERIFY)
+    - name: COMMIT_SHA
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Could you explain why we want this?

ralphbean:
Checking the help for `buildah manifest add`:
```
      --all                           add all of the list's images if the image is a list
```

I suppose then that this helps when making an image index of image indices.

And, although I wasn't a part of the development of this, I can imagine what it's for. For [parts of the instructlab project](https://github.com/containers/ai-lab-recipes/), they want to build disk images with https://github.com/konflux-ci/build-definitions/pull/1059, where (I'm guessing) each build task produces an image index, which then needs to be concatenated onto a longer image index here.

chmeliik (approved):

ralphbean:
Thanks!

PR #1043, opened 2024-05-29
Fix non-hermetic OCI TA builds

stuartwdouglas:


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -267,7 +267,7 @@ spec:
       echo $container > /var/workdir/container_name
 
       # Save the SBOM produced by Cachi2 so it can be merged into the final SBOM later
-      if [ -d "/var/workdir/cachi2" ]; then
+      if [ -f "/var/workdir/cachi2/cachi2.env" ]; then
```
arewm, line 270: If we are checking for files, should we not be checking for `output/bom.json` instead of `cachi2.env`?

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -359,7 +359,7 @@ spec:
   - name: merge-cachi2-sbom
     image: quay.io/redhat-appstudio/cachi2:0.7.0@sha256:1fc772aa3636fd0b43d62120d832e5913843e028e8cac42814b487c3a0a32bd8
     script: |
-      if [ -d "/var/workdir/cachi2" ]; then
+      if [ -f "/var/workdir/cachi2/cachi2.env" ]; then
```
arewm, line 362: Similarly, should we just check for the existence of `sbom-cachi2.json` here?

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
stuartwdouglas: I saw the build fail due to a lack of /cachi2/cachi2.env in the dockerfile, so that was the file I checked for. Would it make any functional difference?

chmeliik: Does `output/bom.json exists == cachi2.env exists`? I  *think* so, but why take the risk?

+1 to Andrew's suggestion, let's check the existence of the file we care about.

chmeliik (commented):
Please update buildah.yaml as well, otherwise https://github.com/konflux-ci/build-definitions/pull/1039 will revert your changes I think

chmeliik:
Could you explain in the commit message why non-hermetic OCI TA builds were broken and how this fixes them?

arewm (approved):
LGTM, thanks for the changes!

PR #1041, opened 2024-05-27
Start pushing all images to quay.io/konflux-ci

chmeliik:
[STONEBLD-2339](https://issues.redhat.com//browse/STONEBLD-2339)

All the images published from this repo will now be pushed to `quay.io/konflux-ci/*`. Some will continue getting pushed to their old location.

| Old (continues getting updates?) | New |
| --- | --- |
| ✅ `redhat-appstudio-tekton-catalog/task-*` | `konflux-ci/tekton-catalog/task-*` |
| ✅ `redhat-appstudio-tekton-catalog/pipeline-*` | `konflux-ci/tekton-catalog/pipeline-*` |
| ✅ `redhat-appstudio-tekton-catalog/data-acceptable-bundles` | `konflux-ci/tekton-catalog/data-acceptable-bundles` |
| ❌ `redhat-appstudio-tekton-catalog/pull-request-builds` | `konflux-ci/pull-request-builds:(task\|pipeline)-*` |
| ❌ `redhat-appstudio/appstudio-utils` | `konflux-ci/appstudio-utils` |
| ❌ `redhat-appstudio/pull-request-builds:build-definitions-utils-*` | `konflux-ci/pull-request-builds:appstudio-utils-*` |

Also, `data-acceptable-bundles` will now contain bundle references for both `redhat-appstudio-tekton-catalog/task-*` and `konflux-ci/tekton-catalog/task-*`.

The full list of repos that we need created in <https://quay.io/konflux-ci>:

<details>
<summary>click me</summary>

```
tekton-catalog/task-acs-deploy-check
tekton-catalog/task-acs-image-check
tekton-catalog/task-acs-image-scan
tekton-catalog/task-apply-tags
tekton-catalog/task-buildah
tekton-catalog/task-buildah-10gb
tekton-catalog/task-buildah-20gb
tekton-catalog/task-buildah-24gb
tekton-catalog/task-buildah-6gb
tekton-catalog/task-buildah-8gb
tekton-catalog/task-buildah-oci-ta
tekton-catalog/task-buildah-remote
tekton-catalog/task-buildah-rhtap
tekton-catalog/task-build-image-manifest
tekton-catalog/task-clair-scan
tekton-catalog/task-clamav-scan
tekton-catalog/task-deprecated-image-check
tekton-catalog/task-download-sbom-from-url-in-attestation
tekton-catalog/task-ecosystem-cert-preflight-checks
tekton-catalog/task-fbc-related-image-check
tekton-catalog/task-fbc-validation
[...]

PR #1044, opened 2024-05-29
Add buildah-remote-oci-ta

stuartwdouglas:


stuartwdouglas:
Note that this is just an initial implementation, when all involved images are multi-platform we can likely do a much more efficient implementation that does not rely on rsync.

brianwcook:
@ralphbean  should the build for this be pushed to https://quay.io/organization/redhat-appstudio-tekton-catalog or to the new quay.io Konflux org?

stuartwdouglas:
We should probably just wait until https://github.com/konflux-ci/build-definitions/pull/1041 is in, and then only push it to konflux-ci

brianwcook:
1041 is in

arewm (approved):

PR #1045, opened 2024-05-29
Support matrix results in build-image-manifest

zregvart:
Using matrix[1] to fan out TaskRuns on a list of parameter combinations greatly improves the maintainability of the Pipeline definition. For example adding or removing support for a platform is a couple of lines changed, compared to maintaining several copies of the task.

The results of the matrix-run Tasks can only be referenced in whole, that is as an array of results that contains all the results of the single result of one of the matrix-run Tasks.

In our case we have the `IMAGE_URL` and `IMAGE_DIGEST` results from the single buildah Task invocation, and if run as a matrix the results can be obtained by accessing the `IMAGE_DIGEST[*]` array and the `IMAGE_DIGEST[*]` array separately.

Given that, when aggregating the results for the `IMAGES` parameter of the build-image-manifest Task, some of the elements might not contain the digests.

For example if the parameters are passed as:

    - $(tasks.build-container-amd64.results.IMAGE_URL)@$(tasks.build-container-amd64.results.IMAGE_DIGEST)
    - $(tasks.build-container-multiarch.results.IMAGE_URL[*])
    - $(tasks.build-container-multiarch.results.IMAGE_DIGEST[*])

The variable substitution would provide the args value as:

    - registry.io/repository/image-amd64@sha256:...
    - registry.io/repository/image-arm64
    - registry.io/repository/image-ppc64le
    - ...
    - @sha256:[arm64]
    - @sha256:[ppc64le]
    - ...

This change tries to support that case. When an argument is encountered that does not contain the digest, the digest is read from the trailing list of digests. The order of the elements from the results of the matrix-run Tasks is deterministic -- sorted by TaskRun name; so given
the convention of digests following the image references this should be fairly safe.

[1] https://tekton.dev/docs/pipelines/matrix/# Before you complete this pull request ...

### DIFF
--- a/task/build-image-manifest/0.1/build-image-manifest.yaml
+++ b/task/build-image-manifest/0.1/build-image-manifest.yaml
@@ -69,17 +69,62 @@ spec:
 
       sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
 
-      buildah manifest create "$IMAGE"
-      for i in $@
-      do
-        TOADD="$i"
-        if [[ $(echo $i | tr -cd ":" | wc -c) == 2 ]]; then
-          #we need to remove the tag, and just reference the digest
-          #as tag + digest is not supported
-          TOADD="$(echo $i | cut -d: -f1)@sha256:$(echo $i | cut -d: -f3)"
+      buildah manifest create "${IMAGE}"
+
+      # collect arguments in two arrays, one containing images with digests and
+      # one containing only digests, it is expected that all images without
+      # digests are followed with their digests in the same exact order, this
+      # aligns with how Tekton results from matrix Tasks are provided, e.g.
+      # given args as:
+      # - $(tasks.build-container-amd64.results.IMAGE_URL)@$(tasks.build-container-amd64.results.IMAGE_DIGEST)
+      # - $(tasks.build-container-multiarch.results.IMAGE_URL[*])
+      # - $(tasks.build-container-multiarch.results.IMAGE_DIGEST[*])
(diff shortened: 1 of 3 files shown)


zregvart:
cc @stuartwdouglas

task/build-image-manifest/0.1/build-image-manifest.yaml
```diff
@@ -69,17 +69,62 @@ spec:
...
-          TOADD="$(echo $i | cut -d: -f1)@sha256:$(echo $i | cut -d: -f3)"
+      buildah manifest create "${IMAGE}"
+
+      # collect arguments in two arrays, one containing images with digests and
+      # one containing only digests, it is expected that all images without
+      # digests are followed with their digests in the same exact order, this
+      # aligns with how Tekton results from matrix Tasks are provided, e.g.
+      # given args as:
+      # - $(tasks.build-container-amd64.results.IMAGE_URL)@$(tasks.build-container-amd64.results.IMAGE_DIGEST)
+      # - $(tasks.build-container-multiarch.results.IMAGE_URL[*])
```
stuartwdouglas, line 80: Can we just assume that all the images are in the same repo as the output image? Especially if we are using a matrix build?

If so we could just take the digests, and combine this with the output-image repo to get the full image.

zregvart: I would be worried that the blobs might get garbage collected in that case.

zregvart:
@stuartwdouglas @konflux-ci/build-maintainers what do you think would be a way forward here?

zregvart:
Closing, perhaps #1111 would gather some feedback

PR #1048, opened 2024-05-30
Use Konflux git-clone image

gbenhaim:
A git-clone image built by Konflux to be used in the git-clone task.


gbenhaim:
the image is missing the `find` command. I'm going to update it.

PR #1054, opened 2024-06-05
fix(KFLUXBUGS-1328): pull secrets in rpm-ostree

scoheb:
- copy functionality used in buildah-remote task

### DIFF
--- a/task/rpm-ostree/0.1/rpm-ostree.yaml
+++ b/task/rpm-ostree/0.1/rpm-ostree.yaml
@@ -121,6 +121,7 @@ spec:
       ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp"
 
       rsync -ra $(workspaces.source.path)/ "$SSH_HOST:$BUILD_DIR/workspaces/source/"
+      rsync -ra "$HOME/.docker/" "$SSH_HOST:$BUILD_DIR/.docker/"
       cat >scripts/script-build.sh <<'REMOTESSHEOF'
       #!/bin/sh
       set -o verbose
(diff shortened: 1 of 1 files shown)


chmeliik (approved):
LGTM, thanks :+1:

task/rpm-ostree/0.1/rpm-ostree.yaml
```diff
@@ -173,6 +174,7 @@ spec:
           -e COMMIT_SHA="$COMMIT_SHA" \
           --rm \
           -v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z" \
+          -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
```
cgwalters, line 177: I'm a bit confused by this one, is it really necessary?  It seems to expose the pull secret to the container itself, but I don't think we need that here FWIW.

chmeliik: Oh good catch, this is a different use case than the buildah-remote task. This one wants to use the credentials to pull the BUILDER_IMAGE itself, and there's no need to pass the credentials to that image again

I think we need to rsync the docker credentials to the $HOME of the VM user

cgwalters:
There's a more medium term thing here that I think we can actually kill the rpm-ostree task in favor of buildah-remote.

task/rpm-ostree/0.1/rpm-ostree.yaml
stuartwdouglas: You need to either sync the credentials to the $HOME/.docker dir, or explicitly tell podman which credentials file to use.

stuartwdouglas (changes requested):

lmilbaum:
This PR is a blocker for a downstream MR. Would you be able to prioritize this?

chmeliik:
looks abandoned, can we close this?

scoheb:
Yep. We can close.

PR #1056, opened 2024-06-05
fix(STONEBLD-2496, KFLUXBUGS-1317) Ensure that the base image digest result is present

arewm:
If there is no base image (i.e. if the image is only FROM scratch), then the file will not be present and therefore step-create-base-images-sbom will fail.

Resolves: https://issues.redhat.com/browse/STONEBLD-2496
Resolves: https://issues.redhat.com/browse/KFLUXBUGS-1317


arewm:
This PR will need to modify the code added from https://github.com/konflux-ci/build-definitions/pull/1044 after that merges.

chmeliik (approved):
LGTM, thanks!

PR #1062, opened 2024-06-07
KFLUXBUGS-1338 fix workspace path

stuartwdouglas:


task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -189,6 +189,8 @@ spec:
       fi
 
       rsync -ra $(workspaces.source.path)/ "$SSH_HOST:$BUILD_DIR/workspaces/source/"
+      rsync -ra /entitlement/ "$SSH_HOST:$BUILD_DIR/volumes/etc-pki-entitlement/"
+      rsync -ra /mnt/trusted-ca/ "$SSH_HOST:$BUILD_DIR/volumes/trusted-ca/"
```
chmeliik, line 193: Is it intentional that `buildah-remote` mounts the trusted-ca but `buildah-remote-oci-ta` doesn't?

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -363,8 +365,8 @@ spec:
       buildah images
       buildah tag localhost/rhtap-final-image "$IMAGE"
       container=$(buildah from --pull-never "$IMAGE")
-      buildah mount "$container" | tee /var/workdir/container_path
-      echo $container > /var/workdir/container_name
+      buildah mount "$container" | tee /workspace/container_path
+      echo $container > /workspace/container_name
```
chmeliik, line 369: This one uses `syft dir:$(cat /var/workdir/container_path) --output cyclonedx-json=/var/workdir/sbom-image.json` though

task/buildah-remote/0.1/buildah-remote.yaml
stuartwdouglas: All of these changes are generated by changes in the generator in the multi-platform controller. 

We should probably consider moving that code to here, so that it is all in the right place. 

In this case the reason appears to be that the trusted-ca mount is missing from buildah-oci-ta, so the generator matches the original source.

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
stuartwdouglas: Ugh, I think these both need a change.

The /workspace thing is a tekton implementation detail, we should not be relying on it as it is not documented AFAIK. 

I will change both of these to use a /shared volume mount.

ralphbean (approved):

PR #1063, opened 2024-06-08
add entitlement support

brianwcook:
this PR ensures that entitlement certificates are copied over and used for multi-platform controller builds. It was tested with a Containerfile build using ubi9 that installed kernel-devel. Builds with and without entitlement secrets will work properly.


brianwcook:
dropping this because it has to be done in the generate step

PR #1065, opened 2024-06-10
Move shared to step template

stuartwdouglas:
This lets the volume sync properly for the oci ta remote task


mmorhun (approved):

zregvart:
Why not use the existing workdir volume? Also this seems to be causing issues, see https://github.com/konflux-ci/build-trusted-artifacts/pull/98/checks?check_run_id=25998888462 and https://github.com/enterprise-contract/golden-container/pull/591/checks?check_run_id=25887803018

PR #1068, opened 2024-06-11
STONEBLD-2525 Move remote task generator

stuartwdouglas:
This moves the remote task generator to the build-definitions repo. It means that changes to the generator can no go into the same commit as the corresponding remote task update.


go.mod
```diff
@@ -0,0 +1,86 @@
+module github.com/konflux-ci/build-definitions
```
chmeliik, line 1: Could you move the go.mod and go.sum to the remote-task-generator subdirectory?

chmeliik (commented):
Nice, this will feel much better :taco:

go.mod
zregvart: :+1: Yeah, that would help with EC-690, move of tash for #1039 in here.

stuartwdouglas: I might rename it to task-generator then, once this is in we can look at moving tash

stuartwdouglas:
Should be ready now

PR #1071, opened 2024-06-12
Don't use variable substitution in script block

zregvart:
This is considered to be a bad practice[1]. Also breaks trusted artifacts due to the script being reformatted and an extra space added around the parentheses.

[1] https://tekton.dev/docs/pipelines/tasks/#substituting-in-script-blocks

### DIFF
--- a/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
+++ b/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
@@ -175,6 +175,10 @@ spec:
           value: $(workspaces.basic-auth.bound)
         - name: WORKSPACE_BASIC_AUTH_DIRECTORY_PATH
           value: $(workspaces.basic-auth.path)
+        - name: COMMIT_PATH_RESULT
+          value: $(results.commit.path)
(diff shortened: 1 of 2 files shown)


lcarva:
Not sure about this. I believe the bad practice being called out is regarding using values from `$(params.X)` since a user could set that to whatever. `$(results.X.path)` is managed by Tekton. A Task user can't really exploit it by just running the Task. There are a **lot** of instances across the Tasks in this repo that use this pattern. I'd be cautious about making a fix at this scope if it's not really necessary.

Obviously, we need to fix the issue with the `git-clone-oci-ta.yaml` Task.

simonbaird:
Griping in general, not about this PR: Wtf does the bash formatter want to add that space? :wtf:

simonbaird:
I think we could merge this as a pragmatic solution then consider the other options later.

Do we think there are other tasks impacted apart from git-clone?

simonbaird (approved):

lcarva:
https://github.com/konflux-ci/build-definitions/pull/1074 supersedes this. The fix was to tweak the tash tool to not add the space.

PR #1075, opened 2024-06-14
Add tests for Tasks

lcarva:
This commit adds tests for some of the Tasks in this repo. The tests are implemented as Go tests. The test files reside alongside the Task definitions for convenience. See README.md for how to run them.

The tests execute against a Kubernetes cluster, including a kind[1] cluster.

[1] https://kind.sigs.k8s.io/


lcarva:
This is draft because I think a couple of things need to happen before merging this:

* [ ] Agreement that this is a good idea :smile: 
* [ ] Test against an OpenShift cluster. I only tried on a kind cluster. It works wonderfully there.
* [ ] Add a Task to the build Pipeline to execute the tests. Setting `TEST_KUBE_NAMESPACE` to `build-templates-e2e` might just work. We may also already have a repo we can use for `TEST_OCI_REPO`.
* [ ] Adjust the parallel parameter for the tests. We may have have to tone this down a bit when running in CI, but we should just use whatever the default value is in other cases.

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
```diff
@@ -0,0 +1,86 @@
...
+			matched, err := regexp.MatchString(fmt.Sprintf("^%s[a-f0-9]{64}$", expected), result.Value.StringVal)
+			require.NoError(t, err)
+			require.True(t, matched)
+			sourceArtifactResultFound = true
+		}
+	}
+	require.True(t, commitResultFound)
+	require.True(t, urlResultFound)
+	require.True(t, sourceArtifactResultFound)
+}
```
chmeliik, line 86: The tests are too verbose for my taste (though maybe that's just Go being Go)

If writing simple tests is more involved than something like this, I will not want to write them :smile: 

```
taskRun = makeTaskRun(taskSpec, params, workspaces)
run(taskRun)
assert(taskRun, expectedResults, expectedLogs)
```

I wonder if Go has something like https://github.com/samuelcolvin/dirty-equals so that you could actually just declare the expected results and wouldn't need all those `*found` booleans and switch-cases and a bunch of logic for verifying each individual result (does Gomega get close?). (And I also wonder if Go is the best choice for these tests)

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
```diff
@@ -0,0 +1,86 @@
...
+			},
+			TaskSpec: taskSpec,
+		},
+	}
+
+	k, err := kube.NewClient()
+	require.NoError(t, err)
+
+	require.NoError(t, k.Create(ctx, &taskRun))
+	defer k.Delete(ctx, &taskRun)
```
chmeliik, line 56: What does it look like when a test fails? One of the big problems with our e2e-tests is that most of the time, the error message is pretty much useless on its own. If we don't leave the TaskRun around to debug, dealing with failing tests will be hard

Maybe the tests could capture the TaskRun yaml, logs, events and output/save them somewhere for debugging?

chmeliik (commented):
Integration tests for individual tasks would be great :taco:

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
lcarva: Yeah. This is probably something that needs further refinement. My hope is that this is significantly easier to run than the current e2e tests. In the case of a failure, it should be very simple to re-run the test to reproduce the issue, and allow one to attach a debugger which is much better than whatever data is printed out.

I would like to try this as is and see if dumping all the additional information is really needed.

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
lcarva: We can definitely improve on this. I'm just a bit wary of providing too many levels of indirection that will actually make it harder to read the tests.

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
tkdchen: For such kind of tests, it should be fine if saving TaskRun logs only. PR authors should be responsible for the correction of the proposed changes. From my point of view, a tests pass in the CI is just one of the acceptance criteria for approval. So, if there is any error causing the failure, PR authors can always re-run the test by themselves and they will have everything for debug.

task/git-clone-oci-ta/0.1/git_clone_oci_ta_test.go
tkdchen: To be honest, for my taste, the assertions writing in Golang is not easier-to-read-and-understand than in Python. :smile:

These lines of assertions are ok to me. The assertion style choice depends on the preference of PR authors/task owners/maintenance team, etc. We can't force a unified assertion style for build-definitions, unless we want to. I think we can recommend or even force to use one assertion library in our teams and encourage custom task authors to use that.

tkdchen:
There is a story [KONFLUX-2735](https://issues.redhat.com/browse/KONFLUX-2735) for decentralizing the tasks. Not clear so far in my mind how suitable this mechanism is for the future possible change to build-definitions. Would a mechanism without compilation be more flexible?

lcarva:
> There is a story [KONFLUX-2735](https://issues.redhat.com/browse/KONFLUX-2735) for decentralizing the tasks. Not clear so far in my mind how suitable this mechanism is for the future possible change to build-definitions. Would a mechanism without compilation be more flexible?

@tkdchen, what do you mean by mechanism without compilation?

tkdchen:
> > There is a story [KONFLUX-2735](https://issues.redhat.com/browse/KONFLUX-2735) for decentralizing the tasks. Not clear so far in my mind how suitable this mechanism is for the future possible change to build-definitions. Would a mechanism without compilation be more flexible?
> 
> @tkdchen, what do you mean by mechanism without compilation?

I meant some way by using interpreted programming languages.

lcarva:
There's a renewed effort to approach this differently. Let's close this out as it has served its purpose.

PR #1082, opened 2024-06-20
Add CAP_SYS_ADMIN param to buildah task

ralphbean:
This instructs buildah to use --cap-add=SYS_ADMIN.

The capability is not generally safe and should be restricted by both cluster SecurityContextConstraints and supply chain policy for most situations.

The param is present for unusual situations where it is required.


chmeliik:
#1017 does it more generically (though it would need a slightly more complex policy check I suppose). I'd prefer that one if we're ready to go through with this

ralphbean:
You're right! I'm going to sponsor getting that one merged and in and I'll make sure to follow up with EC.

ralphbean:
I pushed a rebase of it at https://github.com/konflux-ci/build-definitions/pull/1084.

ralphbean:
Merging https://github.com/konflux-ci/build-definitions/pull/1084.

PR #1085, opened 2024-06-21
Add Checkton workflow

chmeliik:
This PR demonstrates [Checkton](https://github.com/chmeliik/checkton)'s features, such as...
* When a file is renamed, do not treat it as a completely new file. Report only new issues introduced in the PR
* When a file is copied (e.g. creating a new version of a task), report only new issues compared to the source file


chmeliik:
Adding for real in #1183

PR #1086, opened 2024-06-21
fix: expose image manifest digest of all copied artifacts

ralphbean:
See https://github.com/enterprise-contract/ec-cli/issues/1698

ec expects to validate a signature for every image manifest referred to
in a image index. By exposing this additional result, we induce chains
to sign each one.

### DIFF
--- a/task/oci-copy/0.1/oci-copy.yaml
+++ b/task/oci-copy/0.1/oci-copy.yaml
@@ -23,6 +23,8 @@ spec:
       name: IMAGE_DIGEST
     - description: Repository where the artifact was pushed
       name: IMAGE_URL
+    - description: URIs for all image manifests published, for signing
+      name: IMAGES
   stepTemplate:
     env:
       - name: OCI_COPY_FILE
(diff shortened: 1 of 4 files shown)


task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -100,14 +102,29 @@ spec:
...
+    - name: artifact-manifest-generate
+      image: quay.io/konflux-ci/yq:latest@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430
+      script: |
+        MANIFEST_DIGESTS=$(cat "$(workspaces.source.path)"/image-manifests | yq -r '.manifests[].digest')
+        for MANIFEST_DIGEST in $MANIFEST_DIGESTS; do
+          reference="$IMAGE@$MANIFEST_DIGEST, "
+          echo "Writing ${reference} to $(results.IMAGES.path)"
+          echo -n "${reference}" >> $(results.IMAGES.path)
+        done
+        echo -n "$IMAGE@$(cat $(results.IMAGE_DIGEST.path))" >> $(results.IMAGES.path)
```
chmeliik, line 123: This imposes a pretty nasty limit on the number of images that this task can copy (before the result breaks the 4kb size limit). If the general expectation is that this task will be used for copying, say, 10 artifacts, we should be good. But maybe EC could (should it?) support signed index images referencing unsigned manifests?

PR #1088, opened 2024-06-24
fix(kfluxbugs-1142) make related image check work with yaml and json

14rcole:
Use `opm render` to gather related images instead of `yq` and `jq`.  This streamlines the code and makes the related images check compatible with edge cases such as a `config.json` that contains yaml or a document with yaml and json nested within one another.


jsztuka (approved):
lgtm

kasemAlem (approved):

kasemAlem:
approved, 
just requires a rebase @14rcole

Josh-Everett (approved):

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -34,25 +34,20 @@ spec:
         trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
 
         FAILEDIMAGES=""
-        catalog="$(find $(workspaces.workspace.path)/hacbs/fbc-validation/ -name catalog.yaml)"
+        catalog="$(opm render configs/)"
```
dirgim, line 37: The previous version of this line checked the `fbc-validation` task workspace where the catalog image is expected to beextracted. 

This version seems to assume that the catalog files will be in this task's workspace which I don't think is right, if I'm not missing anything.

14rcole: That was a copy/paste error from local testing.  Not sure why e2e tests are passing.

dirgim: Ok, that makes more sense. Follow up question - should the name of the directory be `confdir` based on [this line](https://github.com/konflux-ci/build-definitions/blob/83a22dc82f7ae6ce8868ad2594aec8641884505f/task/fbc-validation/0.1/fbc-validation.yaml#L130)?

14rcole: I'm not sure, but the related images check was already using the configs directory.  Wouldn't we be seeing other failures if that was incorrect?

dirgim: In most cases the directory will be originally named `configs`. The issue is that it's not guaranteed because it can be named anything as long as it's specified in the `operators.operatorframework.io.index.configs.v1` label, hence why the directory name is extracted in [this line here](https://github.com/konflux-ci/build-definitions/blob/83a22dc82f7ae6ce8868ad2594aec8641884505f/task/fbc-validation/0.1/fbc-validation.yaml#L102C9-L102C19).

That's why the `fbc-validate` task copies the configuration directiory into a standardized directory named in [this line](https://github.com/konflux-ci/build-definitions/blob/83a22dc82f7ae6ce8868ad2594aec8641884505f/task/fbc-validation/0.1/fbc-validation.yaml#L130).

14rcole: Got it, updated to use `confdir`

dirgim: Thanks, looks good!

dirgim (approved):
LGTM

PR #1089, opened 2024-06-24
Refactor oci-copy task

ralphbean:
Update this to produce a single OCI artifact with a layer for each file, rather than an image index with an OCI artifact for each file.


ralphbean:
FYI, I've got a successful test of this with a user repo at https://gitlab.com/redhat/rhel-ai/models/granite-8b-code-instruct/-/merge_requests/107

mmorhun (approved):

PR #1091, opened 2024-06-24
fix: remove duplicate volumeMounts from oci-copy-oci-ta

14rcole:
The oci-copy-oci-ta task had two volumeMounts with the same name and mounting location, which would cause an error when the task was applied to a cluster.  This change removes the duplicate

# Before you complete this pull request ...

Look for any open pull requests in the repository with the title "e2e-tests update" and 
see if there are recent e2e-tests updates that will be applicable to your change.

PR #1093, opened 2024-06-25
fix: drop use of /var/workdir in oci-copy task

ralphbean:
This is another take on https://github.com/konflux-ci/build-definitions/pull/1091, but it starts by modifying the oci-copy task first and then running hack/generate-ta-tasks.sh.


chmeliik (approved):

chmeliik (commented):
Just noticed that oci-copy.yaml still declares
```yaml
  volumes:
    - emptyDir: {}
      name: workdir
```
^ maybe that can be dropped now?

chmeliik:
@ralphbean could you rebase this PR?

ralphbean:
Checkton is unhappy. I'll separate those changes into a separate commit.

PR #1099, opened 2024-06-27
feat: add buildah param to inject secrets into the build

ralphbean:
The secret is made available only to RUN lines that are invoked with `--mount`, like this:

```
RUN --mount=type=secret,id={secret-name} cat /run/secrets/{secret-name}
```

See https://docs.podman.io/en/latest/markdown/podman-build.1.html#secret-id-id-src-path

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -67,6 +67,11 @@ spec:
         hours, days, and weeks, respectively.
       type: string
       default: ""
+    - name: OPTIONAL_SECRET
+      description: Name of a secret which will be made available to the build
+        with 'buildah build --secret' at /run/secrets/$OPTIONAL_SECRET
+      type: string
+      default: ""
     - name: PREFETCH_INPUT
       description: In case it is not empty, the prefetched content should
         be made available to the build.
@@ -135,6 +140,10 @@ spec:
       secret:
         optional: true
         secretName: $(params.ENTITLEMENT_SECRET)
+    - name: optional-secret
+      secret:
+        optional: true
+        secretName: $(params.OPTIONAL_SECRET)
     - name: shared
       emptyDir: {}
     - name: trusted-ca
@@ -172,6 +181,8 @@ spec:
         value: $(params.SQUASH)
       - name: SQUASH_ALL
         value: $(params.SQUASH_ALL)
+      - name: OPTIONAL_SECRET
+        value: $(params.OPTIONAL_SECRET)
       - name: STORAGE_DRIVER
         value: vfs
       - name: TARGET_STAGE
@@ -206,6 +217,8 @@ spec:
           name: varlibcontainers
         - mountPath: /entitlement
           name: etc-pki-entitlement
+        - mountPath: /optional-secret
+          name: optional-secret
         - mountPath: /mnt/trusted-ca
           name: trusted-ca
           readOnly: true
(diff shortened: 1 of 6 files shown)


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -314,6 +327,13 @@ spec:
           echo "Adding the entitlement to the build"
         fi
 
+        OPTIONAL_SECRET_PATH="/optional-secret"
+        if [ -d "$OPTIONAL_SECRET_PATH" ]; then
+          cp -r --preserve=mode "$OPTIONAL_SECRET_PATH" /tmp/optional-secret
+          BUILDAH_ARGS+=("--secret=id=${OPTIONAL_SECRET},src=/tmp/optional-secret")
+          echo "Adding the secret ${OPTIONAL_SECRET} to the build, available at /run/secrets/${OPTIONAL_SECRET}"
+        fi
```
chmeliik, line 335: Technically, this allows the user to inject arbitrary content into the build. Although the same could be said for every volume mount unless the content comes from the output of a trusted task. The likelihood of this being misused seems slightly higher than for the other secrets, though still fairly low.

I'm not _too_ concerned over adding this, just raising an observation

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -125,6 +130,10 @@ spec:
       secret:
         optional: true
         secretName: $(params.ENTITLEMENT_SECRET)
+    - name: optional-secret
+      secret:
+        optional: true
+        secretName: $(params.OPTIONAL_SECRET)
```
chmeliik, line 136: I suspect the test failures may be caused by 

```yaml
    - name: optional-secret
      secret:
        optional: true
        secretName: ""  # the default $(params.OPTIONAL_SECRET) is ""
```

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -67,6 +67,11 @@ spec:
         hours, days, and weeks, respectively.
       type: string
       default: ""
+    - name: OPTIONAL_SECRET
+      description: Name of a secret which will be made available to the build
+        with 'buildah build --secret' at /run/secrets/$OPTIONAL_SECRET
```
chmeliik, line 72: Btw, what is the reasoning behind the name OPTIONAL_SECRET?

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
ralphbean: Attempted providing a default value in ef5f5de.

chmeliik (approved):
LGTM

For posterity, could you describe what immediate use case this is meant to address?

mkosiarc (approved):

ralphbean:
Yes, it's amazing. In [this line of this Containerfile](https://github.com/containers/ai-lab-recipes/blob/aee1c80b7c1444a40ab45950f30b79b375cf7bfc/training/nvidia-bootc/Containerfile#L186) the instructlab team is adding a "[physically bound](https://github.com/containers/bootc/issues/644)" container, embedding it in another container.

Upstream, that works as is. In private environments, users need auth to pull that inner image during the build process of the outer image.

This is incompatible with hermetic mode.

I can imagine a user might someday want to use this in combination with hermetic mode in order to decrypt some content, encrypted at rest, during their build. But, that is an imagined use case - not real.

PR #1103, opened 2024-07-01
Incorporate auth param for oci-copy task

ralphbean:
The use case for this is copying artifacts into the OCI registry from locations that require some authentication, in particular https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -13,6 +13,11 @@ spec:
   description: Given a file in the user's source directory, copy content from
     arbitrary urls into the OCI registry.
   params:
+    - name: BEARER_TOKEN_SECRET_NAME
+      description: Name of a secret which will be made available to the build
+        as an Authorization header
+      type: string
+      default: ""
     - name: IMAGE
       description: Reference of the image we will push
       type: string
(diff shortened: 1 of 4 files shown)


task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -71,16 +75,33 @@ spec:
...
+
+        set -u
+
         for varfile in /var/workdir/vars/*; do
           echo "Reading $varfile"
           source $varfile
 
           echo "Downloading $OCI_SOURCE to $OCI_FILENAME"
-          curl --silent --show-error --location $OCI_SOURCE -o $OCI_FILENAME
+          curl "${CURL_ARGS[@]}" --silent --show-error --location $OCI_SOURCE -o $OCI_FILENAME
```
chmeliik, line 104: Not directly related to this PR, but I noticed we're not passing `--fail` to curl so we can end up writing an HTTP error response into the output file and proceeding. Although that would be caught by a (somewhat confusing) checksum mismatch

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
```diff
@@ -85,16 +90,33 @@ spec:
...
       script: |
-        set -eu
+        set -e
         set -o pipefail
 
+        CURL_ARGS=()
+        if [ -n "${BEARER_TOKEN}" ]; then
+          echo "Found bearer token. Using it for authentication."
+          CURL_ARGS+=(-H "Authorization: Bearer ${BEARER_TOKEN}")
+        else
```
chmeliik, line 108: It doesn't feel right to send the same Bearer token for all the requests considering that the oci-copy file can reference artifacts coming from multiple servers

If the huggingface token can also be used with HTTP Basic, maybe this task could use the same approach as the prefetch task (a .netrc file). Otherwise, IMO we at least need to document that the same token will be sent to all the servers

chmeliik: Or maybe the secret could be in a format like
```
huggingface.co: $token1
example.com: $token2
```

ralphbean: I spent some time looking today for basic auth options, and while the [hugging face docs](https://huggingface.co/docs/hub/security-tokens#how-to-use-user-access-tokens) say that "User Access Tokens can be used in place of a password to access the Hugging Face Hub ... with basic authentication.", I couldn't actually get it to work with curl - trying both the `-u username:password` form and the `~/.netrc` file.

I gave up, and proposed documentation in f61a705.

The specially formatted secret could work, but - I'm not sure it is worth it at this point. Also, introducing *two* new file formats for this task feels excessive.

task/oci-copy/0.1/oci-copy.yaml
ralphbean: Fixed in b0d7434.

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
chmeliik: Rather than a file format, I was thinking the secret would just contain multiple keys, one per hostname. The task would have to parse the hostname from each url and get the bearer token from `cat /tokens-secret/$hostname`

But documenting the limitations should be good enough for now

chmeliik (approved):

ralphbean:
@chmeliik thank you for the thoughtful reviews.

hongweiliu17:
@ralphbean The push plr https://github.com/konflux-ci/build-definitions/runs/27067818667 fails for this latest merged PR https://github.com/konflux-ci/build-definitions/commit/main

ralphbean:
Thanks @hongweiliu17! I just saw it too and made my way back here and found your comment. I'm attempting `/retest` of 12f31d38fcc241598c7abc3e054e0425ee61c0ac (latest on `main`) to analyze.

ralphbean:
Well, that seemed to work: https://github.com/konflux-ci/build-definitions/runs/27098685636

PR #1110, opened 2024-07-02
Iterate over additional secret keys, making all available

ralphbean:
Two bugs are fixed here.

First, the buildah `--secret` arg expects a *file*, not a directory, so every key/value pair in the kubernetes secrets mounted in needs to be supplied as a separate argument to buildah.

Second, buildah won't accept symlinks and all of the individual files in the mounted directory are symlinks. The `-L` option to `cp` addresses that by de-referencing them.

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -342,10 +342,10 @@ spec:
         ADDITIONAL_SECRET_TMP="/tmp/additional-secret"
         if [ -d "$ADDITIONAL_SECRET_PATH" ]; then
           cp -r --preserve=mode -L "$ADDITIONAL_SECRET_PATH" $ADDITIONAL_SECRET_TMP
-          for filename in $(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;); do
+          while read -r filename; do
             echo "Adding the secret ${ADDITIONAL_SECRET}/${filename} to the build, available at /run/secrets/${ADDITIONAL_SECRET}/${filename}"
             BUILDAH_ARGS+=("--secret=id=${ADDITIONAL_SECRET}/${filename},src=$ADDITIONAL_SECRET_TMP/${filename}")
-          done
+          done < <(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;)
         fi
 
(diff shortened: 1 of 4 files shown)


ralphbean:
Related PR: https://github.com/containers/ai-lab-recipes/pull/656

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -339,10 +339,13 @@ spec:
...
+        ADDITIONAL_SECRET_TMP="/tmp/additional-secret"
         if [ -d "$ADDITIONAL_SECRET_PATH" ]; then
-          cp -r --preserve=mode "$ADDITIONAL_SECRET_PATH" /tmp/additional-secret
-          BUILDAH_ARGS+=("--secret=id=${ADDITIONAL_SECRET},src=/tmp/additional-secret")
-          echo "Adding the secret ${ADDITIONAL_SECRET} to the build, available at /run/secrets/${ADDITIONAL_SECRET}"
+          cp -r --preserve=mode -L "$ADDITIONAL_SECRET_PATH" $ADDITIONAL_SECRET_TMP
+          for filename in $(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;); do
+            echo "Adding the secret ${ADDITIONAL_SECRET}/${filename} to the build, available at /run/secrets/${ADDITIONAL_SECRET}/${filename}"
+            BUILDAH_ARGS+=("--secret=id=${ADDITIONAL_SECRET}/${filename},src=$ADDITIONAL_SECRET_TMP/${filename}")
+          done
```
chmeliik, line 348: Technically a `while read` loop is more robust here (https://www.shellcheck.net/wiki/SC2044)

```suggestion
          while read -r filename; do
            echo "Adding the secret ${ADDITIONAL_SECRET}/${filename} to the build, available at /run/secrets/${ADDITIONAL_SECRET}/${filename}"
            BUILDAH_ARGS+=("--secret=id=${ADDITIONAL_SECRET}/${filename},src=$ADDITIONAL_SECRET_TMP/${filename}")
          done < <(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;)
```
(and this form still updates BUILDAH_ARGS in the current shell, unlike `find ... | while read ...` which would try to update them in a subshell)

chmeliik (commented):
LGTM, one bash suggestion

PR #1111, opened 2024-07-03
`IMAGE_REF` result from image building Tasks

zregvart:
This helps in the reuse of the results when using matrix feature of Tekton. Given that the concatenation of two results from matrix-spawned Tasks is not supported, e.g.

    $(tasks.build-container-multiarch.results.IMAGE_URL[*])@$(tasks.build-container-multiarch.results.IMAGE_DIGEST[*])

will not expand correctly.

This produces the image reference in full in the `IMAGE_REF` result, so the result from the matrix-spawned Tasks can be referenced using:

    $(tasks.build-container-multiarch.results.IMAGE_REF[*])

Reference: https://issues.redhat.com/browse/EC-654

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -35,6 +35,8 @@ spec:
   results:
     - name: IMAGE_DIGEST
       description: Digest of the artifact just pushed
+    - name: IMAGE_REF
+      description: Image reference of the built image
     - name: IMAGE_URL
       description: Repository where the artifact was pushed
     - name: SBOM_BLOB_URL
@@ -193,6 +195,7 @@ spec:
         RESULTING_DIGEST=$(oras resolve --registry-config auth.json "${IMAGE}")
         echo -n "$RESULTING_DIGEST" | tee "$(results.IMAGE_DIGEST.path)"
         echo -n "$IMAGE" | tee "$(results.IMAGE_URL.path)"
+        echo -n "${IMAGE}@${RESULTING_DIGEST}" >"$(results.IMAGE_REF.path)"
       computeResources:
         limits:
           memory: 1Gi
--- a/task/rpm-ostree/0.1/README.md
+++ b/task/rpm-ostree/0.1/README.md
@@ -1,0 +1,31 @@
+# rpm-ostree task
+
+RPM Ostree
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|IMAGE|Reference of the image rpm-ostree will produce.||true|
+|BUILDER_IMAGE|The location of the rpm-ostree builder image.|quay.io/redhat-user-workloads/project-sagano-tenant/ostree-builder/ostree-builder-fedora-38:d124414a81d17f31b1d734236f55272a241703d7|false|
+|CONTEXT|Path to the directory to use as context.|.|false|
+|IMAGE_FILE|The file to use to build the image||true|
+|TLSVERIFY|Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)|true|false|
+|IMAGE_EXPIRES_AFTER|Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.|""|false|
+|COMMIT_SHA|The image is built from this commit.|""|false|
+|PLATFORM|The platform to build on||true|
+|CONFIG_FILE|The relative path of the file used to configure the rpm-ostree tool found in source control. See https://github.com/coreos/rpm-ostree/blob/main/docs/container.md#adding-container-image-configuration|""|false|
+|HERMETIC|Determines if build will be executed without network access.|false|false|
+
+## Results
+|name|description|
+|---|---|
+|IMAGE_DIGEST|Digest of the image just built|
+|IMAGE_URL|Image repository where the built image was pushed|
+|IMAGE_REF|Image reference of the built image|
+|BASE_IMAGES_DIGESTS|Digests of the base images used for build|
+|SBOM_BLOB_URL|Reference, including digest to the SBOM blob|
+
+## Workspaces
+|name|description|optional|
+|---|---|---|
+|source|Workspace containing the source code to build.|false|
--- a/task/source-build/0.1/source-build.yaml
+++ b/task/source-build/0.1/source-build.yaml
@@ -29,6 +29,8 @@ spec:
       description: The source image url.
     - name: SOURCE_IMAGE_DIGEST
       description: The source image digest.
+    - name: IMAGE_REF
+      description: Image reference of the built image
   workspaces:
     - name: workspace
       description: The workspace where source code is included.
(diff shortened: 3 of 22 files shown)


task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
```diff
@@ -120,6 +122,7 @@ spec:
         IMAGE_INDEX_DIGEST=$(oras resolve --registry-config auth.json "${IMAGE}")
         echo -n "$IMAGE_INDEX_DIGEST" | tee "$(results.IMAGE_DIGEST.path)"
         echo -n "$IMAGE" | tee "$(results.IMAGE_URL.path)"
+        echo -n "${IMAGE}@${IMAGE_INDEX_DIGEST}" >"$(results.IMAGE_REF.path)"
```
simonbaird, line 125: Could use tee for consistency I guess.

zregvart: Not sure why? The information is already on stdout, why display it the second time concatenated?

task/rpm-ostree/0.1/README.md
```diff
@@ -0,0 +1,31 @@
+# rpm-ostree task
```
simonbaird, line 1: Is there a good reason to have this in the same commit?

zregvart: I regenerated all README.md's for the Tasks I changed. The better question might be why all the other authors did not bother to do that.

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
simonbaird: Yeah, that's true.

task/rpm-ostree/0.1/README.md
simonbaird: Ack, it makes sense, but consider explaining that in the commit message, or putting those changes in another commit.

arewm:
Is the intention of this to make it easier for users to onboard to a multi-arch flow or just to simplify the interface? If the former, then this process can also be simplified when we create a dedicated multi-arch Tekton Pipeline. The yaml files pushed out can look something similar to what I am using here for example: https://github.com/konflux-ci/buildah-container/pull/9/files.

task/rpm-ostree/0.1/README.md
zregvart: I have, it stays in the commit and the message remains the same.

zregvart:
@arewm the intention here is to enable the use of [matrix feature](https://tekton.dev/docs/pipelines/matrix/) which reduces the complexity of a multi-arch Pipeline definition. This was already presented on the May 30th demo call, you can lookup the recording.

task/buildah/0.1/buildah.yaml
```diff
@@ -488,6 +490,10 @@ spec:
 
       cat "$(workspaces.source.path)"/image-digest | tee $(results.IMAGE_DIGEST.path)
       echo -n "$IMAGE" | tee $(results.IMAGE_URL.path)
+      {
+        echo -n "${IMAGE}@"
+        cat "$(workspaces.source.path)/image-digest"
+      } > "$(results.IMAGE_REF.path)"
```
chmeliik, line 496: The Tekton result size limit is a ticking time bomb, I don't think we can add any new result to the buildah task without getting rid of BASE_IMAGES_DIGESTS https://issues.redhat.com/browse/STONEBLD-2326

The last time we added a `${repo}@${digest}` type of result, it was enough to break some builds https://github.com/konflux-ci/build-definitions/pull/654

chmeliik: Or, in this case, not without getting rid of `IMAGE_URL` and `IMAGE_DIGEST`. But that would probably require a huge migration

zregvart: Let's do [STONEBLD-2326](https://issues.redhat.com/browse/STONEBLD-2326) then before this.

zregvart:
Keeping this in draft until we get to [STONEBLD-2326](https://issues.redhat.com/browse/STONEBLD-2326)

arewm:
I just [ran into](https://github.com/konflux-ci/olm-operator-konflux-sample/pull/15/files) an issue related to this. For the OLM konflux sample, I cannot "prepare" all architectures to run and then only enable the ones that I need with the current form. As long as I explicitly include all of the previous tasks in the parameters, the index image generation will be skipped if any of the results are empty.

Changing to this pattern will enable a more user-friendly flow.

zregvart:
Rebased on buildah 0.2 changes, should be ready for another round of reviews now.

task/source-build/0.1/source-build.yaml
```diff
@@ -157,5 +159,6 @@ spec:
 
         cat "$RESULT_FILE" | jq -j ".image_url" >"$RESULT_SOURCE_IMAGE_URL"
         cat "$RESULT_FILE" | jq -j ".image_digest" >"$RESULT_SOURCE_IMAGE_DIGEST"
+        jq -r '"\(.image_url)@\(.image_digest)"' "${RESULT_FILE}" >"$(results.IMAGE_REF.path)"
```
chmeliik, line 162: This will include a newline at the end (I think the previous jq calls use `-j` to avoid that)

task/tkn-bundle/0.1/README.md
```diff
@@ -1,51 +1,23 @@
...
-    │   │   │   └── task_run_subject.sh
-    │   │   ├── test1.yaml
-    │   │   ├── test2.yml
-    │   │   ├── test3.yaml
-    │   │   └── tkn-bundle_spec.sh
-    │   ├── TESTING.md
-    │   └── tkn-bundle.yaml
-    └── OWNERS
-
-Only the `0.1/tkn-bundle.yaml` file will be included in the bundle.
```
chmeliik, line 42: It would be a shame to lose this description, we can probably ignore what the generate-readme script wants to do here

task/tkn-bundle/0.1/tkn-bundle.yaml
```diff
@@ -134,7 +136,9 @@ spec:
         $(printf ' -f %s' "${FILES[@]}") \
         |tee /proc/self/fd/3)"
       echo -n "$IMAGE" > $(results.IMAGE_URL.path)
-      echo -n "${OUT#*Pushed Tekton Bundle to *@}" > $(results.IMAGE_DIGEST.path)
+      digest="${OUT#*Pushed Tekton Bundle to *@}""
```
chmeliik, line 139:
```suggestion
      digest="${OUT#*Pushed Tekton Bundle to *@}"
```

task/tkn-bundle/0.1/README.md
zregvart: Yeah, bit lazy on my side...

task/tkn-bundle/0.1/tkn-bundle.yaml
zregvart: OMG, thanks!

arewm:
@zregvart, were you intending to change the generation of the image index in this PR as well or will that be up to users to change in their own pipelines?

chmeliik (approved):

zregvart:
> @zregvart, were you intending to change the generation of the image index in this PR as well or will that be up to users to change in their own pipelines?

I think this is entirely up to the users and out of scope of this PR. If we had a multi-arch template pipeline then I would update that to use the new results. Follow [EC-714](https://issues.redhat.com/browse/EC-714) for more on that.

PR #1112, opened 2024-07-03
Combine three bundle build tasks into one

zregvart:
Seems that scheduling Task pods often runs into issues, most likely the inability to mount persistent volumes caused by "volume node affinity conflict". To minimize this issue three separate bundle build Tasks are now combined into a single Tasks as three Steps instead.

Reference: https://issues.redhat.com/browse/EC-705

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -42,20 +42,6 @@ spec:
           - name: output
             workspace: workspace
 
-      - name: clone-repository-to-redhat-appstudio-workspace
-        params:
-          - name: url
-            value: $(params.git-url)
-          - name: revision
-            value: "$(params.revision)"
-          - name: depth
-            value: "0"
-        taskRef:
-          name: git-clone
-        workspaces:
-          - name: output
-            workspace: workspace-redhat-appstudio
-
       - name: ec-task-checks
         runAfter:
           - clone-repository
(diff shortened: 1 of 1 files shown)


simonbaird:
The diff is not easy to review by eye, but I like the concept. If this can solve or even reduce the frequency of the intermittent failures described in EC-705 that would be fantastic.

.tekton/push.yaml
```diff
@@ -213,50 +134,44 @@ spec:
...
-                - $(workspaces.artifacts-redhat-appstudio.path)/task-bundle-list
-                - $(workspaces.artifacts-redhat-appstudio.path)/pipeline-bundle-list
+                - $(workspaces.source.path)/task-bundle-list-konflux-ci
+                - $(workspaces.source.path)/pipeline-bundle-list-konflux-ci
+                - $(workspaces.source.path)/task-bundle-list-appstudio
+                - $(workspaces.source.path)/pipeline-bundle-list-appstudio
               volumeMounts:
                 - mountPath: /root/.docker/config.json
                   subPath: .dockerconfigjson
                   name: quay-secret
```
chmeliik, line 153: I don't think the task will be able to push both to quay.io/konflux-ci and to quay.io/redhat-appstudio-tekton-catalog with same dockerconfig

.tekton/push.yaml
```diff
@@ -78,126 +64,61 @@ spec:
...
-              volumeMounts:
-                - mountPath: /root/.docker/config.json
-                  subPath: .dockerconfigjson
-                  name: quay-secret
-          volumes:
-          - name: quay-secret
-            secret:
-              secretName: redhat-appstudio-tekton-catalog-build-definitions-pull-secret
-          workspaces:
-            - name: source
```
chmeliik, line 122: I may have missed it, but isn't this volume declaration missing now?

.tekton/push.yaml
```diff
@@ -213,50 +134,44 @@ spec:
...
-                set -euo pipefail
-                set -x
-
-                DATA_BUNDLE_KONFLUX_CI=quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles
-
-                skopeo copy \
-                  "docker://$DATA_BUNDLE_RH_APPSTUDIO:$DATA_BUNDLE_TAG" \
-                  "docker://$DATA_BUNDLE_KONFLUX_CI:$DATA_BUNDLE_TAG"
+          - build-bundles-redhat-appstudio
+          - build-bundles-konflux-ci
```
chmeliik, line 160:
```suggestion
          - build-bundles
```

chmeliik (approved):
LGTM :+1: 

it does look a lot simpler than before, hopefully it will also behave better

zregvart:
Thanks for the review, I'll start the merge now and watch out for any issues.

zregvart:
Found an issue:
```
cat: /workspace/source/task-bundle-list-konflux-ci: No such file or directory
cat: /workspace/source/task-bundle-list-appstudio: No such file or directory
```
I'll do a followup PR with the fix

PR #1117, opened 2024-07-03
Ignore bundle_values.env

ralphbean:
This gets generated as a side-effect when you run hack/build-and-push.sh

Is there any value in *not* ignoring it?


tisutisu (approved):
LGTM

PR #1121, opened 2024-07-04
Switch to a new buildah image source

arewm:
Some changes:
* The image is built from https://github.com/konflux-ci/buildah-container
* The image is now built and released from Konflux so now we can easily track the provenance of the image
* The buildah image will just update the floating latest tag
* Buildah is now multi-arch
* Buildah has been updated to a recent commit and is therefore a newer verison

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -19,6 +19,7 @@
         "quay.io/konflux-ci/pull-request-builds",
         "quay.io/redhat-appstudio/github-app-token",
         "quay.io/konflux-ci/appstudio-utils",
+        "quay.io/konflux-ci/buildah",
         "quay.io/konflux-ci/source-container-build",
         "quay.io/redhat-appstudio/e2e-tests",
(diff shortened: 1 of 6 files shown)


rcerven (approved):

mkosiarc (approved):

mmorhun (approved):

tkdchen (commented):
I think we also need to update renovate.json in order to get update pr.

tkdchen (approved):

arewm:
Seeing an issue in the e2e tests
```
container step-build: 
  chown: changing ownership of '/var/lib/containers': Operation not permitted
```

Do we have documentation for this image? I built it as a `buildah` image, but I probably need to change https://github.com/konflux-ci/buildah-container/blob/main/Containerfile#L31-L35 to remove the `buildah` user and set the entrypoint to `bash`.

PR #1129, opened 2024-07-05
feat: push Dockerfile to registry

tkdchen:
[STONEBLD-2522](https://issues.redhat.com//browse/STONEBLD-2522)

Add a new task push-dockerfile to push Dockerfile to registry as an OCI
artifact.

Task is configurable to accept varaints of Dockerfile with different
names, like Containerfile.

Use `oras pull' to get the image.

This new task is added to docker-build pipeline and enabled by default.

Trusted Artifacts version of push-dockerfile is created accordingly and
updated into the docker-build-oci-ta pipeline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 pipeline-bundle-list
 task-bundle-list
 bundle_values.env
+
+*~
+*.swp
--- a/pipelines/docker-build-oci-ta/patch.yaml
+++ b/pipelines/docker-build-oci-ta/patch.yaml
@@ -24,6 +24,7 @@
 #  9  clamav-scan
 # 10  sbom-json-check
 # 11  apply-tags
+# 12  push-dockerfile
 
 # clone-repository Task
 - op: replace
@@ -111,6 +112,18 @@
       value: $(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)
 - op: remove
   path: /spec/tasks/8/workspaces/0
+
+# push-dockerfile
+- op: replace
+  path: /spec/tasks/12/taskRef/name
+  value: push-dockerfile-oci-ta
+- op: add
+  path: /spec/tasks/12/params/-
+  value:
+    name: SOURCE_ARTIFACT
+    value: $(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)
+- op: remove
+  path: /spec/tasks/12/workspaces/0
 
 # Order of finally Tasks from the base docker-build Pipeline:
 # $ kustomize build pipelines/docker-build | yq .spec.finally.[].name | nl -v 0
--- a/pipelines/docker-build-rhtap/patch.yaml
+++ b/pipelines/docker-build-rhtap/patch.yaml
@@ -84,6 +84,8 @@
 #      8  sast-snyk-check
 #      9  clamav-scan
 #      10 sbom-json-check
+#      11 apply-tags
+#      12 push-dockerfile
 - op: replace
   path: /spec/tasks/3/runAfter/0
   value: clone-repository
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -240,6 +240,24 @@ spec:
       params:
       - name: IMAGE
         value: $(tasks.build-container.results.IMAGE_URL)
+    - name: push-dockerfile
+      runAfter:
+      - build-container
+      taskRef:
+        name: push-dockerfile
+        version: "0.1"
+      params:
+      - name: IMAGE
+        value: $(tasks.build-container.results.IMAGE_URL)
+      - name: IMAGE_DIGEST
+        value: $(tasks.build-container.results.IMAGE_DIGEST)
+      - name: DOCKERFILE
+        value: $(params.dockerfile)
+      - name: CONTEXT
+        value: $(params.path-context)
+      workspaces:
+      - name: workspace
+        workspace: workspace
 
   finally:
     - name: show-sbom
--- a/task/push-dockerfile-oci-ta/0.1/README.md
+++ b/task/push-dockerfile-oci-ta/0.1/README.md
@@ -1,0 +1,20 @@
+# push-dockerfile-oci-ta task
+
+Discover Dockerfile from source code and push it to registry as an OCI artifact.
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|ARTIFACT_TYPE|Artifact type of the Dockerfile image.|application/vnd.konflux.dockerfile|false|
+|CONTEXT|Path to the directory to use as context.|.|false|
+|DOCKERFILE|Path to the Dockerfile.|./Dockerfile|false|
+|IMAGE|The built binary image. The Dockerfile is pushed to the same image repository alongside.||true|
+|IMAGE_DIGEST|The built binary image digest, which is used to construct the tag of Dockerfile image.||true|
+|SOURCE_ARTIFACT|The Trusted Artifact URI pointing to the artifact with the application source code.||true|
+|TAG_SUFFIX|Suffix of the Dockerfile image tag.|.dockerfile|false|
+
+## Results
+|name|description|
+|---|---|
+|IMAGE_REF|Digest-pinned image reference to the Dockerfile image.|
+
--- a/task/push-dockerfile-oci-ta/0.1/push-dockerfile-oci-ta.yaml
+++ b/task/push-dockerfile-oci-ta/0.1/push-dockerfile-oci-ta.yaml
@@ -1,0 +1,112 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: push-dockerfile-oci-ta
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: docker
+spec:
+  description: Discover Dockerfile from source code and push it to registry
+    as an OCI artifact.
+  params:
+    - name: ARTIFACT_TYPE
+      description: Artifact type of the Dockerfile image.
+      type: string
+      default: application/vnd.konflux.dockerfile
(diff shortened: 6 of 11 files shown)


task/build-image-manifest/0.1/build-image-manifest.yaml
```diff
@@ -105,3 +105,26 @@ spec:
...
+    env:
+    - name: IMAGE_DIGEST_FILE
+      value: $(results.IMAGE_DIGEST.path)
+    script: |
+      suffix=".dockerfile"
+      image_without_tag=${IMAGE%:*}
+      # remove the potential existing newline
+      manifest_digest=$(cat "${IMAGE_DIGEST_FILE}" | tr -d '\r\n')
+      df_image=${image_without_tag}:${manifest_digest/:/-}${suffix}
+      artype=application/vnd.konflux.dockerfile
```
mmorhun, line 120: Could you expand `ar` in the variable name?

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -544,3 +548,23 @@ spec:
...
+        - name: IMAGE_DIGEST_FILE
+          value: $(results.IMAGE_DIGEST.path)
+        - name: DOCKERFILE_PATH
+          value: /shared/Dockerfile
+      script: |
+        suffix=".dockerfile"
+        image_without_tag=${IMAGE%:*}
+        # remove the potential existing newline
+        manifest_digest=$(cat "${IMAGE_DIGEST_FILE}" | tr -d '\r\n')
+        df_image=${image_without_tag}:${manifest_digest/:/-}${suffix}
```
mmorhun, line 563: Please use full word, it makes code easier to read

mmorhun (approved):

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
tkdchen: Updated.

task/build-image-manifest/0.1/build-image-manifest.yaml
tkdchen: Updated.

task/build-image-manifest/0.1/build-image-manifest.yaml
```diff
@@ -105,3 +105,38 @@ spec:
...
+      retry oras pull --output "$output_dir" "${image_without_tag}:${arch_image_digest/:/-}${suffix}"
+
+      echo "Selecting auth for $IMAGE"
+      auth_json=$(mktemp)
+      select-oci-auth $IMAGE >"$auth_json"
+
+      # Do not assume the actual file name. Generally, it could be Dockerfile or Containerfile.
+      cd "$output_dir"
+      n=1
+      find . -type f | while read filename; do
```
chmeliik, line 134:
```suggestion
      find . -type f | while read -r filename; do
```
otherwise `read` treats some characters specially

chmeliik (commented):
I'll also note that it will not be possible for the release service (or any other consumer) to verify that the content of the `.dockerfile` image is still the same as what the build pipeline pushed. Which is probably fine, IIUC the Dockerfile image is kind of just a nice-to-have informational kind of thing

tkdchen:
I haven't seen any necessity to make a separate task. As an associated artifacts images like .sbom and .att, it is natural to push it as a step of the build.

chmeliik:
> I haven't seen any necessity to make a separate task. As an associated artifacts images like .sbom and .att, it is natural to push it as a step of the build.

`.att` and `.sig` are pushed by Tekton Chains, not by the build task. I would argue `.sbom` shouldn't be pushed by the build task either. In the current situation, we have ~9 build tasks that all have to copy-paste the same logic because they have to fulfill this undefined "interface" for what a build task should do. When some of those things don't actually make sense.

Our tasks don't really make sense as upstream parts of the Tekton ecosystem. Now we're injecting this functionality into every build task because that's more convenient for the Red Hat container catalog.

On the other hand, there is no good mechanism for sharing artifacts between tasks - other than Trusted Artifacts, which involves pushing to the registry in some form anyway.

mkosiarc (approved):

tkdchen:
PTAL. The previous changes are reworked by adding a new Tekton task `push-build-file`. Once review is done, I'll squash all the commits before merge.

.gitignore
```diff
@@ -3,3 +3,6 @@
 pipeline-bundle-list
 task-bundle-list
 bundle_values.env
+
+*~
```
chmeliik, line 7: Out of curiosity, what are these?

task/push-build-file-oci-ta/0.1/push-build-file-oci-ta.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: "docker"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "image-build, appstudio"
+  name: push-build-file-oci-ta
```
chmeliik, line 10: Could you add the `recipe.yaml` (every oci-ta task should have one) for this task, so that auto-generation works properly?

task/build-image-manifest/0.1/build-image-manifest.yaml
```diff
@@ -104,38 +104,3 @@ spec:
       capabilities:
         add:
           - SETFCAP
-
-  - name: push-dockerfile
-    image: quay.io/konflux-ci/oras@sha256:5d0a8a5535fcc4ba467264cacbdeab2fb8662a538a61cb7fc8b3155e3f20fa39
-    env:
-    - name: IMAGE_DIGEST_FILE
-      value: $(results.IMAGE_DIGEST.path)
```
chmeliik: The commits will need to be squashed before merging

chmeliik (commented):
Nice, this looks better to me :+1:

chmeliik:
FYI, I think the failing ec-tasks-check should be resolved by rebasing on main

task/build-image-manifest/0.1/build-image-manifest.yaml
chmeliik: just noticed your comment

.gitignore
tkdchen: These are temporary and backup files created by some text editors like vim.

task/push-build-file-oci-ta/0.1/push-build-file-oci-ta.yaml
tkdchen: Updated.

tkdchen:
Rebased. Added recipel.yaml and add a task result `IMAGE_URL`.

task/push-build-file-oci-ta/0.1/push-build-file-oci-ta.yaml
```diff
@@ -0,0 +1,112 @@
...
+      description: The Trusted Artifact URI pointing to the artifact with
+        the application source code.
+      type: string
+    - name: TAG_SUFFIX
+      description: Suffix of the build file image tag.
+      type: string
+      default: .dockerfile
+  results:
+    - name: IMAGE_URL
+      description: Digest-pinned image URL.
```
chmeliik, line 47: Let's maybe use `IMAGE_REF`? Build tasks use `IMAGE_URL` as the digest-less repo url

task/push-build-file/0.1/push-build-file.yaml
```diff
@@ -0,0 +1,97 @@
...
+    description: Suffix of the build file image tag.
+    type: string
+    default: .dockerfile
+  - name: ARTIFACT_TYPE
+    description: Artifact type of the build file image.
+    type: string
+    default: application/vnd.konflux.dockerfile
+  results:
+  - name: IMAGE_URL
+    description: Digest-pinned image URL.
```
chmeliik, line 39: Let's maybe use `IMAGE_REF`? Build tasks use `IMAGE_URL` as the digest-less repo url

task/push-build-file/0.1/push-build-file.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+      build_file_image=${IMAGE%:*}:${IMAGE_DIGEST/:/-}${TAG_SUFFIX}
+
+      cd "$(dirname $build_file_path)"
+      retry oras push --no-tty \
+        --format json \
+        --registry-config "$auth_json" \
+        --artifact-type "$ARTIFACT_TYPE" \
+        "$build_file_image" "$(basename $build_file_path)" \
+        | yq '.reference' >"$IMAGE_URL_RESULT"
```
chmeliik, line 93: Doesn't this cause an extra `\n` at the end of the image url result?

tkdchen:
The task `push-build-file` is renamed to `push-dockerfile`. PTAL.

task/push-build-file/0.1/push-build-file.yaml
tkdchen: Maybe. But, I'm thinking if we should use consistent result name, since we haven't had separated concept for digest-pinned and digest-less image name yet.

chmeliik: There's been a proposal to add an IMAGE_REF result to the buildah task (a digest-pinned url): https://github.com/konflux-ci/build-definitions/pull/1111

I think `IMAGE_URL` and `IMAGE_DIGEST` also have special meaning to Tekton Chains

task/push-build-file/0.1/push-build-file.yaml
tkdchen: Nice catch. Updated.

task/push-build-file/0.1/push-build-file.yaml
tkdchen: Oh, haven't noticed that PR. Give me a sec, let me update this PR.

tkdchen: PTAL.

chmeliik (approved):

tkdchen:
Squashed all commits and rebased.

PR #1130, opened 2024-07-06
Refactor oci-copy to be more efficient

ralphbean:
Originally, this task would download all artifacts requested in the input file, check them all, and then upload them all to the registry in one invocation of "oras push".

This had two problems. First, if "oras push" flaked out part way through and the user needed to retry their pipeline, the entire download section would need to be run again needlessly. Second, for extremely large artifacts with lots of medium-sized files, an enormous PVC would be needed to hold all of them between download and push to the registry.

The change here addresses both problems.

First, files are downloaded, checked, pushed to the registry and then deleted from local storage - one at a time. This obviates the need for a large volume to store all files at once, since only enough storage is needed to store one file, not all of them.

Second, as files are considered, first the registry is checked to see if the blob has already been pushed there. If it has, then skip the download step. This has the effect of greatly improving the runtime for artifacts where only one or two of many files have changed since the last taskrun.

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -114,36 +114,75 @@ spec:
 
         set -u
 
-        for varfile in /var/workdir/vars/*; do
-          echo "Reading $varfile"
-          source $varfile
-
-          echo "Downloading $OCI_SOURCE to $OCI_FILENAME"
-          curl "${CURL_ARGS[@]}" --fail --silent --show-error --location $OCI_SOURCE -o $OCI_FILENAME
-
-          echo "Confirming that digest of $OCI_FILENAME matches expected $OCI_ARTIFACT_DIGEST"
-          echo "$OCI_ARTIFACT_DIGEST $OCI_FILENAME" | sha256sum --check
-
-          echo "Appending to arguments for $OCI_FILENAME of type $OCI_ARTIFACT_TYPE"
-          args+=("${OCI_FILENAME}:${OCI_ARTIFACT_TYPE}")
-        done
-
-        if [ -z "${args}" ]; then
-          echo "No files found. Something is very wrong. Skipping upload."
-          exit 1
-        fi
+        echo "Selecting auth for $IMAGE"
+        select-oci-auth $IMAGE >auth.json
 
         echo "Extracting artifact_type"
         ARTIFACT_TYPE=$(cat "$(pwd)/source/$OCI_COPY_FILE" | yq '.artifact_type')
 
-        echo "Selecting auth for $IMAGE"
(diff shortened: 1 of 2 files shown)


task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
```diff
@@ -114,36 +114,75 @@ spec:
...
-          echo "Confirming that digest of $OCI_FILENAME matches expected $OCI_ARTIFACT_DIGEST"
-          echo "$OCI_ARTIFACT_DIGEST $OCI_FILENAME" | sha256sum --check
+            echo "Confirming that digest of $OCI_FILENAME matches expected $OCI_ARTIFACT_DIGEST"
+            echo "$OCI_ARTIFACT_DIGEST $OCI_FILENAME" | sha256sum --check
 
-          echo "Appending to arguments for $OCI_FILENAME of type $OCI_ARTIFACT_TYPE"
-          args+=("${OCI_FILENAME}:${OCI_ARTIFACT_TYPE}")
-        done
+            echo "Pushing blob of $OCI_FILENAME of type $OCI_ARTIFACT_TYPE"
+            oras blob push --registry-config auth.json quay.io/redhat-user-workloads/rhel-ai-tenant/models/mixtral-8x7b-instruct-v0-1 --media-type ${OCI_ARTIFACT_TYPE} ${OCI_FILENAME}
```
scoheb, line 160: hardcoded pullspec?

ralphbean: Ugh, thank you!

scoheb (approved):

task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -99,36 +99,75 @@ spec:
 
         set -u
 
+        echo "Selecting auth for $IMAGE"
+        select-oci-auth $IMAGE > auth.json
+
+        echo "Extracting artifact_type"
+        ARTIFACT_TYPE=$(cat "$(pwd)/source/$OCI_COPY_FILE" | yq '.artifact_type')
+
+        REPO=$(echo ${IMAGE} | awk -F ':' '{print $1}')
```
chmeliik, line 108: In https://github.com/konflux-ci/build-definitions/pull/1129/files and https://github.com/konflux-ci/build-definitions/blob/c316db02a8fe4f67ba7fd36c021e367ebad0b4fc/task/rpm-ostree/0.1/rpm-ostree.yaml#L264-L265 we use this, which theoretically works if the IMAGE reference contains a port number
```suggestion
        REPO=${IMAGE%:*}
```

chmeliik (approved):
Nice solution, if somewhat low-level 

Just one nitpick for extracting the repo from the image ref (mainly for consistency with other tasks)

PR #1131, opened 2024-07-06
Expose SBOM results from oci-copy task

ralphbean:
This is important, because it is how the enterprise contract tooling recognizes that this is an SBOM to be trusted. It won't trust just any SBOM tagged in the registry. It has to be able to find a reference to the SBOM by digest in the provenance record.

https://github.com/enterprise-contract/ec-policies/blob/718386d2239dae85a866f6bcd0adbba036cd1b1d/policy/lib/sbom.rego#L26-L34


ralphbean:
Related: https://github.com/enterprise-contract/ec-policies/pull/1060

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
```diff
@@ -191,3 +193,12 @@ spec:
         - cyclonedx
         - $(params.IMAGE)
       workingDir: /var/workdir
+    - name: report-sbom-url
+      image: quay.io/konflux-ci/yq:latest@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430
+      workingDir: /var/workdir
+      script: |
+        REPO=$(echo ${IMAGE} | awk -F ':' '{print $1}')
```
chmeliik, line 200: Same as https://github.com/konflux-ci/build-definitions/pull/1130#discussion_r1668639651

chmeliik (approved):

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
ralphbean: Revised!

PR #1059, opened 2024-06-06
Add task to build RHEL AI

stuartwdouglas:
# Before you complete this pull request ...

Look for any open pull requests in the repository with the title "e2e-tests update" and 
see if there are recent e2e-tests updates that will be applicable to your change.

PR #1133, opened 2024-07-08
bootc build VM task

ralphbean:
This is a duplicate of https://github.com/konflux-ci/build-definitions/pull/1059. I'm picking that up since stuart is on PTO.

I tried to address @mmorhun's first round of feedback there, here.

I'll close that one so we can focus on this one (although https://github.com/konflux-ci/build-definitions/pull/1042 is still valid).

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -42,6 +42,8 @@ spec:
     volumeMounts:
       - mountPath: "/var/workdir"
         name: workdir
+      - mountPath: "/var/lib/containers/storage"
+        name: varlibcontainers
   steps:
     - name: use-trusted-artifact
       image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:4e39fb97f4444c2946944482df47b39c5bbc195c54c6560b0647635f553ab23d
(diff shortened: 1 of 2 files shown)


ralphbean:
I think this is blocked on https://github.com/konflux-ci/build-definitions/pull/1137.

ralphbean:
Submitted https://github.com/enterprise-contract/hacks/pull/102

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -0,0 +1,263 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: image-build, appstudio, hacbs
```
mmorhun, line 6: Do we still need `hacbs`?

mmorhun (approved):

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: Almost certainly not. I can drop it and add konflux, or we can do a mass `sed` PR later to update it across the board?

ralphbean:
```
❯ git grep tekton.dev/tags | grep hacbs | wc -l
30
```
Let's do it in a separate PR and get them all at once. :)

ralphbean:
Thank you for the review @mmorhun.

PR #1144, opened 2024-07-10
Buffer output from `tash`

zregvart:
This allows us to read the existing `-oci-ta.yaml` file without the redirection truncating the file first.


zregvart:
Need this to be merged before I can merge [a change](https://github.com/enterprise-contract/hacks/pull/103) in `tash`

mmorhun (approved):

PR #1147, opened 2024-07-10
fix: expose image manifest from build-vm-image task                       

ralphbean:
The way this previously worked was that each build-vm-image task would produce an image index (index1), which was later fed into a the build-image-index task to produce a second image index (index2), which (by use of the --all option) would throwaway the vm image's original index image (index1), keeping only its image manifest.

The problem this caused was that the vm image's original index image (index1) was exposed as a result, and not the image manifest. This meant that tekton chains would not see the image manifest, would not generate an attestation for it, and would not sign it.

Later, when trying to validate the aggregate index image (index2), policy checks would fail since the index image (index2) was signed, but non of the image manifests were signed.

The change here modifies things so that the build-vm-image task exposes only an image manifest, which will be attested to and signed. Its exposed pullspect will be fed to the build-image-index task, which will expose its own image index pullspec as a result to be attested to and signed. And in the end, we should have a correct situation with one image index (signed) referring to $N image manifests (also signed).

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -148,7 +148,7 @@ spec:
         # form the --type arguments
         IMAGE_TYPE_ARGUMENT=" --type $IMAGE_TYPE "
 
-        # this heredoc allows expansions for the image name
+        # this unquoted heredoc allows expansions for the image name
         cat >scripts/script-build.sh <<REMOTESSHEOF
         #!/bin/sh
         set -e
@@ -166,7 +166,7 @@ spec:
 
         REMOTESSHEOF
 
-        # no expansions in this one, the env vars are evaluated on the remote vm
+        # this quoted heredoc prevents expansions and command substitutions. the env vars are evaluated on the remote vm
         cat >>scripts/script-build.sh <<'REMOTESSHEOF'
         echo >config.toml <<EOF
         [[blueprint.customizations.user]]
@@ -203,9 +203,17 @@ spec:
         REMOTESSHEOF
 
         # script-push.sh script is intended run _inside_ podman on the ssh host and requires sudo
+        # this unquoted heredoc allows expansions for the image name
         cat >scripts/script-push.sh <<REMOTESSHEOF
         #!/bin/bash
         set -ex
+
+        export OUTPUT_IMAGE="$OUTPUT_IMAGE"
+
+        REMOTESSHEOF
+
+        # this quoted heredoc prevents expansions and command substitutions. the env vars are evaluated on the remote vm
+        cat >>scripts/script-push.sh <<'REMOTESSHEOF'
         dnf -y install buildah pigz jq
 
         # Build an image index of length 1 referring to an image manifest with the content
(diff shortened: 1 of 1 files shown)


task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -203,10 +203,20 @@ spec:
...
+        export OUTPUT_IMAGE="$OUTPUT_IMAGE"
+
+        REMOTESSHEOF
+
+        # this quoted heredoc prevents expansions and command substitutions. the env vars are evaluated on the remote vm
+        cat >>scripts/script-push.sh <<'REMOTESSHEOF'
+        dnf -y install buildah pigz jq
+
+        # Build an image index of length 1 referring to an image manifest with the content
         buildah --storage-driver=vfs manifest create "$OUTPUT_IMAGE"
```
arewm, line 220: We use environment variables in other buildah tasks to specify the storage driver. Should we change this task to use that too? (and should the storage driver be vfs or overlay as we are building in VMs and not within k8s?)

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -226,8 +236,14 @@ spec:
           buildah --storage-driver=vfs manifest add --arch $(arch) --os linux --artifact --artifact-type application/vnd.diskimage.iso.gzip $OUTPUT_IMAGE /output/bootiso/install.iso.gz
         fi
         buildah --storage-driver=vfs manifest push --digestfile image-digest --authfile /.docker/config.json --all $OUTPUT_IMAGE
+
+        # At this point, we have pushed an image index of length 1 to the registry.
+        # Next, extract a reference to the image manifest and expose that, throwing away the image index.
+        IMAGE_INDEX_DIGEST=$(cat image-digest)
+        MANIFEST_DIGEST=$(buildah manifest inspect --authfile /.docker/config.json $OUTPUT_IMAGE@$IMAGE_INDEX_DIGEST | jq '.manifests[0].digest')
```
arewm, line 243: It is a simpler change to continue to use buildah to push the artifacts. I first proposed the use of buildah so that we could generate the Image Index in the task as we were going to produce all artifacts together in a single task.

Since we have settled on this task to only produce a single disk image, does it make more sense to use a different tool that will _only_ push the image manifest and not the manifest + index? The closest approximation that buildah has is `buildah source` ... I have used that in the past but it doesn't allow for specifying the properties like artifacttype and platform.

arewm (approved):
I am fine with the change. The comments are closer to being nitpicks.

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: Yeah, I went down the same path thinking through the change. The usage of `oras` in the `oci-copy` task is probably pretty close to what we want to do here.

It'll be worth following up to see if that can make the implementation simpler.

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: Yeah, let's follow upon it.

PR #1152, opened 2024-07-11
source-build: read base image from SBOM

chmeliik:
[STONEBLD-2607](https://issues.redhat.com//browse/STONEBLD-2607)

When the BASE_IMAGES param is empty, fall back to reading the base image from the SBOM of the BINARY_IMAGE.

This removes the source-build task's dependency on the BASE_IMAGES_DIGESTS result, finally allowing us to stop returning this result from build tasks.

### DIFF
--- a/task/source-build-oci-ta/0.1/source-build-oci-ta.yaml
+++ b/task/source-build-oci-ta/0.1/source-build-oci-ta.yaml
@@ -12,10 +12,11 @@ spec:
   description: Source image build.
   params:
     - name: BASE_IMAGES
-      description: Base images used to build the binary image. Each image
-        per line in the same order of FROM instructions specified in a multistage
-        Dockerfile. Default to an empty string, which means to skip handling
-        a base image.
+      description: By default, the task inspects the SBOM of the binary image
+        to find the base image. With this parameter, you can override that
+        behavior and pass the base image directly. The value should be a newline-separated
+        list of images, in the same order as the FROM instructions specified
+        in a multistage Dockerfile.
       type: string
       default: ""
     - name: BINARY_IMAGE
@@ -42,6 +43,11 @@ spec:
     - name: workdir
       emptyDir: {}
   stepTemplate:
+    env:
+      - name: BASE_IMAGES_FILE
+        value: /var/source-build/base-images.txt
+      - name: BINARY_IMAGE
+        value: $(params.BINARY_IMAGE)
     volumeMounts:
       - mountPath: /var/workdir
         name: workdir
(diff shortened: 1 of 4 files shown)


chmeliik:
Tested in https://github.com/redhat-appstudio/rh-syft/pull/82 (https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-build/applications/rh-syft/pipelineruns/rh-syft-on-pull-request-t2qb9)

chmeliik:
> timed out when waiting for image-controller annotations to be updated on component test-comp-sgjs in namespace build-templates-e2e. component: {

/retest

chmeliik:
> timed out when waiting for image-controller annotations to be updated on component test-comp-ilms

again :disappointed: 
looking

chmeliik:
Rebased on main. That should fix it :facepalm:

task/source-build-oci-ta/0.1/source-build-oci-ta.yaml
```diff
@@ -52,16 +58,55 @@ spec:
...
+            image_without_digest=${BINARY_IMAGE%@*}
+            image_without_tag=${image_without_digest%:*}
+            image=${image_without_tag}@${manifest_digest}
+        else
+            # The image is a single manifest
+            image=$BINARY_IMAGE
+        fi
+
+        echo "Downloading SBOM for $image"
+        sbom=$(cosign download sbom "$image")
```
mmorhun, line 93: Should we gracefully handle errors? Do we need a retry mechanism?

mmorhun (approved):

task/source-build-oci-ta/0.1/source-build-oci-ta.yaml
chmeliik: :+1: added basic retry mechanism. Not sure if we should let the source build proceed when we fail to download the SBOM. It is true that the source build task is generally very tolerant to failing to find base image source. WDYT @tkdchen ?

tkdchen: Yeah. Any failure during handling the base image sources should not break the whole build. I would prefer keeping the graceful handle with meaningful logs.

task/source-build/0.1/source-build.yaml
```diff
@@ -34,7 +35,69 @@ spec:
...
+            exit
+        fi
+
+        echo "BASE_IMAGES param is empty, inspecting the SBOM instead"
+
+        raw_inspect=$(skopeo inspect --raw "docker://$BINARY_IMAGE")
+        if manifest_digest=$(jq -e -r '.manifests[0].digest' <<< "$raw_inspect"); then
+            # The BINARY_IMAGE is an index image, each manifest in the list has its own SBOM.
+            # We're gonna assume the base images are the same or similar enough in all the SBOMs.
+            # How would we even handle a build where each manifest in the list is built from different base images?
```
tkdchen, line 69: This can be one issue of supporting source container build for multi-arch builds. But, what would be the case this question mentions? Is there any example or real build?

task/source-build/0.1/source-build.yaml
```diff
@@ -34,7 +35,69 @@ spec:
...
+
+        echo -n "Extracting base images from SBOM"
+        echo " (looking for .formulation[].components[] with 'konflux:container:is_base_image' property)"
+        jq -r '
+            .formulation[]?
+            | .components[]?
+            | select(any(.properties[]?; .name == "konflux:container:is_base_image"))
+            | (.purl | capture("^pkg:oci/.*?@(?<digest>.*?:[a-f0-9]*)")) as $matched
+            | .name + "@" + $matched.digest
+        ' <<< "$sbom" | tee "$BASE_IMAGES_FILE"
```
tkdchen, line 99: It would be good to have a comment telling that the order of base image components has been ensured during the SBOM generation.

task/source-build/0.1/source-build.yaml
```diff
@@ -34,7 +35,69 @@ spec:
...
+        for i in {1..5}; do
+            echo "Downloading SBOM for $image (attempt $i)"
+            sbom=$(cosign download sbom "$image") && break
+            [[ "$i" -lt 5 ]] && sleep 1
+        done
+
+        if [[ -z "$sbom" ]]; then
+            echo "Failed to download SBOM after 5 attempts"
+            exit 1
+        fi
```
tkdchen, line 89: In addition to this message, log something else to tell the build keeps running and the expected status of including base image sources in the result?

tkdchen (commented):
LGTM. There are a few minor comments.

task/source-build/0.1/source-build.yaml
chmeliik: Technically, the images grouped together in a manifest list don't really need to have anything in common with each other. But I'm not aware of any real use cases. For the usage we expect, taking any one image from the list should be fine

task/source-build/0.1/source-build.yaml
chmeliik: We don't really depend on the order here, because we're only taking the one image with 
```json
{
  "name": "konflux:container:is_base_image"
  "value": "true"
}
```
the others have
```json
{
  "name": "konflux:container:is_builder_image:for_stage"
  "value": "<index of stage in Dockerfile>"
}
```
I'll add a comment about that

tkdchen: I see.

task/source-build-oci-ta/0.1/source-build-oci-ta.yaml
chmeliik: Changed it to proceed even if the SBOM can't be downloaded

task/source-build/0.1/source-build.yaml
chmeliik: Removed the last sentence in the comment

task/source-build/0.1/source-build.yaml
chmeliik: Changed

chmeliik:
Re-generated the oci-ta variant

task/source-build/0.1/source-build.yaml
tkdchen: Seems `./generate-ta-tasks.sh` was not run. `source-build-oci-ta` task does not have that WARNING.

tkdchen (approved):
:+1:

PR #1153, opened 2024-07-11
SDCICD-1337: custom task to generate an olm bundle

jbpratt:
Add a new custom task to generate the OLM bundle from the provided input
directory. This generates the on-disk representation of an operator (a
bundle) including the CSV, CRDs, manifests, additional metadata and the
Dockerfile to be built. Once built, it can be submitted to a catalog and
deployed to a cluster.

```
$ operator-sdk generate bundle \
    --overwrite \
    --version 4.16.265-f15333d \
    --input-dir deploy/ \
    --package osd-example-operator-bundle
Generating bundle version 4.16.265-f15333d
Generating bundle manifests
Building a ClusterServiceVersion without an existing base
Bundle manifests generated successfully in bundle
Generating bundle metadata
INFO[0000] Creating bundle.Dockerfile
INFO[0000] Creating bundle/metadata/annotations.yaml
INFO[0000] Bundle metadata generated successfully
```

This is different from the traditional flow of an operator where the
bundle metadata and manifests are stored somewhere to be built from or
actively generated on new versions. SRE-P operators do not actively
version their operators or maintain an upgrade graph so this allows
dynamically generating a bundle at build time based on the current state
of the repository.

Signed-off-by: Brady Pratt <bpratt@redhat.com>

---

This new task is highly specific to the way that SRE-P builds and deploys
operators, so it may not be generally useful to others without additional
changes. I'm more than open to a better name for the task or any additional
context needed to ensure it isn't misused or leads to confusion when someone
else is onboarding a standard OLM operator.

/hold

The only things I want to make sure of are the two TODOs left. First is
ensuring this image is actually built in Konflux and can be used. Second is a
parameter to set the image based on targeted version.

https://issues.redhat.com/browse/SDCICD-1337

### DIFF
--- a/task/operator-sdk-generate-bundle/0.1/README.md
+++ b/task/operator-sdk-generate-bundle/0.1/README.md
@@ -1,0 +1,16 @@
+# operator-sdk-generate-bundle task
+
+Generate an OLM bundle using the operator-sdk
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|input-dir|Directory to read cluster-ready operator manifests from|deploy|false|
+|channels|Comma-separated list of channels the bundle belongs to|alpha|false|
+|version|Semantic version of the operator in the generated bundle||true|
+|package-name|Bundle's package name||true|
+
+## Workspaces
+|name|description|optional|
+|---|---|---|
+|source|Workspace with the source code|false|
--- a/task/operator-sdk-generate-bundle/0.1/operator-sdk-generate-bundle.yaml
+++ b/task/operator-sdk-generate-bundle/0.1/operator-sdk-generate-bundle.yaml
@@ -1,0 +1,37 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: operator-sdk-generate-bundle
+spec:
+  description: Generate an OLM bundle using the operator-sdk
+  params:
+    - name: input-dir
+      description: Directory to read cluster-ready operator manifests from
+      default: deploy
+    - name: channels
+      description: Comma-separated list of channels the bundle belongs to
+      default: alpha
+    - name: version
+      description: Semantic version of the operator in the generated bundle
+    - name: package-name
+      description: Bundle's package name
+  workspaces:
+    - name: source
+      description: Workspace with the source code
+  steps:
+    - name: operator-sdk-generate-bundle
+      image: "registry.redhat.io/openshift4/ose-operator-sdk-rhel9:v4.16"
+      workingDir: $(workspaces.source.path)/source
+      args:
+        - generate
+        - bundle
+        - --overwrite
+        - --channels
+        - $(params.channels)
+        - --input-dir
+        - $(params.input-dir)
+        - --version
+        - $(params.version)
+        - --package
+        - $(params.package-name)
--- a/task/operator-sdk-generate-bundle/OWNERS
+++ b/task/operator-sdk-generate-bundle/OWNERS
@@ -1,0 +1,6 @@
+approvers:
+- jbpratt
+- gurnben
+reviewers:
+- jbpratt
+- gurnben


task/generate-olm-bundle/OWNERS
jbpratt, line 1: I want to add owners here, are they just github usernames and do they require to be in the `konflux-ci` org?

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
```diff
@@ -0,0 +1,40 @@
...
+      description: Bundle's package name
+  workspaces:
+    - name: source
+      description: Workspace with the source code
+  steps:
+    - name: generate-olm-bundle
+      # TODO: ensure this image is built within konflux
+      # TODO: can this be configurable as a parameter? it should be specific to
+      # the target version
+      image: "registry.redhat.io/openshift4/ose-operator-sdk-rhel9:v4.16"
```
jbpratt, line 27: Is it valid for this to be a parameter? This should be set to whatever version of OCP is being targetted

gurnben (approved):
LGTM!

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
arewm: I am not sure I understand. The bundle version is agnostic to the OCP version targeted. 

We prefer to not make the image a parameter as it makes the result of the task nondeterministic as people can substitute _and_ image in there.

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - name: version
+      description: Semantic version of the operator in the generated bundle
+    - name: package-name
+      description: Bundle's package name
+  workspaces:
+    - name: source
+      description: Workspace with the source code
+  steps:
+    - name: generate-olm-bundle
+      # TODO: ensure this image is built within konflux
```
arewm, line 24: What are you trying to ensure is built? The operator-sdk image?

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
arewm, line 27: We should not use pullspecs for registry.redhat.io as these definitions are available to the community to use. They might not have Red Hat entitlements.

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
jbpratt: Yes! I haven't found the pipeline files for this so I'm just going off of what Joe said.

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
jbpratt: Sorry, I confused this one with the catalog task (https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1720615804304029?thread_ts=1720609182.337489&cid=C04PZ7H0VA8) which will have this problem

jbpratt: I'm afraid I don't understand. `fbc-validation` uses `registry.redhat.io/openshift4/ose-operator-registry-rhel9`

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
```diff
@@ -0,0 +1,40 @@
...
+        - bundle
+        - --overwrite
+        - --channels
+        - $(params.channels)
+        - --input-dir
+        - $(params.input-dir)
+        - --version
+        - $(params.version)
+        - --package
+        - $(params.package-name)
```
arewm, line 40: I understand that you are trying to build a task that solves a specific use case but this feels too specific as the task is worded in a way that it should be general purpose.

jbpratt: > This new task is highly specific to the way that SRE-P builds and deploys
operators, so it may not be generally useful to others without additional
changes. I'm more than open to a better name for the task or any additional
context needed to ensure it isn't misused or leads to confusion when someone
else is onboarding a standard OLM operator.

:smile: I don't know of a usecase outside of how we are using it, I'm open to any recommendations on how to generalize it more so it is useful for others.

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
gurnben: @arewm in this case, any suggestions on non registry.redhat.io images that fulfill the same use-case?  We were recommended to use this image by the OLM team if I recall!

jbpratt:
Hey @arewm, just wanted to follow up on this to discuss the review

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
arewm: We do not have a better solution for this today, so I am going to resolve this thread for now. If Konflux users do not have Red Hat entitlements then they wouldn't be able to use a couple of the tasks.

arewm: We should remove this TODO then.

task/generate-olm-bundle/0.1/generate-olm-bundle.yaml
arewm: Do we still need this TODO?

jbpratt:
Hey @arewm would it be appropriate to output the generated Dockerfile path as a result of this task?

task/generate-olm-bundle/OWNERS
tkdchen: AFAIK, there is no specific rules for writing content of OWNERS files. The content can be in arbitrary strings but I think it should be clear enough for describing the owners of the task.

chmeliik: The rules would be https://www.kubernetes.dev/docs/guide/owners/, but build-definitions doesn't always follow them

jbpratt: Updated with owners! Thank you both

tkdchen (approved):
Where is the generated OLM bundle stored?

jbpratt:
> Where is the generated OLM bundle stored?

The on disk representation is never stored anywhere, it is immediately built into a container image using buildah right after this task runs.

jbpratt:
@chmeliik @tkdchen :eyes: could y'all take a look at this again?

chmeliik (approved):

manstis:
Hey @jbpratt great work.. I think this is going to be very helpful for me.

Do you know of, or plans for, a Task that can create an Operator Catalog too?

Thank-you.

jbpratt:
hey @manstis, #1170 is the PR for the catalog generation! The use case is a bit different than traditional OLM as this maintains no upgrade graph, just as an FWIW.

PR #1158, opened 2024-07-13
buildah: Don't include oci-archive: FROM lines in BASE_IMAGES

owtaylor:
In certain cases, we might have a FROM line which is from an archive exported by a pevious build stage; when we are pre-pulling images for a hermetic build, skip these images.

(Even if the oci-archive wasn't generated during the build, there's no point in pre-pulling a local URL.)

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -267,7 +267,7 @@ spec:
 
         BUILDAH_ARGS=()
 
-        BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}')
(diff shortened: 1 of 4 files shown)


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -267,7 +267,7 @@ spec:
 
         BUILDAH_ARGS=()
 
-        BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}')
+        BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}' | grep -v ^oci-archive:)
```
chmeliik, line 270: Technically, if all the FROM lines are `FROM oci-archive:...` then grep will exit with 1 and kill the build

owtaylor: Sharp eyes! Don't think that's likely, but pushed a revised version, in case this code gets cut-and-pasted somwhere else.

chmeliik (approved):

PR #1159, opened 2024-07-15
Comment out how template-build is intended to be processed

tkdchen:
Someone may be confused when processing the build pipeline template YAML by kustomize directly and does not easily know what is the purpose of having fields, e.g. version, that are not valid field of tekton taskRef.

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -1,3 +1,6 @@
+# This build pipeline template is intended to be processed by scripts
+# under hack/ directory rather than by kustomize directly.
+
 apiVersion: tekton.dev/v1
 kind: Pipeline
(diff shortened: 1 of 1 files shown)


chmeliik (approved):

cmoulliard:
`# A pointer for referencing the correct version of task in the built bundle.` What is a `built` bundle ? That should be clarified and ideally pointing to the documentation explaining how the rendering is taking place

tkdchen:
@cmoulliard

I'll replace the "built bundle" with "built pipeline bundles".

AFAIK, there is no specific document explaining how the rendering is taking place. At least two scripts handle the version field. The message at the top of the YAML tells that. Developers can refer to the scripts implementation for detailed information.

PR #1161, opened 2024-07-15
.tekton/push: add missing runAfter relation

chmeliik:
The build-bundles step uses the image built by the build-container step.

Without the runAfter relation, the step would start before the image has been built and eventually give up on trying to pull the not-yet-existing image.


zregvart (approved):
Thanks!

PR #1164, opened 2024-07-16
buildah: Handle missing components property in syft-generated SBOM

owtaylor:
When the Syft SBOM didn't have the components property, the output was generated incorrectly because - the default empty list needed to be added back to image_sbom.


owtaylor:
I noticed this because it caused a backtrace in merge_syft_boms.py: https://github.com/containerbuildsystem/cachi2/pull/576  - but they are actually two independent bugs.

chmeliik (approved):
Nice catch, thanks!

chmeliik:
Just a nitpick: I think the commit message is missing a verb ("handle"?)

owtaylor:
Pushed a new version with a fixed up Subject for the commit message

PR #1167, opened 2024-07-16
Tag bundles after acceptable bundles list pushed

joejstuart:
This change addresses the issue where the acceptable bundles list does not contain new bundle updates. The issue is addressed by tagging new bundles after the acceptable bundle list is pushed. This will ensure renovate  does not trigger until after the acceptable bundles list has been updated.
    
 https://issues.redhat.com/browse/EC-627

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -125,6 +125,8 @@ spec:
                 value: "$(params.revision)"
               - name: GIT_URL
                 value: "$(params.git-url)"
+              - name: OUTPUT_TASK_BUNDLE_LIST
+                value: $(workspaces.source.path)/full-bundle-list
               script: |
                 #!/bin/bash
                 set -euo pipefail
@@ -138,7 +140,7 @@ spec:
                   [[ -f "$f" ]] && list+=("$f")
                 done
 
-                .tekton/scripts/build-acceptable-bundles.sh "${list[@]}"
+                hack/build-acceptable-bundles.sh "${list[@]}"
 
                 echo -n "${DATA_BUNDLE_TAG}" > acceptable_bundle_tag
               args:
@@ -168,6 +170,14 @@ spec:
                 - mountPath: /root/.docker/config.json
                   subPath: .dockerconfigjson
                   name: quay-secret
+            - name: tag-bundles-konflux-ci
+              image: quay.io/konflux-ci/appstudio-utils:{{ revision }}
+              workingDir: $(workspaces.source.path)/source
+              command: ["./hack/push-and-tag.sh"]
+              env:
+              - name: OUTPUT_TASK_BUNDLE_LIST
+                value: $(workspaces.source.path)/full-bundle-list
+
           volumes:
           - name: quay-secret
             secret:
--- a/hack/build-acceptable-bundles.sh
+++ b/hack/build-acceptable-bundles.sh
@@ -3,9 +3,26 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+
+# Function to remove the sha and digest from the image name
+# from: quay.io/konflux-ci/task1:0.1-1234@sha256:5678 to quay.io/konflux-ci/task1:0.1
+strip_image_tag() {
+  sed 's/\(:[^-]*\).*/\1/' <<< "$1"
+}
+
 # helps with debugging
 DATA_BUNDLE_REPO="${DATA_BUNDLE_REPO:-quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles}"
 mapfile -t BUNDLES < <(cat "$@")
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+# File containing the list of images
+OUTPUT_TASK_BUNDLE_LIST="${OUTPUT_TASK_BUNDLE_LIST-${SCRIPTDIR}/../task-bundle-list}"
+for i in "${!BUNDLES[@]}"; do
+    original_line="${BUNDLES[$i]}"
+    modified_line=$(strip_image_tag "$original_line")
+    BUNDLES[$i]="$modified_line"
+    echo "$original_line,$modified_line" >> "$OUTPUT_TASK_BUNDLE_LIST"
+done
 
 # store a list of changed task files
 task_records=()
@@ -24,11 +41,15 @@ for path in $(git log -m -1 --name-only --pretty="format:" "${REVISION}"); do
     fi
 done
 
-echo "Tasks to be added:"
-printf '%s\n' "${task_records[@]}"
+if [ ${#task_records[@]} -gt 0 ]; then
+    echo "Tasks to be added:"
+    printf '%s\n' "${task_records[@]}"
+fi
 
-echo "Bundles to be added:"
-printf '%s\n' "${BUNDLES[@]}"
+if [ ${#BUNDLES[@]} -gt 0 ]; then
+    echo "Bundles to be added:"
+    printf '%s\n' "${BUNDLES[@]}"
+fi
 
 # The OPA data bundle is tagged with the current timestamp. This has two main
 # advantages. First, it prevents the image from accidentally not having any tags,
@@ -44,7 +65,6 @@ fi
 mapfile -t -d ' ' BUNDLES_PARAM < <(printf -- '--bundle=%s ' "${BUNDLES[@]}")
 
 PARAMS=("${TASK_PARAM[@]}" "${BUNDLES_PARAM[@]}")
-
 ec track bundle --debug \
     --input "oci:${DATA_BUNDLE_REPO}:latest" \
     --output "oci:${DATA_BUNDLE_REPO}:${DATA_BUNDLE_TAG}" \
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -20,6 +20,7 @@ tkn_bundle_push() {
     local retry=0
     local -r interval=${RETRY_INTERVAL:-5}
     local -r max_retries=5
+
     while true; do
         tkn bundle push "$@" && break
         status=$?
@@ -42,7 +43,6 @@ tkn_bundle_push() {
 function save_ref() {
     local output
     output="$(< /dev/stdin)"
-    echo "${output}"
     local digest
     digest="$(echo "${output}" | grep -Po '@\K(sha256:[a-f0-9]*)')"
 
@@ -51,7 +51,6 @@ function save_ref() {
     local refFile
     refFile="$2"
     echo "${tagRef}@${digest}" >> "${refFile}"
-    echo "Created:"
     echo "${tagRef}@${digest}"
 }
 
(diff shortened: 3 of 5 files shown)


spec/hack/build-and-push_spec.sh
```diff
@@ -0,0 +1,46 @@
...
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+eval "$(shellspec - -c) exit 1"
+
+check_lines() {
+  while read -r line; do
+    if [[ "$line" == quay.io/* ]] && [[ ! "$line" =~ ([0-9]+\.[0-9]-1234)?@sha256:5678$ ]]; then
```
zregvart, line 11: Nitpick: I'd be a bit more strict here (regex is most likely wrong, typing from the top of my head...

```suggestion
    if [[ "$line" == quay.io/* ]] && [[ ! "$line" =~ ^quay.io/(?:[a-z\-/]+){2}:[0-9]+\.[0-9]-1234@sha256:5678$ ]]; then
```

zregvart (commented):
Thanks for writing the tests :partying_face:. I have few uncertainties, I might be misunderstanding bits here...

hack/build-and-push.sh
```diff
@@ -114,20 +113,18 @@ do
         echo Unknown task in "$task_dir"
         continue
     fi
+
+    # the task files have been copied to the $prepared_task_file location at this point
     repository=${TEST_REPO_NAME:-task-${task_name}}
     tag=${TEST_REPO_NAME:+${task_name}-}${task_version}
-    task_bundle=quay.io/$QUAY_NAMESPACE/${repository}:${tag}
+    task_bundle=quay.io/$QUAY_NAMESPACE/${repository}:${tag}-${task_file_sha}
```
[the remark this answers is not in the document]
joejstuart: Yes, good grief. Thanks @zregvart, should be easy to fix.

.tekton/push.yaml
```diff
@@ -137,7 +139,7 @@ spec:
                   [[ -f "$f" ]] && list+=("$f")
                 done
 
-                .tekton/scripts/build-acceptable-bundles.sh "${list[@]}"
+                hack/build-acceptable-bundles.sh "${list[@]}"
```
lcarva, line 142: Why move this file? I think it makes sense to live under `.tekton/scripts` since it's meant to be used during the CI of this repo. `hack/push-and-tag.sh` should probably go there as well.

hack/build-acceptable-bundles.sh
```diff
@@ -3,10 +3,34 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# Function to remove everything after @ in the image reference
+remove_digest() {
+  local input="$1"
+  echo "$input" | sed 's/@.*//'
+}
+
+# Function to remove the checksum and leading hyphen from the image reference
```
lcarva, line 12: Leading hyphen?

hack/build-and-push.sh
```diff
@@ -42,7 +43,6 @@ tkn_bundle_push() {
 function save_ref() {
     local output
     output="$(< /dev/stdin)"
-    echo "${output}"
```
lcarva, line 45: Any particular reason to remove these statements? They seem helpful for debugging purposes.

lcarva (commented):
Maybe I'm being naive, but can't we just push the acceptable bundle list image in all cases, but only tag it with `latest` at the end of the Pipeline? It's already tagged with a timestamp so it should be pretty simple to make this work.

zregvart:
> Maybe I'm being naive, but can't we just push the acceptable bundle list image in all cases, but only tag it with `latest` at the end of the Pipeline? It's already tagged with a timestamp so it should be pretty simple to make this work.

The way we have it right now is that Task bundles are pushed and version-tagged before the acceptable bundle is pushed (or tagged), so when the acceptable bundle fails pushing to quay and folk upgrade to the newer versions of Tasks they start seeing untrusted Task violations. We'd like to version-tag the Task bundles after we have pushed the updated acceptable bundle to quay.

hack/build-and-push.sh
```diff
@@ -114,20 +113,18 @@ do
...
         continue
     fi
+
+    # the task files have been copied to the $prepared_task_file location at this point
     repository=${TEST_REPO_NAME:-task-${task_name}}
     tag=${TEST_REPO_NAME:+${task_name}-}${task_version}
-    task_bundle=quay.io/$QUAY_NAMESPACE/${repository}:${tag}
+    task_bundle=quay.io/$QUAY_NAMESPACE/${repository}:${tag}-${task_file_sha}
 
     if digest=$(skopeo inspect --no-tags --format='{{.Digest}}' docker://"${task_bundle}-${task_file_sha}" 2>/dev/null); then
```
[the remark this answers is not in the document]
joejstuart: Does it need to work with `hex-hex`?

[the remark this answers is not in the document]
zregvart: Oh, I wasn't really clear, I thought it use (i.e. work with) tags as `:0.1-hex-hex`, because on line 120 the `$task_bundle` is `:0.1-hex` and here it would yet again add the `$task_file_sha` to become `:0.1-hex-hex`.

.tekton/push.yaml
joejstuart: I moved it because `build-and-push.sh` is under hack and it just made it easier if all three were in the same location, since they all create and work with files as input. I can move them under `.tekton` if you feel strongly about it.

hack/build-acceptable-bundles.sh
joejstuart: It removes the digest and file sha. I made it clearer in this commit https://github.com/konflux-ci/build-definitions/pull/1167/commits/691da0685efcabafa4a267caf6aa522a2daf9942

hack/build-and-push.sh
joejstuart: No, I must have removed it while modifying the script. I added it back.

hack/build-acceptable-bundles.sh
```diff
@@ -3,10 +3,27 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+
+# Function to remove the sha and digest from the image name
+# from: quay.io/konflux-ci/task1:0.1-1234@sha256:5678 to quay.io/konflux-ci/task1:0.1
+strip_image_tag() {
+  echo "$1" | sed 's/\(:[^-]*\).*/\1/'
```
simonbaird, line 10: I learned this recently:

```suggestion
  sed 's/\(:[^-]*\).*/\1/' <<< "$1"
```

hack/build-and-push.sh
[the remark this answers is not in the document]
joejstuart: Makes sense now. I removed the redundant sha. thanks

hack/build-acceptable-bundles.sh
joejstuart: That's cool. Updated.

spec/hack/build-and-push_spec.sh
joejstuart: done

hack/build-and-push.sh
joejstuart: I actually am keeping that one removed. It only prints the file location which is noise. The tasks being added is more helpful which is printed here. https://github.com/konflux-ci/build-definitions/pull/1167/files#diff-78a47706e6b9aca3579f93bc34adac628f9c102de7257a2fe14939130786a807R126

hack/push-and-tag.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+# File containing the list of images
+OUTPUT_TASK_BUNDLE_LIST="${OUTPUT_TASK_BUNDLE_LIST-${SCRIPTDIR}/../task-bundle-list}"
+
+# Read the file and process each line
+while IFS=, read -r original_image new_image; do
```
zregvart, line 8: I'm a bit confused about setting IFS to `,`, should it be `""`, and why have the `new_image` on this line?

hack/push-and-tag.sh
```diff
@@ -0,0 +1,17 @@
...
+    # Remove the quotes from the strings
+    original_image=$(echo "$original_image" | tr -d '"' | xargs)
+    new_image=$(echo "$new_image" | tr -d '"' | xargs)
```
zregvart, line 11: There will be quotes? I thought his used the `$OUTPUT_TASK_BUNDLE_LIST` that was placed by `save_ref`. I tried locally, got the `task-bundle-list` and there are no quotes within it, single reference per line, from what I can see. What am I missing?

Shouldn't this compute the `new_image`, something similar to what `strip_image_tag` is doing in `build-acceptable-bundles.sh` is doing?

hack/push-and-tag.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
zregvart, line 1: My mind broke on this script. I thought this would process `$OUTPUT_TASK_BUNDLE_LIST` where image references in the form of `quay.io/konflux-ci/task-xyz:0.1-task-hash@sha256:hash`, remove the `-task-hash` and push, i.e. release the task. I must be missing something.

hack/build-acceptable-bundles.sh
```diff
@@ -3,10 +3,27 @@ set -o errexit
...
 # helps with debugging
 DATA_BUNDLE_REPO="${DATA_BUNDLE_REPO:-quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles}"
 mapfile -t BUNDLES < <(cat "$@")
 
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+# File containing the list of images
+OUTPUT_TASK_BUNDLE_LIST="${OUTPUT_TASK_BUNDLE_LIST-${SCRIPTDIR}/../task-bundle-list}"
+for i in "${!BUNDLES[@]}"; do
+    original_line="${BUNDLES[$i]}"
+    modified_line=$(strip_image_tag "$original_line")
```
zregvart, line 22: I just realized that we pull the image in `ec track bundle`, so the stripping of the task hash from the tag will break `ec track bundle`, see https://github.com/enterprise-contract/ec-cli/blob/5cf4a7ff5d29729ce42150b030f2f1f2b4140b5d/internal/tracker/tracker.go#L160 and https://github.com/enterprise-contract/ec-cli/blob/5cf4a7ff5d29729ce42150b030f2f1f2b4140b5d/internal/tracker/bundle_info.go#L30.

That is the image reference here needs to be pushed already with the release tag for `ec track bundle` to work.

zregvart (changes requested):
I hope I'm wrong, the way we have it here: pushing the tag with the task hash (e.g. `0.1-task-hash`) first, then running `ec track bundle` then re-tagging with the release version (e.g. `0.1`) is not going to work. For `ec track bundle` the reference needs to be pushed to the registry, so it would need to work on `0.1-task-hash` tag, then that tag would end up in acceptable bundles; we try to strip the `-task-hash` here first, but then `ec track bundle` would not be able to pull the task bundle image.

hack/push-and-tag.sh
joejstuart: There will not be quotes. I originally had written the images to the file with quotes, but they were not needed so I removed them and didn't clean this. I'll push an update. 
It could compute `new_image` the same as `strip_image_tag`, but I thought it was better to have `build-acceptable-bundles` do it and pass that information to `push-and-tag`.

hack/push-and-tag.sh
zregvart: I'm [seeing it](https://github.com/konflux-ci/build-definitions/pull/1167/files#diff-06250a838abc96be6c8e0b4d81d7eab197b36fbb32a775795ef0c6e68a40ac9eR24) now, for me this needed to be in the `hack/build-and-push.sh` via `save_ref` function; this can work too.

hack/build-acceptable-bundles.sh
zregvart: Given that the digests are present in the image references, this will work even without the images being tagged with the release tag (e.g. `0.X`).

zregvart (approved):
The only issue that remains seems to be the order of records in the acceptable bundles when `--freshen` is used, I don't think this can be addressed with this PR -- a change is needed in the `ec track bundle`. I think we might be done here.

PR #1168, opened 2024-07-17
source-build-oci-ta: fix non-existent filepath

chmeliik:
The oci-ta variant of the task doesn't have the /var/source-build volume
mount


chmeliik:
This will fail CI until https://github.com/enterprise-contract/hacks/pull/106 is merged and a new tash released

zregvart (approved):

PR #1169, opened 2024-07-17
Support cachi2 config passing in prefetch-dependencies

eskultety:
Allow consumers to pass a cachi2 config to the tool in the prefetch task.

### DIFF
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -22,6 +22,11 @@ spec:
   - description: Set cachi2 log level (debug, info, warning, error)
     name: log-level
     default: "info"
+  - description: |
+      Pass configuration to cachi2.
+      Note this needs to be passed as a YAML-formatted config dump, not as a file path!
+    name: config-file-content
+    default: ""
   - name: caTrustConfigMapName
     type: string
     description: The name of the ConfigMap to read CA bundle data from.
@@ -30,7 +35,27 @@ spec:
     type: string
     description: The name of the key in the ConfigMap that contains the CA bundle data.
     default: ca-bundle.crt
+
+  stepTemplate:
+    env:
+      - name: CONFIG_FILE_CONTENT
+        value: $(params.config-file-content)
+    volumeMounts:
+      - name: config
+        mountPath: /mnt/config
+
   steps:
+  - name: sanitize-cachi2-config-file-with-yq
+    image: quay.io/konflux-ci/yq@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430
+    script: |
+      if [ -n "${CONFIG_FILE_CONTENT}" ]
+      then
+        # we need to drop 'goproxy_url' for safety reasons until cachi2 decides what the SBOM
+        # impact of this configuration option will be:
+        # https://github.com/containerbuildsystem/cachi2/issues/577
+        yq 'del(.goproxy_url)' <<< "${CONFIG_FILE_CONTENT}" > /mnt/config/config.yaml
+      fi
+
   - image: quay.io/redhat-appstudio/cachi2:0.8.0@sha256:5cf15d6f3fb151a3e12c8a17024062b7cc62b0c3e1b165e4a9fa5bf7a77bdc30
     # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
     # the cluster will set imagePullPolicy to IfNotPresent
@@ -60,6 +85,12 @@ spec:
         # Confirm input was provided though it's likely the whole task would be skipped if it wasn't
         echo "No prefetch will be performed because no input was provided for cachi2 fetch-deps"
         exit 0
+      fi
+
+      if [ -f /mnt/config/config.yaml ]; then
+        config_flag=--config-file=/mnt/config/config.yaml
+      else
+        config_flag=""
       fi
 
       if [ "$DEV_PACKAGE_MANAGERS" = "true" ]; then
@@ -97,7 +128,7 @@ spec:
         update-ca-trust
       fi
 
-      cachi2 --log-level="$LOG_LEVEL" fetch-deps \
+      cachi2 --log-level="$LOG_LEVEL" $config_flag fetch-deps \
       $dev_pacman_flag \
       --source=$(workspaces.source.path)/source \
       --output=$(workspaces.source.path)/cachi2/output \
@@ -133,3 +164,5 @@ spec:
           - key: $(params.caTrustConfigMapKey)
             path: ca-bundle.crt
         optional: true
+    - name: config
+      emptyDir: {}
--- a/task/prefetch-dependencies-oci-ta/0.1/README.md
+++ b/task/prefetch-dependencies-oci-ta/0.1/README.md
@@ -11,6 +11,7 @@ https://github.com/containerbuildsystem/cachi2#basic-usage.
 |SOURCE_ARTIFACT|The Trusted Artifact URI pointing to the artifact with the application source code.||true|
 |caTrustConfigMapKey|The name of the key in the ConfigMap that contains the CA bundle data.|ca-bundle.crt|false|
 |caTrustConfigMapName|The name of the ConfigMap to read CA bundle data from.|trusted-ca|false|
+|config-file-content|Pass configuration to cachi2. Note this needs to be passed as a YAML-formatted config dump, not as a file path! |""|false|
 |dev-package-managers|Enable in-development package managers. WARNING: the behavior may change at any time without notice. Use at your own risk. |false|false|
 |input|Configures project packages that will have their dependencies prefetched.||true|
 |log-level|Set cachi2 log level (debug, info, warning, error)|info|false|
--- a/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
@@ -28,6 +28,11 @@ spec:
       description: The name of the ConfigMap to read CA bundle data from.
       type: string
       default: trusted-ca
+    - name: config-file-content
+      description: |
+        Pass configuration to cachi2.
+        Note this needs to be passed as a YAML-formatted config dump, not as a file path!
+      default: ""
     - name: dev-package-managers
       description: |
         Enable in-development package managers. WARNING: the behavior may change at any time without notice. Use at your own risk.
@@ -56,6 +61,8 @@ spec:
         the application source code.
       type: string
   volumes:
+    - name: config
+      emptyDir: {}
     - name: trusted-ca
       configMap:
         items:
@@ -79,7 +86,12 @@ spec:
         performing http(s) requests.
       optional: true
   stepTemplate:
+    env:
+      - name: CONFIG_FILE_CONTENT
+        value: $(params.config-file-content)
     volumeMounts:
+      - mountPath: /mnt/config
+        name: config
       - mountPath: /var/workdir
         name: workdir
   steps:
@@ -104,6 +116,15 @@ spec:
       args:
         - use
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
+    - name: sanitize-cachi2-config-file-with-yq
+      image: quay.io/konflux-ci/yq@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430
+      script: |
+        if [ -n "${CONFIG_FILE_CONTENT}" ]; then
+          # we need to drop 'goproxy_url' for safety reasons until cachi2 decides what the SBOM
+          # impact of this configuration option will be:
+          # https://github.com/containerbuildsystem/cachi2/issues/577
+          yq 'del(.goproxy_url)' <<<"${CONFIG_FILE_CONTENT}"  >/mnt/config/config.yaml
+        fi
     - name: prefetch-dependencies
       image: quay.io/redhat-appstudio/cachi2:0.8.0@sha256:5cf15d6f3fb151a3e12c8a17024062b7cc62b0c3e1b165e4a9fa5bf7a77bdc30
       volumeMounts:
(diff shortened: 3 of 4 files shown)


chmeliik:
The cachi2 config doesn't contain any options that would compromise SBOM accuracy, right?

Maybe the [goproxy_url](https://github.com/containerbuildsystem/cachi2/blob/34b72cc4beddbae6295ba56c66aa0ff8668bf73c/cachi2/core/config.py#L14), since that would allow the user to make cachi2 download through a [private proxy](https://go.dev/ref/mod#private-module-proxy-private). But I think the purl spec doesn't account for that anyway :-/

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -125,6 +137,10 @@ spec:
       Workspace containing a .netrc file. Cachi2 will use the credentials in this file when
       performing http(s) requests.
     optional: true
+  - name: prefetch-config
+    description: |
+      Workspace containing a cachi2 YAML config file to further tweak its behaviour.
```
chmeliik, line 142: Passing the config as a workspace may not be ideal. The user would essentially have to create a ConfigMap or a Secret in their namespace and pass it to the pipeline as a workspace. At which point the behavior of the build is not configurable in the repo, it depends on an external Kubernetes object.

IMO it would be preferable to do one of
* pass the content of the config file as a param
* pass the path to the config file as a param, the user commits it into their repo (although at that point it would make more sense if the user could also incorporate (an equivalent of) the JSON input for fetch-deps into that file)
* pass the config as an array param of `NAME=value` strings, the task converts them to environment variables (after https://github.com/containerbuildsystem/cachi2/issues/573)

eskultety:
> The cachi2 config doesn't contain any options that would compromise SBOM accuracy, right?
> 
> Maybe the [goproxy_url](https://github.com/containerbuildsystem/cachi2/blob/34b72cc4beddbae6295ba56c66aa0ff8668bf73c/cachi2/core/config.py#L14), since that would allow the user to make cachi2 download through a [private proxy](https://go.dev/ref/mod#private-module-proxy-private). But I think the purl spec doesn't account for that anyway :-/

From PURL spec you wouldn't really know. As for the options, I'd imagine the idea to be that options can only modify the behaviour without deliberately causing a regressive behaviour on the output, IOW I can imagine existence of options improving the quality of the SBOM, not decreasing it, especially as a side effect.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
eskultety: > pass the config as an array param of NAME=value strings, the task converts them to environment variables (after https://github.com/containerbuildsystem/cachi2/issues/573)

^This was my intention all along, this PR is supposed to be just a workaround (or a band aid if you will) to unblock existing consumers, so I'm wondering if for the time being (i.e. before https://github.com/containerbuildsystem/cachi2/issues/573 lands) the consumers could live with a temporary suboptimal user experience, or whether we do want to spend more time with this, especially since with my nonexistent knowledge of Konflux I can't objectively estimate how much work would the following pose:

> pass the content of the config file as a param

chmeliik: Easier than the workspace solution, conceptually something like this

```diff
diff --git a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
index b993d3e..91970f5 100644
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -12,6 +12,8 @@ spec:
     Task that uses Cachi2 to prefetch build dependencies.
     See docs at https://github.com/containerbuildsystem/cachi2#basic-usage.
   params:
+  - name: config-file-content
+    description: cachi2 config file content
   - description: Configures project packages that will have their dependencies prefetched.
     name: input
   - description: >
@@ -50,6 +52,8 @@ spec:
       value: $(workspaces.netrc.bound)
     - name: WORKSPACE_NETRC_PATH
       value: $(workspaces.netrc.path)
+    - name: CONFIG_FILE_CONTENT
+      value: $(params.config-file-content)
     volumeMounts:
       - name: trusted-ca
         mountPath: /mnt/trusted-ca
@@ -68,6 +72,13 @@ spec:
         dev_pacman_flag=""
       fi
 
+      if [ -n "$CONFIG_FILE_CONTENT" ]; then
+        printf "%s" "$CONFIG_FILE_CONTENT" > /tmp/cachi2.yaml
+        configfile_flag=--config-file=/tmp/cachi2.yaml
+      else
+        configfile_flag=''
+      fi
+
       # Copied from https://github.com/konflux-ci/build-definitions/blob/main/task/git-clone/0.1/git-clone.yaml
       if [ "${WORKSPACE_GIT_AUTH_BOUND}" = "true" ] ; then
         if [ -f "${WORKSPACE_GIT_AUTH_PATH}/.git-credentials" ] && [ -f "${WORKSPACE_GIT_AUTH_PATH}/.gitconfig" ]; then

```

chmeliik:
> > The cachi2 config doesn't contain any options that would compromise SBOM accuracy, right?
> > Maybe the [goproxy_url](https://github.com/containerbuildsystem/cachi2/blob/34b72cc4beddbae6295ba56c66aa0ff8668bf73c/cachi2/core/config.py#L14), since that would allow the user to make cachi2 download through a [private proxy](https://go.dev/ref/mod#private-module-proxy-private). But I think the purl spec doesn't account for that anyway :-/
> 
> From PURL spec you wouldn't really know. As for the options, I'd imagine the idea to be that options can only modify the behaviour without deliberately causing a regressive behaviour on the output, IOW I can imagine existence of options improving the quality of the SBOM, not decreasing it, especially as a side effect.

So what does all this mean for the goproxy_url option?

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
eskultety: lovely, thank you @acmiel let me steal and apply that!

eskultety:
> > > The cachi2 config doesn't contain any options that would compromise SBOM accuracy, right?
> > > Maybe the [goproxy_url](https://github.com/containerbuildsystem/cachi2/blob/34b72cc4beddbae6295ba56c66aa0ff8668bf73c/cachi2/core/config.py#L14), since that would allow the user to make cachi2 download through a [private proxy](https://go.dev/ref/mod#private-module-proxy-private). But I think the purl spec doesn't account for that anyway :-/
> > 
> > 
> > From PURL spec you wouldn't really know. As for the options, I'd imagine the idea to be that options can only modify the behaviour without deliberately causing a regressive behaviour on the output, IOW I can imagine existence of options improving the quality of the SBOM, not decreasing it, especially as a side effect.
> 
> So what does all this mean for the goproxy_url option?

What it probably means is that cachi2 will either have to abuse the `repository_url/download_url` qualifier in the PURL if `goproxy_url` has been set OR have to invent a custom qualifier, e.g. `proxy=` in the Golang PURL (slightly preferable). However, just like with any privately hosted artifacts, cachi2 is only ever going to be able to report the "actual" locations where a machinery will need to be in place to convert them to "apparent" locations so as not to disclose private unresolvable URLs.

Zooming out slightly, should the `goproxy_url` be considered a blocker for this PR?

chmeliik:
> Zooming out slightly, should the `goproxy_url` be considered a blocker for this PR?

Not a blocker as long as it's tracked. Removing the option could be a good start :slightly_smiling_face:

eskultety:
> > Zooming out slightly, should the `goproxy_url` be considered a blocker for this PR?
> 
> Not a blocker as long as it's tracked. Removing the option could be a good start 🙂

Okay, I can 'sed' it out for now.

chmeliik:
> > > Zooming out slightly, should the `goproxy_url` be considered a blocker for this PR?
> > 
> > 
> > Not a blocker as long as it's tracked. Removing the option could be a good start 🙂
> 
> Okay, I can 'sed' it out for now.

I meant removing it from cachi2 until it's needed / until cachi2 can report it. But yeah, trying to avoid usage of it until it can be removed from cachi2 is probably a good idea.

eskultety:
> > > > Zooming out slightly, should the `goproxy_url` be considered a blocker for this PR?
> > > 
> > > 
> > > Not a blocker as long as it's tracked. Removing the option could be a good start 🙂
> > 
> > 
> > Okay, I can 'sed' it out for now.
> 
> I meant removing it from cachi2 until it's needed / until cachi2 can report it. But yeah, trying to avoid usage of it until it can be removed from cachi2 is probably a good idea.

Oh, then I misunderstood. That would break backwards incompatibility though, in general we don't know whether any option is in active use anywhere (like yeah, nobody is likely using this one, but still...), so disabling it in the tekton task seems like a more practical workaround that can be replaced with a proper solution at any time.

chmeliik:
> Oh, then I misunderstood. That would break backwards compatibility though, in general we don't know whether any option is in active use anywhere (like yeah, nobody is likely using this one, but still...), so disabling it in the tekton task seems like a more practical workaround that can be replaced with a proper solution at any time.

Yeah, that's fine. I'll just mention that the sed command will have to be pretty gnarly if we want it robust (one could e.g. use YAML multiline strings to try and avoid detection, pass the file formatted as JSON, ...). Maybe just don't bother with it and accept that it's an inherent shortcoming in the current version of cachi2

eskultety:
> > Oh, then I misunderstood. That would break backwards compatibility though, in general we don't know whether any option is in active use anywhere (like yeah, nobody is likely using this one, but still...), so disabling it in the tekton task seems like a more practical workaround that can be replaced with a proper solution at any time.
> 
> Yeah, that's fine. I'll just mention that the sed command will have to be pretty gnarly if we want it robust (one could e.g. use YAML multiline strings to try and avoid detection, pass the file formatted as JSON, ...). 

Hmm, is `yq` pre-installed in the prefetch env? Because if it is and provided the file is formatted correctly within the format spec, how about this:

```
$ cat $config_file
requests_timeout: |
  1200
  foo 300

foo: >
  bar
  baz

$ yq --output-format props $config_file | sed '/requests_timeout/d' | yq --input-format props --output-format yaml > config.yaml
```

That way we're relying on `yq` doing the heavy lifting and with `props` output format it seems we should always get 1 line per option. In case you want some extra handling of random garbage being passed then I guess we could prepend
```
yq -pa e $config_file
```
and see if `yq` can chew it and spit out something, not sure how far we want to take this though. Oh yeah, and this all just assumes we have more than just basic shell builtins and commands available.

chmeliik:
> Hmm, is `yq` pre-installed in the prefetch env?

yq would make this very easy, yeah: `yq < $config_file 'del(.goproxy_url)'`

But it is not installed in the prefetch env (the cachi2 container) https://github.com/konflux-ci/build-definitions/blob/a326151eb9bc540ed6c0a2c2cad0629c2ebf99df/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L34

There is a konflux-ci/yq container though: https://github.com/konflux-ci/build-definitions/blob/a326151eb9bc540ed6c0a2c2cad0629c2ebf99df/task/build-vm-image/0.1/build-vm-image.yaml#L67

To make use of it, you would have to:
* add a step that uses this image, deletes the option from the config file and writes it to a shared directory
  * first, define that shared directory, as is done for example here (the `volumes` section declares the volume, `stepTemplate` makes sure it gets mounted to every step): https://github.com/konflux-ci/build-definitions/blob/a326151eb9bc540ed6c0a2c2cad0629c2ebf99df/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml#L58-L84
* in the second step (the current `prefetch-dependencies` step), pass `--config-file=<path in shared directory>`

eskultety:
@acmiel thanks for the guidance, I tried adjusting the task as advised.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -30,7 +34,27 @@ spec:
...
     description: The name of the key in the ConfigMap that contains the CA bundle data.
     default: ca-bundle.crt
+
+  stepTemplate:
+    env:
+    - name: CONFIG_FILE
+      value: $(params.config-file)
+    volumeMounts:
+      - name: config
+        mountPath: /mnt/config
```
chmeliik, line 44: The declaration of the volume is missing:
```suggestion
        mountPath: /mnt/config

  volumes:
    - name: config
      emptyDir: {}
```

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -30,7 +34,27 @@ spec:
...
+
   steps:
+  - name: sanitize cachi2 config file with yq
```
chmeliik, line 47:
```suggestion
  - name: sanitize-cachi2-config-file-with-yq
```
I think step names must be valid kubernetes names

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -62,6 +86,12 @@ spec:
         exit 0
       fi
 
+      if [ -f /mnt/config/config.yaml ]; then
+        config_flag=--config-flag /mnt/config/config.yaml
```
chmeliik, line 90:
```suggestion
        config_flag=--config-file=/mnt/config/config.yaml
```
- typo in name
- I think the space makes this a `SOME_VARIABLE=foo some-command` situation (sets a variable and invokes a command, doesn't set a variable that contains a space)

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -22,6 +22,10 @@ spec:
   - description: Set cachi2 log level (debug, info, warning, error)
     name: log-level
     default: "info"
+  - description: |
+      Pass configuration to cachi2.
+      Note this needs to be passed as a YAML-formatted config dump, not as a file path!
+    name: config-file-content
```
chmeliik, line 28:
```suggestion
    name: config-file-content
    default: ""
```
forgot we need a default value

chmeliik (approved):
LGTM :+1:

PR #1171, opened 2024-07-17
Enable users to be able to provide their own config.toml

arewm:
This removes the default configuration (which also removes a default user and password from being injected into all images), instead deferring to the config.toml which is present in the source repository.

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -26,6 +26,10 @@ spec:
       default: bib.yaml
       type: string
       description: The config file specifying what to build and the builder to build it with
+    - name: CONFIG_TOML_FILE
+      default: ""
+      type: string
+      description: The path for the config.toml file within the source repository
     - default: etc-pki-entitlement
       description: Name of secret which contains the entitlement certificates
       name: ENTITLEMENT_SECRET
@@ -45,6 +49,8 @@ spec:
         value: $(params.OUTPUT_IMAGE)
       - name: BIB_CONFIG_FILE
         value: $(params.BIB_CONFIG_FILE)
+      - name: CONFIG_TOML_FILE
+        value: $(params.CONFIG_TOML_FILE)
       - name: IMAGE_TYPE
         value: $(params.IMAGE_TYPE)
       - name: ENTITLEMENT_SECRET
@@ -148,6 +154,22 @@ spec:
         echo "$BUILD_DIR"
         ssh -v $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp" "$BUILD_DIR/tekton-results" "$BUILD_DIR/entitlement"
 
+        if [ ! -n "${CONFIG_TOML_FILE}" ]; then
+          echo "No CONFIG_TOML_FILE specified"
+          export CONFIG_TOML_FILE=config.toml
+          if [ -f /var/workdir/source/config.toml ]; then
+            echo "Using the config.toml file found in the repository root!"
+            echo "  Remove the config.toml file or set params.CONFIG_TOML_FILE to another file to prevent using config.toml."
+          else
+            echo "No config.toml file found. Using an empty configuration."
+            touch /var/workdir/source/$CONFIG_TOML_FILE
+          fi
+        fi
+        echo "Using the following config.toml file $CONFIG_TOML_FILE:"
+        cat /var/workdir/source/$CONFIG_TOML_FILE
+
+
+        rsync -ra "/var/workdir/source/$CONFIG_TOML_FILE" "$SSH_HOST:$BUILD_DIR/config.toml"
         rsync -ra "$HOME/.docker/" "$SSH_HOST:$BUILD_DIR/.docker/"
         rsync -ra /entitlement/ "$SSH_HOST:$BUILD_DIR/entitlement/"
 
@@ -175,12 +197,6 @@ spec:
 
         # this quoted heredoc prevents expansions and command substitutions. the env vars are evaluated on the remote vm
         cat >>scripts/script-build.sh <<'REMOTESSHEOF'
-        echo >config.toml <<EOF
-        [[blueprint.customizations.user]]
-        name = "user"
-        password = "pass"
-        groups = ["wheel"]
-        EOF
 
         mkdir output
         echo "PULLING BUILDER IMAGE"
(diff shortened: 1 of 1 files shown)


task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,9 @@ spec:
         echo "$BUILD_DIR"
         ssh -v $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp" "$BUILD_DIR/tekton-results" "$BUILD_DIR/entitlement"
 
+        # ensure that a config toml file is present in case one is not provided
+        touch /var/workdir/source/$CONFIG_TOML_FILE
```
pastequo, line 158: would that work if someone specify `CONFIG_TOML_FILE=a/b/c/config.toml` but none of that exist ?

It should probably fail anyway

arewm: It will not work, no. Not without the `mkdir -p`.

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -26,6 +26,10 @@ spec:
       default: bib.yaml
       type: string
       description: The config file specifying what to build and the builder to build it with
+    - name: CONFIG_TOML_FILE
+      default: "config.toml"
+      type: string
+      description: The name of the config.toml file within the source repository's root
```
chmeliik, line 32: It doesn't actually need to in the repo root, does it?

task/build-vm-image/0.1/build-vm-image.yaml
chmeliik, line 158: If the user does have a config file but messes up the file path, their error will be silently ignored. Let's at least log a warning?

arewm: What do you mean? If they mess up the file path then the config file will be created and used. I will print out the config that we are using after the file is touched.

chmeliik: If I have a `cool-config.toml` that actually sets some options and I forget to override the default `config.toml`, the task will create a completely empty config and I'll be wondering why my config is being ignored

arewm: How should we cover that? I added a cat of the toml file, do you think that should be sufficient?

chmeliik: That's probably fine. Maybe also log a big WARNING when we create a previously non-existent file

Alternatively, the option could be smarter
* Empty string by default
  * But if `config.toml` exists use it anyway, otherwise create an empty config
* If not empty string, don't create an empty config. The user really wants their config file to be used

arewm: Resolved in the latest revision.

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,24 @@ spec:
...
+            echo "Using the config.toml file found in the repository root!"
+            echo "  Remove the config.toml file or set params.CONFIG_TOML_FILE to an invalid path to prevent its use."
+          else
+            echo "No config.toml file found. Set params.CONFIG_TOML_FILE to use one in the repository."
+            export CONFIG_TOML_FILE=config.toml
+          fi
+        fi
+        # ensure that a config toml file is present in case one is not provided or the path is invalid
+        mkdir -p /var/workdir/source/$(dirname $CONFIG_TOML_FILE)
+        touch /var/workdir/source/$CONFIG_TOML_FILE
```
chmeliik, line 169: > or the path is invalid

The point of the changes suggested in https://github.com/konflux-ci/build-definitions/pull/1171#discussion_r1682771804 was to not do this (if the user specifies a wrong path, the task should fail). Otherwise the current code is just a more complex way to achieve the same behavior as before :smile:

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,23 @@ spec:
         echo "$BUILD_DIR"
         ssh -v $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp" "$BUILD_DIR/tekton-results" "$BUILD_DIR/entitlement"
 
+        if [! -n "${CONFIG_TOML_FILE}" ]
```
pastequo, line 157:
```suggestion
        if [! -n "${CONFIG_TOML_FILE}" ]; then
```

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,22 @@ spec:
         echo "$BUILD_DIR"
         ssh -v $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/tmp" "$BUILD_DIR/tekton-results" "$BUILD_DIR/entitlement"
 
+        if [! -n "${CONFIG_TOML_FILE}" ]; then
```
chmeliik, line 157:
```suggestion
        if [ ! -n "${CONFIG_TOML_FILE}" ]; then
```
otherwise, I think this is a `[!: command not found`

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,22 @@ spec:
...
+          echo "No CONFIG_TOML_FILE specified"
+          export CONFIG_TOML_FILE=config.toml
+          if [ -f /var/workdir/source/config.toml ]; then
+            echo "Using the config.toml file found in the repository root!"
+            echo "  Remove the config.toml file or set params.CONFIG_TOML_FILE to an invalid path to prevent its use."
```
chmeliik, line 162: setting CONFIG_TOML_FILE to an invalid path is no longer an option

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -148,6 +154,22 @@ spec:
...
+          else
+            echo "No config.toml file found. Set params.CONFIG_TOML_FILE to use an empty configuration."
```
chmeliik, line 164:
```suggestion
            echo "No config.toml file found. Using an empty configuration."
```

chmeliik (commented):
LGTM apart from one bash error and some log output details

Would be nice to squash the commits into logical units (probably just one commit)

task/build-vm-image/0.1/build-vm-image.yaml
arewm: You are right. They would have to create an empty config file and point to that.

chmeliik: Or just remove the default config.toml (unless for whatever reason they want to keep it and make the build ignore it)

chmeliik (approved):

MartinBasti:
https://github.com/konflux-ci/build-definitions/pull/1165/checks?check_run_id=2785204219 

I thinks that PR causes that all PRs are failing on linter error ^, I don't know why this one passed the test though

IMO this is the faulty line and linter makes it as false positive report
https://github.com/arewm/build-definitions/blob/c8f32dff3fa102786d30175e10f66eaa9260f29a/task/build-vm-image/0.1/build-vm-image.yaml#L162

arewm:
@chmeliik has a fix for that in #1189.

PR #1172, opened 2024-07-17
prevent using host subscription

rcerven:
# Before you complete this pull request ...

Look for any open pull requests in the repository with the title "e2e-tests update" and 
see if there are recent e2e-tests updates that will be applicable to your change.

PR #1173, opened 2024-07-18
add support for subscription activation keys to trusted artifacts

brianwcook:
… tasks, including buildah-remoate-oci-ta. It also disables podman subscription-manager integration on all remote builders so that subscription-manager commands can be used to configure which repos are enabled.  

i had recently had a couple interactions with people smart about subscription mgmt and they all said that using certs the way we are is wrong and eventually will lead to random build failures due to revoked certs. Activation keys solves that problems, and it solves the problem that Camilla was having in this thread, so I just went ahead and did it.

ref slack thread https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1721226430548109

### DIFF
--- a/task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
+++ b/task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
@@ -59,6 +59,10 @@ spec:
     description: Name of secret which contains the entitlement certificates
     name: ENTITLEMENT_SECRET
     type: string
+  - default: activation-key
+    description: Name of secret which contains subscription activation key
+    name: ACTIVATION_KEY
+    type: string
   - default: "false"
     description: Determines if build will be executed without network access.
     name: HERMETIC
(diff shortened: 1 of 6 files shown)


brianwcook:
basic documentation is here: https://docs.google.com/document/d/1mBCAaIU_4XmvCRoXe-Thg3uIEDAs5yPrWVCx3KPopWY/edit

brianwcook:
I guess the changes need to be moved from TA tasks to the PVC based ones.

brianwcook:
@rcerven if you are able to help get this in shape to merge, I think it will solve a lot of problems. thanks :)

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -413,13 +427,15 @@ spec:
       chmod +x scripts/script-build.sh
       rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
       ssh $SSH_ARGS "$SSH_HOST" $PORT_FORWARD podman  run $PODMAN_PORT_FORWARD \
+       --tmpfs /run/secrets \
```
chmeliik, line 430: Same as https://github.com/konflux-ci/build-definitions/pull/1172#discussion_r1682659433, this breaks the ADDITIONAL_SECRET feature

chmeliik: Ah, nevermind - this is a mount for the podman container that runs the build, not for the build itself

chmeliik (approved):

PR #1178, opened 2024-07-22
rpm-ostree: drop BASE_IMAGES_DIGESTS result

chmeliik:
[STONEBLD-2326](https://issues.redhat.com//browse/STONEBLD-2326)

For consistency with the buildah tasks, don't return BASE_IMAGES_DIGESTS from rpm-ostree either.

### DIFF
--- a/task/rpm-ostree/0.2/MIGRATION.md
+++ b/task/rpm-ostree/0.2/MIGRATION.md
@@ -1,0 +1,43 @@
+# Migration from 0.1 to 0.2
+
+Version 0.2:
+
+* Removes the `BASE_IMAGES_DIGESTS` result. Please remove all the references to this
+  result from your pipeline.
+  * Note that the result never provided any value. It did not contain base image
+    digests, it contained a local-only reference for the **output** image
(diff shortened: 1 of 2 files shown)


zregvart (approved):

chmeliik:
Note to self: will need to rebase once #1111 makes it through the merge queue

zregvart (approved):

chmeliik:
There's currently an internal Slack thread ongoing about EC blocking releases of images that don't have a BASE_IMAGES_DIGESTS result and have no base images in the SBOM. I'll mark this as draft until that's resolved

chmeliik:
> There's currently an internal Slack thread ongoing about EC blocking releases of images that don't have a BASE_IMAGES_DIGESTS result and have no base images in the SBOM (which would be the expected state for this task in 0.2). I'll mark this as draft until that's resolved

This was a different issue (missing `vendor="Red Hat, inc."` label, due to which EC didn't read the SBOM embedded in the image).

Rebased on main, un-drafted

PR #1183, opened 2024-07-23
Add Checkton workflow

chmeliik:
Checkton is a GitHub action that runs ShellCheck on scripts embedded in YAML files (https://github.com/chmeliik/checkton)

### DIFF
--- a/.github/workflows/checkton.yaml
+++ b/.github/workflows/checkton.yaml
@@ -1,0 +1,33 @@
+name: Checkton
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          # Differential Checkton requires full git history
+          fetch-depth: 0
+
+      - name: Run Checkton
+        id: checkton
+        uses: chmeliik/checkton@v0.2.1
+        with:
+          # Set to false when re-enabling SARIF uploads
+          fail-on-findings: true
+          find-copies-harder: true
+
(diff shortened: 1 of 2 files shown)


chmeliik:
FYI, I've opened a proposal for Checkton to move to the tektoncd org: https://github.com/tektoncd/community/issues/1156

But until that's resolved, let's start using it already?

chmeliik:
The output looks like this:

image

We could also upload the generated SARIF file to make GitHub show the warnings inline. But currently, there is an annoying bug (or feature?) where the warnings never go away even if you fix them - https://github.com/orgs/community/discussions/132787.

zregvart (approved):

chmeliik:
Tests timed out while waiting for the enterprise-contract pipeline to pass

Tekton was very lazy today, just refusing to schedule any steps. Even the Tekton built-in ones

```
[verify-enterprise-contract : prepare] 2024/07/23 11:36:28 Entrypoint initialization

[verify-enterprise-contract : place-scripts] 2024/07/23 12:04:28 Decoded script /tekton/scripts/script-1-dfsnc

[verify-enterprise-contract : validate] time="2024-07-23T12:16:24Z"
```

tkdchen (approved):

tkdchen:
It doesn't need to run the Konflux pipeline when changes are made to `.github/` and `hack/`.

chmeliik:
Rebased on main

> It doesn't need to run the Konflux pipeline when changes are made to .github/ and hack/.

True, but we can't merge without it. I'll see if we can make it not run when those are the only two dirs that changed. There was an attempt to do this by enumerating all the directories for which we *do* want to run the pipeline (https://github.com/konflux-ci/build-definitions/pull/524), but that seemed very fragile.

PR #1185, opened 2024-07-23
Add an architecture suffix to images pushed for multi-platform if missing

arewm:
In order to reduce the likelihood of users accidentally forgetting to specify unique tags for each architecture, we can add a suffix to the pushed image

### DIFF
--- a/task-generator/remote/main.go
+++ b/task-generator/remote/main.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
@@ -33,9 +34,11 @@ import (
 func main() {
 	var buildahTask string
 	var buildahRemoteTask string
+	var taskVersion string
 
 	flag.StringVar(&buildahTask, "buildah-task", "", "The location of the buildah task")
 	flag.StringVar(&buildahRemoteTask, "remote-task", "", "The location of the buildah-remote task to overwrite")
+	flag.StringVar(&taskVersion, "task-version", "", "The version of the task to overwrite")
 
 	opts := zap.Options{
 		Development: true,
@@ -43,8 +46,8 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	klog.InitFlags(flag.CommandLine)
 	flag.Parse()
-	if buildahTask == "" || buildahRemoteTask == "" {
-		println("Must specify both buildah-task and remote-task params")
+	if buildahTask == "" || buildahRemoteTask == "" || taskVersion == "" {
+		println("Must specify both buildah-task, remote-task, and task-version params")
 		os.Exit(1)
 	}
 
@@ -53,7 +56,7 @@ func main() {
 
 	decodingScheme := runtime.NewScheme()
 	utilruntime.Must(tektonapi.AddToScheme(decodingScheme))
-	convertToSsh(&task)
+	convertToSsh(&task, taskVersion)
 	y := printers.YAMLPrinter{}
 	b := bytes.Buffer{}
 	_ = y.PrintObj(&task, &b)
@@ -87,7 +90,7 @@ func streamFileYamlToTektonObj(path string, obj runtime.Object) runtime.Object {
 	return decodeBytesToTektonObjbytes(bytes, obj)
 }
 
-func convertToSsh(task *tektonapi.Task) {
+func convertToSsh(task *tektonapi.Task, taskVersion string) {
 
 	builderImage := ""
 	syncVolumes := map[string]bool{}
@@ -96,14 +99,44 @@ func convertToSsh(task *tektonapi.Task) {
 			syncVolumes[i.Name] = true
 		}
 	}
+	// The images produced in multi-platform builds need to have unique tags in order
+	// to prevent them from getting garbage collected before generating the image index.
+	// We can simplify this process, preventing the need for users to manually specify
+	// the image by auto-appending the architecture from the PLATFORM parameter. For
+	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. Since
+	// we cannot modify the parameter itself, this replacement needs to happen in any task
+	// step where the IMAGE parameter is used.
+	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
+	// not be appended again based on the PLATFORM.
+	adjustRemoteImage := `if [[ "${IMAGE##*-}" != "${PLATFORM##*/}" ]]; then
+  export IMAGE="${IMAGE}-${PLATFORM##*/}"
+fi
+`
+
 	for stepPod := range task.Spec.Steps {
+		ret := ""
 		step := &task.Spec.Steps[stepPod]
-		if step.Name != "build" {
+		if step.Script != "" && taskVersion != "0.1" && step.Name != "build" {
+			scriptHeaderRE := regexp.MustCompile(`^#!/bin/bash\nset -e\n`)
+			if scriptHeaderRE.FindString(step.Script) != "" {
+				ret = scriptHeaderRE.ReplaceAllString(step.Script, "")
+			} else {
+				ret = step.Script
+			}
+			if !strings.HasPrefix(ret, "#!") {
+				// If there is a shebang, it is explicitly non-bash, so don't adjust the image
+				ret = "#!/bin/bash\nset -e\n" + adjustRemoteImage + ret
+			}
+			step.Script = ret
+			continue
+		} else if step.Name != "build" {
 			continue
 		}
 		podmanArgs := ""
 
-		ret := `set -o verbose
+		ret = `#!/bin/bash
+set -e
+set -o verbose
 mkdir -p ~/.ssh
 if [ -e "/ssh/error" ]; then
   #no server could be provisioned
@@ -130,7 +163,9 @@ PORT_FORWARD=" -L 80:$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR:80"
 PODMAN_PORT_FORWARD=" -e JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR=localhost"
 fi
 `
-
+		if taskVersion != "0.1" {
+			ret += adjustRemoteImage
+		}
 		env := "$PODMAN_PORT_FORWARD \\\n"
 
 		// disable podman subscription-manager integration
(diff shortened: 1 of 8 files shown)


task/buildah-remote/0.1/buildah-remote.yaml
```diff
@@ -238,9 +240,7 @@ spec:
       rsync -ra "$HOME/.docker/" "$SSH_HOST:$BUILD_DIR/.docker/"
       rsync -ra "/tekton/results/" "$SSH_HOST:$BUILD_DIR/tekton-results/"
       cat >scripts/script-build.sh <<'REMOTESSHEOF'
-      #!/bin/bash
```
MartinBasti, line 241: This shouldn't be removed, it's part of script on the remote host

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -242,6 +242,8 @@ spec:
         - name: COMMIT_SHA
           value: $(params.COMMIT_SHA)
       script: |
+        #!/bin/bash
+        set -e
```
ralphbean, line 246: Why the addition of `-e`?

task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml
```diff
@@ -251,8 +258,8 @@ spec:
       rsync -ra "/tekton/results/" "$SSH_HOST:$BUILD_DIR/tekton-results/"
       cat >scripts/script-build.sh <<'REMOTESSHEOF'
       #!/bin/bash
-      set -o verbose
       set -e
+      set -o verbose
```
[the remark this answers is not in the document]
arewm: The easiest way to move this up is if I move out the addition of this line from the remote build generation and instead include it in the base buildah tasks. Is there a reason that we would only want this in the remote tasks?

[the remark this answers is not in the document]
arewm: This is resolved. I will no longer add setting verbose output. If you want verbose output in the remote script, you will need to add it in the base script.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
arewm: Ah, this is re-added by the trusted artifact generation. :)

task/buildah/0.2/buildah.yaml
```diff
@@ -524,6 +526,7 @@ spec:
         echo -n "${IMAGE}@"
         cat "$(workspaces.source.path)/image-digest"
       } > "$(results.IMAGE_REF.path)"
+      cp "$(results.IMAGE_REF.path)" /shared/image_ref
```
chmeliik, line 529: Is the copy needed?

task-generator/remote/main.go
```diff
@@ -18,6 +18,7 @@ import (
 	"flag"
```
chmeliik, line 18: I'm having a hard time trying to understand the task-generator changes. Could you explain them in the commit message?

task/buildah/0.2/buildah.yaml
arewm: In testing, the `results.IMAGE_REF.path` wasn't available at the later task step so I copied the reference to the shared workspace so I could use it later.

I cannot find specific documentation for this behavior in Tekton.

chmeliik: Hmm, I don't think Tekton behaves like that. Could you test it again? Maybe the nonexistent filepath was just caused by incorrect attempts at shell quoting

task-generator/remote/main.go
arewm: Sure, I can add some comments into the code.

task/buildah/0.2/buildah.yaml
arewm: Yes, certainly!

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
brianwcook: We have seen scripts encounter errors but not fail when 'set -e' is not there.

task-generator/remote/main.go
```diff
@@ -96,14 +99,34 @@ func convertToSsh(task *tektonapi.Task) {
 			syncVolumes[i.Name] = true
 		}
 	}
+	// The images produced in multi-platform builds need to have unique tags in order
+	// to prevent them from getting garbage collected before generating the image index.
+	// We can simplify this process, preventing the need for users to manually specify
+	// the image by auto-appending the architecture from the PLATFORM parameter. For
+	// example, this will append -arm64 if PLATFORM is linux/arm64. Since we cannot
+	// modify the parameter itself, this replacement needs to happen in any task step
+	// where the IMAGE parameter is used.
```
brianwcook, line 108: so for existing builds, will we end up with $image-s390x-s390x (since they must have manually added the suffix already?) I suppose that is ugly but harmless?

arewm: No, I made the comment clearer. If the suffix does not exist on the `IMAGE` then we will add it. If the suffix does exist, we will not duplicate it.

task-generator/remote/main.go
```diff
@@ -160,9 +187,19 @@ fi
 		script := "scripts/script-" + step.Name + ".sh"
 
 		ret += "\ncat >" + script + " <<'REMOTESSHEOF'\n"
-		if !strings.HasPrefix(step.Script, "#!") {
+
+		// The base task might now be using a bash shell, so we need to make sure
+		// that we only have one shebang declaration. If there is a shebang declaration,
+		// we should also consolidate the set declarations.
+		reShebang := regexp.MustCompile(`(#!.*\n)(set -.*\n)*`)
```
chmeliik, line 194: This is for the script that ends up getting executed on the VM, right? I don't think we should mess with its shebang :thinking: 

Only if it doesn't have one at all, in which case we should prepend the same thing that Tekton would
```
#!/bin/sh
set -e
```

arewm: Without this change, the `cd` operation is in the wrong location:
```
       cat >scripts/script-build.sh <<'REMOTESSHEOF'
+      cd /var/workdir
       #!/bin/bash
       set -e
-      cd /var/workdir
```

arewm: This is because we are adding the shebang to the beginning of the script in the base task so we can no longer simply concatenate it after setting the workingDir.

chmeliik: Ah, right

task-generator/remote/main.go
```diff
@@ -96,14 +99,36 @@ func convertToSsh(task *tektonapi.Task) {
 			syncVolumes[i.Name] = true
 		}
 	}
+	// The images produced in multi-platform builds need to have unique tags in order
+	// to prevent them from getting garbage collected before generating the image index.
+	// We can simplify this process, preventing the need for users to manually specify
+	// the image by auto-appending the architecture from the PLATFORM parameter. For
+	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. Since
+	// we cannot modify the parameter itself, this replacement needs to happen in any task
+	// step where the IMAGE parameter is used.
```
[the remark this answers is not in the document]
arewm: I have changed it to add the modification to any script which isn't explicitly defined as non-bash (i.e. not to python scripts)

task/buildah/0.2/buildah.yaml
arewm: I have removed this.

task-generator/remote/main.go
```diff
@@ -96,14 +99,42 @@ func convertToSsh(task *tektonapi.Task) {
...
+	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
+	// not be appended again based on the PLATFORM.
+	adjustRemoteImage := "if [[ \"${IMAGE##*-}\" != \"${PLATFORM##*/}\" ]]; then"
+	adjustRemoteImage += "\n  export IMAGE=\"${IMAGE}-${PLATFORM##*/}\"\nfi\n"
+
 	for stepPod := range task.Spec.Steps {
+		ret := ""
 		step := &task.Spec.Steps[stepPod]
-		if step.Name != "build" {
+		if step.Script != "" && taskVersion == "0.2" && step.Name != "build" {
```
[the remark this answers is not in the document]
arewm: We cannot easily add this to v0.1 because we don't have `IMAGE_REF` to fall back on to when we change the image tag that is produced.

task-generator/remote/main.go
```diff
@@ -96,14 +99,42 @@ func convertToSsh(task *tektonapi.Task) {
...
+			scriptHeaderRE := regexp.MustCompile(`#!/bin/bash\nset -e\n`)
+			if scriptHeaderRE.FindString(step.Script) != "" {
+				ret = scriptHeaderRE.ReplaceAllString(step.Script, "")
+			} else {
+				ret = step.Script
+			}
+			if !strings.HasPrefix(ret, "#!") {
+				// If there is a shebang, it is explicitly non-bash, so don't adjust the image
```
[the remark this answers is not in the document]
arewm: If it is not `bash`, we cannot inject the remote image adjustment. We cannot design for all potential cases that we might encounter so we just bail out and then can address the cases in the future.

task-generator/remote/main.go
```diff
@@ -96,14 +99,42 @@ func convertToSsh(task *tektonapi.Task) {
...
+	// to prevent them from getting garbage collected before generating the image index.
+	// We can simplify this process, preventing the need for users to manually specify
+	// the image by auto-appending the architecture from the PLATFORM parameter. For
+	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. Since
+	// we cannot modify the parameter itself, this replacement needs to happen in any task
+	// step where the IMAGE parameter is used.
+	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
+	// not be appended again based on the PLATFORM.
+	adjustRemoteImage := "if [[ \"${IMAGE##*-}\" != \"${PLATFORM##*/}\" ]]; then"
+	adjustRemoteImage += "\n  export IMAGE=\"${IMAGE}-${PLATFORM##*/}\"\nfi\n"
```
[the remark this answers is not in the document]
arewm: Done

task-generator/remote/main.go
```diff
@@ -96,14 +99,42 @@ func convertToSsh(task *tektonapi.Task) {
...
+
 	for stepPod := range task.Spec.Steps {
+		ret := ""
 		step := &task.Spec.Steps[stepPod]
-		if step.Name != "build" {
+		if step.Script != "" && taskVersion == "0.2" && step.Name != "build" {
+			scriptHeaderRE := regexp.MustCompile(`#!/bin/bash\nset -e\n`)
+			if scriptHeaderRE.FindString(step.Script) != "" {
+				ret = scriptHeaderRE.ReplaceAllString(step.Script, "")
```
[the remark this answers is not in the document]
arewm: Just in case that happens, we shouldn't try to do anything with it. I added a `^`.

task-generator/remote/main.go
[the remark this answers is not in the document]
lcarva: Can we mark the other versions as deprecated?

We did this [here](https://github.com/konflux-ci/build-definitions/blob/193d62a02899e25ddd92b3335f54a1110f757de2/task/deprecated-image-check/0.2/deprecated-image-check.yaml#L9).

That makes EC not trust those Tasks any longer (at that given time). This should be sufficient to make users upgrade.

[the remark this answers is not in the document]
arewm: Interesting. We should probably mark all tasks which produce or consume the `BASE_IMAGE_DIGESTS` as deprecated together. I think that should be done outside of the current PR.

mmorhun (approved):

PR #1188, opened 2024-07-24
.tekton/pull-request: run conditionally

chmeliik:
Don't run the pipeline if the only changed files are

- `*.md`
- `.github/*`
- `renovate.json`

Note that a similar attempt was made in https://github.com/konflux-ci/build-definitions/pull/524, which took the approach of enumerating every file/directory which _should_ trigger the pipeline (maybe `files.all` wasn't available at the time). This PR takes the opposite approach.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -4,7 +4,15 @@ kind: PipelineRun
 metadata:
   name: build-definitions-pull-request
   annotations:
-    pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
+    pipelinesascode.tekton.dev/on-cel-expression: |
+      (
+          // PR to "main" branch, not draft
+          event == "pull_request" && target_branch == "main"
+          && (!has(body.pull_request) || !body.pull_request.draft)
+      ) || (
+          // PR enters the merge queue
+          event == "push" && target_branch.startsWith("gh-readonly-queue/main/")
+      )
     pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/yaml-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.1/sast-snyk-check.yaml]"
     pipelinesascode.tekton.dev/task-2: "yaml-lint"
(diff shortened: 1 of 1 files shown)


chmeliik:
Tested the expression in https://playcel.undistro.io/ with variations of the following input

```yaml
# Here is the input data in YAML or JSON format.

event: push
target_branch: gh-readonly-queue/main/

body:
  pull_request:
    draft: false

files:
  all:
  - "README.md"
  - "renovate.json"
  - ".github/workflows/checkton.yaml"
```

.tekton/pull-request.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
       ) && (
           // Don't run pipeline if only these files changed. Note that 'files.all' is an object
           // coming from PaC (list of all changed files) while the second 'all' is a CEL macro.
-          !files.all.all(x, x.matches(r"^(renovate\.json|.*\.md|\.github/.*)$"))
+          !files.all.all(x, x.matches(r"^(renovate\.json|.*\.md|\.github/.*|\.tekton/.*)$"))
```
MartinBasti, line 20: Why do we want to exclude .tekton?

I'd like to see if changes in tekton are valid and will pass testing before merging them

chmeliik: Yeah I'm just testing if the exclusion works (and found out that `Red Hat Konflux / build-definitions-pull-request` somehow became a required check, so we'll need to disable that)

chmeliik: \o/ thanks @arewm for removing the required check, now PRs wil no longer be blocked in cases where the .tekton/pull-request pipeline doesn't need to run

chmeliik: Dropped the `temp` commit, ready for review

chmeliik:
As @arewm found out, making the Konflux pipeline a not-required check means that if you press the `Merge when ready` button, github will not wait for the pipeline to finish.

There have been many requests for GitHub to make the "required check is skippable" behavior sane (https://github.com/orgs/community/discussions/13690), but until that happens, we can't go with this approach.

zregvart:
Perhaps this could be an enhancement in Pipelines as Code, mark skipped Pipelines as successful?

zregvart:
https://github.com/openshift-pipelines/pipelines-as-code/issues/1746 if anyone wants to follow up with comments.

chmeliik:
> [openshift-pipelines/pipelines-as-code#1746](https://github.com/openshift-pipelines/pipelines-as-code/issues/1746) if anyone wants to follow up with comments.

Nice, thanks for filing that. For our use case, we can probably add something to the pipeline to skip just the e2e-tests task, as that's the main one we really care about.

But this does seem like it could be a useful generic PaC feature

PR #1189, opened 2024-07-24
CI: avoid false-positive $(params.*) detection

chmeliik:
Previously, the yaml-lint task (the custom script that checks for $(params.*) usage) would flag false positives such as

    echo "set params.SOME_PARAM to do X"

Make the regex match actual param usage more closely

### DIFF
--- a/.tekton/tasks/yaml-lint.yaml
+++ b/.tekton/tasks/yaml-lint.yaml
@@ -38,7 +38,7 @@ spec:
       script: |
         #!/bin/bash
         for task in $(find task -name '*.yaml'); do
-          if yq '.spec?.steps[] | .script' $task | grep -q 'params\.'; then
+          if yq '.spec?.steps[] | .script' $task | grep -q '\$(params\.'; then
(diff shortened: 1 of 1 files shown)


chmeliik:
Yaml-lint (well, the yaml-lint task, not actual yaml-lint) was flagging https://github.com/konflux-ci/build-definitions/blob/ef146c45dbdc04bad190e3c8538fb73da48f94a9/task/build-vm-image/0.1/build-vm-image.yaml#L162 as param usage

(But only sometimes. The check often doesn't work for some unfathomable reason https://issues.redhat.com/browse/STONEBLD-2167)

arewm (approved):

MartinBasti (approved):

MartinBasti:
/retest

Maybe deleting deleted pipeline shouldn't trigger error
```
      failed to delete PipelineRun "test-comp-jdfh-qxt2s" from "build-templates-e2e" namespace with error: pipelineruns.tekton.dev "test-comp-jdfh-qxt2s" not found

```

PR #1192, opened 2024-07-24
Create the SBOM_BLOB_URL for v0.2 buildah tasks

arewm:
Now that the BASE_IMAGE_DIGESTS result has been removed, there should be enough room for us to re-add the SBOM_BLOB_URL. This will enable EC verification of the SBOM based on the digest which is recorded in the provenenance. It will prevent supply-chain attacks which are driven by modifying the floating tag of the uploaded SBOM.

Even with the referrer's API, the digest can be used to identify which SBOM should be the one built from Konflux.

This was added in #645 and then removed in #654 due to the limitation of Tekton results' size.

### DIFF
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -111,6 +111,9 @@ spec:
     name: IMAGE_URL
   - description: Image reference of the built image
     name: IMAGE_REF
+  - name: SBOM_BLOB_URL
+    description: Reference of SBOM blob digest to enable digest-based verification from provenance
+    type: string
   - name: SBOM_JAVA_COMPONENTS_COUNT
     description: The counting of Java components by publisher in JSON format
     type: string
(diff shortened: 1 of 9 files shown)


chmeliik (commented):
I'm not convinced this solves any immediate issues, since EC also reads the SBOM embedded in the image.

But adding the SBOM_BLOB_URL does sound good to me. Can we also stop injecting the SBOM into the image (maybe bump the version to 0.3 for responsible versioning?)

task/buildah/0.2/buildah.yaml
```diff
@@ -111,6 +111,9 @@ spec:
     name: IMAGE_URL
   - description: Image reference of the built image
     name: IMAGE_REF
+  - name: SBOM_BLOB_URL
+    description: Reference of SBOM digest to enable digest-based verification from provenance
```
chmeliik, line 115:
```suggestion
    description: Reference of SBOM blob to enable digest-based verification from provenance
```
Maybe? "Reference of SBOM digest" doesn't fully make sense to me

arewm: Updated it to the blob digest

chmeliik:
>  Can we also stop injecting the SBOM into the image (maybe bump the version to 0.3 for responsible versioning?)

For open-source-ness: discussed on internal slack, decided to defer this to a separate change because it could have some implications

rcerven (approved):

PR #1195, opened 2024-07-25
buildah-rhtap: drop BASE_IMAGES_DIGESTS

chmeliik:
STONEBLD-2326

The result is a Tekton anti-pattern (an unbounded list of items). It can easily cause builds to hit the result size limit.

In the sample RHTAP pipeline, nothing uses this result. Drop it. (In version 0.2 of the task).

### DIFF
--- a/task/buildah-rhtap/0.2/MIGRATION.md
+++ b/task/buildah-rhtap/0.2/MIGRATION.md
@@ -1,0 +1,6 @@
+# Migration from 0.1 to 0.2
+
+Version 0.2:
+
+* Removes the `BASE_IMAGES_DIGESTS` result. Please remove the references to this
+  result (if any) from your pipeline.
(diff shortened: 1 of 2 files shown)


chmeliik:
Note that unlike the regular buildah task, this one doesn't put base images in the SBOM so there's no replacement. But the EC policies in RHTAP don't check base images at all, so that's probably fine.

tkdchen (approved):

tkdchen:
@chmeliik Is it time to merge this pr?

chmeliik:
Ah, I forgot about this one. Actually, let's wait a bit - RHTAP is currently doing the 1.2 release so I'd rather not introduce any changes at the moment

lcarva:
This PR is now obsolete.

PR #1197, opened 2024-07-25
Readme: prefetch-deps: example of usage of configfile

MartinBasti:
Add example how config file should be used

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.1/README.md
+++ b/task/prefetch-dependencies-oci-ta/0.1/README.md
@@ -5,6 +5,24 @@ application source code are stored as a trusted artifact in the provided OCI rep
 For additional info on Cachi2, see docs at
 https://github.com/containerbuildsystem/cachi2#basic-usage.
 
+## Configuration
+
+Config file must be passed as a YAML string. For all available config options please check
+[available configuration parameters] page.
+
+Example of setting timeouts:
+
+```yaml
+params:
+  - name: config-file-content
+    value: |
+      ---
+      requests_timeout: 300
+      subprocess_timeout: 3600
+```
+
+[available configuration parameters]: https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#available-configuration-parameters
+
 ## Parameters
 |name|description|default value|required|
 |---|---|---|---|
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -11,6 +11,24 @@ spec:
   description: |-
     Task that uses Cachi2 to prefetch build dependencies.
     See docs at https://github.com/containerbuildsystem/cachi2#basic-usage.
+
+    ## Configuration
+
+    Config file must be passed as a YAML string. For all available config options please check
+    [available configuration parameters] page.
+
+    Example of setting timeouts:
+
+    ```yaml
+    params:
+      - name: config-file-content
+        value: |
+          ---
+          requests_timeout: 300
+          subprocess_timeout: 3600
+    ```
+
+    [available configuration parameters]: https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#available-configuration-parameters
   params:
   - description: Configures project packages that will have their dependencies prefetched.
     name: input
--- a/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
@@ -14,6 +14,24 @@ spec:
     application source code are stored as a trusted artifact in the provided OCI repository.
     For additional info on Cachi2, see docs at
     https://github.com/containerbuildsystem/cachi2#basic-usage.
+
+    ## Configuration
+
+    Config file must be passed as a YAML string. For all available config options please check
+    [available configuration parameters] page.
+
+    Example of setting timeouts:
+
+    ```yaml
+    params:
+      - name: config-file-content
+        value: |
+          ---
+          requests_timeout: 300
+          subprocess_timeout: 3600
+    ```
+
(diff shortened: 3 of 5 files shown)


task/prefetch-dependencies-oci-ta/0.1/README.md
```diff
@@ -18,6 +18,21 @@ https://github.com/containerbuildsystem/cachi2#basic-usage.
 |ociArtifactExpiresAfter|Expiration date for the trusted artifacts created in the OCI repository. An empty string means the artifacts do not expire.|""|false|
 |ociStorage|The OCI repository where the Trusted Artifacts are stored.||true|
 
+### Config file usage
```
MartinBasti, line 21: `hack/generate-ta-tasks.sh` tries to remove it from here

brunoapimentel:
Not sure if it's just for me, but when I check the README I don't see the link in `available configuration parameters` being rendered: 
https://github.com/konflux-ci/build-definitions/blob/6d5f1541435af4b363dc3914594cfc21978abb35/task/prefetch-dependencies-oci-ta/0.1/README.md

task/prefetch-dependencies-oci-ta/0.1/README.md
chmeliik: I think `generate-ta-tasks` uses the generate-readme script to generate the README. (And the next time someone runs that script manually, it would remove your changes). You would need to add this to the `description` of the task, not directly to the README

task/prefetch-dependencies-oci-ta/0.1/README.md
```diff
@@ -18,6 +18,21 @@ https://github.com/containerbuildsystem/cachi2#basic-usage.
...
+Config file must be passed as a YAML string. For all available config options please check [available configuration parameters]("https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#available-configuration-parameters") page.
+
+Example of setting timeouts:
+
+```yaml
+params:
+    - name: config-file-content
+      value: |
+         ---
+         requests_timeout: 300
```
chmeliik, line 32: Just an FYI (not a request to change anything either in cachi2 or the docs): the timeout has a very different meaning for `requests` and for `aiohttp`.

For `requests`, it's the timeout for establishing the connection and the timeout between receiving some bytes from the server https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

For `aiohttp`, it's the overall timeout for the whole download

As I found out in this flame-war of a thread https://github.com/aio-libs/aiohttp/issues/2249

eskultety: Yes it is, but I really didn't want to bother with connection and read timeout values in the aiohttp structure, because realistically the user cares only about the operation as a unit and from that POV the timeout here behaves the same in both cases, it's a cumulative value.

chmeliik: > (is it that very different though?)

My understanding was that `requests` with a timeout of 1 minute will happily keep downloading for an hour if the server sends something every minute

While `aiohttp` will abort after a minute

MartinBasti: This is wrong PR to discuss this :D

MartinBasti:
> Not sure if it's just for me, but when I check the README I don't see the link in `available configuration parameters` being rendered: https://github.com/konflux-ci/build-definitions/blob/6d5f1541435af4b363dc3914594cfc21978abb35/task/prefetch-dependencies-oci-ta/0.1/README.md

you are right, it worked only in my IDE :D

task/prefetch-dependencies-oci-ta/0.1/README.md
MartinBasti: I'm not very happy about adding it into description, formatting cannot be so rich and also describing something that's defined in later section seems off.

MartinBasti: I also cannot add example into param description as there are tools which are using that field and doesn't parse markdown

I can probably just keep it documented in original task only

MartinBasti: We need readme templating :D

chmeliik: Works fine for https://github.com/konflux-ci/build-definitions/blob/c22a0bb0d4bcbaebb27fd4ceb317ae71b3ddae85/task/upload-sbom-to-trustification/0.1/upload-sbom-to-trustification.yaml#L8-L30

But yes, we do need README templating

MartinBasti: How does it look in konflux webUI?

chmeliik: You can view task descriptions in Konflux web UI? :open_mouth:

MartinBasti: image
 it seems that it can parse markdown

MartinBasti: Updated, I think that this could work

chmeliik: Nice

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -11,6 +11,24 @@ spec:
...
+
+    Example of setting timeouts:
+
+    ```yaml
+    params:
+      - name: config-file-content
+        value: |
+          ---
+          requests_timeout: 300
+          subprocess_timeout:3600
```
chmeliik, line 28:
```suggestion
          subprocess_timeout: 3600
```

chmeliik (approved):
LGTM, one formatting issue

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
MartinBasti: oh I failed horribly, fixed

PR #1198, opened 2024-07-25
Ensure that /shared/base_images_digests exists

arewm:
When there are no base images (i.e. it is FROM scratch), we need to make sure that the base_images_digests file exists.


rcerven (approved):

chmeliik:
Thanks for fixing, I created a regression when dropping the BASE_IMAGES_DIGESTS results

PR #1203, opened 2024-07-26
Attach a patch for needed updates to the TA Tasks

zregvart:
When the workflow fails it should attach a patch that folk can apply to get the needed changes on their PR.

Reference: https://issues.redhat.com/browse/EC-769

### DIFF
--- a/.github/workflows/check-ta.yaml
+++ b/.github/workflows/check-ta.yaml
@@ -10,4 +10,11 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
       - name: Check Trusted Artifact variants
+        id: check
         run: hack/generate-ta-tasks.sh
+      - name: Attach patch
+        if: ${{ always() && steps.check.conclusion == 'failure' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: Trusted artifacts patch
+          path: ./ta.patch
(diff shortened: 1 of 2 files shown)


.github/workflows/check-ta.yaml
```diff
@@ -11,3 +11,9 @@ jobs:
         uses: actions/checkout@v4
       - name: Check Trusted Artifact variants
         run: hack/generate-ta-tasks.sh
+      - name: Attach patch
```
arewm, line 14: Where does this attachment happen? Can you trigger a test on this PR or will the PR only use the workflows from main?

zregvart: The patch will be uploaded to GitHub and served from the "Summary" page, [here is one](https://github.com/konflux-ci/build-definitions/actions/runs/10113623566?pr=1203) from a previous run that I had intentionally made to fail.

chmeliik: I wouldn't know to go to the Summary page. I think I'd be confused about where to get the `ta.patch`.

For users of the `gh` CLI, the message could give executable instructions:

```bash
gh run download "$GITHUB_RUN_ID" -n 'Trusted artifacts patch'
git apply ta.patch
```

(GITHUB_RUN_ID should be injected automatically into the GH runner environment)

For others, telling them to go to the Summary page would help

lcarva (approved):

robnester-rh (approved):
lgtm

PR #1204, opened 2024-07-26
Ensure that Chains sees all image manifests referenced

arewm:
Due to a [bug in Chains](https://github.com/tektoncd/chains/issues/1164), it does not currently see all images produced within a matrix. In order to work around this, we can expose all images that we include in our Image Index so that Chains can generate the provenance for all of them.

This fix is also relevant if any of the images added are themselves an image index due to https://github.com/tektoncd/chains/issues/1070.


zregvart (approved):
LGTM

PR #1205, opened 2024-07-26
prefetch-task-rhsm-integration

brianwcook:
These changes are to accompany https://github.com/containerbuildsystem/cachi2/pull/580 and **that PR needs to be merged first.**

The cachi2 changes in that PR enable usage of client certificates for accessing Red Hat content from subscription protected repos, but it is out of scope for cachi2 how those certificates are obtained. They cachi2 find the files via environment variables.

This PR causes the prefetch task to react to the same ACTIVATION_KEY parameter that is used for non-hermetic builds. The container will use the pipeline-provided activation key to register the STEP container and set the proper environment variables before executing cachi2.

The following points are pertinent:
* the RHSM (Red Hat subscription management) files generated by `subscription-manager register` are generated by this task, immediately before running cach2
* they are never used again

Therefore this implementation is safe from certificate revocation / rotation behavior of RHSM.

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
@@ -144,7 +144,7 @@ spec:
           yq 'del(.goproxy_url)' <<<"${CONFIG_FILE_CONTENT}"  >/mnt/config/config.yaml
         fi
     - name: prefetch-dependencies
-      image: quay.io/redhat-appstudio/cachi2:0.9.1@sha256:df67f9e063b544a8c49a271359377fed560562615e0278f6d0b9a3485f3f8fad
+      image: quay.io/bcook/cachi2@sha256:cd0ee8284eb41838071e7987644fcfe05cef77f89ac426c14a2fa4ed5bd1b5ac
       volumeMounts:
         - mountPath: /mnt/trusted-ca
           name: trusted-ca
@@ -212,8 +212,27 @@ spec:
           update-ca-trust
         fi
 
-        cachi2 --log-level="$LOG_LEVEL" $config_flag fetch-deps \
-          $dev_pacman_flag \
+        cachi2 --log-level="$LOG_LEVEL" fetch-deps \
+          ACTIVATION_KEY_PATH="/activation-key"
+        if [ -d "$ACTIVATION_KEY_PATH" ]; then
+            echo "Detected activation key, registering with RHSM."
+
+            # todo: make sure orgid key is consistent with buildah task and docs
+            subscription-manager register \
+              --org $(cat "/activation-key/orgid") \
+              --activationkey $(cat "/activation-key/activationkey")
+
+            # detect entitlement certs and setup environment variables
+            ls /etc/pki/entitlement/
+
+            export RHSM_ID=$(ls /etc/pki/entitlement/ | grep key | cut -d - -f 1)
+            echo $RHSM_ID
+            export C2_CLIENT_CERT="/etc/pki/entitlement/$RHSM_ID.pem"
+            export C2_CLIENT_KEY="/etc/pki/entitlement/$RHSM_ID-key.pem"
+            echo "Using client certificate $C2_CLIENT_CERT and key $C2_CLIENT_KEY."
+        fi
+
+        $dev_pacman_flag \
           --source=/var/workdir/source \
           --output=/var/workdir/cachi2/output \
           "${INPUT}"
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -74,7 +74,7 @@ spec:
         yq 'del(.goproxy_url)' <<< "${CONFIG_FILE_CONTENT}" > /mnt/config/config.yaml
       fi
 
-  - image: quay.io/redhat-appstudio/cachi2:0.9.1@sha256:df67f9e063b544a8c49a271359377fed560562615e0278f6d0b9a3485f3f8fad
+  - image: quay.io/bcook/cachi2@sha256:cd0ee8284eb41838071e7987644fcfe05cef77f89ac426c14a2fa4ed5bd1b5ac
     # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
     # the cluster will set imagePullPolicy to IfNotPresent
     name: prefetch-dependencies
@@ -146,7 +146,27 @@ spec:
         update-ca-trust
       fi
 
-      cachi2 --log-level="$LOG_LEVEL" $config_flag fetch-deps \
+      cachi2 --log-level="$LOG_LEVEL" fetch-deps \
+      ACTIVATION_KEY_PATH="/activation-key"
+      if [ -d "$ACTIVATION_KEY_PATH" ]; then
+          echo "Detected activation key, registering with RHSM."
+        
+          # todo: make sure orgid key is consistent with buildah task and docs
+          subscription-manager register \
+            --org $(cat "/activation-key/orgid") \
+            --activationkey $(cat "/activation-key/activationkey")
+
+          # detect entitlement certs and setup environment variables
+          ls /etc/pki/entitlement/
+
+          export RHSM_ID=$(ls /etc/pki/entitlement/ | grep key | cut -d - -f 1)
+          echo $RHSM_ID
+          export C2_CLIENT_CERT="/etc/pki/entitlement/$RHSM_ID.pem"
+          export C2_CLIENT_KEY="/etc/pki/entitlement/$RHSM_ID-key.pem"
+          echo "Using client certificate $C2_CLIENT_CERT and key $C2_CLIENT_KEY."
+        fi
+
+      
       $dev_pacman_flag \
       --source=$(workspaces.source.path)/source \
       --output=$(workspaces.source.path)/cachi2/output \


task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -212,8 +212,27 @@ spec:
...
-          $dev_pacman_flag \
+        cachi2 --log-level="$LOG_LEVEL" fetch-deps \
+          ACTIVATION_KEY_PATH="/activation-key"
+        if [ -d "$ACTIVATION_KEY_PATH" ]; then
+            echo "Detected activation key, registering with RHSM."
+
+            # todo: make sure orgid key is consistent with buildah task and docs
+            subscription-manager register \
+              --org $(cat "/activation-key/orgid") \
+              --activationkey $(cat "/activation-key/activationkey")
```
eskultety, line 223: Is the user running inside the container privileged?

> Error: this command requires root access to execute

eskultety: Default UBI container.

brianwcook: the buildah task it running as user 0 already so imo this is a non-issue.

eskultety: Checked the buildah task, there's nothing going on in terms of user setting really (apart from ID mapping), so by default this runs as root. I guess the container itself isn't privileged, so while not ideal, I guess it's acceptable, so consider my initial comment retracted.

However, that makes things easier for cachi2 then and since we're assuming the default root user inside the container, this whole RHSM registration should IMO be baked into cachi2 rpm-dnf backend rather than in the tekton task for proper integration - we could also leverage dbus to communicate with RHSM for better error handling rather than dealing with shell.

brianwcook:
IMHO the way I have implemented it is the most versatile option.

- If your task is running on a Jenkins or gitlab runner,set your environment variables and pull content
- If your task is running on Openshift with insights operator, set your environment variables and pull content.
- If you are running in an unregistered environment like Konflux is, you can register a container with activation key and then set variables.
- In addition if you want to run your own client-cert protected yum repo, my code also works for that.

eskultety:
> IMHO the way I have implemented it is the most versatile option.
> 
>     * If your task is running on a static Jenkins or gitlab runner which is registered with RHSM,set your environment variables and pull content
> 
>     * If your task is running on Openshift with insights operator, set your environment variables and pull content.
> 
>     * If you are running in an unregistered environment like Konflux is, you can register a container with activation key and then set variables and pull content.
> 
>     * In addition if you want to run your own client-cert protected yum repo, my code also works for that.
> 
> 
> It also introduces no new dependencies on yum, dnf or subscription manager and so is insulated from changes there. The client certificate scheme hasn't changed in over a decade and should be quite stable.

Which of ^these use cases does the following not comply with, rendering it less versatile?

>  this whole RHSM registration should IMO be baked into cachi2 rpm-dnf backend rather than in the tekton task

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -146,7 +146,27 @@ spec:
...
+          ls /etc/pki/entitlement/
+
+          export RHSM_ID=$(ls /etc/pki/entitlement/ | grep key | cut -d - -f 1)
+          echo $RHSM_ID
+          export C2_CLIENT_CERT="/etc/pki/entitlement/$RHSM_ID.pem"
+          export C2_CLIENT_KEY="/etc/pki/entitlement/$RHSM_ID-key.pem"
+          echo "Using client certificate $C2_CLIENT_CERT and key $C2_CLIENT_KEY."
+        fi
+
+
```
eskultety, line 169: I don't see any unregistration here which is a good habit in general to do once we're done using a service to prevent a plethora of automated single-use registrations lingering in the subscription service.

brianwcook:
e2e failing on 
`[e2e-tests : e2e-test]   No RHSM ID found.
[e2e-tests : e2e-test]   Preprocessing result: gomod
[e2e-tests : e2e-test]   Traceback (most recent call last):
[e2e-tests : e2e-test]     File "/tekton/scripts/script-3-9ml5w", line 77, in <module>
[e2e-tests : e2e-test]       with open('/shared/rhsm/preprocessed_input', 'w') as f:
[e2e-tests : e2e-test]   FileNotFoundError: [Errno 2] No such file or directory: '/shared/rhsm/preprocessed_input'

`

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -46,11 +50,6 @@ spec:
       description: The name of the ConfigMap to read CA bundle data from.
       type: string
       default: trusted-ca
-    - name: config-file-content
```
MartinBasti, line 49: Why is config file being removed?

brianwcook: hm.. its not intentional. It looks like this change just didn't get pulled in via rebase for some reason. Will add it back.

brianwcook: should be resolved

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
brianwcook: unregitration is now handled.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
+          value: $(params.ACTIVATION_KEY)
+      script: |
+        #!/bin/bash
+        if [ -f /shared/skip ]; then
+          echo "Skipping."
+          exit 0
+        fi
+
+        echo "false" >/shared/registered
+        ACTIVATION_KEY_PATH="/activation-key"
```
eskultety, line 192: Noob question - how is the `ACTIVATION_KEY` parameter even used, it doesn't seem to since you're hardcoding the volume path here? What if someone changes the value, how do we probe the right file system location? I'm clearly missing something here making me confused.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -74,13 +79,164 @@ spec:
...
+      mkdir -p /shared/rhsm/consumer
+
+      if [ -e /activation-key/org ]; then
+        cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
+
+        echo "Registering with Red Hat subscription manager."
+        subscription-manager register --org "$(cat /tmp/activation-key/org)" --activationkey "$(cat /tmp/activation-key/activationkey)"
+
+        # copy generated certificates to /shared/rhsm
+        cp /etc/pki/entitlement/*.pem /shared/rhsm/entitlement/
```
eskultety, line 130: Related to my earlier comment on not splitting this among different steps, I'm confused why we'd want a shared volume full of these default pre-populated subman system locations with data instead of handling this within the same context and referencing only these standard and expected system locations instead of shared volumes.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -74,13 +79,164 @@ spec:
...
+
+          print("Preprocessing result: %s" % input)
+          with open('/shared/rhsm/preprocessed_input', 'w') as f:
+              f.write(input)
+
+
+  - name: prefetch-dependencies
+    image: quay.io/redhat-appstudio/cachi2@sha256:eb34cfe3fea20997eebd8164dc93eedb2fd7a60dc1fb4afcc1b1ff43df9d6667
+    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+    # the cluster will set imagePullPolicy to IfNotPresent
```
eskultety, line 238: What does this commentary explain to me? Should it be part of a separate commit with more context mentioned in the commit message?

pipelines/docker-build-multi-platform-oci-ta/README.md
```diff
@@ -135,6 +135,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 ### prefetch-dependencies-oci-ta:0.1 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
+|ACTIVATION_KEY| Name of secret which contains subscription activation key| activation-key| |
```
eskultety, line 138: I think this option is wort mentioning and better describing its usefulness in the commit msg.

eskultety, line 138: **Commit message:**

s/THe/The

> THe input is modified, injecting the entitlement certs

How is the input modified? Yes, we copy stuff to a shared volume, most importantly the certs that get passed to cachi2 to use for TLS auth, but other than that, how is the user input modified?

eskultety (commented):
@brianwcook Terribly sorry to have gone so long without a review, it has now become my top-most priority to help you finish the work.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
       image: quay.io/redhat-appstudio/cachi2:0.13.0@sha256:eb34cfe3fea20997eebd8164dc93eedb2fd7a60dc1fb4afcc1b1ff43df9d6667
+      env:
+        - name: INPUT
+          value: $(params.input)
+      script: |
+        if [ -z "${INPUT}" ]; then
+          # Confirm input was provided though it's likely the whole task would be skipped if it wasn't
+          echo "No prefetch will be performed because no input was provided for cachi2 fetch-deps"
+          echo "skip" >/shared/skip
+        fi
```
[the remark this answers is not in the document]
brianwcook: this is just preserving existing behavior.  https://github.com/konflux-ci/build-definitions/blob/6f4a50deb863050574a5422e89279aa43b1399ae/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L101

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
+        - name: INPUT
+          value: $(params.input)
+        - name: ACTIVATION_KEY
+          value: $(params.ACTIVATION_KEY)
+      script: |
+        #!/bin/bash
+        if [ -f /shared/skip ]; then
+          echo "Skipping."
+          exit 0
+        fi
```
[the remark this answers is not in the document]
brianwcook: yes, when prefetch-input is not provided, it just preserves the existing behavior which is to quickly exit the task.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: The path _is_ hardcoded but the secret that is populated to that path has a default value of 'activation-key' in the task parameter declarations and can be overridden in the params passed to task invocation. It is the secret name the user needs to be concerned with, the path is an internal implementation detail.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
+
+        echo "false" >/shared/registered
+        ACTIVATION_KEY_PATH="/activation-key"
+
+        mkdir -p /shared/rhsm/entitlement
+        mkdir -p /shared/rhsm/consumer
+
+        if [ -e /activation-key/org ]; then
+          cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
```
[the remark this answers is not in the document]
brianwcook: It could be possible to read it directly from the volume but   I tried to keep this code in sync with the buildah task which does a copy.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
brianwcook: I suspect that this bit of code just evolved this way but I tried to keep it consistent between the buildah task and this one. If we want to try to remove the copying and just use the volume directly I would prefer to handle it in a followup PR in both this task _and_ the buildah task. Side note: this could be a candidate for use with future stepActions.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
brianwcook: this is a preexisting message (https://github.com/konflux-ci/build-definitions/blob/6f4a50deb863050574a5422e89279aa43b1399ae/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L78) - not sure what its purpose is but I didn't want to remove it.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
eskultety: > I suspect that this bit of code just evolved this way but I tried to keep it consistent between the buildah task and this one.

Is this: https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml#L436-L462 the bit of buildah task work you're referring to?

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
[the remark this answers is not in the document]
eskultety: ...GitHub is either acting up or I can't mark the thread as resolved. Maybe because I'm not part of the org? @brianwcook if you can do that, then you can resolve this thread.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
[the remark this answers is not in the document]
eskultety: ->Resolved.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
eskultety: Should have done my due diligence on grepping the code base first. No idea why it's needed, but it can be dropped in a follow up if needed -> Resolved.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
+
+          # copy generated certificates to /shared/rhsm
+          cp /etc/pki/entitlement/*.pem /shared/rhsm/entitlement/
+          cp /etc/pki/consumer/*.pem /shared/rhsm/consumer/
+
+          file="$(find /shared/rhsm/entitlement -regextype egrep -regex '.*[0-9]+\.pem' -printf %f)"
+          echo "file: $file"
+          basename "$file" .pem >/shared/RHSM_ID
+          echo "./RHSM_ID:"
+          cat /shared/RHSM_ID
```
eskultety, line 211: This is IMO not needed at all, IIUC you only added it to report the RHSM_ID in the logs, but I wonder what value that brings compared to reporting the TLS auth credentials as part of cachi2's own debug logging system.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -143,15 +157,159 @@ spec:
...
+            if type(input) is dict:
+                input = [input]
+            return json.dumps(input)
+
+
+        def inject_certs(input: str, rhsm_id: str):
+            input_list: list = json.loads(input)
+
+            cert = ("/shared/rhsm/entitlement/%s.pem" % rhsm_id)
+            key = ("/shared/rhsm/entitlement/%s-key.pem" % rhsm_id)
```
eskultety, line 250: Your code already assumes there's only ever going to be a single set of entitlement key-cert pair of files which is a reasonable assumption ATM, so you don't really need this RHSM_ID filename construction logic, one file is going to be named `xyz-key.pem` the other `xyz.pem` no RHSM involvement needed at all.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: 208, 210 and 211 for visibility and we could remove them but it would make any necessary debugging harder. 207 and 209 are crucial functionality.

the "RHSM_ID" is not known before `subsription-manager register` is run. 
That is why the code has to pase this value from the file names. Those files _also_ don't exist before `subscrip-manager-register` is run.  What is happening (and there is no alternative) is this
1. user provides org id and activation key values in Kube secret
2. code runs subscription-manager register --org [org] --key key
3. Two files are generated during registration,  `[rhsm_id.pem]` and `[rhsm_id].key
4. prefetch input has to be modified on the fly adding certificate options with those file namesfor _all_ the occurences of the rpm manager in the prefetch input - even if all the user passed as input is `rpm`.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: The construction is necessary. Even though the code limits the user to one set of certificates, the certificate names are not known until subscription-manager register is run.

eskultety: > The construction is necessary. Even though the code limits the user to one set of certificates, the certificate names are not known until subscription-manager register is run.

It's true that you don't know the actual names, but I don't believe the construction is really necessary, regardless of whether this is Bash or Python, you can get extract the right names dynamically without knowing RHSM_ID at the time of use, because you have the most important bit of information - the files are named the same sans the `-key` suffix. 
What I'm trying to say is, we should only be using and passing around data what we absolutely have to and need instead of generating/passing around data we don't.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
eskultety: > the "RHSM_ID" is not known before `subsription-manager register` is run. That is why the code has to pase this value from the file names. Those files _also_ don't exist before `subscrip-manager-register` is run. What is happening (and there is no alternative) is this
> 
>     1. user provides org id and activation key values in Kube secret
> 
>     2. code runs subscription-manager register --org [org] --key key
> 
>     3. Two files are generated during registration,  `[rhsm_id.pem]` and `[rhsm_id].key
> 
>     4. prefetch input has to be modified on the fly adding certificate options with those file namesfor _all_ the occurences of the rpm manager in the prefetch input - even if all the user passed as input is `rpm`.

We can extract the file names on demand right before executing cachi2. The intermediate RHSM_ID handling is just opaque and makes it visibly hard to read, why complicating stuff? :)

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: ...but the names of the keys need to be input into the python step so it can be injected into input JSON.

eskultety: Why? IIUC the shared volume `entitlement`directory contains only 2 files from which you can derive the intent of each of those.

brianwcook:
if the user input is simply `rpm` and an activation key is presented, the input string has to be transformed before being passes to cachi2 to:
`
'[{"type": "rpm", "options": {"ssl": {"client_key": null, "client_cert": null, "ca_bundle": null, "verify": 1}}}]'
`
And then the client_key and client_cert values need to be populated with the name of the pem file generated by `subscription-manager register`

pipelines/docker-build-multi-platform-oci-ta/README.md
brianwcook: done

brianwcook: added example of how the input is modified.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: if the user input is simply rpm and an activation key is presented, the input string has to be transformed before being passed to cachi2 to:
'[{"type": "rpm", "options": {"ssl": {"client_key": null, "client_cert": null, "ca_bundle": null, "verify": 1}}}]'
And then the client_key and client_cert values need to be populated with the name of the pem file generated by subscription-manager register

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
brianwcook: the general pattern even predates that section - [this bit](https://github.com/konflux-ci/build-definitions/blob/bf3a7f2c00c3863ca68566942d1a40c43175bcb5/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml#L470) existed before activation keys were supported at all. There was some issue with trying to mount directly but I cannot recall what it was.

I also think that this code can be consolidated into a [stepAction](https://tekton.dev/vault/pipelines-v0.53.x-lts/stepactions/) in the future and to do that we 1) need to keep the data on a shared emptyDir as it is now and 2) would benefit from keeping the code in sync as best we can between prefetch and buildah tasks.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -53,6 +53,10 @@ spec:
     type: string
     description: The name of the key in the ConfigMap that contains the CA bundle data.
     default: ca-bundle.crt
+  - name: ACTIVATION_KEY
```
MartinBasti, line 56: Missing README update

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
brianwcook: you cannot extract the names only once, modify the input in a separate task and omit passing the names between steps.

Passing data between steps is a common pattern to save Tekton results space. It is not unexpected in this context or abnormally complicated. If you want to avoid it for some reason then we have to extract the names from the files twice. In any case, I don't think the difference between the two approaches is worth the time spent discussing it and I find parsing the name once to be the simpler way (hence the way I went).

MartinBasti (approved):

PR #1207, opened 2024-07-26
Use tmpfs for pki entitlements

brianwcook:
only allow 1 method for subscription certs at a time; ensure certificates are not included in final image 

this commit updates activation-keys work to mount a temporary directory to /etc/pki/entitlement during the build. **This ensure that any certificates generated by "subscription-manager register" are not included in the final build.**

The subscription manager team has raised that new processes will revoke generated client subscriptions. On a normal machine using subscription-manager or insights, there is a regeneration process that runs and this is not an issue. In Konflux, when certificates stored as secrets are revoked nothing will regenerate them and it will result in a broken build.

For this reason, they asked us to push people to using activation-keys. This change will prefer activation keys when both are present. Docs updates will also be added to recommend using activation and we would be well positioned to deprecated the etc-pki-entitlement secret in the future.

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -363,18 +363,21 @@ spec:
         [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
         [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
 
+        ACTIVATION_KEY_PATH="/activation-key"
         ENTITLEMENT_PATH="/entitlement"
-        if [ -d "$ENTITLEMENT_PATH" ]; then
+
+        # do not enable activation key and entitlement at same time. If both vars are provided, prefer activation key.
+        # when activation keys are used, an empty directory on shared emptydir volume to /etc/pki/entitlement to prevent certificates from being included in the produced container.
+
+        if [ -d "$ACTIVATION_KEY_PATH" ]; then
+          cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
+          mkdir /shared/rhsm-tmp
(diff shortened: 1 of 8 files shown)


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -363,20 +363,23 @@ spec:
         [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
         [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
 
+        ACTIVATION_KEY_PATH="/activation-key"
         ENTITLEMENT_PATH="/entitlement"
-        if [ -d "$ENTITLEMENT_PATH" ]; then
+
+        # do not enable activation key and entitlement at same time. If both vars are provided, prefer activation key.
+        # when activation keys are used, an empty directory on shared emptydir volume to /etc/pki/entitlement to prevent certificates from being included in the produced container.
```
chmeliik, line 370: Do you specifically intend to make the methods mutually exclusive, or just to `ensure certificates generated by "subscription-manager register" are not included in the final build.`

I think we can achieve that ^ without making the mutually exclusive

chmeliik: And if you do intend to make them mutually exclusive, should we prefer certs? IIUC activation keys require extra handling in the Containerfile, while the certs "just work"

brianwcook: I think they should be made mutually exclusive and that is what I did here intentionally. I'd like to document that the the etc-pki-entitlement method of storing certs is discouraged and prone to random certificate revocation. Including activation keys now intentionally causes the certs generated by that method to be preferred.

brianwcook: updated the PR with more info about my intentions here.

arewm (approved):

task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
arewm: Since the intended direction is activation keys, I think it is fine for that to be the overriding path even if it is more work than certs. The workaround is to remove the configuration for the unused method which should be a best practice anyway.

PR #1208, opened 2024-07-28
Add S3 bucket authentication to oci-copy

ralphbean:
The existing bearer token auth is not sufficient for pulling from an S3 bucket. This enables just that: pulling content from an s3 bucket.

Policy mechanisms (already in place and merged) prevent redistributing content from buckets that are not explicitly allowed.

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -13,6 +13,12 @@ spec:
   description: Given a file in the user's source directory, copy content from
     arbitrary urls into the OCI registry.
   params:
+    - name: AWS_SECRET_NAME
+      description: Name of a secret which will be made available to the build
+        to construct Authorization headers for requests to Amazon S3. If specified,
+        this will take precedence over BEARER_TOKEN_SECRET_NAME.
+      type: string
+      default: does-not-exist
     - name: BEARER_TOKEN_SECRET_NAME
       description: Name of a secret which will be made available to the build
         as an Authorization header. Note, the token will be sent to all servers
@@ -69,6 +75,7 @@ spec:
       image: quay.io/konflux-ci/yq:latest@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430
       workingDir: /var/workdir
       script: |
+        #!/bin/bash
         set -eu
         set -o pipefail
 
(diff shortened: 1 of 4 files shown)


task/oci-copy/0.1/README.md
```diff
@@ -14,6 +14,7 @@ Note: the bearer token secret, if specified, will be sent to **all servers liste
 |IMAGE|Reference of the image buildah will produce.||true|
 |OCI_COPY_FILE|Path to the oci copy file.|./oci-copy.yaml|false|
 |BEARER_TOKEN_SECRET_NAME|Name of a secret which will be made available to the build as an Authorization header. Note, the token will be sent to all servers found in the oci-copy.yaml file. If you do not wish to send the token to all servers, different taskruns and therefore different oci artifacts must be used.|"does-not-exist"|false|
+|AWS_SECRET_NAME|Name of a secret which will be made available to the build to construct Authorization headers for requests to Amazon S3. If specified, this will take precedence over BEARER_TOKEN_SECRET_NAME.|does-not-exist|false|
```
chmeliik, line 17: Hmm, so it's not possible to do both? I.e. all the authenticated parts must come _either_ from S3 _or_ from servers that use bearer auth?

I guess that's fine if it meets the current use cases. Wondering what would be the path to make this task more generically usable though

task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -89,17 +96,47 @@ spec:
...
+        valueFrom:
+          secretKeyRef:
+            name: $(params.AWS_SECRET_NAME)
+            key: aws_access_key_id
+            optional: true
+      - name: AWS_SECRET_ACCESS_KEY
+        valueFrom:
+          secretKeyRef:
+            name: $(params.AWS_SECRET_NAME)
+            key: aws_secret_access_key
```
chmeliik, line 109: It would be good to document the expected content of the Secret

task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -89,17 +96,47 @@ spec:
...
+          if [ -n "${AWS_ACCESS_KEY_ID}" ] && [ -n "${AWS_SECRET_ACCESS_KEY}" ]; then
+            echo "Found both aws credentials secret with both aws_access_key_id and aws_secret_access_key. Assuming S3 bucket"
+            path=$(echo "$url" | cut -d/ -f4-)
+            echo "Bucket path is $path"
+            date="$(date -u '+%a, %e %b %Y %H:%M:%S +0000')"
+            printf -v string_to_sign "%s\n\n\n%s\n%s" "$method" "$date" "/$path"
+            echo "String to sign is $string_to_sign"
+            signature=$(echo -n "$string_to_sign" | openssl dgst -sha1 -binary -hmac "${AWS_SECRET_ACCESS_KEY}" | openssl base64)
+            authorization="AWS ${AWS_ACCESS_KEY_ID}:${signature}"
+            curl "${curl_args[@]}" -H "Date: ${date}" -H "Authorization: ${authorization}" --location "$url" -o "$file"
```
chmeliik, line 131: Do you have a link to some docs that explain the S3 auth? These seem like pretty low-level AWS details

task/oci-copy/0.1/oci-copy.yaml
```diff
@@ -89,17 +96,47 @@ spec:
...
+          url="$1"
+          file="$2"
+          method="GET"
+
+          curl_args=(--fail --silent --show-error)
+          if [ -n "${AWS_ACCESS_KEY_ID}" ] && [ -n "${AWS_SECRET_ACCESS_KEY}" ]; then
+            echo "Found both aws credentials secret with both aws_access_key_id and aws_secret_access_key. Assuming S3 bucket"
+            path=$(echo "$url" | cut -d/ -f4-)
+            echo "Bucket path is $path"
+            date="$(date -u '+%a, %e %b %Y %H:%M:%S +0000')"
```
chmeliik, line 126: `+0000` - are we assuming that the task runs on a node in the UTC+0 timezone?

task/oci-copy/0.1/oci-copy.yaml
ralphbean: Yeah, it's here. https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html

I'll add it as an inline comment.

task/oci-copy/0.1/oci-copy.yaml
ralphbean: Hm, yes. I guess - whatever format can be used so long as the date is provided via the date header to the s3 provider. The s3 provider just has to be able to parse it so they can use it in verification of the signature.

chmeliik: ~Ack, so if the date is inaccurate it's not a big deal?~

Nevermind, I missed the `-u` - the date should be correct for UTC+0

task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
```diff
@@ -104,17 +113,48 @@ spec:
...
-        fi
+        download() {
+          url="$1"
+          file="$2"
+          method="GET"
+
+          curl_args=(--fail --silent --show-error)
+          if [ -n "${AWS_ACCESS_KEY_ID}" ] && [ -n "${AWS_SECRET_ACCESS_KEY}" ]; then
+            echo "Found both aws credentials secret with both aws_access_key_id and aws_secret_access_key. Assuming S3 bucket"
+            # This implements https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html
```
chmeliik, line 141: Hmm, this links to the V4 of the docs, but I think what is implemented here is V2 as described in https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html

The note at the top is a little concerning
> This topic explains authenticating requests using Signature Version 2. Amazon S3 now supports the latest Signature Version 4. This latest signature version is supported in all regions and any new regions after January 30, 2014 will support only Signature Version 4.

ralphbean: Oof, you're right. I need to re-do this.

It works with the actual bucket I need for my use case today, but we should future-proof.

ralphbean: Well - I have a failing e2e test which I assume is unrelated. I'm going to rebase on `main` first but then I intend to merge.

PR #1216, opened 2024-07-29
Enable remote tasks to be run in cluster

arewm:
By default, we should run builds matching the local architecture in-cluster to reduce the overhead of provisioning platforms. This will enable a fully matrixed build for all images using only the remote builds. This change will require the multi-platform controller to set the /ssh/host to localhost in order for the builds to run in-cluster.

In a change from the prior behavior of auto-appending the architecture to the tag, we will now append a sanitized version of the entire PLATFORM to the image tag upon request. This behavior is now needs to be explicitly requested which fixes a bug where users cannot specify the specific tag desired when using a remote build.

NOTE: There are not currently any local platforms configured. These need to be added before this PR have an effect in production.

### DIFF
--- a/task-generator/remote/main.go
+++ b/task-generator/remote/main.go
@@ -102,14 +102,16 @@ func convertToSsh(task *tektonapi.Task, taskVersion string) {
 	// The images produced in multi-platform builds need to have unique tags in order
 	// to prevent them from getting garbage collected before generating the image index.
 	// We can simplify this process, preventing the need for users to manually specify
-	// the image by auto-appending the architecture from the PLATFORM parameter. For
-	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. Since
-	// we cannot modify the parameter itself, this replacement needs to happen in any task
-	// step where the IMAGE parameter is used.
-	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
-	// not be appended again based on the PLATFORM.
-	adjustRemoteImage := `if [[ "${IMAGE##*-}" != "${PLATFORM##*/}" ]]; then
-  export IMAGE="${IMAGE}-${PLATFORM##*/}"
+	// the image by auto-appending the a sanitized PLATFORM parameter. For example, this
+	// will append linux-arm64 if PLATFORM is linux/arm64 and IMAGE_APPEND_PLATFORM is true.
+	// Many special characters are not allowed in tags so we will replace anything that
+	// isn't alphanumeric with a "-" to be safe. Since we cannot modify the parameter itself,
+	// this replacement needs to happen in any task step where the IMAGE parameter is used.
+	// IMAGE_APPEND_PLATFORM will be set to "false" by default so appending the platform is
+	// and explicit opt-in.
+	adjustRemoteImage := `if [ "${IMAGE_APPEND_PLATFORM}" == "true" ]; then
+  IMAGE="${IMAGE}-${PLATFORM//[^a-zA-Z0-9]/-}"
+  export IMAGE
 fi
 `
 
@@ -142,55 +144,65 @@ if [ -e "/ssh/error" ]; then
   #no server could be provisioned
   cat /ssh/error
   exit 1
+fi
+export SSH_HOST=$(cat /ssh/host)
+
+if [ "$SSH_HOST" == "localhost" ] ; then
+  IS_LOCALHOST=true
+  echo "Localhost detected; running build in cluster"
 elif [ -e "/ssh/otp" ]; then
- curl --cacert /ssh/otp-ca -XPOST -d @/ssh/otp $(cat /ssh/otp-server) >~/.ssh/id_rsa
- echo "" >> ~/.ssh/id_rsa
+  curl --cacert /ssh/otp-ca -XPOST -d @/ssh/otp $(cat /ssh/otp-server) >~/.ssh/id_rsa
+  echo "" >> ~/.ssh/id_rsa
 else
   cp /ssh/id_rsa ~/.ssh
 fi
-chmod 0400 ~/.ssh/id_rsa
-export SSH_HOST=$(cat /ssh/host)
-export BUILD_DIR=$(cat /ssh/user-dir)
-export SSH_ARGS="-o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10"
-mkdir -p scripts
-echo "$BUILD_DIR"
-ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/volumes"
 
-PORT_FORWARD=""
-PODMAN_PORT_FORWARD=""
-if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] ; then
-PORT_FORWARD=" -L 80:$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR:80"
-PODMAN_PORT_FORWARD=" -e JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR=localhost"
-fi
+mkdir -p scripts
+
+if ! [[ $IS_LOCALHOST ]]; then
+  chmod 0400 ~/.ssh/id_rsa
+  export BUILD_DIR=$(cat /ssh/user-dir)
+  export SSH_ARGS="-o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10"
+  echo "$BUILD_DIR"
+  ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/volumes"
+
+  PORT_FORWARD=""
+  PODMAN_PORT_FORWARD=""
+  if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] ; then
+    PORT_FORWARD=" -L 80:$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR:80"
+    PODMAN_PORT_FORWARD=" -e JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR=localhost"
+  fi
 `
-		if taskVersion != "0.1" {
-			ret += adjustRemoteImage
-		}
 		env := "$PODMAN_PORT_FORWARD \\\n"
 
 		// disable podman subscription-manager integration
-		env += " --tmpfs /run/secrets \\\n"
+		env += "    --tmpfs /run/secrets \\\n"
 
 		// Before the build we sync the contents of the workspace to the remote host
 		for _, workspace := range task.Spec.Workspaces {
-			ret += "\nrsync -ra $(workspaces." + workspace.Name + ".path)/ \"$SSH_HOST:$BUILD_DIR/workspaces/" + workspace.Name + "/\""
-			podmanArgs += " -v \"$BUILD_DIR/workspaces/" + workspace.Name + ":$(workspaces." + workspace.Name + ".path):Z\" \\\n"
+			ret += "\n  rsync -ra $(workspaces." + workspace.Name + ".path)/ \"$SSH_HOST:$BUILD_DIR/workspaces/" + workspace.Name + "/\""
+			podmanArgs += "    -v \"$BUILD_DIR/workspaces/" + workspace.Name + ":$(workspaces." + workspace.Name + ".path):Z\" \\\n"
 		}
 		// Also sync the volume mounts from the template
 		for _, volume := range task.Spec.StepTemplate.VolumeMounts {
-			ret += "\nrsync -ra " + volume.MountPath + "/ \"$SSH_HOST:$BUILD_DIR/volumes/" + volume.Name + "/\""
-			podmanArgs += " -v \"$BUILD_DIR/volumes/" + volume.Name + ":" + volume.MountPath + ":Z\" \\\n"
+			ret += "\n  rsync -ra " + volume.MountPath + "/ \"$SSH_HOST:$BUILD_DIR/volumes/" + volume.Name + "/\""
+			podmanArgs += "    -v \"$BUILD_DIR/volumes/" + volume.Name + ":" + volume.MountPath + ":Z\" \\\n"
 		}
 		for _, volume := range step.VolumeMounts {
 			if syncVolumes[volume.Name] {
-				ret += "\nrsync -ra " + volume.MountPath + "/ \"$SSH_HOST:$BUILD_DIR/volumes/" + volume.Name + "/\""
-				podmanArgs += " -v \"$BUILD_DIR/volumes/" + volume.Name + ":" + volume.MountPath + ":Z\" \\\n"
+				ret += "\n  rsync -ra " + volume.MountPath + "/ \"$SSH_HOST:$BUILD_DIR/volumes/" + volume.Name + "/\""
+				podmanArgs += "    -v \"$BUILD_DIR/volumes/" + volume.Name + ":" + volume.MountPath + ":Z\" \\\n"
 			}
 		}
-		ret += "\nrsync -ra \"$HOME/.docker/\" \"$SSH_HOST:$BUILD_DIR/.docker/\""
-		podmanArgs += " -v \"$BUILD_DIR/.docker/:/root/.docker:Z\" \\\n"
-		ret += "\nrsync -ra \"/tekton/results/\" \"$SSH_HOST:$BUILD_DIR/tekton-results/\""
-		podmanArgs += " -v \"$BUILD_DIR/tekton-results/:/tekton/results:Z\" \\\n"
+		ret += "\n  rsync -ra \"$HOME/.docker/\" \"$SSH_HOST:$BUILD_DIR/.docker/\""
+		podmanArgs += "    -v \"$BUILD_DIR/.docker/:/root/.docker:Z\" \\\n"
+		ret += "\n  rsync -ra \"/tekton/results/\" \"$SSH_HOST:$BUILD_DIR/results/\""
+		podmanArgs += "    -v \"$BUILD_DIR/results/:/tekton/results:Z\" \\\n"
+		ret += "\nfi\n"
+
+		if taskVersion != "0.1" {
+			ret += adjustRemoteImage
+		}
 
 		script := "scripts/script-" + step.Name + ".sh"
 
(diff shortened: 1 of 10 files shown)


task-generator/remote/main.go
```diff
@@ -125,55 +125,62 @@ if [ -e "/ssh/error" ]; then
   #no server could be provisioned
   cat /ssh/error
   exit 1
+fi
+export SSH_HOST=$(cat /ssh/host)
+IS_LOCALHOST=$([ "$SSH_HOST" == "localhost" ] && echo true || echo false)
```
ifireball, line 130: instead of using a true/false string, consider using set/has value v.s. unset/empty string. That way a "true" set value is detected with:
```
if [[ $IS_LOCALHOST ]] ...
```
There as the "false" unset value is detected with:
```
if [[ -z $IS_LOCALHOST ]]
```
or
```
if ! [[ $IS_LOCALHOST ]]
```

task-generator/remote/main.go
```diff
@@ -125,55 +125,62 @@ if [ -e "/ssh/error" ]; then
...
-echo "$BUILD_DIR"
-ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/volumes"
 
-PORT_FORWARD=""
-PODMAN_PORT_FORWARD=""
-if [ -n "$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR" ] ; then
-PORT_FORWARD=" -L 80:$JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR:80"
-PODMAN_PORT_FORWARD=" -e JVM_BUILD_WORKSPACE_ARTIFACT_CACHE_PORT_80_TCP_ADDR=localhost"
-fi
+if [[ $IS_LOCALHOST == false ]]; then
```
ifireball, line 144: it would be safer to check for `[[ $IS_LOCALHOST != true ]]` here rather then assume it could only be `"true"` or  `"false"`. Alternatively consider my suggestion above.

task-generator/remote/main.go
```diff
@@ -125,55 +125,62 @@ if [ -e "/ssh/error" ]; then
...
-		podmanArgs += " -v \"$BUILD_DIR/.docker/:/root/.docker:Z\" \\\n"
-		ret += "\nrsync -ra \"/tekton/results/\" \"$SSH_HOST:$BUILD_DIR/tekton-results/\""
-		podmanArgs += " -v \"$BUILD_DIR/tekton-results/:/tekton/results:Z\" \\\n"
+		ret += "\n  rsync -ra \"$HOME/.docker/\" \"$SSH_HOST:$BUILD_DIR/.docker/\""
+		podmanArgs += "    -v \"$BUILD_DIR/.docker/:/root/.docker:Z\" \\\n"
+		ret += "\n  rsync -ra \"/tekton/results/\" \"$SSH_HOST:$BUILD_DIR/results/\""
+		podmanArgs += "    -v \"$BUILD_DIR/results/:/tekton/results:Z\" \\\n"
+		ret += "\nfi"
 
 		script := "scripts/script-" + step.Name + ".sh"
```
ifireball, line 186: This lines below here take the original build task's script and pack it into a file that is then pushed to the remote server and execute it.

Since the remote pushing and execution had now become optional, we need to either add more conditions here to execute the original code instead of writing it to file, or change the condition below (line 215) so that is will invoke the script file locally in the local case.

Without this the task is effectively a big noop in the local case.

ifireball (changes requested):
I found what seems to be one major issue where the task essentially becomes a noop in the local case.

Also since the `buildah-remote` task is now gaining local capabilities, it may be desirable to stop using the `task-generator` script and instead switch to maintaining the generated code directly. We can do this under the assumption that the remote-capable task will eventually replace the local-only task. Doing this will IMO simplify maintenance going forward.

task-generator/remote/main.go
arewm: Thanks for catching that. I had intended to put in a command to run the script locally. It got lost in my head during all of the changes.
```
"else\n  bash" + containerScript
```

chmeliik:
> Also since the buildah-remote task is now gaining local capabilities, it may be desirable to stop using the task-generator script and instead switch to maintaining the generated code directly.

+1, the task-generator is quickly becoming very difficult to understand (the diff here is kind of impossible to review)

Can we instead add the remote capabilities to regular `buildah`
* and make `buildah-remote` an alias for buildah (the same way as all the `buildah-*gb` variants, but with no patch.yaml)
* or deprecate buildah-remote altogether?

arewm:
We certainly can consolidate the tasks. Changes like the one that I proposed in https://github.com/konflux-ci/build-definitions/pull/1185 will become much simpler as well. 

If we consolidate, however, we will have to ensure that we support builds when the multi-platform-controller is not enabled. I don't think that we will be able to consolidate before the controller supports local builds.

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -214,41 +214,53 @@ spec:
     image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f@sha256:246adeaaba600e207131d63a7f706cffdcdc37d8f600c56187123ec62823ff44
```
arewm, line 214: @ifireball @chmeliik, I just realized there is a remaining issue that we will need to resolve with this PR.
These remote jobs are not run using a buildah image for the container. When we run locally, we will need to either embed our buildah binary into the multi-platform runner or pull a buildah image to run the build in.

arewm: Ah, buildah is already in it: https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/multi-platform-runner/Dockerfile but we will need to ensure that this image is up to date since it hasn't changed

ifireball: Can we switch the the image the local task is using? It does not seem like the remote commands require stuff that I wouldn't expect to find in any image

arewm: The buildah image that we are currently using does not install `rsync`, `openssh-clients` and it does not have `kubectl` either. 

Something will need to be done as I don't think that the runner image is sufficient as it also has an older buildah version than what we currently use (1.29.1 vs 1.38.0-dev).

arewm: The easiest thing for the short term may be to update the runner to be based on the current buildah image that we use. This will change its parent image from being UBI-based to Fedora based.

task-generator/remote/main.go
arewm: I have made this change here and all other locations where I perform this check.

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
arewm: Resolving this thread now that I have changed to all tasks using https://github.com/konflux-ci/buildah-container/blob/main/Containerfile.task (which is the buildah-task image)

task-generator/remote/main.go
```diff
@@ -103,12 +103,13 @@ func convertToSsh(task *tektonapi.Task, taskVersion string) {
...
-	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. Since
-	// we cannot modify the parameter itself, this replacement needs to happen in any task
+	// example, this will append -arm64 if PLATFORM is linux/arm64 if not present. If PLATFORM
+	// does not contain a "/" then the entire platform will be appended to IMAGE (i.e. "localhost").
+	// Since we cannot modify the parameter itself, this replacement needs to happen in any task
 	// step where the IMAGE parameter is used.
 	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
 	// not be appended again based on the PLATFORM.
-	adjustRemoteImage := `if [[ "${IMAGE##*-}" != "${PLATFORM##*/}" ]]; then
+	adjustRemoteImage := `if [ -n "${APPEND_PLATFORM_ARCH}" ] &&  [ "${IMAGE##*-}" != "${PLATFORM##*/}" ]; then
```
arewm, line 112: One potential concern with #1185 is that the suffix will always be added if the remote task is used. Now that we are changing the remote task to be a general-purpose task (i.e. being able to run locally as well), I modified the behavior so that this task _can_ be used as the only build task in a single-arch mode without affecting the image tag used.

task-generator/remote/main.go
```diff
@@ -102,14 +102,16 @@ func convertToSsh(task *tektonapi.Task, taskVersion string) {
...
-	// we cannot modify the parameter itself, this replacement needs to happen in any task
-	// step where the IMAGE parameter is used.
-	// If a user defines the IMAGE parameter with an -arm64 suffix, the arm64 suffix will
-	// not be appended again based on the PLATFORM.
-	adjustRemoteImage := `if [[ "${IMAGE##*-}" != "${PLATFORM##*/}" ]]; then
-  export IMAGE="${IMAGE}-${PLATFORM##*/}"
+	// the image by auto-appending the a sanitized PLATFORM parameter. For example, this
+	// will append linux-arm64 if PLATFORM is linux/arm64 and IMAGE_APPEND_PLATFORM is true.
+	// Many special characters are not allowed in tags so we will replace anything that
+	// isn't alphanumeric with a "-" to be safe. Since we cannot modify the parameter itself,
```
mmorhun, line 108: Should we fail the pipeline if users specify invalid tag? Then we don't have to check / replace in every task.

task-generator/remote/main.go
```diff
@@ -142,55 +144,65 @@ if [ -e "/ssh/error" ]; then
   #no server could be provisioned
   cat /ssh/error
   exit 1
+fi
+export SSH_HOST=$(cat /ssh/host)
+[ "$SSH_HOST" == "localhost" ] && IS_LOCALHOST=true
```
mmorhun, line 149: nitpick: just if condition would be easier to read than a functional style

chmeliik:
> > This change will require the multi-platform controller to set
> > the /ssh/host to localhost in order for the builds to run in-cluster.
> 
> Ah, so we can't really replace replace the regular buildah task with this one if we don't get rid of this requirement. The basic buildah use case shouldn't require multi-platform-controller

Another problem for unifying the tasks is that the user doesn't get to decide whether to build in-cluster or on a VM - the multi-platform-controller config does.

task-generator/remote/main.go
```diff
@@ -142,55 +144,65 @@ if [ -e "/ssh/error" ]; then
...
+  fi
 `
-		if taskVersion != "0.1" {
-			ret += adjustRemoteImage
-		}
 		env := "$PODMAN_PORT_FORWARD \\\n"
 
 		// disable podman subscription-manager integration
-		env += " --tmpfs /run/secrets \\\n"
+		env += "    --tmpfs /run/secrets \\\n"
```
mmorhun, line 179: What's the purpose of extra spaces?

chmeliik:
> Another problem for unifying the tasks is that the user doesn't get to decide whether to build in-cluster or on a VM - the multi-platform-controller config does.

Now that I think about it, won't these changes together with https://github.com/redhat-appstudio/infra-deployments/pull/4329/files break the option to run x86 builds on a VM?

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -420,48 +438,52 @@ spec:
       # Needed to generate base images SBOM
       echo "$BASE_IMAGES" > /shared/base_images_from_dockerfile
 
-      buildah push "$IMAGE" oci:rhtap-final-image
+      buildah push "$IMAGE" "oci:konflux-final-image:$IMAGE"
```
chmeliik, line 441: why the switch to `oci:$path:$image`?

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -207,38 +214,49 @@ spec:
...
+        rsync -ra /activation-key/ "$SSH_HOST:$BUILD_DIR/volumes/activation-key/"
+        rsync -ra /additional-secret/ "$SSH_HOST:$BUILD_DIR/volumes/additional-secret/"
+        rsync -ra /mnt/trusted-ca/ "$SSH_HOST:$BUILD_DIR/volumes/trusted-ca/"
+        rsync -ra "$HOME/.docker/" "$SSH_HOST:$BUILD_DIR/.docker/"
+        rsync -ra "/tekton/results/" "$SSH_HOST:$BUILD_DIR/results/"
       fi
-      if [[ "${IMAGE##*-}" != "${PLATFORM##*/}" ]]; then
-        export IMAGE="${IMAGE}-${PLATFORM##*/}"
+      if [ "${IMAGE_APPEND_PLATFORM}" == "true" ]; then
+        IMAGE="${IMAGE}-${PLATFORM//[^a-zA-Z0-9]/-}"
```
chmeliik, line 256: I notice this uses a different platform tag than before, e.g. it will be `linux-x86-64` instead of `x86_64` (I think)

Could you explain the reasoning in the commit message or put the change in a separate commit?

chmeliik (commented):
Looks reasonable, my main concern is if we're not removing the option to run x86 builds on a VM

mmorhun (approved):

arewm:
> Another problem for unifying the tasks is that the user doesn't get to decide whether to build in-cluster or on a VM - the multi-platform-controller config does.

> Now that I think about it, won't these changes together with https://github.com/redhat-appstudio/infra-deployments/pull/4329/files break the option to run x86 builds on a VM?

The user decides what platform they want to build by specifying the parameter. The configuration determines whether that should happen locally or on a VM if necessary. If it doesn't matter whether builds happen locally or remotely, then this implementation is left up to the system. One challenge for the host configs/MPC controller will be to identify the best way to migrate workloads from a VM-configured platform to a local-configured platform (or vice versa). I think that these challenges exist outside of the scope of the currently proposed changes to buildah-remote.

For now, there are only AMD nodes in clusters, but there might be ARM nodes on the cluster in the future. With the changes in redhat-appstudio/infra-deployments/pull/4329, users can still run x86 builds on a VM, they just need to select one of the configured non-local hosts (i.e. using any of the [`*/amd64` platforms](https://github.com/redhat-appstudio/infra-deployments/pull/4329/files#diff-aa7873886e13f9f939b03fc2748a7b7ca96886f0ded13f12582e2f02f27730d9R16-R34). There is a desire to move `linux/amd64` to be run in-cluster but this will be a breaking change and we need to migrate any current users off of it onto the new `linux-mlarge/amd64` platform first (this change will happen in the future if it happens at all).

Today, `localhost` can be easily used to indicate that you want to build in-cluster as I used in [my tests](https://github.com/spork-madness/multi-platform-test-stage/pull/12/files#diff-d4a636169037563d1c4da8af737ec8ea468242b306236874db909158e1d11ff1R31-R34). When multiple node platforms are available, we will have to figure out how to enable the local builds to be scheduled to the right nodes.

task-generator/remote/main.go
arewm: This environment is used inside the if block of running the build script remotely so I added additional spaces to maintain the cohesion of the block.

chmeliik:
> users can still run x86 builds on a VM, they just need to select one of the configured non-local hosts (i.e. using any of the [*/amd64 platforms](https://github.com/redhat-appstudio/infra-deployments/pull/4329/files#diff-aa7873886e13f9f939b03fc2748a7b7ca96886f0ded13f12582e2f02f27730d9R16-R34)

> There is a desire to move linux/amd64 to be run in-cluster but this will be a breaking change and we need to migrate any current users off of it onto the new linux-mlarge/amd64 platform first

Ah right, it's amd64, not x86_64. `local-platforms: "linux/x86_64,` in that PR confused me. This will need good docs :sweat_smile:

task/buildah-remote/0.2/buildah-remote.yaml
arewm: One side effect of changing the step image from `quay.io/redhat-appstudio/multi-platform-runner` to `quay.io/konflux-ci/buildah-task` is that we have picked up a newer version of `buildah`

<details>

```
$ podman run --tls-verify=false --rm -it --platform=linux/arm64 quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f  buildah version
WARNING: image platform (linux/amd64) does not match the expected platform (linux/arm64)
Version:         1.29.1
Go Version:      go1.19.6
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.24.1
Git Commit:
Built:           Mon Apr  3 08:34:42 2023
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
$ podman run --tls-verify=false --rm -it --platform=linux/arm64 quay.io/konflux-ci/buildah-task:latest buildah version
Version:         1.38.0-dev
Go Version:      go1.21.11 (Red Hat 1.21.11-1.el9_4)
Image Spec:      1.1.0
Runtime Spec:    1.2.0
CNI Spec:        1.1.0
libcni Version:  v1.2.3
image Version:   5.32.0
Git Commit:
Built:           Wed Aug  7 13:15:20 2024
OS/Arch:         linux/arm64
BuildPlatform:   linux/arm64/v8
```

</details>

This included a change https://github.com/containers/common/commit/08fc0b450 where file paths are not valid tags. This resulted in the image being pulled but no repository/tags getting pulled with it. To resolve the issue, we had to properly push and pull the oci artifact's tag from its file path.

I can move this change to the commit where I change the buildah image if it would make it clearer.

task-generator/remote/main.go
arewm: The users are not providing a tag. They are providing the PLATFORM and we are sanitizing that variable so that we can use it as a tag suffix.

task/buildah-remote/0.2/buildah-remote.yaml
chmeliik: > I can move this change to the commit where I change the buildah image if it would make it clearer.

that would be nice :+1:

task/buildah-remote/0.2/buildah-remote.yaml
chmeliik: Ah, the reasoning is in the generator script

arewm: I made the change to make fewer assumptions about the value of the PLATFORM configured since all might not conform to `os/arch` (i.e. with `localhost` or if variants are added after the arch).

I'll can add that to the first commit message.

task/buildah-remote/0.2/buildah-remote.yaml
arewm: This is now in the second commit and I have added more information to the commit message as well.

task/buildah-remote/0.2/buildah-remote.yaml
arewm: Yeah. I also added more information to the first commit message.

chmeliik (approved):

task-generator/remote/main.go
arewm: I have changed this.

ifireball:
> > Another problem for unifying the tasks is that the user doesn't get to decide whether to build in-cluster or on a VM - the multi-platform-controller config does.
> 
> Now that I think about it, won't these changes together with https://github.com/redhat-appstudio/infra-deployments/pull/4329/files break the option to run x86 builds on a VM?

They still get to decide, by a platform name that is explicit about it such as `linux-mlarge/amd64` as opposed to assuming `linux/amd64` would end on a VM, which is something they shouldn't care about if they only want to build for a foreign platform as opposed to explicitly asking to be on a VM.

PR #1219, opened 2024-07-30
Sprinkle some log statements in inject-and-push

ralphbean:
Just to help users see what's going on.

### DIFF
--- a/task/buildah/0.1/buildah.yaml
+++ b/task/buildah/0.1/buildah.yaml
@@ -502,10 +502,14 @@ spec:
         echo "WARNING: provided deprecated BUILDER_IMAGE parameter has no effect."
       fi
 
+      echo "Inspecting $IMAGE to find its base image name and digest"
       base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $IMAGE | cut -f1 -d'@')
       base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $IMAGE)
+
       container=$(buildah from --pull-never $IMAGE)
+      echo "Injecting sbom files into /root/buildinfo/content_manifests/ of $IMAGE"
       buildah copy $container sbom-cyclonedx.json sbom-purl.json /root/buildinfo/content_manifests/
+      echo "Re-applying base image annotations and labels"
       buildah config -a org.opencontainers.image.base.name=${base_image_name} -a org.opencontainers.image.base.digest=${base_image_digest} $container
 
       BUILDAH_ARGS=()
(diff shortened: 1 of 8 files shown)


chmeliik (commented):
LGTM, one small nitpick

And in my opinion it's not necessary to update the older versions of tasks. Though we don't have any versioning policy yet.

task/buildah/0.2/buildah.yaml
```diff
@@ -488,17 +488,22 @@ spec:
...
       #!/bin/bash
       set -e
+      echo "Inspecting $IMAGE to find its base image name and digest"
       base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $IMAGE | cut -f1 -d'@')
       base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $IMAGE)
+
       container=$(buildah from --pull-never $IMAGE)
+      echo "Injecting sbom files into /root/buildinfo/content_manifests/ of $IMAGE"
       buildah copy $container sbom-cyclonedx.json sbom-purl.json /root/buildinfo/content_manifests/
+      echo "Re-applying base image annotations and labels"
```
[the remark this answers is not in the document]
ralphbean: Ack, that's right.

task/buildah/0.1/buildah.yaml
```diff
@@ -502,17 +502,22 @@ spec:
...
       fi
 
+      echo "Inspecting $IMAGE to find its base image name and digest"
       base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' $IMAGE | cut -f1 -d'@')
       base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' $IMAGE)
+
       container=$(buildah from --pull-never $IMAGE)
+      echo "Injecting sbom files into /root/buildinfo/content_manifests/ of $IMAGE"
       buildah copy $container sbom-cyclonedx.json sbom-purl.json /root/buildinfo/content_manifests/
+      echo "Re-applying base image annotations and labels"
```
chmeliik, line 512:
```suggestion
      echo "Re-applying base image annotations"
```
Here as well, if we're keeping the `0.1` changes

ralphbean: Doh, how did I miss that. Thanks!

ralphbean: Regarding old task versions, I'd like to add it anyways since - it's for me. I need the visual aid when reviewing the output of buildah tasks and trying to confirm that manifests are being injected where they should be.

(Long story, but the ai-lab images are so huge that ec times out trying to verify that the files exist in /root/buildinfo/, but it appears to suppress the timeout and instead just reports that the files are missing. Related to https://github.com/enterprise-contract/ec-cli/issues/1788)

chmeliik (approved):

PR #1220, opened 2024-07-30
feat(RHTAPWATCH-1011): support custom certificate in buildah

yftacherzog:
The build step already included support for mounting custom ca-bundle. This change introduces similar capabilities also to steps inject-sbom-and-push and upload-sbom.


avi-biton (approved):
/lgtm

PR #1221, opened 2024-07-30
feat(RHTAPWATCH-1019): support custom certificate in show-sbom

yftacherzog:
Support mounting a custom ca-bundle to allow the show-sbom task to use a registry with a self-signed certificate.

### DIFF
--- a/task/show-sbom/0.1/show-sbom.yaml
+++ b/task/show-sbom/0.1/show-sbom.yaml
@@ -20,6 +20,14 @@ spec:
         refers to a multi-arch image and this parameter is empty, the task will default to use "linux/amd64".
       type: string
       default: "linux/amd64"
+    - name: CA_TRUST_CONFIG_MAP_NAME
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: trusted-ca
+    - name: CA_TRUST_CONFIG_MAP_KEY
+      type: string
+      description: The name of the key in the ConfigMap that contains the CA bundle data.
+      default: ca-bundle.crt
   steps:
   - name: show-sbom
     image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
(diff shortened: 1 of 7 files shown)


task/show-sbom/0.1/show-sbom.yaml
```diff
@@ -68,3 +76,15 @@ spec:
       else
         download_sbom_with_retry " --platform=$PLATFORM "
       fi
+    volumeMounts:
+    - name: trusted-ca
+      mountPath: /etc/ssl/certs
+      readOnly: true
```
chmeliik, line 82: Is there a reason not to do it using `update-ca-trust` like most other tasks?

yftacherzog: I wanted to avoid copying things around inside the task itself. In hindsight, I think the mount path should be `/etc/pki/tls/certs/ca-custom-bundle.crt` and should mount only the specific file, as `/etc/ssl/certs` is a link on that image.

chmeliik: Speaking of symlinks:

```bash
podman run --rm -ti quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9 ls -l /etc/pki/tls/certs/ca-bundle.crt   
lrwxrwxrwx. 1 root root 49 Aug 29  2023 /etc/pki/tls/certs/ca-bundle.crt -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
```

The actual location is /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem. IMO this is why it's better to use `update-ca-trust` when possible

yftacherzog: That's the reason I gave it a different name and I put it on that location.
I think it leaves the task itself cleaner and avoids writing into the container filesystem.

Anyhow, let me know if you want it changed.
Next, I'm planning to change some additional tasks as well for similar reasons, so I'll repeat the same pattern, we agree upon here, there as well.

yftacherzog: Yes. I have a local Konflux setup in which I have a CA bundle created with cert-manager that includes the default CA alongside a self-signed certificate that is used by an instance of [docker registry](https://hub.docker.com/_/registry).

I can confirm that the task is able to fetch the sbom when the certificate is mounted this way and cannot fetch it if I don't mount it (i.e. when it uses just the certificates bundled by default inside the image).

chmeliik (approved):

yftacherzog:
@chmeliik do you think we can have this one and https://github.com/konflux-ci/build-definitions/pull/1220 merged?

chmeliik:
> @chmeliik do you think we can have this one and #1220 merged?

My bad, I thought you had merge permissions. Added both to the merge queue

PR #1222, opened 2024-07-30
fix: clean up logic around config.toml handling

ralphbean:


ralphbean:
See also https://gitlab.com/redhat/rhel-ai/disk-images/nvidia-bootc/-/merge_requests/56

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -155,19 +155,19 @@ spec:
...
-          fi
         fi
-        echo "Using the following config.toml file $CONFIG_TOML_FILE:"
-        cat /var/workdir/source/$CONFIG_TOML_FILE
 
+        if [ -f "/var/workdir/source/${CONFIG_TOML_FILE}" ]; then
+          echo "Using the ${CONFIG_TOML_FILE} file found in the repository root!"
+        else
+          echo "No ${CONFIG_TOML_FILE} file found in the repository. Set params.${CONFIG_TOML_FILE}_FILE to a valid config.toml file."
+          exit 1
```
arewm, line 166: This will make it impossible for the task to run if there is no file regardless as to whether the parameter was specified or not. Is that desired?

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -155,19 +155,19 @@ spec:
...
-          else
-            echo "No config.toml file found. Using an empty configuration."
-            touch /var/workdir/source/$CONFIG_TOML_FILE
-          fi
         fi
-        echo "Using the following config.toml file $CONFIG_TOML_FILE:"
-        cat /var/workdir/source/$CONFIG_TOML_FILE
 
+        if [ -f "/var/workdir/source/${CONFIG_TOML_FILE}" ]; then
+          echo "Using the ${CONFIG_TOML_FILE} file found in the repository root!"
```
arewm, line 163: This is not necessarily in the repository root as a user can provide a path to a subdirectory.

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -155,19 +155,19 @@ spec:
...
+        else
+          echo "No ${CONFIG_TOML_FILE} file found in the repository. Set params.${CONFIG_TOML_FILE}_FILE to a valid config.toml file."
```
arewm, line 165:
```suggestion
          echo "No ${CONFIG_TOML_FILE} file found in the repository. Set params.CONFIG_TOML_FILE to a valid config.toml file."
```
?

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: Right, but - if the provided var is `all-my-tomls/config-qcow2.toml` then the print statement will still make sense:

> Using the all-my-tomls/config-qcow2.toml file found in the repository root!

I dunno, maybe just change it to say "found in the repository"?

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: I think so. Right? Like, what's the behavior of bib if there is literally no config.toml at all?

arewm: This is effectively a breaking change. If we want to make it explicit like this then will we have to bump the version?

pastequo: I don't know if the file is missing, but bib is fine if the file is empty

pastequo:
Maybe the PR title doesn't reflect what you want to achieve, but the CONFIG_TOML_FILE is taken into account (at least in our pipeline it seems ok)

chmeliik:
> Maybe the PR title doesn't reflect what you want to achieve, but the CONFIG_TOML_FILE is taken into account (at least in our pipeline it seems ok)

Looking at https://gitlab.com/redhat/rhel-ai/disk-images/nvidia-bootc/-/merge_requests/56, it's not up to date with the current build-definitions main. Maybe the behavior is already fixed?

A very nasty bug in the gitlab version is
```bash
if [! -n "${CONFIG_TOML_FILE}" ]; then
```
This is just a much harder to spot `if nonexistent-command`

```bash
#!/bin/bash
set -e

if [! -n "" ]; then
    echo 'true'
else
    echo 'false'
fi
```

```
$ bash /tmp/x.bash
/tmp/x.bash: line 4: [!: command not found
false
```

task/build-vm-image/0.1/build-vm-image.yaml
ralphbean: OK - I revised this to drop the `exit 1` and restore the `touch` line. (Now I understand what that line was for. I assumed bib would balk at an empty config.toml file.)

arewm (approved):

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -155,19 +155,19 @@ spec:
...
         fi
-        echo "Using the following config.toml file $CONFIG_TOML_FILE:"
-        cat /var/workdir/source/$CONFIG_TOML_FILE
 
+        if [ -f "/var/workdir/source/${CONFIG_TOML_FILE}" ]; then
+          echo "Using the ${CONFIG_TOML_FILE} file found in the repository."
+        else
+          echo "No ${CONFIG_TOML_FILE} file found. Using an empty configuration."
+          touch "/var/workdir/source/${CONFIG_TOML_FILE}"
+        fi
```
chmeliik, line 167: With the current version of the task, explicitly passing a non-existent CONFIG_TOML_FILE would result in a failure (intentionally) rather than ignoring the problem and creating an empty config.

IMO that's the saner behavior that one would expect from most tools, but there seems to be a natural pull towards creating the empty config instead (both in your and in Andrew's PRs :smile: )

I'm fine with it either way, the log message should make it clear why e.g. a mis-spelled config file path is being ignored

arewm: Aha. Now that you say that, I do remember those iterations two weeks back. I think that the reason that both @ralphbean and myself did that is because it enables cleaner/clearer logic. 

Once I tried to allow failures when a nonexistent file is specified, the logic became more confusing.

arewm: See the table for intended behavior from the prior PR: https://github.com/konflux-ci/build-definitions/pull/1171#discussion_r1686588637

chmeliik: I wouldn't say it _enables_ cleaner logic, it's just a touch easier to make it clean. Can still be clean and behave nicely though

```bash
if [ -z "$CONFIG_TOML_FILE" ]; then
    CONFIG_TOML_FILE=config.toml
    touch "$CONFIG_TOML_FILE"
elif ! [ -f "$CONFIG_TOML_FILE" ]; then
    echo "Config file doesn't exist: $CONFIG_TOML_FILE"
    exit 1
fi
```

ralphbean: I fear I've made a mess. I wasn't aware of the conversations you guys had the other week about this and I assumed this section of code was just carelessly written in the first place during fast-hacking to make vm building work in early June, when in reality you guys had really thought it through. The code was written more recently than I knew!

My real goal here was to get https://gitlab.com/redhat/rhel-ai/disk-images/nvidia-bootc/-/commit/3ab9e272069e9469abc2ace36bf11b63a438f456 to happen - and it did.

I'm inclined to _drop_ this PR without merging it, just to let it exist as you guys defined it in https://github.com/konflux-ci/build-definitions/pull/1171#discussion_r1686588637. WDYT?

chmeliik: :+1:

arewm: @ralphbean , @chmeliik 's suggestion above is similar to what you proposed. It still defines the failure criteria and clearly describes why a failure happens instead of falling back to other processes to fail (i.e. rsync). I think this would still be an improvement.

ralphbean: Ack, if we adopt that - we'll need a version bump per [this comment](https://github.com/konflux-ci/build-definitions/pull/1222#discussion_r1697548103), true?

arewm: No, If the only failure is a user specifying a file path that does not exist then that is consistent. The breaking change previously is if no file was specified. We would not create an empty file for them.

ralphbean:
Hey guys, I'm going to drop this. Focusing on other things.

PR #1229, opened 2024-08-01
chore: add jencull to the integration team

dirgim:


kasemAlem (approved):

dheerajodha (commented):
@dirgim Should we also add Sonam?

dirgim:
> @dirgim Should we also add Sonam?

Good idea, added.

PR #1226, opened 2024-07-31
Adding trusted artifacts multi arch docker build pipeline

joejstuart:
This change adds the multi arch oci-ta docker build pipeline.

https://issues.redhat.com/browse/EC-714

PR #1236, opened 2024-08-02
DO NOT MERGE: Adding multi-arch pipeline

arewm:
This is a draft PR for a multi-arch pipeline with matrix builds. It cannot be merged yet as there are still a couple of outstanding issues with running a matrix build including:

* [ ] Multi-platform controller does not support running workloads on the local cluster
* [ ] Support in the buildah task for localhosts to be run in cluster instead of remotely: #1216 
* [ ] Automatically appending an architecture suffix to the image tags: #1216 
* [ ] Support for referencing single-element matrix results by array: [SRVKP-5882](https://issues.redhat.com/browse/SRVKP-5882)


zregvart (approved):
I like this

pipelines/docker-build-oci-ta/patch.yaml
```diff
@@ -12,19 +12,20 @@
...
-#  8  sast-snyk-check
-#  9  clamav-scan
-# 10  sbom-json-check
-# 11  apply-tags
-# 12  push-dockerfile
+#   0  init
+#   1  clone-repository
+#   2  prefetch-dependencies
+#   3  build-container
+#   4  build-image-index
```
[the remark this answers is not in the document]
arewm: We don't use it anywhere for the "single-arch" pipeline, but encouraging its use is beneficial to ensure consistency if and when pipelines might change to multi-arch.

Tasks should universally work with image indexes IMO. I think that this approach is better than what I previously explored in #776.

An added benefit of putting the image index in the oci-ta pipeline is that it makes it easy to create the multi-arch pipeline with tasks in a logical order.

pipelines/docker-build-multi-arch-oci-ta/patch.yaml
```diff
@@ -0,0 +1,73 @@
...
+# change build-image-index to build image index
+# change JAVA_COMMUNITY_DEPENDENCIES value to "$(tasks.build-containers.results.JAVA_COMMUNITY_DEPENDENCIES[0])"
+
+# build-container
+- op: replace
+  path: /spec/tasks/3/name
+  value: build-containers
+- op: replace
+  path: /spec/tasks/3/taskRef/name
+  value: buildah-remote-oci-ta
```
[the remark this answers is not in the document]
arewm: Well, we use the same task for everything. The intention behind some of the linked issues in the beginning is to allow these to be a "pass through" to an in-cluster build without changing the task used.

pipelines/docker-build-multi-arch-oci-ta/patch.yaml
```diff
@@ -0,0 +1,73 @@
...
+#     10  clamav-scan
+#     11  sbom-json-check
+#     12  apply-tags
+#     13  push-dockerfile
+# change name to build-containers
+# add build-platforms parameter
+# add matrix to run based on build-platforms
+# change build-image-index to use results from build-containers as well as run after it
+# change build-image-index to build image index
+# change JAVA_COMMUNITY_DEPENDENCIES value to "$(tasks.build-containers.results.JAVA_COMMUNITY_DEPENDENCIES[0])"
```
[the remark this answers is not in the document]
arewm: This has been the pattern that we have maintained when exploring multi-arch pipelines. It can be revisited when the java build pipeline is reassessed for multi-arch support.

pipelines/docker-build-multi-arch-oci-ta/patch.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+# Add a pipeline definition parameter to customize the build platforms
+- op: add
+  path: /spec/params/-
+  value:
+    name: build-platforms
+    description: List of platforms to build the container images on. The available set of values is determined by the configuration of the multi-platform-controller.
+    type: array
+    default:
+      - "local/amd64"
```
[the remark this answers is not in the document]
arewm: @ifireball and I have been discussing what approach we might want to take for this. The note about being consistent for `os/arch` is valid. The value of this parameter will be a decision for the configuration of available platforms and is irrelevant for the MPC functionality change: https://github.com/konflux-ci/multi-platform-controller/pull/277

pipelines/docker-build-multi-arch-oci-ta/patch.yaml
```diff
@@ -0,0 +1,73 @@
...
+# add build-platforms parameter
+# add matrix to run based on build-platforms
+# change build-image-index to use results from build-containers as well as run after it
+# change build-image-index to build image index
+# change JAVA_COMMUNITY_DEPENDENCIES value to "$(tasks.build-containers.results.JAVA_COMMUNITY_DEPENDENCIES[0])"
+
+# build-container
+- op: replace
+  path: /spec/tasks/3/name
+  value: build-containers
```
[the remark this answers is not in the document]
arewm: I like this change. I'll resolve this and push the change as it needs to be updated more than just here.

pipelines/docker-build-multi-arch-oci-ta/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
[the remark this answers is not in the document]
arewm: I see this missing in other files too so I assume not.

pipelines/docker-build-rhtap/patch.yaml
```diff
@@ -8,6 +8,37 @@
     "pipelines.openshift.io/used-by": "build-cloud"
     "pipelines.openshift.io/runtime": "generic"
     "pipelines.openshift.io/strategy": "docker"
+# Remove unused parameters from the template
```
arewm, line 11: @chmeliik @mmorhun , I have removed unneeded parameters from the RHTAP pipeline in this PR. Let me know if you want to get these in with a different PR.

pipelines/docker-build-multi-arch-oci-ta/README.md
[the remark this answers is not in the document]
arewm: I had a missing `-` so it was actually complaining since it couldn't process an expected list.

ifireball:
Is this an implementation of [STONEBLD-2567](https://issues.redhat.com/browse/STONEBLD-2567)?

arewm:
@ifireball, yes. As is https://github.com/konflux-ci/build-definitions/pull/1226.

simonbaird:
See also https://github.com/konflux-ci/build-definitions/pull/1226 which may have some overlap.

arewm:
@simonbaird, This is not blocked on the Chains bug fix. I have worked around that with https://github.com/konflux-ci/build-definitions/pull/1204

simonbaird:
> @simonbaird, This is not blocked on the Chains bug fix. I have worked around that with #1204

Ack, so I think we're likely to abandon #1226, which (AIUI) tries to implement a multi-arch pipeline without matrix support. Do you agree @joejstuart ?

arewm:
All dependent issues are resolved except for supporting a matrix of size 1. This issue will be most likely to be resolved by upgrading to OCP 4.15 which will start in the next sprint. In order to work around this limitation, I have proposed to add a default set of two platforms to build on -- the localhost and arm.

pipelines/docker-build-oci-ta/patch.yaml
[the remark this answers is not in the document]
arewm: This can also be a way that users can create a single-arch image index before we get support for a matrix of size 1.

zregvart:
> I have proposed to add a default set of two platforms to build on -- the localhost and arm.

Let's not do `localhost`, let's set them to proper values, e.g. `amd64`, as long as I'm nitpicking, "platform" is "os+architecture", what we have here is just "architecture", we might want to future proof our terminology and Tasks.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -7,29 +7,33 @@ metadata:
...
     type: array
-  - default: ""
-    description: Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
-    name: IMAGE_EXPIRES_AFTER
+  - name: IMAGE_EXPIRES_AFTER
+    description: Delete image tag after specified time resulting in garbage collection of the digest. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.
     type: string
+    default: ""
+  - name: BUILD_IMAGE_INDEX
+    description: Create an OCI image index referencing all passed params.IMAGES. This allows us to always include this task for a consistent pipeline even if a user does not want to generate the image index. If the image index generation is skipped, the task will forward values for params.IMAGES[0] to results.IMAGE_*.
```
chmeliik, line 34: What does this simplify compared to not including this task in a single-arch pipeline?

I'm concerned about adding very specific behavior to tasks completely in service of one specific pipeline if that behavior doesn't make sense on its own. Our tasks should make sense standalone, not just in the sample pipeline

task/build-image-index/0.1/README.md
chmeliik, line 1: There's two READMEs here, one in the `0.1` directory and one in the directory above. I think we normally put READMEs in the version directories

pipelines/fbc-builder/patch.yaml
```diff
@@ -31,30 +31,37 @@
 # Remove tasks
 # Example - yq .spec.tasks.[].name ../build-definitions/pipelines/template-build/template-build.yaml | nl -v 0
 # to compute offsets
```
mmorhun, line 33: This could be in sync with other similar comments

pipelines/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
 - docker-build
 - docker-build-oci-ta
+- docker-build-multi-arch-oci-ta
```
mmorhun, line 6: We have both `multi-arch` and `multi-platform`. Let's stick to one of them.

task/build-image-index/0.1/build-image-index.yaml
arewm: Many of the builds that use the pipelines will compile binaries which cannot be run by multiple host architectures. Without an Image Index, the host will not have the ability to selectively pull the image that will work on the system. By adding the image index task by default, we enable users who only want to build in-cluster to still be conformant to the [image-spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md)

> `platform` object
> 
> This OPTIONAL property describes the minimum runtime requirements of the image. This property SHOULD be present if its target is platform-specific.

pipelines/template-build/template-build.yaml
```diff
@@ -121,6 +125,28 @@ spec:
       workspaces:
         - name: source
           workspace: workspace
+    - name: build-image-index
```
mmorhun, line 128: Not sure it belongs here. To me it's more related to the docker-build pipeline.

task/build-image-index/0.1/README.md
arewm: One of them was accidentally added by me recently: https://github.com/konflux-ci/build-definitions/pull/1204/files

pipelines/template-build/template-build.yaml
arewm: `template-build` is used as a template to create many of the different pipelines which generate container images. I think it makes sense to be here, but should we rename the template to be more precise about how it is intended?

The pipelines that use `template-build` are:
* docker-build
* docker-build-rhtap
* fbc-builder
* java-builder
* nodejs-builder
* tekton-bundle-builder (this is the outlier)

pipelines/kustomization.yaml
arewm: I will change to multi-platform. The remaining use of the term `multi-arch` in the repository comes from the `show-sbom` task.

task/build-image-index/0.1/build-image-index.yaml
chmeliik: I get why we would want the image-index task in a single-arch pipeline, but why do we want a parameter that makes the task not do the only thing it's meant to do?

chmeliik: Because it would be too hard for users to add/remove the task? Because there's no good way in Tekton to skip a task and still have the pipeline function? :disappointed:

chmeliik: Maybe a param that would make more sense would be something like

```
ALWAYS_BUILD_INDEX: Build an image index even if IMAGES is of length 1. Default true.
```

arewm: All of the above. By including the task as running by default (as a pass-through) then there is consistency in the result that contains the image generated. If users skip the task then they would have to change the path to the result.

I like the proposed parameter change. It also makes sense to set the default to `true` in the task and just override this as `false` in the specific pipelines for backwards compatibility. As a side note, this functionality would also enable _all_ pipelines for OLM operators to share a common pipeline definition since the same matrix pipeline can be used for single-arch builds without an image index as well as multi-arch with one (after any clusters are upgraded to OCP 4.15 so that the single-element fix is available).

arewm: Resolved in the latest commit.

mmorhun (approved):

pipelines/template-build/template-build.yaml
arewm: I think that the proper resolution here is likely to move multiple tasks out of the `template-build` and into the `docker-build` pipeline. I think that this is a larger change and it should happen in a new PR to not grow this PR any larger.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -79,10 +85,19 @@ spec:
...
+          TOADD_REPOSITORY="$(echo "$i" | cut -d: -f1)"
+          TOADD_DIGEST="sha256:$(echo "$i" | cut -d: -f3)"
+          TOADD="${TOADD_REPOSITORY}@${TOADD_DIGEST}"
+        fi
+        if [[ "$ALWAYS_BUILD_INDEX" != "true" ]]; then
+          echo "Skipping image index generation. Returning results for $TOADD"
+          echo -n "${TOADD_REPOSITORY}" > "$(results.IMAGE_URL.path)"
+          echo -n "${TOADD_DIGEST}" > "$(results.IMAGE_DIGEST.path)"
+          echo -n "${TOADD}" > "$(results.IMAGES.path)"
+          exit 0
```
chmeliik, line 100: IMO `ALWAYS_BUILD_INDEX=false` should be ignored if there's multiple IMAGES

chmeliik: Or we can have a `BUILD_INDEX=false` and raise an error if there's multiple IMAGES

arewm: Sounds good. We can always revisit this in the future as loosening of the restriction won't be a breaking change but adding functionality.

chmeliik:
> Is it still a requirement for all tasks to be in both catalogs (that doesn't seem to be a requirement for the pipelines)? I cannot create `redhat-appstudio-tekton-catalog/task-build-image-index`.

It's a requirement that all the tasks which already exist in `redhat-appstudio-tekton-catalog` keep getting updated (or we have to migrate every Konflux user to `konflux-ci/tekton-catalog`

For new tasks, yeah we don't need to push them to `redhat-appstudio-tekton-catalog`. Unfortunately, the `hack/build-and-push.sh` script is not smart enough to be able to handle that, but we can make that improvement if requested.

chmeliik:
> For new tasks, yeah we don't need to push them to `redhat-appstudio-tekton-catalog`. Unfortunately, the `hack/build-and-push.sh` script is not smart enough to be able to handle that, but we can make that improvement if requested.

Although maybe nothing would break _too hard_ if we just remove the repo existence check. It's possible that `hack/build-and-push.sh` would just create the repos for new tasks as private in `redhat-appstudio-tekton-catalog`. Which is not ideal, but not the end of the world

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -74,15 +80,29 @@ spec:
...
-          TOADD="$(echo $i | cut -d: -f1)@sha256:$(echo $i | cut -d: -f3)"
+          TOADD_REPOSITORY="$(echo "$i" | cut -d: -f1)"
+          TOADD_DIGEST="sha256:$(echo "$i" | cut -d: -f3)"
+          TOADD="${TOADD_REPOSITORY}@${TOADD_DIGEST}"
+        fi
+        if [[ "$ALWAYS_BUILD_INDEX" != "true" ]]; then
+          echo "Skipping image index generation. Returning results for $TOADD"
+          echo -n "${TOADD_REPOSITORY}" > "$(results.IMAGE_URL.path)"
+          echo -n "${TOADD_DIGEST}" > "$(results.IMAGE_DIGEST.path)"
+          echo -n "${TOADD}" > "$(results.IMAGES.path)"
```
chmeliik, line 104: Now that I look at it again, I noticed the script only initializes `TOADD_REPOSITORY` and `TOADD_DIGEST` in the `if [[ $(echo "$i" | tr -cd ":" | wc -c) == 2 ]]` case. Other cases will be handled wrong (either cause a failure with un-initialized vars or use vars that were set earlier in the for loop)

arewm: good catch! I'll fix that really quick.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -74,15 +80,33 @@ spec:
...
+
       image_manifests=""
       buildah manifest create "$IMAGE"
       for i in $@
       do
         TOADD="$i"
-        if [[ $(echo $i | tr -cd ":" | wc -c) == 2 ]]; then
+        TOADD_REPOSITORY="$(echo "$i" | cut -d@ -f1)"
+        TOADD_DIGEST="$(echo "$i" | cut -d@ -f2)"
+        TOADD_TAG="$(echo "$IMAGE" | cut -d: -f2)"
```
chmeliik, line 95: Shouldn't we be extracting the tag from `$i` rather than `$IMAGE`? Do we want to ignore the actual tag in `IMAGES` for some reason?

Note: this code wouldn't work as is, the tag extraction doesn't handle digests (and port numbers in the registry host, but that's a problem across almost every task in the pipeline)

chmeliik (approved):
I think this works, though the build-image-index code is a bit mind-bending in its handling of tags

task/buildah/0.2/buildah.yaml
```diff
@@ -107,7 +107,7 @@ spec:
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
-  - description: Image repository where the built image was pushed
+  - description: Image repository and tag where the built image was pushed
```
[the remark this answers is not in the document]
arewm: If no tag is set than an implicit tag of `latest` will be used.

task/build-image-index/0.1/build-image-index.yaml
arewm: I added this tag logic because the e2e tests were failing since no tag was getting set. After looking at the pipeline definitions again, this is because I was passing the `IMAGE_REF` from the container build instead of `IMAGE_URL`. The result of this is that the IMAGE_URL that was returned ended up stripping off the tag.

For a matrix build, we need to pass `IMAGE_REF`, but for non-matrix builds, however, `$(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)` should be used to define the input `IMAGES` to allow the pass-through functionality.

task/buildah/0.2/buildah.yaml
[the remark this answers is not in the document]
chmeliik: Used yes, but not returned in IMAGE_URL

task/build-image-index/0.1/build-image-index.yaml
chmeliik: I'm not sure I understood that, but it would be great if the task worked in general, not only for some possible configurations of a pipeline

chmeliik (approved):

PR #1240, opened 2024-08-02
Use a var for the images file path in the RHTAP gather-deploy-images task

simonbaird:
Short explanation: Clean up the code a little by using a variable.

Long explanation: This task gets converted to a bash script and wrapped by a Jenkins shared library in the brand new RHTAP Jenkins pipeline sample templates. When testing the RHTAP Jenkins pipeline I found the default Jenkins user could not write to `/tmp/`, so I wanted to change the location of the file. This refactor makes that just a single line change, rather than having to change the path in three places.

See also
https://github.com/redhat-appstudio/tssc-sample-jenkins/pull/1/ where this PR was suggested by @mmorhun, and also
https://github.com/jduimovich/tssc-jenkins/pull/2 where the cosign, Enterprise Contract, and promote pipeline support was being added to the RHTAP Jenkins templates.

### DIFF
--- a/task/gather-deploy-images/0.1/gather-deploy-images.yaml
+++ b/task/gather-deploy-images/0.1/gather-deploy-images.yaml
@@ -40,6 +40,7 @@ spec:
       set -euo pipefail
 
       IMAGE_PATH='.spec.template.spec.containers[0].image'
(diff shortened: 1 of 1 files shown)


lcarva (approved):
/lgtm

PR #1247, opened 2024-08-05
feat(KONFLUX-1503): Set some requests and limits so pods can spread across cluster nodes

jhutar:
IMO this needs to happen before https://github.com/redhat-appstudio/infra-deployments/pull/4275

Some support for these values: https://docs.google.com/document/d/1SvR7AP1UCU67bV0UvYhn6b8U30LCoPu9zcD5Wd1uAxk/edit?pli=1#heading=h.7kx4nf5n5g

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -422,10 +422,11 @@ spec:
         echo "$BASE_IMAGES" >/shared/base_images_from_dockerfile
       computeResources:
         limits:
-          memory: 4Gi
+          cpu: "4"
+          memory: 8Gi
         requests:
-          cpu: 250m
-          memory: 512Mi
+          cpu: "1"
+          memory: 2Gi
       securityContext:
         capabilities:
           add:
@@ -444,6 +445,13 @@ spec:
         find $(cat /shared/container_path) -xtype l -delete
         echo "Running syft on the image filesystem"
         syft dir:$(cat /shared/container_path) --output cyclonedx-json=/var/workdir/sbom-image.json
+      computeResources:
+        limits:
+          cpu: "2"
+          memory: 4Gi
+        requests:
+          cpu: 500m
+          memory: 1Gi
     - name: analyse-dependencies-java-sbom
       image: quay.io/redhat-appstudio/hacbs-jvm-build-request-processor:127ee0c223a2b56a9bd20a6f2eaeed3bd6015f77
       volumeMounts:
@@ -458,6 +466,13 @@ spec:
         else
           touch $(results.JAVA_COMMUNITY_DEPENDENCIES.path)
         fi
+      computeResources:
+        limits:
+          cpu: 200m
+          memory: 512Mi
+        requests:
+          cpu: 100m
+          memory: 256Mi
       securityContext:
         runAsUser: 0
     - name: merge-syft-sboms
@@ -492,6 +507,13 @@ spec:
         # write the CycloneDX unified SBOM
         with open("./sbom-cyclonedx.json", "w") as f:
           json.dump(image_sbom, f, indent=4)
+      computeResources:
+        limits:
+          cpu: 200m
+          memory: 512Mi
+        requests:
+          cpu: 100m
+          memory: 256Mi
       securityContext:
         runAsUser: 0
     - name: merge-cachi2-sbom
@@ -505,6 +527,13 @@ spec:
         else
           echo "Skipping step since no Cachi2 SBOM was produced"
         fi
+      computeResources:
+        limits:
+          cpu: 200m
+          memory: 512Mi
+        requests:
+          cpu: 100m
+          memory: 256Mi
       securityContext:
         runAsUser: 0
     - name: create-purl-sbom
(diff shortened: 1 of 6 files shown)


task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml
```diff
@@ -204,10 +204,11 @@ spec:
...
     computeResources:
       limits:
-        memory: 4Gi
+        cpu: "4"
+        memory: 8Gi
       requests:
-        cpu: 250m
-        memory: 512Mi
+        cpu: "1"
+        memory: 2Gi
```
hugares, line 211: I am OK to keep this in the change but I want to mention that for remote build, the pod in ocp does not do much as the build is happening in a remote VM.  We can keep it like that as it will help the scheduler to spread pods on nodes.

hugares (approved):
/lgtm

task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml
jhutar: Oh, right. We should set requests/limits for buildah-remote task in build-definitions so we are not wasting resources.

jhutar:
Fixed a typo found by PR checks

hugares (approved):

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -418,10 +418,11 @@ spec:
...
       computeResources:
         limits:
-          memory: 4Gi
+          cpu: "4"
+          memory: 8Gi
         requests:
-          cpu: 250m
-          memory: 512Mi
+          cpu: "1"
+          memory: 2Gi
```
lcarva, line 425: For memory, limit and request should be the same value. This article does a good explaining why: https://home.robusta.dev/blog/kubernetes-memory-limit

jhutar: I would love to have limits == requests, but I assume that is for far bigger fight (and might not make sense from cost point of view).

hugares: I also think we should have limits=requests but I would do this in steps. Already this PR is increasing the requests for memory and CPU and this will cause pods to be more spread. I would like to see the effect on the worker pools before doing the other step of making memory request=limits.

lcarva: As I understand it, if those values are different we are shooting ourselves in the foot. CPU and memory are different. My concern is only with memory. From the link I shared:

> CPU is fundamentally different than memory. CPU is a compressible resource and memory is not. In simpler terms, you can give someone spare CPU in one moment when it's free, but that does not obligate you to continue giving them CPU in the next moment when another pod needs it. There is no downside to giving away idle CPU, because it's easy and non-violent to reclaim it.

That does not apply to memory.

lcarva: > I also think we should have limits=requests but I would do this in steps

Sure. We do that for the cases where compute resources is already set and this PR is simply adjusting the values. But I think we should make the values match where we are defining a new compute resource.

hugares: Kubesaw is setting a limit range in tenant namespace so we currently have memory/cpu requests/limits set for all the containers. This PR is setting new compute resource but effectively, it is  changing the value from what the limitrange is setting, which are cpu/memory requests/limits: 10m/2  CPU and 256Mi/2Gi

hugares: @lcarva Are you OK with proceeding with this change as first step or you think we should do memory requests=limits as part of this first step?

lcarva: @hugares, apologies for not coming back to this sooner. How about we go ahead with this but have a tracker for a follow up?

lcarva: Perfect. Thank you so much.

rcerven (changes requested):
please add resource request only to the task and not steps for sake of all us :)

jhutar:
> please add resource request only to the task and not steps for sake of all us :)

Hello @rcerven . For sake of me, please could you give more info on why? I assume you meant this functionality: https://tekton.dev/docs/pipelines/compute-resources/#task-level-compute-resources-configuration but it is marked as Beta, so not sure if it is OK to use it. I think it is better to use more granular settings. You are concerned about maintainability of these settings?

If I read the docs for Task-level resources feature correctly:

> If users specify a Task-level resource limit, no Step may use more than that amount of resources.

all the containers will run with whatever I configure on Task-level, so if I would use 1/2Gi requests (as that is biggest one used for `build` step), it will run all the containers with that, these 9 build pod containers would account for 9/18Gi requests that might be a significant hit for cluster resources (at least compared to what we have now). With limits, we would run out of node size.

I'm happy to do so as it will just mean more compute plane nodes created by node autoscaler and better distribution of tasks, but I thought what I propose is already big change. Also I would like to be sure it is a general consensus before doing these changes.

jhutar:
FYI rebased.

rcerven (approved):

rcerven (approved):

PR #1250, opened 2024-08-05
Fix default value in verify-signed-rpms

lcarva:
The value of the `FAIL_UNSIGNED` parameter of the `verify-signed-rpms` Task is used when calling the `rpm_verifier` binary. An empty string value is not a valid value result in this error:

```
Usage: rpm_verifier [OPTIONS]
Try 'rpm_verifier --help' for help.

Error: Invalid value for '--fail-unsigned': '' is not a valid boolean.
```

This commit changes the default value to `"false"`. This is both valid and inline with [ADR-14](https://konflux-ci.dev/architecture/ADR/0014-let-pipelines-proceed.html).

### DIFF
--- a/task/verify-signed-rpms/0.1/verify-signed-rpms.yaml
+++ b/task/verify-signed-rpms/0.1/verify-signed-rpms.yaml
@@ -11,7 +11,7 @@ spec:
     - name: FAIL_UNSIGNED
       type: string
       description: "[true | false] If true fail if unsigned RPMs were found"
-      default: ""
+      default: "false"
     - name: WORKDIR
       type: string
       default: /tmp


yftacherzog:
This task is maintained in tekton-tools and replicated here, so changes should be made there, where it's also tested.
We should probably mention that on the README.
https://github.com/redhat-appstudio/tekton-tools/blob/main/tasks/verify-signed-rpms/0.1/verify-signed-rpms.yaml

lcarva:
> This task is maintained in tekton-tools and replicated here, so changes should be made there, where it's also tested. We should probably mention that on the README. https://github.com/redhat-appstudio/tekton-tools/blob/main/tasks/verify-signed-rpms/0.1/verify-signed-rpms.yaml

Thanks for the info! That is a private repo with forking disabled. I cannot contribute.

yftacherzog:
> > This task is maintained in tekton-tools and replicated here, so changes should be made there, where it's also tested. We should probably mention that on the README. https://github.com/redhat-appstudio/tekton-tools/blob/main/tasks/verify-signed-rpms/0.1/verify-signed-rpms.yaml
> 
> Thanks for the info! That is a private repo with forking disabled. I cannot contribute.

I added you. Let me know if you're now able to create a branch on the repo itself and create a PR.

lcarva:
Let's follow up in https://github.com/redhat-appstudio/tekton-tools/pull/131

PR #1248, opened 2024-08-05
WIP: try reporting results to snyk.io

chmeliik:
# Before you complete this pull request ...

Look for any open pull requests in the repository with the title "e2e-tests update" and 
see if there are recent e2e-tests updates that will be applicable to your change.

PR #1256, opened 2024-08-06
Upload Snyk report to snyk.io on push

chmeliik:
The PR pipeline already runs Snyk, but doesn't upload the results
anywhere. Run Snyk in the push pipeline as well and upload the results
to snyk.io.

Note: we should not upload to Snyk from the PR pipeline. Each PR would
overwrite the Snyk results from other PRs. By uploading only in the push
pipeline, the results will at least always reflect the state in 'main'.

The results can be found in the 'konflux-ci/build-definitions' project
in the Snyk organization associated with the Snyk token used by the
pipeline (currently the 'developer-red-hat-trusted-application-pipeline'
organization).

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -49,9 +49,6 @@ spec:
         - name: args
           value: [".", "-c", "/workspace/shared-workspace/source/.yamllint"]
       - name: sast-snyk-check
-        params:
-          - name: SNYK_SECRET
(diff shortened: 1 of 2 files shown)


chmeliik:
> Note: we should not upload to Snyk from the PR pipeline. Each PR would overwrite the Snyk results from other PRs.

Tested in #1248

mmorhun (approved):

Issue #1249, opened 2024-08-05
etc-pki-entitlement secret ignored after #1207


owtaylor:
I have a component that has a etc-pki-secret for accessing RHEL repositories but no activation-key secret.

With the d4262b93dceaef41f8cbad313006fc236a0eb321 version of the buildah task, I see in my logs:

```
[build] Adding the entitlement to the build
[build] Adding activation key to the build
```

With 209998831f1c2fd7d51f10e9def92276848d0c95 ([failing PR for my component](https://gitlab.com/flatpak-container/flatpak-container-tools/-/merge_requests/8#note_2021217698)), I see only:

```
[build] Adding activation key to the build
```

I believe the checks in the code:

```
      if [ -d "$ACTIVATION_KEY_PATH" ]; then
        cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
        mkdir /shared/rhsm-tmp
        VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume /tmp/activation-key:/activation-key -v /shared/rhsm-tmp:/etc/pki/entitlement:Z"
        echo "Adding activation key to the build"
      elif [ -d "$ENTITLEMENT_PATH" ]; then
        cp -r --preserve=mode "$ENTITLEMENT_PATH" /tmp/entitlement
        VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume /tmp/entitlement:/etc/pki/entitlement"
        echo "Adding the entitlement to the build"
      fi
```

are faulty because if the optional secrets are missing, the mount points are still created, they just are empty directories. [The Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-secret-must-exist) doesn't specify what happens for a missing optional secret, but see [this stack exchange question](https://stackoverflow.com/questions/48222827/optional-volume-secret-volume-in-kubernetes).

PR #1269, opened 2024-08-08
issue-1249: entitlements cert not working using buildah task with support for activation keys

brianwcook:
closes https://github.com/konflux-ci/build-definitions/issues/1249 

When an activation key secret is not provided, Kubernetes still mounts a volume. This volume is also not truly empty but contains some symlinks in it. To ensure that the fallback to mounted certificates works for backwards compatibility, activation keys will only be mounted now if the 'org' key is present in the key/value secret. This is tested by checking for a file-like-object (-e) in the dir where the secret is mounted.

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -171,8 +171,6 @@ spec:
       emptyDir: {}
   stepTemplate:
     env:
-      - name: ACTIVATION_KEY
-        value: $(params.ACTIVATION_KEY)
       - name: ADDITIONAL_SECRET
         value: $(params.ADDITIONAL_SECRET)
       - name: ADD_CAPABILITIES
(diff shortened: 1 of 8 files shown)


arewm:
I have no issue with this pull request but I cannot comment on the technical details of it as I don't have experience with activation keys. 

Do we have documentation for activation keys that indicate the required org key? I assume that this is more of a bug fix than a breaking change so no version bump would be needed.

brianwcook:
this is indeed a bugfix. I have tested it pretty thoroughly. Without this fix anyone mounting entitlement certificates will see PR failures on newer versions of the buildah task. 

I am holding off on posting activation keys docs (but I have a draft) because there is a chance that there will be a follow up PR to move the 'subscription-manager register' command into a STEP script so that it is not required to be in the Dockerfile. This is being asked for by a number of people already. I have a meeting to discuss it with another team this coming week. I'd like to minimize the number of people who need to make yet another change to how they obtain content access.

arewm (approved):

PR #1271, opened 2024-08-08
Don't verify redhat-appstudio/dance-bootstrap-app

simonbaird:
This is the upstream change related to the following downstream changes to make sure everything stays in sync.

- https://github.com/redhat-appstudio/tssc-sample-jenkins/pull/5
- https://github.com/redhat-appstudio/tssc-sample-pipelines/pull/54

Explanation: For various reasons quay.io/redhat-appstudio/dance-bootstrap-app will never pass an EC check, so try to validate it.

This was causing problems during QE testing of the Jenkins pipeline in RHTAP 1.2.

Ref: https://issues.redhat.com/browse/RHTAPBUGS-1284

### DIFF
--- a/task/gather-deploy-images/0.1/gather-deploy-images.yaml
+++ b/task/gather-deploy-images/0.1/gather-deploy-images.yaml
@@ -53,6 +53,13 @@ spec:
             # don't check images that didn't change between the current revision and the target branch
             continue
           fi
+
+        fi
+
+        # Workaround for RHTAPBUGS-1284
+        if [[ "$image" =~ quay.io/redhat-appstudio/dance-bootstrap-app ]]; then
+          # Don't check the dance-bootstrap-app image
+          continue
         fi
 
         printf "%s\n" "$image"


simonbaird:
Hmm, it says "Remove quotes from right-hand side of =~ to match as a regex rather than literally" but I kinda want it to match literally.

I'll see if I can convert it to regex..

simonbaird:
Seems to work fine without the quotes :shrug: .

jduimovich (approved):
LGTM

jduimovich (approved):
LGTM

chmeliik (commented):
Unless the user somehow "promotes" the bootstrap image to stage or prod, the promotion pipeline will not try to validate the bootstrap image. Why is this change needed?

chmeliik:
> Hmm, it says "Remove quotes from right-hand side of =~ to match as a regex rather than literally" but I kinda want it to match literally.
> 
> I'll see if I can convert it to regex..

Just FYI: to match literally, you want `=` rather than `=~`

simonbaird:
> > Hmm, it says "Remove quotes from right-hand side of =~ to match as a regex rather than literally" but I kinda want it to match literally.
> > I'll see if I can convert it to regex..
> 
> Just FYI: to match literally, you want `=` rather than `=~`

I want it to match with `:latest` or `@sha256:abcdef123` at the end, so that's why I'm chosing a regex. (Could maybe use `== "quay.io/redhat-appstudio/dance-bootstrap-app"* ]]` to achieve that, but I think `=~` is fine.)

chmeliik:
> Unless the user somehow "promotes" the bootstrap image to stage or prod, the promotion pipeline will not try to validate the bootstrap image. Why is this change needed?

After looking at the Jenkins pipeline briefly, I think the root issue is that it doesn't set the TARGET_BRANCH variable so the task doesn't look for changed images, it always returns all the images. https://github.com/redhat-appstudio/tssc-sample-templates/blob/bed5fcaf8eb225dcf5ab17b09a4fb3be8dad4309/skeleton/ci/gitops-template/jenkins/rhtap/env.sh#L32

https://github.com/konflux-ci/build-definitions/blob/43cf5b8a617c87f7472b4f76542f8b7a428c511f/task/gather-deploy-images/0.1/gather-deploy-images.yaml#L11-L16

So I don't think we need the "ignore dance-boostrap-app" workaround upstream

simonbaird:
Thanks @chmeliik , I think you're probably right. I'm happy to abandon this and also https://github.com/redhat-appstudio/tssc-sample-pipelines/pull/54 .

PR #1274, opened 2024-08-09
Add ssh keepalives into buildah-remote to prevent (or at least debug) the ssh conection interruptions

mshaposhnik:
It is reported some ssh conection interruptions during the long running remote build tasks. It is not yet 100% clear what causes it, but sending those keepalives might help to keep the channel up, or understood which side is closing the connections.
See https://issues.redhat.com/browse/KFLUXBUGS-1523 for more details

### DIFF
--- a/task-generator/remote/main.go
+++ b/task-generator/remote/main.go
@@ -118,7 +118,7 @@ fi
 chmod 0400 ~/.ssh/id_rsa
 export SSH_HOST=$(cat /ssh/host)
 export BUILD_DIR=$(cat /ssh/user-dir)
-export SSH_ARGS="-o StrictHostKeyChecking=no"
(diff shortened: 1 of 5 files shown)


mshaposhnik:
@chmeliik  @brianwcook pls TAL

ifireball (changes requested):
you need to update the `task-generator/remote` tool to generate the remote tasks with this change. While we may get rid of it soon, we still use it, so without updating it your changes will be lost when the next person comes along and runs it.

mshaposhnik:
@ifireball yes i figured that out and did so, please make sure you're on the latest version :)

ifireball (approved):
ok,
/lgtm

chmeliik (approved):

PR #1282, opened 2024-08-12
ROX-23812: Add the acs-determine-image-tag task for ACS build pipelines

tommartensen:
## Description

`acs-determine-image-tag` Task is one of the custom tasks we (ACS) team use in our build pipeline. 
It runs a script in our repository to see if there are modified files in the workspace (`fail-build-if-git-is-dirty.sh`).
Then, depending on the `IMAGE_TAG_STYLE`, it runs the correct `make tag` command.
The resulting string is captured as output for use in following tasks.

The existing task is at https://github.com/stackrox/stackrox/blob/master/.tekton/determine-image-tag-task.yaml

## Validation

Validated in https://github.com/stackrox/stackrox/pull/12334/, e.g. https://console.redhat.com/application-pipeline/ns/rh-acs-tenant/pipelinerun/central-db-build-sd8b9 and https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/taskruns/central-db-build-sd8b9-determine-image-tag. 

I used `tkn bundle push quay.io/tommartensen/rhtap-tasks:acs-determine-image-tag -f acs-determine-image-tag.yaml` to build the Task bundle for the validation.

### DIFF
--- a/task/acs-determine-image-tag/0.1/README.md
+++ b/task/acs-determine-image-tag/0.1/README.md
@@ -1,0 +1,23 @@
+# acs-determine-image-tag task
+
+## Description
+
+The `acs-determine-image-tag` Task will determine the tag for the output image using the StackRox convention from 'make tag' output.
+
+## Params
+
+| name            | description                                                                         |
+|-----------------|-------------------------------------------------------------------------------------|
+| IMAGE_TAG_STYLE | Image Tag style to be used, valid options are 'main' or 'operator'.                 |
+| SOURCE_ARTIFACT | The Trusted Artifact URI pointing to the artifact with the application source code. |
+| TAG_SUFFIX      | Suffix to add to the make tag output.                                               |
+
+## Results
+
+| name              | description                   |
+|-------------------|-------------------------------|
+| IMAGE_TAG | Image Tag determined by custom logic. |
+
+## Additional links
+
+- [stackrox/stackrox](https://github.com/stackrox/stackrox)
--- a/task/acs-determine-image-tag/0.1/acs-determine-image-tag.yaml
+++ b/task/acs-determine-image-tag/0.1/acs-determine-image-tag.yaml
@@ -1,0 +1,62 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: acs-determine-image-tag
+  labels:
+    app.kubernetes.io/version: "0.1"
+spec:
+  description: The acs-determine-image-tag Task will determine the tag for the
+    output image using the StackRox convention from 'make tag' output.
+  params:
+  - name: TAG_SUFFIX
+    description: Suffix to append to generated image tag.
+    type: string
+  - name: SOURCE_ARTIFACT
+    description: The Trusted Artifact URI pointing to the artifact with
+      the application source code. This should be the result of the git-clone task,
+      results from other tasks might fail as dirty.
+    type: string
+  - name: IMAGE_TAG_STYLE
+    description: Image Tag style to be used, valid options are 'main' or 'operator'.
+    type: string
+    default: main
+  results:
+  - name: IMAGE_TAG
+    description: Image Tag determined by custom logic.
+  volumes:
+  - name: workdir
+    emptyDir: {}
+  stepTemplate:
+    volumeMounts:
+    - mountPath: /var/workdir
+      name: workdir
+  steps:
+  - name: use-trusted-artifact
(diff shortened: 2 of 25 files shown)


tommartensen:
Questions for reviewers: which annotations should we set? Saw different styles used in the repository.

task/acs-determine-image-tag/0.1/acs-determine-image-tag.yaml
```diff
@@ -0,0 +1,62 @@
...
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+    workingDir: /var/workdir/source
+    script: |
+      #!/usr/bin/env bash
+      set -euo pipefail
+
+      .konflux/scripts/fail-build-if-git-is-dirty.sh
+      image_tag=""
+      image_tag_style="$(params.IMAGE_TAG_STYLE)"
+      case "$image_tag_style" in
```
msugakov, line 50: Can swap image tag style for directory where to run make https://github.com/stackrox/stackrox/pull/12350

task/acs-determine-image-tag/0.1/README.md
```diff
@@ -0,0 +1,23 @@
+# acs-determine-image-tag task
```
msugakov, line 1: Obviously, it would be good to find out a convention suggested by the Konflux team about the way to store tenant-specific trusted tasks so that they don't show up in the list of shared tasks and don't confuse other tenants. Here, additionally, the task starts with `acs-` just as well as the existing `acs-deploy-check`, `acs-image-check` and `acs-image-scan`. The thing is that `acs-determine-image-tag` in this PR is _tenant-specific_ one which we need in order to have _our_ ACS builds compliant, the other three are _shared_ tasks that any tenant can use to leverage ACS features in their pipeline. The fact that they all will sit in the same directory with the same `acs-` prefix could be confusing.

chmeliik: Ideally, we don't want these in build-definitions at all. It should now be possible to put custom, "untrusted" tasks in your pipeline and pass EC checks as long as those tasks do not modify the source code on the way to the build task.

That requires using a Trusted Artifacts-based pipeline. Example pipeline here: https://github.com/konflux-ci/olm-operator-konflux-sample/blob/main/.tekton/single-arch-build-pipeline.yaml

@konflux-ci/mota could probably provide more info.

msugakov: This task does not modify the source code. It gets the source code on the input and outputs an image tag that we put on the resulting images.

We already use oci-ta tasks and not using workspaces to pass the source code around.

We, however, have some other tasks that download blobs that are included in the source code and then included in the resulting containers.

What would be the plan? Should we skip this task but open PRs for the others that download blobs?

chmeliik: > Should we skip this task but open PRs for the others that download blobs?

If it's generically re-usable, that may be reasonable. But the blobs downloaded by such a task would likely bypass source containers and SBOMs, wouldn't they?

The prefetch task should eventually gain the ability to download arbitrary blobs (while also allowing one to reference the source code for those blobs): https://issues.redhat.com/browse/KONFLUX-2390

chmeliik: Until then, you might as well make your build non-hermetic

msugakov: > the blobs downloaded by such a task would likely bypass source containers and SBOMs, wouldn't they?

During ACS CPaaS onboarding, we had a session or two about these blobs. I could find [this doc](https://docs.google.com/document/d/1u9E9r-qeZyUZjfveMd6-yBalQu_5lNkAkPnflbs3JLQ/edit) but I think there should be more records and I can dig it out if needed.
Basically, we concluded that since these blobs carry data and no code, we can not include them in source containers.
In present Konflux time, the arrangements might be different and it seems Curlito could fit us.

> https://issues.redhat.com/browse/KONFLUX-2390

Thanks, I subscribed to it and linked to ACS Enablement ticket (KONFLUX-258).

chmeliik:
Since you already use Trusted Artifacts (https://github.com/konflux-ci/build-definitions/pull/1282#discussion_r1714075606), you should be able to add custom tasks without compromising the trust of your pipeline (an EC verification should pass). Can we close this PR?

tommartensen:
Retracting this PR, because determine-image-tag is not on the critical path and not mentioned in the EC violations.

PR #1288, opened 2024-08-12
feat(RHTAPWATCH-1038): Support custom certificate in ec pipeline

hmariset:
Add ConfigMap parameters to support custom certificate in enterprise-contract pipeline.

Jira-Url: https://issues.redhat.com/browse/RHTAPWATCH-1038

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -51,6 +51,14 @@ spec:
       type: string
       description: Timeout setting for `ec validate`.
       default: "5m0s"
+    - name: CA_TRUST_CONFIGMAP_NAME
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: "trusted-ca"
+    - name: CA_TRUST_CONFIG_MAP_KEY
+      type: string
+      description: The name of the key in the ConfigMap that contains the CA bundle data.
+      default: "ca-bundle.crt"
   results:
     - name: TEST_OUTPUT
       value: "$(tasks.verify.results.TEST_OUTPUT)"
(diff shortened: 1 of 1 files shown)


pipelines/enterprise-contract.yaml
```diff
@@ -75,6 +83,10 @@ spec:
           value: "true"
         - name: TIMEOUT
           value: "$(params.TIMEOUT)"
+        - name: CA_TRUST_CONFIGMAP_NAME
```
lcarva, line 86: The `ec-task-bundle` needs to be updated to include a version of the EC Task that accepts these parameters. The current reference does not:

```bash
🐚 tkn bundle list -o yaml quay.io/enterprise-contract/ec-task-bundle:snapshot@sha256:294b14582fa0e44f42c7e0651915ca67425488527fa7d9ecb49c3974ede028fc | yq .spec.params[].name
DEFINITION
POLICY_SOURCE
NAMESPACE
POLICY_LIB
POLICY_DATA
HOMEDIR
```

yftacherzog: @lcarva If I understand correctly, that command returns the `.spec.params[].name` of the last spec it finds, which is the spec of the `verify-definition` task.
I don't think this is actually a valid yaml so yq might not be the tool to process it.

Looking at the output, without yq, I do see those parameters under the `verify-enterprise-contract` task.

pipelines/enterprise-contract.yaml
```diff
@@ -51,6 +51,14 @@ spec:
       type: string
       description: Timeout setting for `ec validate`.
       default: "5m0s"
+    - name: CA_TRUST_CONFIGMAP_NAME
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: "trusted-ca"
```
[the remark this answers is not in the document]
yftacherzog: Not necessarily, but it's optional on the task, so if it doesn't exist on the namespace, then the task should not hang. The volume mount creates a file that does not exist on the container, so it should not conflict with existing certificates.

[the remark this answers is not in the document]
lcarva: Perfect. Thank you!

pipelines/enterprise-contract.yaml
hmariset: Thanks @lcarva and @yftacherzog!

lcarva: FYI https://github.com/tektoncd/cli/issues/2091

PR #1291, opened 2024-08-13
Merge sbom manipulation steps

mkosiarc:
The steps merge-syft-sboms, merge-cachi2-sbom, create-purl-sbom, create-base-images-sbom were merged into single "prepare-sboms" step. This is done because the tekton results size for a task depends on the number of steps. By reducing the number of the steps we increase the limit.

The scripts from the individual steps were moved to the build-tasks-dockerfiles repo that also contains the Dockerfile for the sbom-utility-scripts image.

[STONEBLD-2608](https://issues.redhat.com//browse/STONEBLD-2608)

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -475,96 +475,24 @@ spec:
           memory: 256Mi
       securityContext:
         runAsUser: 0
-    - name: merge-syft-sboms
-      image: registry.access.redhat.com/ubi9/python-39:1-192.1722518946@sha256:0176b477075984d5a502253f951d2502f0763c551275f9585ac515b9f241d73d
+    - name: prepare-sboms
+      image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:064eeec1cfb75abd21b533bf0d9870a7a3fb138f30192596dce28061798cbf06
       workingDir: /var/workdir
       script: |
-        #!/bin/python3
-        import json
+        echo "Merging contents of sbom-source.json and sbom-image.json into sbom-cyclonedx.json"
+        python3 /scripts/merge_syft_sboms.py
 
-        # load SBOMs
-        with open("./sbom-image.json") as f:
-          image_sbom = json.load(f)
-
-        with open("./sbom-source.json") as f:
-          source_sbom = json.load(f)
-
-        # fetch unique components from available SBOMs
-        def get_identifier(component):
-          return component["name"] + '@' + component.get("version", "")
-
-        image_sbom_components = image_sbom.setdefault("components", [])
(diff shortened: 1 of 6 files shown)


mkosiarc:
Here is the associated PR https://github.com/konflux-ci/build-tasks-dockerfiles/pull/154

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -475,96 +475,24 @@ spec:
...
-    - name: merge-syft-sboms
-      image: registry.access.redhat.com/ubi9/python-39:1-192.1722518946@sha256:0176b477075984d5a502253f951d2502f0763c551275f9585ac515b9f241d73d
+    - name: prepare-sboms
+      image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:53a3041dff341b7fd1765b9cc2c324625d19e804b2eaff10a6e6d9dcdbde3a91
       workingDir: /var/workdir
       script: |
-        #!/bin/python3
-        import json
+        echo "Merging contents of sbom-source.json and sbom-image.json into sbom-cyclonedx.json"
+        python3 /scripts/merge_syft_sboms.py
```
chmeliik, line 483: As the next improvement, we should make these scripts accept params for the sbom paths, similar to the merge_cachi2_sboms.py script.

In the current state, the scripts that don't take params just look like magic :D

mkosiarc: I agree, let's do that as a next improvement (not part of this round of PRs)

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -475,96 +475,24 @@ spec:
           memory: 256Mi
       securityContext:
         runAsUser: 0
-    - name: merge-syft-sboms
-      image: registry.access.redhat.com/ubi9/python-39:1-192.1722518946@sha256:0176b477075984d5a502253f951d2502f0763c551275f9585ac515b9f241d73d
+    - name: prepare-sboms
+      image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:064eeec1cfb75abd21b533bf0d9870a7a3fb138f30192596dce28061798cbf06
```
mmorhun, line 479: I'm not sure we need `-image` suffix in the mage name.

mmorhun (approved):

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
mkosiarc: We don't, but not sure if it is worth it renaming it now. I would have to create a new repo in quay, create additional PR in build-task-dockerfiles and update it in this one as well.

tkdchen (approved):

PR #1295, opened 2024-08-14
STONEBLD-2649: add github action workflow to check task owners

tisutisu:
- This PR adds a github workflow to check task owners file is present and having atleast  1 approver and 1 reviewer 
- Also adds missing owners file

### DIFF
--- a/hack/check-task-owners.sh
+++ b/hack/check-task-owners.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+check_result=$(mktemp)
+
+# Check the OWNERS file is present for each task
+find task/ -mindepth 1 -maxdepth 1 -type d | \
+    while read -r task_dir; do
+        owners_file="$task_dir/OWNERS"
+        if [ ! -e "$owners_file" ]; then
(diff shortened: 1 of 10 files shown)


hack/check-task-owners.sh
```diff
@@ -0,0 +1,31 @@
...
+        fi
+    done
+
+if [ -s "$check_result" ]; then
+    cat "$check_result"
+    echo "Please add OWNERS file with atleast 1 approver and 1 reviewer"
+fi
+
+# Exit with non zero error code if OWNERS file is missing
+if [ "$owners_missing" -eq 1 ]; then
```
mkosiarc, line 28: Do we need this additional check and variable? Can't we exit in the previous "if" block? If the check_result is not empty, we are also setting the variable owners_missing=1 at the same time. So they will both be true/false at the same time, so it might be redundant if I did not miss anything.

tisutisu: Good catch. Agree it can be done in a single if block as well. Fixed it now.

mmorhun (approved):

PR #1299, opened 2024-08-15
Move the `tash` tool here

zregvart:
Moves the current state of the `tash` code from hacks[^1] and adjusts the `hack/generate-ta-tasks.sh` script to use `go run` and run the local version of it.

Resolves: https://issues.redhat.com/browse/EC-690

[^1]: https://github.com/enterprise-contract/hacks/tree/main/tash


hack/generate-ta-tasks.sh
```diff
@@ -1,29 +1,23 @@
...
 set -o errexit
+set -o errtrace
 set -o nounset
 set -o pipefail
 set -o posix
 
 shopt -s globstar nullglob
 
+command -v go &> /dev/null || { echo Please install golang to run this tool; exit 1; }
+[[ "$(go env GOVERSION)" == @(go1|go1.[1-9]+(|.*|rc*|beta*)|go1.1[0-9]+(|.*|rc*|beta*)|go1.20*) ]] && { echo Please install golang 1.21.0 or newer; exit 1; }
```
chmeliik, line 12: On my Fedora, by default I get
```
go: go.mod requires go >= 1.22.3 (running go 1.21.12; GOTOOLCHAIN=local)
```

Now, I happen to know that I need to run the script like this:

```bash
GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh
```

But perhaps the script could try to detect this problem and advise what to do?

hack/generate-ta-tasks.sh
```diff
@@ -1,29 +1,23 @@
...
-  echo INFO: downloading from ${tash_url} to "${tash_dir}"
-  curl --no-progress-meter --location --output "${tash_dir}/tash" "${tash_url}"
-  echo INFO: SHA256: "$(sha256sum "${tash_dir}/tash")"
-  chmod +x "${tash_dir}/tash"
-  tash() {
-    "${tash_dir}/tash" "$@"
-  }
-fi
+tash() {
+  go run -C "${ROOT_DIR}/ta-generator/" github.com/konflux-ci/build-definitions/ta-generator "$@"
```
chmeliik, line 19: This seems significantly slower than pre-compiling

```bash
time GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh
GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh  7.21s user 1.48s system 168% cpu 5.168 total
```

vs. compiling once per script run

```bash
time GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh
GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh  1.61s user 0.37s system 166% cpu 1.192 total
```

chmeliik (commented):
Nice :tada: 

Some usability suggestions for the script, otherwise LGTM

hack/generate-ta-tasks.sh
```diff
@@ -44,7 +38,8 @@ fi
 cd "${TASK_DIR}"
 for recipe_path in **/recipe.yaml; do
     task_path="${recipe_path%/recipe.yaml}/$(basename "${recipe_path%/*/*}").yaml"
-    cat <<< "$(tash "${recipe_path}")" > "${task_path}"
+    sponge=$(tash "${TASK_DIR}/${recipe_path}")
+    echo "${sponge}" > "${task_path}"
```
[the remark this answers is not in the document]
zregvart: I could not get the error handling to work, one of many joys of writing bash code. I was hoping that turning `errtrace` on would bubble up to parent shell but it does not in this case. This explains a lot of it https://mywiki.wooledge.org/BashFAQ/105, but basically:

```
#!/bin/bash

set -o errexit
set -o errtrace
set -o nounset
set -o pipefail
set -o posix

fail() {
  echo hi
  exit 1
}

cat <<< "$(fail)" | cat

echo no!
```

When run produces:
```
/x.sh 
hi
no!
```

hack/generate-ta-tasks.sh
zregvart: I think I might have `GOTOOLCHAIN` turned on by default, I'll update the script to be a bit more helpful. I could also set `GOTOOLCHAIN=auto` in the script?

hack/generate-ta-tasks.sh
zregvart: Yeah, do we care? I guess we might?

chmeliik: Not in CI, but locally it can get annoying if you're trying to fix a recipe and regenerating files often

zregvart: Huh, I see the opposite (ran a couple of times, but getting consistent results):

`go run` version:
```
$ time hack/generate-ta-tasks.sh
hack/generate-ta-tasks.sh  12.54s user 2.80s system 170% cpu 8.990 total

$ time hack/generate-ta-tasks.sh
hack/generate-ta-tasks.sh  12.60s user 2.76s system 169% cpu 9.040 total
```

`go build` once and execute version:
```
$ time hack/generate-ta-tasks.sh
hack/generate-ta-tasks.sh  19.23s user 3.26s system 179% cpu 12.553 total

$ time hack/generate-ta-tasks.sh
hack/generate-ta-tasks.sh  19.24s user 3.13s system 181% cpu 12.347 total
```

I'll push the build & execute version, perhaps you can spot something

hack/generate-ta-tasks.sh
chmeliik: Setting the vars in the script sounds okay. We'd need both GOTOOLCHAIN and GOSUMDB, otherwise I get

```
go: golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64: verifying module: checksum database disabled by GOSUMDB=off
```

I think the Fedora default is GOTOOLCHAIN=local, GOSUMBD=off (at least I don't override them in my shell rcs)

hack/generate-ta-tasks.sh
chmeliik: Hmm, it seems to rebuild tash every time, but I can't tell why :thinking: 

```diff
diff --git a/hack/generate-ta-tasks.sh b/hack/generate-ta-tasks.sh
index 8268155..2762c02 100755
--- a/hack/generate-ta-tasks.sh
+++ b/hack/generate-ta-tasks.sh
@@ -16,6 +16,7 @@ ROOT_DIR="$(git rev-parse --show-toplevel)"
 TASK_DIR="$(realpath "${ROOT_DIR}/task")"
 
 tash() {
+  echo building tash >&2
   tashbin="$(mktemp --dry-run)"
   GOTOOLCHAIN=auto go build -C "${ROOT_DIR}/ta-generator/" -o "${tashbin}"
   trap 'rm "${tashbin}"' EXIT
```

```shell
$ GOTOOLCHAIN=auto GOSUMDB=sum.golang.org hack/generate-ta-tasks.sh
building tash
building tash
building tash
building tash
building tash
building tash
building tash
building tash
```

zregvart: It's my shell scripting, I'll figure this out

zregvart: So, fun, `tash` is executed in a subshell on the `sponge=$(tash...` line, so the redefinition of `tash` function is happening in the subshell and it is not propagated to the parent shell, so next time it is invoked the original non-memoized version is called. I've simplified it now, no need to be fancy here...

chmeliik: :exploding_head: yep, bash is fun

chmeliik (approved):

Issue #1200, opened 2024-07-25
RUN <<EOF without an interpreter and cachi2


owtaylor:
```dockerfile
RUN <<EOF
mkdir -p /etc/foo
cp -aR /src/foo /etc/foo
EOF
```

gets turned into:

```dockerfile
RUN . /cachi2/cachi2.env && <<EOF
mkdir -p /etc/foo
cp -aR /src/foo /etc/foo
EOF
```

which is weirdly not a syntax error, but a no-op. The right transformation would be something like:

```dockerfile
RUN . /cachi2/cachi2.env && /bin/sh <<EOF
mkdir -p /etc/foo
cp -aR /src/foo /etc/foo
EOF
```

Once you consider `RUN --mount=type=bin,src=/a,dest=/b <<EOF` wedging this into the current code gets a little tricky.

chmeliik:
@owtaylor I would assume this is not critical since there is a fairly easy workaround (use `/bin/sh <<EOF` rather than just `<<EOF`)?

But we should fix this at some point. A good Dockerfile parser+editor will be needed.

chmeliik:
Just came across https://www.mankier.com/1/cstrans-df-run (in https://github.com/kdudka/konflux-test-multicluster-global-hub/pull/9/commits/88998934c0b4ffeb9a333316ac89037dbfe60963)

It doesn't handle heredocs correctly either, but perhaps a contribution would be reasonable (it's a C++ codebase, haven't looked into it much)

cgwalters:
> which is weirdly not a syntax error, but a no-op. 

Yeah, man I got burned by this bad. It took me quite a while to debug. I guess in bash it's something like passing the heredoc to `&&` which just discards it? Mind blowing...really seems like it should be a syntax error since there's no reason to ever do that.

> Once you consider RUN --mount=type=bin,src=/a,dest=/b <<EOF wedging this into the current code gets a little tricky.

Yes, which we also do.

chmeliik:
This is fixed. If your prefetch-dependencies task is at least version 0.3.1, and your buildah task at least 0.10.3, then the build no longer modifies RUN instructions but instead sets env vars with `--secret` + `--mount`. This supports bare heredocs and exec-form instructions as well.

This change was missing in the changelog, currently being added in https://github.com/konflux-ci/build-definitions/pull/3649

PR #1304, opened 2024-08-15
Replace grep base images parsing with dockerfile-json

mkosiarc:
This is more reliable and allow us to fix bugs where base images were loaded incorrectly.

For example, previously this part in Dockerfile:

LABEL description="this is a build \
                   from single-arch"

Would return "single-arch" as a base image.

Using dockerfile-json also solves the problem of omitting "scratch" from the results.

Another advantage is that when we have something such as:

FROM registry.access.redhat.com/ubi9/ubi:latest as builder ...
FROM builder AS build1

then only the original image
"registry.access.redhat.com/ubi9/ubi:latest" will be reported.

[KFLUXBUGS-1269](https://issues.redhat.com//browse/KFLUXBUGS-1269)

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -288,14 +288,12 @@ spec:
 
         BUILDAH_ARGS=()
 
-        BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}' | (grep -v ^oci-archive: || true))
+        BASE_IMAGES=$(dockerfile-json "$dockerfile_path" | jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName')
         if [ "${HERMETIC}" == "true" ]; then
           BUILDAH_ARGS+=("--pull=never")
           UNSHARE_ARGS="--net"
           for image in $BASE_IMAGES; do
-            if [ "${image}" != "scratch" ]; then
-              unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
-            fi
+            unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
           done
           echo "Build will be executed with network isolation"
         fi
@@ -415,9 +413,7 @@ spec:
 
         touch /shared/base_images_digests
         for image in $BASE_IMAGES; do
-          if [ "${image}" != "scratch" ]; then
-            buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' --filter reference="$image" >>/shared/base_images_digests
-          fi
+          buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' --filter reference="$image" >>/shared/base_images_digests
         done
 
         # Needed to generate base images SBOM
--- a/task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml
+++ b/task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml
@@ -323,14 +323,12 @@ spec:
 
       BUILDAH_ARGS=()
 
-      BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}' | (grep -v ^oci-archive: || true))
+      BASE_IMAGES=$(dockerfile-json "$dockerfile_path" | jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName')
       if [ "${HERMETIC}" == "true" ]; then
         BUILDAH_ARGS+=("--pull=never")
         UNSHARE_ARGS="--net"
         for image in $BASE_IMAGES; do
-          if [ "${image}" != "scratch" ]; then
-            unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
-          fi
+          unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull $image
         done
         echo "Build will be executed with network isolation"
       fi
@@ -450,9 +448,7 @@ spec:
 
       touch /shared/base_images_digests
       for image in $BASE_IMAGES; do
-        if [ "${image}" != "scratch" ]; then
-          buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' --filter reference="$image" >>/shared/base_images_digests
-        fi
+        buildah images --format '{{ .Name }}:{{ .Tag }}@{{ .Digest }}' --filter reference="$image" >>/shared/base_images_digests
       done
 
       # Needed to generate base images SBOM
--- a/task/buildah-remote/0.2/buildah-remote.yaml
+++ b/task/buildah-remote/0.2/buildah-remote.yaml
@@ -305,14 +305,12 @@ spec:
 
       BUILDAH_ARGS=()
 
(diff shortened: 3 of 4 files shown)


mkosiarc:
This has to be merged only after https://github.com/konflux-ci/buildah-container/pull/62 and the image reference has to be replaced.

chmeliik:
When playing with dockerfile-json to see if it could help with https://github.com/konflux-ci/build-definitions/issues/1200, I found out it will probably make things even worse :disappointed: 

```Dockerfile
FROM registry.fedoraproject.org/fedora-minimal:40

RUN /bin/sh <<EOF
mkdir -p /etc/foo
cp -aR /src/foo /etc/foo
EOF
```

```bash
$ dockerfile-json /tmp/Dockerfile
[dockerfile-json 1.0.8] error: parse "/tmp/Dockerfile": dockerfile/instructions.Parse dockerfile parse error on line 4: unknown instruction: mkdir
```

It looks like the latest release of dockerfile-json (from 3 years ago: https://github.com/keilerkonzept/dockerfile-json/releases) uses a version of buildkit that didn't yet support heredocs (and who knows what else).

Then I tried to build dockerfile-json from source to see if it's fixed in main, but it doesn't even compile anymore:

```
# github.com/keilerkonzept/dockerfile-json/pkg/dockerfile
pkg/dockerfile/parse.go:26:46: not enough arguments in call to instructions.Parse
	have (*parser.Node)
	want (*parser.Node, *linter.Linter)
```

All my trust in that project just went out the window

chmeliik:
On the other hand, the fix was pretty simple if we want to fork the project and add some CI so that the auto-merged Renovate updates don't just randomly break main

```diff
diff --git a/pkg/dockerfile/parse.go b/pkg/dockerfile/parse.go
index 44daf75..02dac0a 100644
--- a/pkg/dockerfile/parse.go
+++ b/pkg/dockerfile/parse.go
@@ -23,7 +23,7 @@ func ParseReader(r io.Reader) (*Dockerfile, error) {
 	if err != nil {
 		return nil, fmt.Errorf("dockerfile/parser.Parse %v", err)
 	}
-	stages, metaArgs, err := instructions.Parse(result.AST)
+	stages, metaArgs, err := instructions.Parse(result.AST, nil)
 	if err != nil {
 		return nil, fmt.Errorf("dockerfile/instructions.Parse %v", err)
 	}
```

chmeliik:
> On the other hand, the fix was pretty simple if we want to fork the project and add some CI (or contribute this upstream, but it looks rather abandoned) so that the auto-merged Renovate updates don't just randomly break main
> 
> ```diff
> diff --git a/pkg/dockerfile/parse.go b/pkg/dockerfile/parse.go
> index 44daf75..02dac0a 100644
> --- a/pkg/dockerfile/parse.go
> +++ b/pkg/dockerfile/parse.go
> @@ -23,7 +23,7 @@ func ParseReader(r io.Reader) (*Dockerfile, error) {
>  	if err != nil {
>  		return nil, fmt.Errorf("dockerfile/parser.Parse %v", err)
>  	}
> -	stages, metaArgs, err := instructions.Parse(result.AST)
> +	stages, metaArgs, err := instructions.Parse(result.AST, nil)
>  	if err != nil {
>  		return nil, fmt.Errorf("dockerfile/instructions.Parse %v", err)
>  	}
> ```

mkosiarc:
@chmeliik Thanks for your investigation here. I tried forking the repo, so far just to my account and adding the fix you proposed. Now it can be build from source and the main branch - see https://github.com/konflux-ci/buildah-container/pull/62

Now, the heredocs can be parsed as well, because it uses latest buildkit version. But this concerns me, I wanted to avoid having more work, but if we decide to fork the repo, setup the CI and maintain it, it will require further work down the line. But seems like we do need such a funcionality. The dockerfile-json uses the parsing from the buildkit directly, if we wanted we could build our own tool around it as well.

chmeliik (commented):
LGTM after the temp changes are undone

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -288,14 +288,12 @@ spec:
 
         BUILDAH_ARGS=()
 
-        BASE_IMAGES=$(grep -i '^\s*FROM' "$dockerfile_path" | sed 's/--platform=\S*//' | awk '{print $2}' | (grep -v ^oci-archive: || true))
+        BASE_IMAGES=$(dockerfile-json "$dockerfile_path" | jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName')
```
arewm, line 291: It is now possible to have an empty value here, right whereas before there was always going to be at least one value?

Will this break any functionality elsewhere?

arewm:
Is there a reason that we are only making these changes on the v0.2 tasks?

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
mkosiarc: Yes, previously there was always at least one value - e.g when it was FROM scratch, but now we are omitting that. That's why I could remove all those if conditions that are checking for "scratch". 

I tested it with a couple of builds and the e2e tests are passing. And from the logic of the buildah it does not seem anything should break, since when we are passing that further (for example to the sbom), the "scratch" was omitted anyway.

mkosiarc:
> Is there a reason that we are only making these changes on the v0.2 tasks?

I remember briefly discussing this with Adam, that it does not seem to worth to update the 0.1 tasks, as we are not expecting many users to still use them and this update does not seem that critical to include it in the old tasks. But if you have better knowledge of this and feel like I should update the 0.1 tasks as well then let me know.

mmorhun (approved):

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
tkdchen: An empty Dockerfile should seldom happen. However, if it happens accidentally, dockerfile-json will report an error like `[dockerfile-json 1.0.8] error: parse "./Dockerfile": dockerfile/parser.Parse file with no instructions`. And `set -o pipefail` is not set in the `build` step.

arewm:
> I remember briefly discussing this with Adam, that it does not seem to worth to update the 0.1 tasks, as we are not expecting many users to still use them and this update does not seem that critical to include it in the old tasks. But if you have better knowledge of this and feel like I should update the 0.1 tasks as well then let me know.

The question comes from a maintenance burden. If the change is minimal to make, then it seems like it would be better to make it so that the tasks are the same. 

I am not saying that we should do it, I am just raising the observation that we are not.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
arewm: It is good that the e2e tests are passing but I don't have much faith in the robustness of the e2e tests lately (i.e. do they test an image which is just `FROM scratch`)? Would you be able to test this change in a pipeline that has a simple `FROM scratch` Containerfile?

If you need a file, you can customize this pipeline: https://github.com/konflux-ci/olm-operator-konflux-sample/blob/main/.tekton/gatekeeper-operator-bundle-pull-request.yaml#L35 and just modify the Dockerfile to remove the first stage (or better yet, test it with the first stage and without).

mkosiarc: We do have a FROM scratch test, see https://github.com/konflux-ci/e2e-tests/commit/9c5c74c7b8c1f23b70b2f48aabe86a78ab365425 and https://github.com/konflux-ci/build-definitions/commit/5573eb35e44f486f539152d184f717bedb3b5542. But I tested it anyway, once again, now also with multiple stages as has the dockerfile you linked.

mkosiarc:
> > I remember briefly discussing this with Adam, that it does not seem to worth to update the 0.1 tasks, as we are not expecting many users to still use them and this update does not seem that critical to include it in the old tasks. But if you have better knowledge of this and feel like I should update the 0.1 tasks as well then let me know.
> 
> The question comes from a maintenance burden. If the change is minimal to make, then it seems like it would be better to make it so that the tasks are the same.
> 
> I am not saying that we should do it, I am just raising the observation that we are not.

It might be minimal to make, but it would again take some time to test it. And as far as I understand, the e2e tests here are running only the newer version of tasks.

PR #1325, opened 2024-08-21
Remove java references from pipelines

arewm:
This does not remove any of the java task dependencies but it will reduce a potential migration for new users that onboard. By removing the result, any users that onboard will have an easier migration.


chmeliik (approved):
I still see the result referenced in the `java-builder` pipeline, but I have a feeling we can delete that whole thing in a separate PR

arewm:
Yeah, the java-builder added it on its own, so I just left that as-is thinking that it might eventually be removed wholesale as you suggested.

PR #1331, opened 2024-08-22
feat(KONFLUX-4136): add new reduce-snapshot task

scoheb:
- This task, **reduce-snapshot-to-single-component** does nothing by default.
- If turned on, it will create a new single component Snapshot
  based on which component caused the Snapshot to be built.

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -59,16 +59,43 @@ spec:
       type: string
       description: The name of the key in the ConfigMap that contains the CA bundle data.
       default: "ca-bundle.crt"
+    - name: SINGLE_COMPONENT
+      type: string
+      description: Reduce the Snapshot to only the component whose build caused the Snapshot to be created
+      default: "false"
+    - name: PIPELINERUN_ID
+      type: string
+      description: PipelineRun ID
+      default: "$(context.pipelineRun.name)"
   results:
     - name: TEST_OUTPUT
       value: "$(tasks.verify.results.TEST_OUTPUT)"
   tasks:
+    - name: reduce-snapshot-to-single-component
+      params:
+        - name: SNAPSHOT
+          value: "$(params.SNAPSHOT)"
+        - name: SINGLE_COMPONENT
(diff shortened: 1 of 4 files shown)


pipelines/enterprise-contract.yaml
```diff
@@ -59,16 +59,41 @@ spec:
...
+          value: "$(params.PIPELINERUN_ID)"
+      taskRef:
+        resolver: "git"
+        params:
+          - name: url
+            value: https://github.com/konflux-ci/build-definitions
+          - name: revision
+            value: main
+          - name: pathInRepo
+            value: task/reduce-snapshot/0.1/reduce-snapshot.yaml
```
zregvart, line 90: I think we should do what is done in other pipelines here, the `hack/build-and-push.sh` should replace this with the newly built Task bundle and configure the bundle resolver
```suggestion
      taskRef:
        name: reduce-snapshot
        version: "0.1"
```

pipelines/enterprise-contract.yaml
```diff
@@ -59,16 +59,41 @@ spec:
       type: string
       description: The name of the key in the ConfigMap that contains the CA bundle data.
       default: "ca-bundle.crt"
+    - name: REDUCE_SNAPSHOT
```
zregvart, line 62: Perhaps (I'm not 100%), considering that folk might not know or care about what Application snapshot is:
```suggestion
    - name: SINGLE_COMPONENT
```

zregvart (commented):
LGTM

pipelines/enterprise-contract.yaml
scoheb: I renamed and refactored some of the parameters, task names and paths

pipelines/enterprise-contract.yaml
scoheb: changed. thanks!

scoheb: actually, this does not work since the EC pipeline is executed via a Git resolver from the ITS ... if I use a git resolver, it works.

scoheb:
superceded by #1338

PR #1336, opened 2024-08-26
Use Konflux-built Syft image

chmeliik:
[STONEBLD-2715](https://issues.redhat.com//browse/STONEBLD-2715)

Switch from quay.io/redhat-appstudio/syft to
registry.access.redhat.com/rh-syft-tech-preview/syft-rhel9.

The latter is built and released via Konflux (from
https://github.com/redhat-appstudio/rh-syft/) and is a bit more up to
date.

### DIFF
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -426,7 +426,7 @@ spec:
           add:
             - SETFCAP
     - name: sbom-syft-generate
-      image: quay.io/redhat-appstudio/syft:v0.105.1@sha256:1910b829997650c696881e5fc2fc654ddf3184c27edb1b2024e9cb2ba51ac431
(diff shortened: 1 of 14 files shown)


task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
```diff
@@ -426,7 +426,7 @@ spec:
           add:
             - SETFCAP
     - name: sbom-syft-generate
-      image: quay.io/redhat-appstudio/syft:v0.105.1@sha256:1910b829997650c696881e5fc2fc654ddf3184c27edb1b2024e9cb2ba51ac431
+      image: registry.access.redhat.com/rh-syft-tech-preview/syft-rhel9:1.4.1@sha256:34d7065427085a31dc4949bd283c001b91794d427e1e4cdf1b21ea4faf9fee3f
```
MartinBasti, line 429: if syft is based on ubi9, why it's called syft-rhel9 and not syft-ubi9 ?

ubi and rhel is not the same

mkosiarc (approved):

mmorhun (approved):

tkdchen (approved):

PR #1338, opened 2024-08-26
feat(KONFLUX-4136): new reduce-snapshot task

scoheb:
- This task, reduce-snapshot-to-single-component does nothing by default.
- If turned on, it will create a new single component Snapshot based on which component caused the Snapshot to be built.

### DIFF
--- a/task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
+++ b/task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
@@ -1,0 +1,59 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: reduce-snapshot-to-single-component
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.0.1
+    tekton.dev/tags: konflux
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: docker
+spec:
+  description: Reduce a snapshot to contain the single component that the snapshot was created for.
+  params:
+    - name: SNAPSHOT
+      description: Snapshot to possibly reduce
+      type: string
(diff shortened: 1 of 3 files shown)


lcarva (approved):

task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
```diff
@@ -0,0 +1,59 @@
...
+          value: $(params.SINGLE_COMPONENT)
+        - name: PIPELINERUN_ID
+          value: $(params.PIPELINERUN_ID)
+      script: |
+        #!/usr/bin/env bash
+        set -eu
+
+        echo "Single Component mode? ${SINGLE_COMPONENT}"
+        if [ "${SINGLE_COMPONENT}" == "true" ]; then
+          SNAPSHOT_CREATION_TYPE=$(oc get "pr/$PIPELINERUN_ID" -ojson | jq -rec '.metadata.labels."test.appstudio.openshift.io/type" // ""')
```
lcarva, line 44: This seems a bit fragile. PipelineRuns are garbage collected quite fast. This will likely fail in most cases.

lcarva: Unless it's querying the Pipeline which is running this Task.

lcarva:
FYI, I did check that https://quay.io/repository/konflux-ci/tekton-catalog/task-reduce-snapshot-to-single-component has already been created :+1:

task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
scoheb: we are getting the labels for the currently executing pipeline...therefore it should still be around :)

PR #1339, opened 2024-08-26
feat(KONFLUX-4136): enable reduce-snapshot task

scoheb:
- Add the reduce-snapshot-to-single-component task to the enterprise contract pipeline.
- To enable, users must add the parameter `SINGLE_COMPONENT` with a value of `true`

**NOTE:** Requires #1338 to be merged first before tests can pass and be merged.


pipelines/enterprise-contract.yaml
```diff
@@ -59,16 +59,46 @@ spec:
...
+      params:
+        - name: SNAPSHOT
+          value: "$(params.SNAPSHOT)"
+        - name: SINGLE_COMPONENT
+          value: "$(params.SINGLE_COMPONENT)"
+        - name: PIPELINERUN_ID
+          value: "$(params.PIPELINERUN_ID)"
+      taskRef:
+        resolver: "git"
+        # Since this pipeline is initiated from a git resolved
```
lcarva, line 84: We'll get a bundle once https://github.com/konflux-ci/build-definitions/pull/1338 is merged. Let's use the same approach to avoid having to use a floating reference, e.g. `main`.

lcarva (approved):

pipelines/enterprise-contract.yaml
scoheb: switched to bundle resolver.

PR #1341, opened 2024-08-27
fix(KFLUXBUGS-1554): check if there are 0 FBC packages

dirgim:
* It is now allowed to have more than one olm.package within the FBC fragment
* The only remaining requirement is to have at least one olm.package entry


hongweiliu17 (approved):
LGTM

PR #1344, opened 2024-08-27
remove redhat-appstudio/cosign image

tnevrlka:
Jira: [STONEBLD-2717](https://issues.redhat.com/browse/STONEBLD-2717)

Replace current uses of `redhat-appstudio/cosign` image with `konflux-ci/appstudio-utils` due to deprecation
of `redhat-appstudio` quay org and migration to `konflux-ci` org

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -25,7 +25,6 @@
         "quay.io/redhat-appstudio/buildah",
         "quay.io/redhat-appstudio/hacbs-jvm-build-request-processor",
         "quay.io/redhat-appstudio/build-definitions-source-image-build-utils",
-        "quay.io/redhat-appstudio/cosign",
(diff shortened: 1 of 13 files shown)


chmeliik (approved):

arewm:
In reverting this PR in #1352, I was trying to figure out why the e2e tests passed even if the change broke builds. I think that is because the tests only exercised the latest version. When you resubmit this change, you need to update the `upload-sbom` task steps to be similar to that change that I did previously: 
https://github.com/konflux-ci/build-definitions/commit/420bc92bb798935d84912c88eca05af798baa8ca#diff-7bee4704ad19a975db0ec0efce3b56efdea36b0e9a0b5655600486a87c2f3ff9L557-R568

PR #1348, opened 2024-08-28
chore: replace multi-platform-runner image with buildah-task image

tisutisu:
Part of the story: https://issues.redhat.com/browse/STONEBLD-2716
migrating away from `redhat-appstudio` quay org to `konflux-ci` org and use a similar existing image

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -113,7 +113,7 @@ spec:
         echo "declare TAGGED_AS=${TAGGED_AS}" >> /var/workdir/vars
 
     - name: build
-      image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f@sha256:246adeaaba600e207131d63a7f706cffdcdc37d8f600c56187123ec62823ff44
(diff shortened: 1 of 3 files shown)


task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -113,7 +113,7 @@ spec:
         echo "declare TAGGED_AS=${TAGGED_AS}" >> /var/workdir/vars
 
     - name: build
-      image: quay.io/redhat-appstudio/multi-platform-runner:01c7670e81d5120347cf0ad13372742489985e5f@sha256:246adeaaba600e207131d63a7f706cffdcdc37d8f600c56187123ec62823ff44
+      image: quay.io/konflux-ci/buildah-task:latest@sha256:1301e1a87a44898ab73e5dff8f6ac7499be4cb64eb7300e25d7a20ae266c87d3
```
scoheb, line 116: Why latest?

chmeliik:
Note: compared to the previous image, the new one doesn't contain `podman`. But if I'm reading the tasks right, they never use `podman` directly, do they? Only via SSH. So the container image having podman doesn't matter as long as the remote VM does.

task/build-vm-image/0.1/build-vm-image.yaml
arewm: That is the floating tag that we push for buildah-task.

arewm (approved):

PR #1349, opened 2024-08-28
refactor reduce-snapshot to use workspace

scoheb:
- update the task to pass the reduced snapshot via a workspace.
- this is possible since the EC task supports and optional workspace.
- a follow-on PR will be created to enable this refactored task in the EC pipeline.


task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
```diff
@@ -37,7 +40,10 @@ spec:
           value: $(params.PIPELINERUN_ID)
       script: |
         #!/usr/bin/env bash
-        set -eu
+        set -eux
```
lcarva, line 43: Intentional `-x`?

lcarva (approved):

task/reduce-snapshot-to-single-component/0.1/reduce-snapshot-to-single-component.yaml
scoheb: yeah. helpful for debugging later on.

PR #1352, opened 2024-08-29
Revert "remove redhat-appstudio/cosign image"

arewm:
This reverts commit b26d80895ab291d635381d583379f66889df49aa. as introduced in #1344.

This change was causing issues with the sbom upload build steps. The task was using cosign image which allowed the arguments to be passed directly in the task args. The new image does not use cosign as the entrypoint so the image update breaks the builds.


arewm:
I am going to go ahead an merge this revert as it has already started blocking users.

cgwalters:
Thanks for reverting this. FTR, some conversation now continues on https://github.com/containers/common/pull/1630#issuecomment-2317786520

PR #1355, opened 2024-08-29
Enable Task results check via EC

lcarva:
Since certain EC policy rules rely on the existence of certain Task results, e.g. CLAIR_SCAN_RESULT, let's verify that these are defined during CI.

Ref: https://issues.redhat.com/browse/EC-794

### DIFF
--- a/policies/all-tasks.yaml
+++ b/policies/all-tasks.yaml
@@ -12,9 +12,14 @@ sources:
         - basic-auth
         - ssh-directory
         - netrc
+      required_task_results:
+        # Certain EC rules rely on the presence of this result when validating an image.
+        - task: clair-scan
+          result: CLAIR_SCAN_RESULT
     config:
       include:
         - kind
+        - results
         - step_image_registries
         - trusted_artifacts
       exclude:


policies/all-tasks.yaml
```diff
@@ -12,9 +12,14 @@ sources:
         - basic-auth
         - ssh-directory
         - netrc
+      required_task_results:
+        # Certain EC rules rely on the presence of this result when validating an image.
+        - task: clair-scan
+          result: CLAIR_SCAN_RESULT
```
simonbaird, line 18: In https://issues.redhat.com/browse/KONFLUX-1485 I get the impression that this result name is deprecated. Should we need to switch to the preferred name?

(This might impact the related ec-policies patch also.)

lcarva: It's still the result that is currently used. There hasn't been a lot of progress in changing the name/format even though there's an [ADR](https://konflux-ci.dev/architecture/ADR/0030-tekton-results-naming-convention.html) for it. The new name will be `SCAN_OUTPUT`. 

I actually think requiring `CLAIR_SCAN_OUTPUT` for now will help us in the future when the new result name is used. CI will fail causing the author/reviewer to think about the implications, e.g. is EC ready for this? (it should be) Part of that change could be to also update the ruledata in the policy config to required the new result name instead.

zregvart (approved):

PR #1359, opened 2024-08-30
sast-snyk-check: increased version to 0.3

jperezdealgaba:
Resolves: https://issues.redhat.com/browse/OSH-737

In this version, the severity-threshold argument is introduced and enabled by default to high and the results are parsed with csgrep to be uploaded with the fingerprint. This MR needs to be merged after this one: https://github.com/konflux-ci/konflux-test/pull/292

All changes have been discussed in the provided Jira tracker.

@konflux-team , we created this as a draft PR in order to gather feedback from you. Would this be acceptable? Is something else needed? ...

### DIFF
--- a/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
+++ b/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
@@ -1,0 +1,262 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: sast-snyk-check-oci-ta
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: konflux
+  labels:
+    app.kubernetes.io/version: "0.3"
+spec:
+  description: |-
+    Scans source code for security vulnerabilities, including common issues such as SQL injection, cross-site scripting (XSS), and code injection attacks using Snyk Code, a Static Application Security Testing (SAST) tool.
+
+    Follow the steps given [here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/) to obtain a snyk-token and to enable the snyk task in a Pipeline.
+
+    The snyk binary used in this Task comes from a container image defined in https://github.com/konflux-ci/konflux-test
+
+    See https://snyk.io/product/snyk-code/ and https://snyk.io/ for more information about the snyk tool.
+  params:
+    - name: ARGS
+      description: Append arguments.
+      type: string
+      default: ""
+    - name: CACHI2_ARTIFACT
+      description: The Trusted Artifact URI pointing to the artifact with
+        the prefetched dependencies.
+      type: string
+      default: ""
+    - name: IMP_FINDINGS_ONLY
+      description: Report only important findings. Default is true. To report
+        all findings, specify "false"
+      type: string
+      default: "true"
+    - name: KFP_GIT_URL
+      description: URL from repository to download known false positives files
+      type: string
+      default: ""
+    - name: PROJECT_NAME
+      description: Name of the scanned project, used to find path exclusions.
+        By default, the Konflux component name will be used.
+      type: string
+      default: ""
+    - name: RECORD_EXCLUDED
+      description: Write excluded records in file. Useful for auditing (defaults
+        to false).
+      type: string
+      default: "false"
+    - name: SNYK_SECRET
+      description: Name of secret which contains Snyk token.
+      default: snyk-secret
+    - name: SOURCE_ARTIFACT
+      description: The Trusted Artifact URI pointing to the artifact with
+        the application source code.
+      type: string
+    - name: caTrustConfigMapKey
+      description: The name of the key in the ConfigMap that contains the
+        CA bundle data.
+      type: string
+      default: ca-bundle.crt
+    - name: caTrustConfigMapName
+      description: The name of the ConfigMap to read CA bundle data from.
+      type: string
+      default: trusted-ca
+    - name: image-url
+      description: Image URL.
+      type: string
+      default: ""
+  results:
+    - name: TEST_OUTPUT
+      description: Tekton task test output.
+  volumes:
+    - name: snyk-secret
+      secret:
+        optional: true
+        secretName: $(params.SNYK_SECRET)
+    - name: trusted-ca
+      configMap:
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        name: $(params.caTrustConfigMapName)
+        optional: true
+    - name: workdir
+      emptyDir: {}
+  stepTemplate:
+    volumeMounts:
+      - mountPath: /var/workdir
+        name: workdir
+  steps:
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:a83c92fc0a1e2c785937c6612dc8c8237818535543f00ecaf6b3b77a35f76259
+      args:
+        - use
+        - $(params.SOURCE_ARTIFACT)=/var/workdir/source
+        - $(params.CACHI2_ARTIFACT)=/var/workdir/cachi2
+    - name: sast-snyk-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.7@sha256:cf6808a3bd605630a5d9f20595ff7c43f8645c00381219d32f5a11e88fe37072
+      workingDir: /var/workdir/source
+      volumeMounts:
+        - mountPath: /etc/secrets
+          name: snyk-secret
+          readOnly: true
+        - mountPath: /mnt/trusted-ca
+          name: trusted-ca
+          readOnly: true
+      env:
+        - name: SNYK_SECRET
+          value: $(params.SNYK_SECRET)
+        - name: ARGS
+          value: $(params.ARGS)
+        - name: IMP_FINDINGS_ONLY
+          value: $(params.IMP_FINDINGS_ONLY)
+        - name: KFP_GIT_URL
+          value: $(params.KFP_GIT_URL)
+        - name: PROJECT_NAME
+          value: $(params.PROJECT_NAME)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+        - name: COMPONENT_LABEL
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['appstudio.openshift.io/component']
+      script: |
+        #!/usr/bin/env bash
+
+        set -euo pipefail
+        # shellcheck source=/dev/null
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        if [[ -z "${PROJECT_NAME}" ]]; then
+            PROJECT_NAME=${COMPONENT_LABEL}
+        fi
+
+        echo "The PROJECT_NAME used is: ${PROJECT_NAME}"
+
+        # Installation of Red Hat certificates for cloning Red Hat internal repositories
+        ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+        if [ -f "$ca_bundle" ]; then
+          echo "INFO: Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+          update-ca-trust
+        fi
+
+        SNYK_TOKEN_PATH="/etc/secrets/snyk_token"
+        if [ -f "${SNYK_TOKEN_PATH}" ] && [ -s "${SNYK_TOKEN_PATH}" ]; then
+          # SNYK token is provided
+          SNYK_TOKEN="$(cat ${SNYK_TOKEN_PATH})"
+          export SNYK_TOKEN
+        else
+          # According to shellcheck documentation, the following error can be ignored as it is ignored through indirection: https://www.shellcheck.net/wiki/SC2034
+          # shellcheck disable=SC2034
+          to_enable_snyk='[here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/)'
+          note="Task $(context.task.name) skipped: If you wish to use the Snyk code SAST task, please create a secret name snyk-secret with the key 'snyk_token' containing the Snyk token by following the steps given ${to_enable_snyk}"
+          TEST_OUTPUT=$(make_result_json -r SKIPPED -t "$note")
+          echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+          exit 0
+        fi
+
+        SNYK_EXIT_CODE=0
+        SOURCE_CODE_DIR=/var/workdir
+        SEVERITY_THRESHOLD="high"
+        if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
+          SEVERITY_THRESHOLD="low"
+        fi
+        set +e
+        # We do want to expand ARGS (it can be multiple CLI flags, not just one)
+        # shellcheck disable=SC2086
+        snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output="${SOURCE_CODE_DIR}"/sast_snyk_check_out.json 1>&2 >>stdout.txt
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        (cd  "${SOURCE_CODE_DIR}" && csgrep --mode=json --embed-context=3 "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json) |
+            csgrep --mode=json --strip-path-prefix="source/" \
+            >sast_snyk_check_out_all_findings.json
+
+        echo "Results:"
+        (set -x && csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
+
+        # We check if the KFP_GIT_URL variable is set to apply the filters or not
+        if [[ -z "${KFP_GIT_URL}" ]]; then
+          echo "KFP_GIT_URL variable not defined. False positives won't be filtered"
+          mv sast_snyk_check_out_all_findings.json filtered_sast_snyk_check_out.json
+        else
+          echo "Filtering false positives in results files using csfilter-kfp..."
+
+          CMD=(
+            csfilter-kfp
+            --verbose
+            --kfp-git-url="${KFP_GIT_URL}"
+            --project-nvr="${PROJECT_NAME}"
+          )
+
+          if [ "${RECORD_EXCLUDED}" == "true" ]; then
+            CMD+=(--record-excluded="excluded-findings.json")
+          fi
+
+          set +e
+          "${CMD[@]}" sast_snyk_check_out_all_findings.json >filtered_sast_snyk_check_out.json
+          status=$?
+          set -e
+          if [ "$status" -ne 0 ]; then
+            echo "Error: failed to filter known false positives" >&2
+            return 1
+          else
+            echo "Message: Succeed to filter known false positives" >&2
+          fi
+          echo "Results after filtering:"
+          (set -x && csgrep --mode=evtstat filtered_sast_snyk_check_out.json)
+        fi
+
+        csgrep --mode=sarif filtered_sast_snyk_check_out.json >sast_snyk_check_out.sarif
+        if [[ "$SNYK_EXIT_CODE" -eq 0 ]] || [[ "$SNYK_EXIT_CODE" -eq 1 ]]; then
+          TEST_OUTPUT=
+          parse_test_output "$(context.task.name)" sarif sast_snyk_check_out.sarif  || true
+
+        # When the test is skipped, the "SNYK_EXIT_CODE" is 3 and it can also be 3 in some other situation
+        elif [[ "$test_not_skipped" -eq 0 ]]; then
+          note="Task $(context.task.name) success: Snyk code test found zero supported files."
+          ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
+        else
+          echo "sast-snyk-check test failed because of the following issues:"
+          cat stdout.txt
+          note="Task $(context.task.name) failed: For details, check Tekton task log."
+          ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+        fi
+        echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee "$(results.TEST_OUTPUT.path)"
+    - name: upload
+      image: quay.io/konflux-ci/oras:latest@sha256:f4b891ee3038a5f13cd92ff4f473faad5601c2434d1c6b9bccdfc134d9d5f820
+      workingDir: /var/workdir/source
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+      script: |
+        #!/usr/bin/env bash
+
+        if [ -z "${IMAGE_URL}" ]; then
+          echo 'No image-url provided. Skipping upload.'
+          exit 0
+        fi
+
+        UPLOAD_FILES="sast_snyk_check_out.sarif excluded-findings.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              continue
+          fi
+            if [ "${UPLOAD_FILES}" == "excluded-findings.json" ]; then
+                MEDIA_TYPE=application/json
+          else
+                MEDIA_TYPE=application/sarif+json
+          fi
+            echo "Selecting auth"
+            select-oci-auth "${IMAGE_URL}" >"${HOME}/auth.json"
+            echo "Attaching to ${IMAGE_URL}"
+            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        done
--- a/task/sast-snyk-check/0.3/MIGRATION.md
+++ b/task/sast-snyk-check/0.3/MIGRATION.md
@@ -1,0 +1,13 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `IMP_FINDINGS_ONLY` parameter has been introduced and enabled by default with "true" value. Only high or critical vulnerabilities will be shown. This behavior can be disabled with "false" value.
+- The scan results uploaded in the SARIF format now additionally contain source code snippets and `csdiff/v1` fingerprints for each finding.
+- There are no default arguments as "--all-projects --exclude=test*,vendor,deps" are ignored by Snyk Code
+- SARIF produced by Snyk Code is not included in the CI log.
+- The `KFP_GIT_URL` parameter has been introduced to indicate the repository to filter false positives. If this variable is left empty, the results won't be filtered. At the same time, we can store all excluded findings in a file using the `RECORD_EXCLUDED` parameter and specify a name of project with the `PROJECT_NAME` to use specific filters.
+
+## Action from users
+
+Renovate bot PR will be created with warning icon for a sast-snyk-check which is expected, no action from users are required.
--- a/task/sast-snyk-check/0.3/README.md
+++ b/task/sast-snyk-check/0.3/README.md
@@ -1,0 +1,39 @@
+# sast-snyk-check task
+
+## Description:
+
+The sast-snyk-check task uses Snyk Code tool to perform Static Application Security Testing (SAST) for Snyk, a popular cloud-native application security platform.
+
+Snyk's SAST tool uses a combination of static analysis and machine learning techniques to scan an application's source code for potential security vulnerabilities, including common issues such as SQL injection, cross-site scripting (XSS), and code injection attacks.
+
+> NOTE: This task is executed only if the user provides a Snyk token stored in a secret in their namespace. The name of the secret then needs to be supplied in the `snyk-secret` pipeline parameter.
+
+## Params:
+
+| name                  | description                                                                                                        | default value                                             | required |
+|-----------------------|--------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|----------|
+| SNYK_SECRET           | Name of secret which contains Snyk token.                                                                          | snyk-secret                                               | true     |
+| ARGS                  | Append arguments.                                                                                                  | ""                                                        | false    |
+| IMP_FINDINGS_ONLY     | Report only important findings.  To report all findings, specify "false"                                           | true                                                      | true     |
+| KFP_GIT_URL           | Link to the known-false-positives repository. If left blank, results won't be filtered                             | ""                                                        | false    |
+| PROJECT_NAME          | Name of the scanned project, used to find path exclusions. By default, the Konflux component name will be used.    | ${metadata.labels['appstudio.openshift.io/component']}    | false    |
+| RECORD_EXCLUDED       | Write excluded records in file. Useful for auditing.                                                               | false                                                     | false    |
+
+## How to obtain a snyk-token and enable snyk task on the pipeline:
+
+Follow the steps given [here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/)
+
+## Results:
+
+| name          | description                |
+|---------------|----------------------------|
+| TEST_OUTPUT   | Tekton task test output.   |
+
+## Source repository for image:
+
+https://github.com/konflux-ci/konflux-test
+
+## Additional links:
+
+* https://snyk.io/product/snyk-code/
+* https://snyk.io/
--- a/task/sast-snyk-check/0.3/sast-snyk-check.yaml
+++ b/task/sast-snyk-check/0.3/sast-snyk-check.yaml
@@ -1,0 +1,244 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.3"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "konflux"
+  name: sast-snyk-check
+spec:
+  description: |-
+    Scans source code for security vulnerabilities, including common issues such as SQL injection, cross-site scripting (XSS), and code injection attacks using Snyk Code, a Static Application Security Testing (SAST) tool.
+
+    Follow the steps given [here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/) to obtain a snyk-token and to enable the snyk task in a Pipeline.
+
+    The snyk binary used in this Task comes from a container image defined in https://github.com/konflux-ci/konflux-test
+
+    See https://snyk.io/product/snyk-code/ and https://snyk.io/ for more information about the snyk tool.
+  results:
+    - description: Tekton task test output.
+      name: TEST_OUTPUT
+  params:
+    - name: SNYK_SECRET
+      description: Name of secret which contains Snyk token.
+      default: snyk-secret
+    - name: ARGS
+      type: string
+      description: Append arguments.
+      default: ""
+    - description: Image URL.
+      name: image-url
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - name: caTrustConfigMapName
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: trusted-ca
+    - name: caTrustConfigMapKey
+      type: string
+      description: The name of the key in the ConfigMap that contains the CA bundle data.
+      default: ca-bundle.crt
+    - name: IMP_FINDINGS_ONLY
+      type: string
+      description: Report only important findings. Default is true. To report all findings, specify "false"
+      default: "true"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files
+      # FIXME: Red Hat internal projects will default to https://gitlab.cee.redhat.com/osh/known-false-positives.git when KONFLUX-4530 is resolved
+      default: ""
+    - name: PROJECT_NAME
+      type: string
+      description: Name of the scanned project, used to find path exclusions. By default, the Konflux component name will be used.
+      default: ""
+    - name: RECORD_EXCLUDED
+      type: string
+      description: Write excluded records in file. Useful for auditing (defaults to false).
+      default: "false"
+  volumes:
+    - name: snyk-secret
+      secret:
+        secretName: $(params.SNYK_SECRET)
+        optional: true
+    - name: trusted-ca
+      configMap:
+        name: $(params.caTrustConfigMapName)
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        optional: true
+  steps:
+    - name: sast-snyk-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.7@sha256:cf6808a3bd605630a5d9f20595ff7c43f8645c00381219d32f5a11e88fe37072
+      # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+      # the cluster will set imagePullPolicy to IfNotPresent
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - name: snyk-secret
+          mountPath: "/etc/secrets"
+          readOnly: true
+        - name: trusted-ca
+          mountPath: /mnt/trusted-ca
+          readOnly: true
+      env:
+        - name: SNYK_SECRET
+          value: $(params.SNYK_SECRET)
+        - name: ARGS
+          value: $(params.ARGS)
+        - name: IMP_FINDINGS_ONLY
+          value: $(params.IMP_FINDINGS_ONLY)
+        - name: KFP_GIT_URL
+          value: $(params.KFP_GIT_URL)
+        - name: PROJECT_NAME
+          value: $(params.PROJECT_NAME)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+        - name: COMPONENT_LABEL
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['appstudio.openshift.io/component']
+      script: |
+        #!/usr/bin/env bash
+
+        set -euo pipefail
+        # shellcheck source=/dev/null
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        if [[ -z "${PROJECT_NAME}" ]]; then
+            PROJECT_NAME=${COMPONENT_LABEL}
+        fi
+
+        echo "The PROJECT_NAME used is: ${PROJECT_NAME}"
+
+        # Installation of Red Hat certificates for cloning Red Hat internal repositories
+        ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+        if [ -f "$ca_bundle" ]; then
+          echo "INFO: Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+          update-ca-trust
+        fi
+
+        SNYK_TOKEN_PATH="/etc/secrets/snyk_token"
+        if [ -f "${SNYK_TOKEN_PATH}" ] && [ -s "${SNYK_TOKEN_PATH}" ]; then
+          # SNYK token is provided
+          SNYK_TOKEN="$(cat ${SNYK_TOKEN_PATH})"
+          export SNYK_TOKEN
+        else
+          # According to shellcheck documentation, the following error can be ignored as it is ignored through indirection: https://www.shellcheck.net/wiki/SC2034
+          # shellcheck disable=SC2034
+          to_enable_snyk='[here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/)'
+          note="Task $(context.task.name) skipped: If you wish to use the Snyk code SAST task, please create a secret name snyk-secret with the key 'snyk_token' containing the Snyk token by following the steps given ${to_enable_snyk}"
+          TEST_OUTPUT=$(make_result_json -r SKIPPED -t "$note")
+          echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+          exit 0
+        fi
+
+        SNYK_EXIT_CODE=0
+        SOURCE_CODE_DIR=$(workspaces.workspace.path)
+        SEVERITY_THRESHOLD="high"
+        if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
+          SEVERITY_THRESHOLD="low"
+        fi
+        set +e
+        # We do want to expand ARGS (it can be multiple CLI flags, not just one)
+        # shellcheck disable=SC2086
+        snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output="${SOURCE_CODE_DIR}"/sast_snyk_check_out.json 1>&2>> stdout.txt
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        (cd  "${SOURCE_CODE_DIR}" && csgrep --mode=json --embed-context=3 "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json) \
+          | csgrep --mode=json --strip-path-prefix="source/"  \
+          > sast_snyk_check_out_all_findings.json
+
+        echo "Results:"
+        (set -x && csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
+
+        # We check if the KFP_GIT_URL variable is set to apply the filters or not
+        if [[ -z "${KFP_GIT_URL}" ]]; then
+          echo "KFP_GIT_URL variable not defined. False positives won't be filtered"
+          mv sast_snyk_check_out_all_findings.json filtered_sast_snyk_check_out.json
+        else
+          echo "Filtering false positives in results files using csfilter-kfp..."
+
+          CMD=(
+            csfilter-kfp
+            --verbose
+            --kfp-git-url="${KFP_GIT_URL}"
+            --project-nvr="${PROJECT_NAME}"
+          )
+
+          if [ "${RECORD_EXCLUDED}" == "true" ]; then
+            CMD+=(--record-excluded="excluded-findings.json")
+          fi
+
+          set +e
+          "${CMD[@]}" sast_snyk_check_out_all_findings.json > filtered_sast_snyk_check_out.json
+          status=$?
+          set -e
+          if [ "$status" -ne 0 ]; then
+            echo "Error: failed to filter known false positives" >&2
+            return 1
+          else
+            echo "Message: Succeed to filter known false positives" >&2
+          fi
+          echo "Results after filtering:"
+          (set -x && csgrep --mode=evtstat filtered_sast_snyk_check_out.json)
+        fi
+
+        csgrep --mode=sarif filtered_sast_snyk_check_out.json > sast_snyk_check_out.sarif
+        if [[ "$SNYK_EXIT_CODE" -eq 0 ]] || [[ "$SNYK_EXIT_CODE" -eq 1 ]]; then
+          TEST_OUTPUT=
+          parse_test_output "$(context.task.name)" sarif sast_snyk_check_out.sarif  || true
+
+        # When the test is skipped, the "SNYK_EXIT_CODE" is 3 and it can also be 3 in some other situation
+        elif [[ "$test_not_skipped" -eq 0 ]]; then
+          note="Task $(context.task.name) success: Snyk code test found zero supported files."
+          ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
+        else
+          echo "sast-snyk-check test failed because of the following issues:"
+          cat stdout.txt
+          note="Task $(context.task.name) failed: For details, check Tekton task log."
+          ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+        fi
+        echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee "$(results.TEST_OUTPUT.path)"
+
+    - name: upload
+      image: quay.io/konflux-ci/oras:latest@sha256:f4b891ee3038a5f13cd92ff4f473faad5601c2434d1c6b9bccdfc134d9d5f820
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+      script: |
+        #!/usr/bin/env bash
+
+        if [ -z "${IMAGE_URL}" ]; then
+          echo 'No image-url provided. Skipping upload.'
+          exit 0
+        fi
+
+        UPLOAD_FILES="sast_snyk_check_out.sarif excluded-findings.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              continue
+            fi
+            if [ "${UPLOAD_FILES}" == "excluded-findings.json" ]; then
+                MEDIA_TYPE=application/json
+            else
+                MEDIA_TYPE=application/sarif+json
+            fi
+            echo "Selecting auth"
+            select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
+            echo "Attaching to ${IMAGE_URL}"
+            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        done
+
+  workspaces:
+    - name: workspace
--- a/task/sast-snyk-check-oci-ta/0.3/MIGRATION.md
+++ b/task/sast-snyk-check-oci-ta/0.3/MIGRATION.md
@@ -1,0 +1,13 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `IMP_FINDINGS_ONLY` parameter has been introduced and enabled by default with "true" value. Only high or critical vulnerabilities will be shown. This behavior can be disabled with "false" value.
+- The scan results uploaded in the SARIF format now additionally contain source code snippets and `csdiff/v1` fingerprints for each finding.
+- There are no default arguments as "--all-projects --exclude=test*,vendor,deps" are ignored by Snyk Code
+- SARIF produced by Snyk Code is not included in the CI log.
+- The `KFP_GIT_URL` parameter has been introduced to indicate the repository to filter false positives. If this variable is left empty, the results won't be filtered. At the same time, we can store all excluded findings in a file using the `RECORD_EXCLUDED` parameter and specify a name of project with the `PROJECT_NAME` to use specific filters.
+
+## Action from users
+
+Renovate bot PR will be created with warning icon for a sast-snyk-check which is expected, no action from users are required.
--- a/task/sast-snyk-check-oci-ta/0.3/README.md
+++ b/task/sast-snyk-check-oci-ta/0.3/README.md
@@ -1,0 +1,30 @@
+# sast-snyk-check-oci-ta task
+
+Scans source code for security vulnerabilities, including common issues such as SQL injection, cross-site scripting (XSS), and code injection attacks using Snyk Code, a Static Application Security Testing (SAST) tool.
+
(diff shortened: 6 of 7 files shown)


jperezdealgaba:
@jsztuka Would you mind giving a review for this?

jperezdealgaba:
@jsztuka Does the 👍🏻 mean that it looks good and nothind needs to be modified?

jsztuka (approved):
lgtm

kdudka:
@jsztuka Could you please approve the 6 workflows that are awaiting approval?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,139 @@
...
+      description: Report only vulnerabilities at the specified level or higher (Options are low, medium or high).
+      default: "high"
+  volumes:
+    - name: snyk-secret
+      secret:
+        secretName: $(params.SNYK_SECRET)
+        optional: true
+  steps:
+    - name: sast-snyk-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.5@sha256:801a105ba0f9c7f58f5ba5cde1a3b4404009fbebb1028779ca2c5de211e94940
```
kdudka, line 45: @jperezdealgaba I think we need to update the reference to the image that includes `csdiff`?

jperezdealgaba: You a right. As mentioned in the description, until this MR is not merged: https://github.com/konflux-ci/konflux-test/pull/292 we won't be able to do it.

jperezdealgaba:
Will update lint problems in following commit adding false positives filtering...

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,158 @@
...
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - name: SEVERITY_THRESHOLD
+      type: string
+      description: Report only vulnerabilities at the specified level or higher (Options are low, medium or high).
+      default: "high"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files
+      default: "https://gitlab.cee.redhat.com/jperezde/known-false-positives#ffd64fa4ffd3c66501054ebdbd01f422cb7dcdca"
```
kdudka, line 41: The default `KFP_GIT_URL` should be: https://gitlab.cee.redhat.com/osh/known-false-positives.git

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,158 @@
...
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        set -x
+        csgrep --mode=json --prepend-path-prefix="$SOURCE_CODE_DIR"/ sast_snyk_check_out.json | csgrep --mode=json --embed-context 3 | csgrep --mode=json --strip-path-prefix="$SOURCE_CODE_DIR"/source/  > sast_snyk_check_out_all_findings.json
+        set +x
+
+        echo "Filtering false positives in results files..."
```
kdudka, line 106: According to the acceptance criteria of [OSH-737](https://issues.redhat.com/browse/OSH-737), there should be a way to disable the KFP feature.

So, if `$KFP_GIT_URL` is empty, we should do:
```sh
mv sast_snyk_check_out_all_findings.json filtered_sast_snyk_check_out.json
```
... instead of running `csfilter-kfp`.  This semantics should also be mentioned in the docs.

kdudka (changes requested):

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Added. Documentation has also been updated. If variable is left empty, we don't filter the results.
Example pipeliens:
- Variable set: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-v7h7v (15 findings)
- Variable not set: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-hbpc5 (28 findings)

Results are not filtered

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: FYI: The image has been updated to [version 1.4.6](https://github.com/konflux-ci/konflux-test/releases/tag/v1.4.6). This version contains `csdiff` and `git` packages

jperezdealgaba:
Although the MR  is finished, I will look for ProdSec feedback before taking this out from draft.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: Unfortunately , the updated container image contains `csdiff-3.4.1-1.el9`, which does not include `csfilter-kfp` yet.

jperezdealgaba: How can that be solved? Is the container image needed to be rebuilt again? Do we need to wait until the 3.5.0 version  package is marked as `stable`: https://packages.fedoraproject.org/pkgs/csdiff/csdiff/?

kdudka: Yes, rebuilding the container image after the package appears on mirrors of stable Fedora repositories sounds like the most straightforward approach.  I believe it will be submitted to stable later today as I have asked a few people to test and provide karma at: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2024-b788a2df5c

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,15 @@
...
+
+Version 0.3:
+
+- The `--severity-threshold` argument has been introduced and enabled by default with "high" value. Only high or superior vulnerabilities will be shown. Possible values for this argument are: `low`, `medium` and `high`.
+- The results are parsed using `csgrep` and they are uploaded as results with a generated fingerprint.
+- There are no default arguments as "--all-projects --exclude=test*,vendor,deps" are ignored by Snyk Code
+- The results are uploaded as SARIF files to the registry
+- SARIF produced by Snyk Code is not included in the CI log.
+- The findings are filtered using the known-false-positives repository (https://gitlab.cee.redhat.com/osh/known-false-positives/)
+- The KFP_GIT_URL environment variable has been introduced to indicate the repository to filter false positives. If this variable is left empty, the results won't be filtered.
```
kdudka, line 11: I think we should refer to `KFP_GIT_URL` as a parameter here.

kdudka (commented):
@jperezdealgaba I can see that we still print the full SARIF file into the CI log, which is not much user-friendly and it will cause problems on tasks that produce too much results.  I would suggest to invoke `csgrep --mode=evtstat` instead to provide a useful summary for end users.  In case we are filtering false positives we can run the command twice to record the number of findings that were excluded from the results.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: The package already appears as "stable". I have asked the Konflux team to rebuild the image. Waiting for response

task/sast-snyk-check/0.3/MIGRATION.md
jperezdealgaba: Fixed

jperezdealgaba:
> @jperezdealgaba I can see that we still print the full SARIF file into the CI log, which is not much user-friendly and it will cause problems on tasks that produce too much results. I would suggest to invoke `csgrep --mode=evtstat` instead to provide a useful summary for end users. In case we are filtering false positives we can run the command twice to record the number of findings that were excluded from the results.

The SARIF output is no longer shown and the results are shown in `evtstat` mode. Before and after filtering.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Regarding this. Engineering team just told me:

> Hi, we can't rebuild the release, but we can adopt the changes when they get propagated to the konflux-test image.

kdudka: If building an image with updated `csdiff` is too complicated, we can updated `csdiff` in the CI task as an interim solution.  The package update command can be removed later on when the image gets updated.

jperezdealgaba: Just got:
> I think we should be able to cut a new release tomorrow

So I will wait until the end of this week and if not, I will add the interim soultion

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,199 @@
...
+          echo "No ${UPLOAD_FILE} exists. Skipping upload."
+          exit 0;
+        fi
+
+        echo "Selecting auth"
+        select-oci-auth $IMAGE_URL > $HOME/auth.json
+        echo "Attaching to ${IMAGE_URL} via the OCI 1.1 Referrers API"
+        oras attach --no-tty --registry-config "$HOME/auth.json" --distribution-spec v1.1-referrers-api --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        echo "Attaching to ${IMAGE_URL} via the OCI 1.1 Referrers Tag"
+        oras attach --no-tty --registry-config "$HOME/auth.json" --distribution-spec v1.1-referrers-tag --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
kdudka, line 197: @jperezdealgaba Please include the change from #1435 so that the update from 0.2 to 0.3 does not introduce a regression.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: The new container image was released (V1.4.7) (https://quay.io/repository/redhat-appstudio/konflux-test?tab=tags&tag=latest)! I tested and it is containing `csfilter-kfp` : https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-8wx2v

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Added.

jperezdealgaba:
@kdudka I updated the container image, I added the `PROJECT_NVR` and `RECORD_EXCLUDED` and updated the upload task

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,200 @@
...
+
+        UPLOAD_FILE=sast_snyk_check_out.sarif
+        MEDIA_TYPE=application/sarif+json
+
+        if [ -z "${IMAGE_URL}" ] || [ -z "${IMAGE_DIGEST}"  ]; then
+          echo 'No image-url or image-digest param provided. Skipping upload.'
+          exit 0;
+        fi
+
+        UPLOAD_FILES="sast_snyk_check_out.sarif record_excluded.json"
```
rhyw, line 187: if the param is a a string and could be specified by user, it could be a filename other than `record_excluded.json`.

Hank had an attempt in his [MR](https://github.com/konflux-ci/build-definitions/pull/1401/files#diff-7e973ab9c156c6c0c394a71e757617d0bc77e24d15dd3b7874dbd61a50cedf24R41) by using boolean `RECORD_EXCLUDED` and use this default value. WDYT?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,200 @@
...
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              exit 0;
+            fi
+
+        echo "Selecting auth"
+        select-oci-auth $IMAGE_URL > $HOME/auth.json
+        echo "Attaching to ${IMAGE_URL}"
+        oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
rhyw, line 197: I didnt' try but will it work if the `UPLOAD_FILE` is unspecified?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,200 @@
...
+        if [ -z "${IMAGE_URL}" ] || [ -z "${IMAGE_DIGEST}"  ]; then
+          echo 'No image-url or image-digest param provided. Skipping upload.'
+          exit 0;
+        fi
+
+        UPLOAD_FILES="sast_snyk_check_out.sarif record_excluded.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              exit 0;
```
rhyw, line 191: Should we use `continue` here?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,200 @@
...
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+        - name: IMAGE_DIGEST
+          value: $(params.image-digest)
+      script: |
+        #!/usr/bin/env bash
+
+        UPLOAD_FILE=sast_snyk_check_out.sarif
```
rhyw, line 179: I believe this line is redundant since we already have `UPLOAD_FILES`

rhyw (changes requested):

task/sast-snyk-check/0.3/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Params:
+
+| name                     | description                                                                                                  |
+|--------------------------|--------------------------------------------------------------------------------------------------------------|
+| SNYK_SECRET              | Name of secret which contains Snyk token.                                                                    |
+| ARGS                     | Append arguments.                                                                                            |
+| SEVERITY_THRESHOLD       | Only vulnerabilities of the specified level or higher are reported (Possible values are low, medium or high) |
+| KFP_GIT_URL              | Link to the known-false-positives repository. If left blank, results won't be filtered                       |
+| PROJECT_NVR              | Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional)             |
+| RECORD_EXCLUDED          | File to store all excluded findings to (it is optional)                                                      |
```
kdudka, line 20: This should take `true` or `false`.  There is no reason to configure the file name for `RECORD_EXCLUDED` if the the default list of (not excluded) findings is not configurable.  Let's stick to `excluded-findings.json` for consistency across the SAST scanning CI tasks.

kdudka (changes requested):

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: Yes, it should be a boolean value.  If it is set to `false`, the extra file will not be uploaded.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: Yep, the exit in a `for` loop looks wrong to me.  The semicolon at the end of line is redundant in shell.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: Yes, and the `for` loop should be terminated by `done` so that `bash` can parse the script.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,200 @@
...
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - name: SEVERITY_THRESHOLD
+      type: string
+      description: Report only vulnerabilities at the specified level or higher (Options are low, medium or high).
+      default: "high"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files
+      default: "https://gitlab.cee.redhat.com/osh/known-false-positives.git"
```
kdudka, line 41: Please change the default value of `KFP_GIT_URL` to an empty string for now with the following comment:
```
```

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Fixed!

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Fixed!

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Fixed!

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,207 @@
...
+        if [[ "$SNYK_EXIT_CODE" -eq 0 ]] || [[ "$SNYK_EXIT_CODE" -eq 1 ]]; then
+          TEST_OUTPUT=
+          parse_test_output '$(context.task.name)' sarif sast_snyk_check_out.sarif  || true
+
+        # When the test is skipped, the "SNYK_EXIT_CODE" is 3 and it can also be 3 in some other situation
+        elif [[ "$test_not_skipped" -eq 0 ]]; then
+          note="Task $(context.task.name) success: Snyk code test found zero supported files."
+          ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
+        else
+          note="Task $(context.task.name) failed: For details, check Tekton task log."
```
kdudka, line 172: I think the error from Snyk should still be printed?  Why did we remove the following code?  https://github.com/konflux-ci/build-definitions/blob/365086844919ce7f4a839a73db981b89b9c792c4/task/sast-snyk-check/0.2/sast-snyk-check.yaml#L93

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,207 @@
...
+            --verbose
+            --kfp-git-url="${KFP_GIT_URL}"
+          )
+        
+          if [[ -n "${PROJECT_NVR}" ]]; then
+            CMD+=(--project-nvr="${PROJECT_NVR}")
+          fi
+
+          if [ "${RECORD_EXCLUDED}" == "true" ]; then
+            CMD+=(--record-excluded="${RECORD_EXCLUDED}")
```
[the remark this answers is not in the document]
jperezdealgaba: Fixed

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Added

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,207 @@
...
+        if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
+          SEVERITY_THRESHOLD="low"
+        fi
+        snyk code test $ARGS --severity-threshold=$SEVERITY_THRESHOLD "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output=sast_snyk_check_out.json 1>&2>> stdout.txt || SNYK_EXIT_CODE=$?
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        csgrep --mode=json --prepend-path-prefix="$SOURCE_CODE_DIR"/ sast_snyk_check_out.json \
```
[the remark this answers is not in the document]
jperezdealgaba: Modified

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,207 @@
...
+          value: $(params.PROJECT_NVR)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+      script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        # Installation of Red Hat certificates for cloning Red Hat internal repositories
```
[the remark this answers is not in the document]
jperezdealgaba: Wasn't aware of that. fixed! Thanks! Just modified.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,207 @@
...
+      type: string
+      description: URL from repository to download known false positives files
+      default: "" # FIXME: Red Hat internal projects will default to https://gitlab.cee.redhat.com/osh/known-false-positives.git when KONFLUX-4530 is resolved
+    - name: PROJECT_NVR
+      type: string
+      description: Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional)
+      default: ""
+    - name: RECORD_EXCLUDED
+      type: string
+      description: Write excluded records in file. Useful for debugging (it is optional).
```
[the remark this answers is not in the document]
jperezdealgaba: Modified

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `--severity-threshold` argument has been introduced and enabled by default with "high" value. Only high or superior vulnerabilities will be shown. Possible values for this argument are: `low`, `medium` and `high`.
+- The results are parsed using `csgrep` and they are uploaded as results with a generated fingerprint.
+- There are no default arguments as "--all-projects --exclude=test*,vendor,deps" are ignored by Snyk Code
+- The results are uploaded as SARIF files to the registry
+- SARIF produced by Snyk Code is not included in the CI log.
+- The findings are filtered using the known-false-positives repository (https://gitlab.cee.redhat.com/osh/known-false-positives/)
```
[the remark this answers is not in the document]
jperezdealgaba: Dropped!

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `--severity-threshold` argument has been introduced and enabled by default with "high" value. Only high or superior vulnerabilities will be shown. Possible values for this argument are: `low`, `medium` and `high`.
+- The results are parsed using `csgrep` and they are uploaded as results with a generated fingerprint.
+- There are no default arguments as "--all-projects --exclude=test*,vendor,deps" are ignored by Snyk Code
+- The results are uploaded as SARIF files to the registry
```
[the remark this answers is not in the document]
jperezdealgaba: Dropped

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `--severity-threshold` argument has been introduced and enabled by default with "high" value. Only high or superior vulnerabilities will be shown. Possible values for this argument are: `low`, `medium` and `high`.
+- The results are parsed using `csgrep` and they are uploaded as results with a generated fingerprint.
```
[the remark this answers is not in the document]
jperezdealgaba: Modified

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `--severity-threshold` argument has been introduced and enabled by default with "high" value. Only high or superior vulnerabilities will be shown. Possible values for this argument are: `low`, `medium` and `high`.
```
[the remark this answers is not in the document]
jperezdealgaba: Modified

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,229 @@
...
+        SNYK_EXIT_CODE=0
+        SOURCE_CODE_DIR=$(workspaces.workspace.path)
+        # shellcheck disable=SC2086
+        # We do want to expand ARGS (it can be multiple CLI flags, not just one)
+        SEVERITY_THRESHOLD="high"
+        if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
+          SEVERITY_THRESHOLD="low"
+        fi
+        snyk code test $ARGS --severity-threshold=$SEVERITY_THRESHOLD "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output=sast_snyk_check_out.json 1>&2>> stdout.txt
+        SNYK_EXIT_CODE=$?
```
kdudka, line 133: @jperezdealgaba Sorry, this change needs to be reverted.  I did not realize that the whole script runs with `set -e`.  So the ugly construction was there for a reason.  My bad.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,229 @@
...
+          if [[ -n "${PROJECT_NVR}" ]]; then
+            CMD+=(--project-nvr="${PROJECT_NVR}")
+          fi
+
+          if [ "${RECORD_EXCLUDED}" == "true" ]; then
+            CMD+=(--record-excluded="excluded-findings.json")
+          fi
+
+          "${CMD[@]}" sast_snyk_check_out_all_findings.json > filtered_sast_snyk_check_out.json
+          status=$?
```
kdudka, line 168: See my previous comment. The `status=$?` command will not work with `set -e` because non-zero exit code of the previous command will terminate the shell.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Reverted. I temporarily disable `-e`.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Reverted. I temporarily disable `-e`.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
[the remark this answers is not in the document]
jperezdealgaba: Just in case, the path of the `sast_snyk_check_out.json` file was modified as it couldn't be found after we executed the commands on the `SOURCE_CODE_DIR`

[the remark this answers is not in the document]
kdudka: Right.  I should have used an absolute path to the input file in my example.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,234 @@
...
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 $(workspaces.workspace.path)/hacbs/$(context.task.name)/sast_snyk_check_out.json) \
+        | csgrep --mode=json --strip-path-prefix="source/"  \
+         > sast_snyk_check_out_all_findings.json
```
kdudka, line 143: Please indent both the lines by two spaces to make it obvious they are continuation of the previous command:
```diff
--- a/task/sast-snyk-check/0.3/sast-snyk-check.yaml
+++ b/task/sast-snyk-check/0.3/sast-snyk-check.yaml
@@ -139,8 +139,8 @@ spec:

         # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
         (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 $(workspaces.workspace.path)/hacbs/$(context.task.name)/sast_snyk_check_out.json) \
-        | csgrep --mode=json --strip-path-prefix="source/"  \
-         > sast_snyk_check_out_all_findings.json
+          | csgrep --mode=json --strip-path-prefix="source/"  \
+          > sast_snyk_check_out_all_findings.json

         echo "Results:"
         (set -x; csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
```

task/sast-snyk-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,13 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+- The `IMP_FINDINGS_ONLY` parameter has been introduced and enabled by default with "true" value. Only high or critical vulnerabilities will be shown. This behavior can be disabled with "false" value.
```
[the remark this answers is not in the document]
jperezdealgaba: Fixed!

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Fixed

task/sast-snyk-check/0.3/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+| name              | description                                                                                      | default value | required |
+|-------------------|--------------------------------------------------------------------------------------------------|---------------|----------|
+| SNYK_SECRET       | Name of secret which contains Snyk token.                                                        | snyk-secret   | yes      |
+| ARGS              | Append arguments.                                                                                | ""            | no       |
+| IMP_FINDINGS_ONLY | Report only important findings. Default is true. To report all findings, specify "false"         | true          | yes      |
+| KFP_GIT_URL       | Link to the known-false-positives repository. If left blank, results won't be filtered           | ""            | no       |
+| PROJECT_NVR       | Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional) | ""            | no       |
+| RECORD_EXCLUDED   | Write excluded records in file. Useful for debugging (it is optional). Default is "false"        | false         | no       |
```
jperezdealgaba, line 20: This doesn't match the description from the task parameter. I will update it in a bit...

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,234 @@
...
+        
+        SNYK_TOKEN_PATH="/etc/secrets/snyk_token"
+
+        if [ -f "${SNYK_TOKEN_PATH}" ] && [ -s "${SNYK_TOKEN_PATH}" ]; then
+          # SNYK token is provided
+          SNYK_TOKEN="$(cat ${SNYK_TOKEN_PATH})"
+          export SNYK_TOKEN
+        else
+          to_enable_snyk='[here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/)'
+          note="Task $(context.task.name) skipped: If you wish to use the Snyk code SAST task, please create a secret name snyk-secret with the key "snyk_token" containing the Snyk token by following the steps given ${to_enable_snyk}"
```
hanchuntao, line 118: this `note` may not work when `snyk_token ` using `"`

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,234 @@
...
+          "${CMD[@]}" sast_snyk_check_out_all_findings.json > filtered_sast_snyk_check_out.json
+          status=$?
+          set -e
+        
+          if [ "$status" -ne 0 ]; then
+            echo "Error: failed to filter known false positives" >&2
+            return 1
+          else
+            echo "Message: Succeed to filter known false positives" >&2
+            SCAN_RESULT="filtered_sast_unicode_check_out.json"
```
hanchuntao, line 179: Where using this `SCAN_RESULT` ?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,234 @@
...
+        set +e
+        snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output=sast_snyk_check_out.json 1>&2>> stdout.txt
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 $(workspaces.workspace.path)/hacbs/$(context.task.name)/sast_snyk_check_out.json) \
```
hanchuntao, line 141: In in L75, the task is  at directory `$(workspaces.workspace.path)/hacbs/$(context.task.name)/`, that means `sast_snyk_check_out.json` is in the current directory, why need to add the full path again?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,234 @@
...
+          | csgrep --mode=json --strip-path-prefix="source/"  \
```
hanchuntao, line 142: only strip "source/"  ?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: We are executing that command from $SOURCE_CODE_DIR and not from the task directory

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Snyk findings will be located there and does not analyze the dependencies. What else should I strip?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Removed

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Single-quoted

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,233 @@
...
+      description: Name of secret which contains Snyk token.
+      default: snyk-secret
+    - name: ARGS
+      type: string
+      description: Append arguments.
+      default: ""
+    - description: Image URL.
+      name: image-url
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
```
kdudka, line 27: [The code](https://github.com/konflux-ci/build-definitions/blob/e95aec6a94bf6dcf532cd75de9a6b8b4e7f79fb1/task/sast-snyk-check/0.2/sast-snyk-check.yaml#L113) to skip the upload when `image-url` is empty is already being removed in 0.3 though.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,233 @@
...
+          to_enable_snyk='[here](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/testing_applications/enable_snyk_check_for_a_product/)'
+          note='Task $(context.task.name) skipped: If you wish to use the Snyk code SAST task, please create a secret name snyk-secret with the key "snyk_token" containing the Snyk token by following the steps given ${to_enable_snyk}'
+          TEST_OUTPUT=$(make_result_json -r SKIPPED -t "$note")
+          echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+          exit 0
+        fi
+
+        SNYK_EXIT_CODE=0
+        SOURCE_CODE_DIR=$(workspaces.workspace.path)
+        # shellcheck disable=SC2086
```
kdudka, line 126: The ShellCheck directive is misplaced now.  It needs to be right before the command that expands `$ARGS`.

jperezdealgaba: Moved directive

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: I am not sure If I understand your comment. I added the checks again as I don't know when the Konflux team is going to request `image-url` and `image-digest` to be mandatory. Does that make sense?

kdudka: I was trying to say that the (now reintroduced) check is needed to make the task succeed without the `image-url` parameter set.  The check makes the `upload` step exit with 0 in this case.  Without the check, the step would invoke `select-oci-auth` and `oras attach` with an empty string, which would fail and most likely cause the whole task to be marked as failed.  Looking at the code now, I am not sure why the task actually takes the `image-digest` parameter, which is not used for anything but the check.

kdudka: The last use of the `image-digest` value was removed in d1aa94f80c7b44d37c836ffe63650d15add2d60d.

jperezdealgaba: So it really seems it was removed. I have removed the references and have left the check for `IMAGE_URL`

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -0,0 +1,231 @@
...
+            fi
+            if [ "${UPLOAD_FILES}" == "excluded-findings.json" ]; then
+                MEDIA_TYPE=application/json
+            else
+                MEDIA_TYPE=application/sarif+json
+            fi
+            echo "Selecting auth"
+            select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
+            echo "Attaching to ${IMAGE_URL}"
+            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
jsztuka, line 227: Out of curiosity, what is the endpoint of this `oras` command?

kdudka: Yes, in general, it should be the same endpoint as the one where the binary image built by Konflux resides.  This command was actually reused from the 0.2 version of the `sast-snyk-check` task: https://github.com/konflux-ci/build-definitions/blob/f9b9f507fe54c281648b86dfa1e4f1b2148a0dcd/task/sast-snyk-check/0.2/sast-snyk-check.yaml#L126

jsztuka:
Will approve once those linting issues are resolved.

chipspeak:
Same as above, happy to approve once checks pass.

jperezdealgaba:
@chipspeak @jsztuka Sorry, I was on PTO! The linting errors should have been fixed.

jperezdealgaba:
@lcarva Do you have more information about it, please?

kdudka:
@jperezdealgaba This is tracked at https://issues.redhat.com/browse/OSH-764.

jperezdealgaba:
@kdudka Thank you very much for that information! Wasn't aware of that.

@lcarva I added the oci-ta updated task. Is there any running pipeline using oci-ta that I could check to test mine? I would like to see how the `SOURCE_ARTIFACT` and `CACHI2_ARTIFACT` are populated as they don't contain any default value

jperezdealgaba:
@jsztuka Thanks for enabling the CI. I have run the `hack/generate-ta-tasks.sh` and apply the `ta.patch` as commented. What else is missing?

jsztuka:
That should resolve the issues, I saw some conflicts, but it looks like there were resolved.

jperezdealgaba:
@chmeliik All tests passed. Nevertheless, were my changes in the `Red Hat Konflux / build-definitions-pull-request ` CI tested? Just want to make sure as I can not see the results

jsztuka:
Yes, they were tested successfully, you can merge this one.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -0,0 +1,204 @@
...
+          value: $(params.ARGS)
+        - name: IMP_FINDINGS_ONLY
+          value: $(params.IMP_FINDINGS_ONLY)
+        - name: KFP_GIT_URL
+          value: $(params.KFP_GIT_URL)
+        - name: PROJECT_NVR
+          value: $(params.PROJECT_NVR)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+      script: "#!/usr/bin/env bash\n\nset -euo pipefail\n# shellcheck source=/dev/null\n.
```
kdudka, line 120: @jperezdealgaba I do not think the script should be obfuscated?  It used to be human-readable in the 0.2 version: https://github.com/konflux-ci/build-definitions/blob/31180b4b4c06e971dcaaac34ec2cfa62355469d9/task/sast-snyk-check-oci-ta/0.2/sast-snyk-check-oci-ta.yaml#L78

kdudka (changes requested):

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
jperezdealgaba: Sure. Not sure how that happeneed. It was fixed

jperezdealgaba: Unresolving this. Just found out why this happened. In the "Trusted Artifacts variants" CI task, the error shows this message: 

```
Error: File is out of date, run `hack/generate-ta-tasks.sh` and include the updated file with your changes
Notice: Download the attached `ta.patch` file and run `git apply ta.patch`
```

Once applying the ta.patch, the obfuscated string script is generated. You can see it here: https://github.com/konflux-ci/build-definitions/actions/runs/11362215386/job/31604655663?pr=1359

Which file should I leave?

kdudka: @lcarva Could you please advise how to create the `oci-ta` version properly?

jperezdealgaba: @lcarva Once the ` ta.patch` file is applied, the script section of the `Task` gets obfuscated. That changed is even shown in the pipeline log: https://github.com/konflux-ci/build-definitions/actions/runs/11362708701/job/31605247099?pr=1359.

The previous version of the snyk-oci-ta task are not obfuscated, that's the reason we are asking, as it seems inconsistent with previous versions of the task

lcarva: The generated file might be ok. It's just using an inline string with `\n` instead of the nicer-looking `|`.

lcarva:
> @lcarva I added the oci-ta updated task. Is there any running pipeline using oci-ta that I could check to test mine? I would like to see how the `SOURCE_ARTIFACT` and `CACHI2_ARTIFACT` are populated as they don't contain any default value

@jperezdealgaba, there are probably other examples, but you can make a change [here](https://github.com/enterprise-contract/ec-cli/blob/696c1ddf6f243267ea3e6f607ecb8b46f232f34d/.tekton/cli-build.yaml#L328-L353) and submit a PR to that repo.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
lcarva: The diff, at least, looks reasonable:

```diff
🐚 diff <(< task/sast-snyk-check/0.3/sast-snyk-check.yaml yq '.spec.steps[0].script') <(< task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml yq '.spec.steps[1].script')
32c32
< SOURCE_CODE_DIR=$(workspaces.workspace.path)
---
> SOURCE_CODE_DIR=/var/workdir
40c40
< snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output=sast_snyk_check_out.json 1>&2>> stdout.txt
---
> snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output=sast_snyk_check_out.json 1>&2 >>stdout.txt
48,50c48,50
< (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 "$(workspaces.workspace.path)"/hacbs/"$(context.task.name)"/sast_snyk_check_out.json) \
<   | csgrep --mode=json --strip-path-prefix="source/"  \
<   > sast_snyk_check_out_all_findings.json
---
> (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 "/var/workdir"/hacbs/"$(context.task.name)"/sast_snyk_check_out.json) |
>     csgrep --mode=json --strip-path-prefix="source/" \
>     >sast_snyk_check_out_all_findings.json
53c53,56
< (set -x; csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
---
> ( 
>   set -x
>          csgrep --mode=evtstat sast_snyk_check_out_all_findings.json
> )
77c80
<   "${CMD[@]}" sast_snyk_check_out_all_findings.json > filtered_sast_snyk_check_out.json
---
>   "${CMD[@]}" sast_snyk_check_out_all_findings.json >filtered_sast_snyk_check_out.json
87c90,93
<   (set -x; csgrep --mode=evtstat filtered_sast_snyk_check_out.json)
---
>   ( 
>     set -x
>            csgrep --mode=evtstat filtered_sast_snyk_check_out.json
>   )
90c96
< csgrep --mode=sarif filtered_sast_snyk_check_out.json > sast_snyk_check_out.sarif
---
> csgrep --mode=sarif filtered_sast_snyk_check_out.json >sast_snyk_check_out.sarif

```

lcarva: > But why don't previous versions of the tasks have the same format

I don't know. And it also doesn't happen to the script on the next step. I wonder if it's because the lines in the original script are too long?

lcarva: These lines are causing it:

```
(set -x; csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
```

```
(set -x; csgrep --mode=evtstat filtered_sast_snyk_check_out.json)
```

lcarva: Changing them to `set -x && csgrep` seems to generate a better script.

jperezdealgaba: Thank you very much for that! I will try it with those changes!

jperezdealgaba:
@kdudka I added the new parameters. would you mind giving a review?

P.S.: I am waiting for the CI to start in order to see if the `oci-ta tasks` problems are solved

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -0,0 +1,262 @@
...
+  steps:
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:e0e457b6af10e44ff6b90208a9e69adc863a865e1c062c4cb84bf3846037d74d
+      args:
+        - use
+        - $(params.SOURCE_ARTIFACT)=/var/workdir/source
+        - $(params.CACHI2_ARTIFACT)=/var/workdir/cachi2
+    - name: sast-snyk-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.7@sha256:cf6808a3bd605630a5d9f20595ff7c43f8645c00381219d32f5a11e88fe37072
+      workingDir: /var/workdir/source
```
jperezdealgaba, line 99: This needs to be checked. We are storing results in the same folder as the code, which doesn't seem like a good solution.
If we change this, that means that not all `oci-ta` tasks are similar

kdudka (approved):
@jperezdealgaba The latest changes look good.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -0,0 +1,262 @@
...
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        (cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 "/var/workdir"/hacbs/"$(context.task.name)"/sast_snyk_check_out.json) |
+            csgrep --mode=json --strip-path-prefix="source/" \
+            >sast_snyk_check_out_all_findings.json
```
jperezdealgaba, line 180: @kdudka FYI: The missing space in this line is being done by the `*-ta` CI pipeline. It seems that some cases are not covered by the script.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
zregvart: This can be changed [here](https://github.com/konflux-ci/build-definitions/pull/1359/files#diff-380d45b72d5bd651295c4026654a85b1e08671a85887a51077c790a787095963R29). You can try removing that line.
Might be worth not using the `hacbs` term any more in the new version of the Task.

zregvart: In general, when using TA one can do whatever they want to the source code and the changes will not affect other Tasks, that's the point of using TA.

jperezdealgaba: @zregvart Is there any alternative for `hacbs` or should it be only removed?

jperezdealgaba: @zregvart So according to the `recipe` file, the `var/workdir/source` should be replaced by `$(workspaces.workspace.path)/hacbs/$(context.task.name)`, right?

zregvart: Other way around `hacbs/$(context.task.name)` will be replaced with `source`.

jperezdealgaba: Isn't that contradicting what CI pipeline is suggesting: https://github.com/konflux-ci/build-definitions/actions/runs/11386335463/job/31682867224?pr=1359#step:3:107 ?
It is suggesting replacing `source/` for `/hacbs/"$(context.task.name)"/`

zregvart: The [line](https://github.com/konflux-ci/build-definitions/pull/1359/files#diff-5f4be57cde9a448d68562ae32cf95e031a726998a5fb3fd192736af02e5ada8fR150) in the non-TA version of the Task is:

```
(cd "$SOURCE_CODE_DIR" && csgrep --mode=json --embed-context=3 "$(workspaces.workspace.path)"/hacbs/"$(context.task.name)"/sast_snyk_check_out.json) \
```

It does not match the regex `hacbs/\$\(context.task.name\)`, there is an extra `"` after `hacbs/`. This means that no replacement is made and the line is taken as is from the non-TA version. This is what you see in the diff. You should not edit the generated files (`sast-snyk-check-oci-ta.yaml`) by hand, run `hack/generate-ta-tasks.sh` locally and tweak the `recipe.yaml` until you reach the desired state of `sast-snyk-check-oci-ta.yaml`.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -0,0 +1,262 @@
...
+        # shellcheck disable=SC2086
+        snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$SOURCE_CODE_DIR" --max-depth=1 --sarif-file-output="${SOURCE_CODE_DIR}"/sast_snyk_check_out.json 1>&2 >>stdout.txt
+        SNYK_EXIT_CODE=$?
+        set -e
+        test_not_skipped=0
+        SKIP_MSG="We found 0 supported files"
+        grep -q "$SKIP_MSG" stdout.txt || test_not_skipped=$?
+
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+        csgrep --mode=json --embed-context=3 "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json |
```
jperezdealgaba, line 178: @kdudka After talking with @zregvart we found that modifying the place of the Snyk results file was better later to generate the `oci-ta` task, as the `generate-ta-task.sh` script had some problems parsing the paths and wasn't working as expected.
What do you think? I tested it and it is working as expected: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-4858d

kdudka: Unfortunately, this does not work as expected.  We need to change CWD for `csgrep --embed-context` to work.

I downloaded scan results from the above test and the source code context and, consequently, the `csdiff/v1` fingerprints were missing:
```
% oras pull quay.io/redhat-user-workloads/jperezde-tenant/test-coverity/osh-cli-container-konflux-test-2@sha256:c4035bff42739b9fefc53574a0d0ff3aa04a4758999d0a930a571fecfb49bc6d --registry-config authfile.json 
✓ Pulled      sast_snyk_check_out.sarif                                                                                                                           92.7/92.7 kB 100.00%  101ms
  └─ sha256:e6b94aa3738a92fd8e92532e1d445f103e463d898dcef5a81c7007eeb58abf56
✓ Pulled      application/vnd.oci.image.manifest.v1+json                                                                                                            741/741  B 100.00%   81µs
  └─ sha256:c4035bff42739b9fefc53574a0d0ff3aa04a4758999d0a930a571fecfb49bc6d
Pulled [registry] quay.io/redhat-user-workloads/jperezde-tenant/test-coverity/osh-cli-container-konflux-test-2@sha256:c4035bff42739b9fefc53574a0d0ff3aa04a4758999d0a930a571fecfb49bc6d
Digest: sha256:c4035bff42739b9fefc53574a0d0ff3aa04a4758999d0a930a571fecfb49bc6d

% grep csdiff/ sast_snyk_check_out.sarif
                        "csdiff/v0": "590b97142a3743e2f9da8306af9c76ea2dd5ff8d"
                        "csdiff/v0": "590b97142a3743e2f9da8306af9c76ea2dd5ff8d"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2"
                        "csdiff/v0": "43d83de159411782c26f5ee322e843d474e0e662"
                        "csdiff/v0": "43d83de159411782c26f5ee322e843d474e0e662"
                        "csdiff/v0": "cd1b0cf57ce214aad5256237fd9608f395798d4d"
                        "csdiff/v0": "cd1b0cf57ce214aad5256237fd9608f395798d4d"
                        "csdiff/v0": "8ffe380c399394cd31318bae191b48edc7ab9a30"
                        "csdiff/v0": "e5902a94c4e13072c0b1dc1c125efdc4f86a1aaa"
                        "csdiff/v0": "8c4be3b0e43c0df08cafbe5e007f4537138d6424"
                        "csdiff/v0": "c4427807cc019b5e2266002d1f413d9c5485781c"
                        "csdiff/v0": "299e01d53746ef91c9db6d546629559ff257b186"
                        "csdiff/v0": "8f59b6bbbeb96b88e6ded4bfbe40d5f3f435fff9"
                        "csdiff/v0": "e949d747563df6f42e67a03e8dcb680fd3fd3ea2"
                        "csdiff/v0": "299360d3509b49ff70567e75f970d37cc79c4692"
                        "csdiff/v0": "759db9a2851f60a5e45eba2ce9ab5f51900f949d"
                        "csdiff/v0": "964f33c91777c5c9dfa1677a8ba10242527e2f3d"
                        "csdiff/v0": "66b1b6558c13df870dc1cdf962f42c68dcc59e30"
                        "csdiff/v0": "10464c70caaf9905d10f5018a5f45cfc243a5b62"
                        "csdiff/v0": "3656880aadb13d28ec5a5760d335e9cdd006a484"
                        "csdiff/v0": "3656880aadb13d28ec5a5760d335e9cdd006a484"
                        "csdiff/v0": "1dc7782c811b83befb57afa9c05b2a1fd21ba1f3"
                        "csdiff/v0": "1dc7782c811b83befb57afa9c05b2a1fd21ba1f3"
                        "csdiff/v0": "932cb88e28f436df7187e81f7d9baf72ca5527e6"
                        "csdiff/v0": "932cb88e28f436df7187e81f7d9baf72ca5527e6"
```

jperezdealgaba: Fixed: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-cwjmv

```
% oras pull quay.io/redhat-user-workloads/jperezde-tenant/test-coverity/osh-cli-container-konflux-test-2@sha256:c16b783aebf787a36e6a17c104afd0d44dad3a9059d52a95291d85b6092021d6
✓ Pulled      sast_snyk_check_out.sarif                                                                                                                                                                               173/173 kB 100.00%   48ms
  └─ sha256:83a59f11360df0dde3f50a4da3d2eba446a2c4008c8558819449b90bdce8aa04
✓ Pulled      application/vnd.oci.image.manifest.v1+json                                                                                                                                                              742/742  B 100.00%  501µs
  └─ sha256:c16b783aebf787a36e6a17c104afd0d44dad3a9059d52a95291d85b6092021d6
Pulled [registry] quay.io/redhat-user-workloads/jperezde-tenant/test-coverity/osh-cli-container-konflux-test-2@sha256:c16b783aebf787a36e6a17c104afd0d44dad3a9059d52a95291d85b6092021d6
Digest: sha256:c16b783aebf787a36e6a17c104afd0d44dad3a9059d52a95291d85b6092021d6


% grep csdiff/ sast_snyk_check_out.sarif
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2",
                        "csdiff/v1": "dc7f797833032fdc064541034c2e958e8e95c656"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2",
                        "csdiff/v1": "41fc798a50d6328b2ad099983989f8ad8ffc14c7"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2",
                        "csdiff/v1": "70472ff8830d244f66f940bce9def891a5dcfef7"
                        "csdiff/v0": "2b5f35e7e512ce4ab6cdb5a8fd3892af7f22fed2",
                        "csdiff/v1": "68fb90019efbb3d31917bd313f89e017fbc736bc"
                        "csdiff/v0": "590b97142a3743e2f9da8306af9c76ea2dd5ff8d",
                        "csdiff/v1": "93a8a1a32ed6e15b2f99d02151fb17b722d1216e"
                        "csdiff/v0": "590b97142a3743e2f9da8306af9c76ea2dd5ff8d",
                        "csdiff/v1": "4b7c975c13f16a2551d66b5ed1db165bcbd69d7a"
                        "csdiff/v0": "43d83de159411782c26f5ee322e843d474e0e662",
                        "csdiff/v1": "8999b23d63627bb8723d69deff6ac19f2c5a61ac"
                        "csdiff/v0": "43d83de159411782c26f5ee322e843d474e0e662",
                        "csdiff/v1": "4c017eb2db2e341b1b6dba7befa5dd8fecf65529"
                        "csdiff/v0": "8c4be3b0e43c0df08cafbe5e007f4537138d6424",
                        "csdiff/v1": "850e4bd843625cb718518af0cace0266398683b6"
                        "csdiff/v0": "8ffe380c399394cd31318bae191b48edc7ab9a30",
                        "csdiff/v1": "5b61f9beb51cf9717f1a168958a6907417d91eb3"
                        "csdiff/v0": "e5902a94c4e13072c0b1dc1c125efdc4f86a1aaa",
                        "csdiff/v1": "d9ad35b868453a84984c5fcaa2f3d8f84a005dee"
                        "csdiff/v0": "3656880aadb13d28ec5a5760d335e9cdd006a484",
                        "csdiff/v1": "28f65eab462d4354f7fe4de51c391432d14906c4"
```

jperezdealgaba:
@kdudka Just tested it and technically it is possible. It would be one small change to the `recipe.yaml`. Will wait for an official confirmation to make sure there is not any other reason why this is being handled like that at the moment

zregvart:
> @jperezdealgaba @lcarva @zregvart Why do we maintain two copies of the task description?

@kdudka because the descriptions often differ, if you do not need to modify the description don't specify it in the `recipe.yaml` and it will be copied from the base task.

kdudka:
@zregvart Thanks for confirmation!  That is exactly what the above patch does.

@jperezdealgaba Could you please apply it in this pull request?

jperezdealgaba:
> @jperezdealgaba Could you please apply it in this pull request?

The changes have been added

jperezdealgaba:
Just rebased the branch

jsztuka:
Now is your chance, you can merge it.

jperezdealgaba:
@jsztuka I tried to merge it but it seems that I don't have access.
P.S.: I rebased again

PR #1371, opened 2024-09-05
replace cosign image with appstudio-utils

tnevrlka:
redhat-appstudio/cosign image is getting deprecated due to migration to konflux-ci, replace with konflux-ci/appstudio-utils image

[STONEBLD-2717](https://issues.redhat.com/browse/STONEBLD-2717), related previous reverted attempt https://github.com/konflux-ci/build-definitions/pull/1344

This PR has been tested by running `./hack/test-builds.sh` locally and checking that pipelines pass and SBOMs are pushed to the Quay namespace.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -25,7 +25,6 @@
         "quay.io/redhat-appstudio/buildah",
         "quay.io/redhat-appstudio/hacbs-jvm-build-request-processor",
         "quay.io/redhat-appstudio/build-definitions-source-image-build-utils",
-        "quay.io/redhat-appstudio/cosign",
         "quay.io/redhat-appstudio/cachi2",
         "quay.io/redhat-appstudio/sbom-utility-scripts-image",
         "registry.access.redhat.com/rh-syft-tech-preview/syft-rhel9"
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -315,16 +315,10 @@ spec:
           yq -oj -i '.components += [ {"purl": "'$purl'", "type": "file", "name": "'$OCI_FILENAME'", "hashes": [{"alg": "SHA-256", "content": "'$OCI_ARTIFACT_DIGEST'"}], "externalReferences": [{"type": "distribution", "url": "'$OCI_SOURCE'"}]} ]' sbom-cyclonedx.json
         done
     - name: upload-sbom
-      image: quay.io/redhat-appstudio/cosign:v2.1.1@sha256:c883d6f8d39148f2cea71bff4622d196d89df3e510f36c140c097b932f0dd5d5
-      args:
-        - attach
-        - sbom
-        - --sbom
-        - sbom-cyclonedx.json
-        - --type
-        - cyclonedx
-        - $(params.IMAGE)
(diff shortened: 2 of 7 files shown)


ta-generator/golden/buildah/base.yaml
```diff
@@ -439,16 +439,21 @@ spec:
...
-      - sbom-cyclonedx.json
-      - --type
-      - cyclonedx
-      - $(params.IMAGE)
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
     workingDir: $(workspaces.source.path)
+    volumeMounts:
+      - mountPath: /mnt/trusted-ca
+        name: trusted-ca
+        readOnly: true
```
mkosiarc, line 447: Could you please explain this change? Why do we also need to include this trusted-ca volume when changing image?

tnevrlka: Thank you for raising this. The change doesn't seem to be needed, removed.

task/buildah/0.1/buildah.yaml
```diff
@@ -543,16 +543,10 @@ spec:
...
-      - sbom
-      - --sbom
-      - sbom-cyclonedx.json
-      - --type
-      - cyclonedx
-      - $(params.IMAGE)
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
     workingDir: $(workspaces.source.path)
+    script: |
+      cosign attach sbom --sbom sbom-cyclonedx.json --type cyclonedx "$(cat "$(results.IMAGE_REF.path)")"
```
chmeliik, line 549: I think that the 0.1 version of `buildah*` doesn't have `IMAGE_REF`. You could probably use `IMAGE_URL` and `IMAGE_DIGEST` instead (or just keep using `params.IMAGE` like it was before)

ta-generator/golden/buildah/base.yaml
```diff
@@ -439,16 +439,10 @@ spec:
...
-      - sbom
-      - --sbom
-      - sbom-cyclonedx.json
-      - --type
-      - cyclonedx
-      - $(params.IMAGE)
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
     workingDir: $(workspaces.source.path)
+    script: |
+      cosign attach sbom --sbom sbom-cyclonedx.json --type cyclonedx "$(cat "$(results.IMAGE_REF.path)")"
```
chmeliik, line 445: nitpick: this one doesn't have `IMAGE_REF` either. But it's a test file, you can update it by copying the current state of buildah 0.1 or buildah 0.2 over it

task/buildah/0.1/buildah.yaml
mkosiarc: do we need to update the 0.1 versions? Recently @mmorhun brought up that we would like to deprecate them and them remove them in time anyway.

chmeliik: Probably not. The presence of the quay.io/redhat-appstudio/cosign image in the repo may raise some questions, but a few rounds of justifying it to people should do the trick (if that's easier than making the change)

mmorhun: Let's not change stuff we are gonna remove soon

chmeliik (approved):

PR #1376, opened 2024-09-05
Ensure that we know all image mainfests added to the index

arewm:
Since we are passing --all to the buildah manifest add command, we enable all manifests referenced by an image index to be added to this index. If that happens, then we will only expose the digest of the image index and not all image manifests.

By post-processing the image index to identify all manifests, we can ensure that we have identified all image manifests referenced.

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -85,7 +85,6 @@ spec:
         exit 2
       fi
 
-      image_manifests=""
       buildah manifest create "$IMAGE"
       for i in $@
       do
@@ -107,7 +106,6 @@ spec:
           exit 0
         fi
         echo "Adding $TOADD"
-        image_manifests="${image_manifests} ${TOADD},"
         buildah manifest add $IMAGE "docker://$TOADD" --all
       done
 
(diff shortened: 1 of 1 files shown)


tnevrlka (approved):

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -128,6 +126,14 @@ spec:
           exit 1
       fi
 
+      INDEX_REPOSITORY="$(echo "$IMAGE" | cut -d@ -f1 | cut -d: -f1)"
+      MANIFEST_DIGESTS=$(buildah manifest inspect "$IMAGE" | jq -e ".manifests[].digest")
```
tkdchen, line 130: Missing `-r`, otherwise the resulting digests are quoted by double-quote.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -128,6 +126,14 @@ spec:
...
+      image_manifests=""
+      for i in $MANIFEST_DIGESTS
+      do
+        image_manifests="${image_manifests} ${INDEX_REPOSITORY}@${i},"
+      done
```
tkdchen, line 135: A minor suggestion is using use array and pipe to replace this for-loop. The expansion `${image_manifests:1:-1}` can be then replaced with array expansion and `tr` to construct a string with substrings separated by ",".

tkdchen (approved):

PR #756, opened 2024-01-22
Generate Table of Images, Tasks and Owners

jduimovich:
This pull request adds a script to generate a table of images used in tasks and the set of tasks referencing those images. The table also contains a list of tasks and owners. 
The output is stored in owner.md as a table. This is to support tracking of productization stage of tasks.   

A snippet of sample output 
## Images shared across Tasks
|  Num   |  Image   | Task(s)  | Count   | 
| -------- | -------- | -------  | ------- | 
| 1 | quay.io/openshift-pipeline/openshift-pipelines-cli-tkn:5.0 | tkn-bundle | 1 | 
| 2 | quay.io/redhat-appstudio/build-definitions-source-image-build-utils | source-build | 1 | 
| 3 | quay.io/redhat-appstudio/buildah:v1.31.0 | build-image-manifest, buildah, buildah-10gb, buildah-6gb, buildah-8gb, buildah-remote | 6 | 
|  ... |  ...| ... |
  
## Tasks and Owners 
| Num |  Task  | Owner  |
|  ------- |  ------- | ------- |
|     1	| build-image-manifest | Stonesoup Build Team | 
|     2	| buildah | Stonesoup Build Team | 
|     3	| buildah-10gb | Stonesoup Build Team |  
|  ... |  ...| ... |

PR #1381, opened 2024-09-06
feat: add a script to generate an asciidocs index

ralphbean:
This is here as an experiment at the moment to see what it would look like if we generated a table of available tasks for the docs at https://github.com/konflux-ci/konflux-ci for KONFLUX-1229.

### DIFF
--- a/hack/generate-index.sh
+++ b/hack/generate-index.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+# Build an asciidocs index of all tasks defined in this repo.
+
+set -e -o pipefail
+
+VCS_URL=https://github.com/konflux-ci/build-definitions
+
+# local dev build script
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+WORKDIR=$(mktemp -d --suffix "-$(basename "${BASH_SOURCE[0]}" .sh)")
+
+cat << EOF > "$WORKDIR/tasks.adoc"
+[cols="1,1,1,1,1,1"]
+|===
+|Name|Description|README|Usage|Migration|Troubleshooting
+EOF
+
+# Build tasks
+(
+cd "$SCRIPTDIR/.."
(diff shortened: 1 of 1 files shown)


ralphbean:
Example output at https://pr-129--konflux-docs.netlify.app/tasks/ rendered from https://github.com/konflux-ci/docs/pull/129

arewm:
Thanks for this!

* I am not sure if a table will be the best format. Maybe a structure like what we have for our glossary would be better? https://konflux-ci.dev/docs/glossary/
* Any kustomized tasks do not show a description. Should we make those work better?
* Should we not render any tasks that have the expiration configured like https://github.com/konflux-ci/build-definitions/blob/main/task/sbom-json-check/0.2/sbom-json-check.yaml#L7 ?
* Should we specify an input list (or an exclusion list) so that all tasks are not included or should we instead focus on getting the extra tasks out of the repository?
* We should probably start creating descriptions for our pipelines as well so that we don't have to manually curate a list like this: https://pr-126--konflux-docs.netlify.app/advanced-how-tos/installing/enabling-builds/#available-pipelines

zregvart:
Do we need more bash in this repository? We tackled a very similar issue in EC by using [golang + template](https://github.com/enterprise-contract/ec-cli/tree/main/internal/documentation/asciidoc/tekton).

ralphbean:
A comment from https://issues.redhat.com/browse/KONFLUX-1229:

> And, for the centralized documentation bit, perhaps we should lean on artifacthub for this.
> 
> See https://artifacthub.io/packages/search?repo=tekton-catalog-tasks&sort=relevance&page=1 for an example.
> 
> If we got all of build-definitions and release-service-catalog published there, we could link to pages like https://artifacthub.io/packages/search?repo=konflux-build-definitions&sort=relevance and https://artifacthub.io/packages/search?repo=konflux-release-catalog&sort=relevance as indices.

Maybe we shouldn't build our own list at all?

ralphbean:
I'm going to drop this and move the script over to https://github.com/konflux-ci/docs/pull/129

ralphbean:
Related: https://github.com/konflux-ci/build-definitions/pull/756

PR #1387, opened 2024-09-09
Use ephemeral namespaces for e2e tests

lcarva:


lcarva:
Switching to draft because it looks like SpaceRequests do not work the konflux-ci namespace because that's not a tenant namespace. The controller just completely ignores it.

PR #1398, opened 2024-09-11
fix: remove duplicate workdir volume

tnevrlka:
- The oci-copy-oci-ta task fails during configuration because of the duplicate volume, so local tests run
via `./hack/test-builds.sh` do not start
- Remove the duplicate workdir volume so tests can be run

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -55,8 +55,6 @@ spec:
       emptyDir: {}
     - name: workdir
       emptyDir: {}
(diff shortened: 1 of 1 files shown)


zregvart:
Seems to be fixed via #1093

tnevrlka:
@zregvart Can you confirm that `hack/build-tests.sh` works for you? 

I'm still getting `/.../build-definitions/hack/../task/oci-copy-oci-ta/0.1//oci-copy-oci-ta.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: multiple volumes with same name "workdir": spec.volumes[2].name`

tnevrlka:
Closing because of #1400

PR #1401, opened 2024-09-11
sast: Add sast-unicode-check task for testing

hanchuntao:

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
-    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml]"
+    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml, task/sast-unicode-check/0.1/sast-unicode-check.yaml]"
     pipelinesascode.tekton.dev/max-keep-runs: "5"
 spec:
   params:
@@ -49,6 +49,14 @@ spec:
           - fetch-repository
         taskRef:
           name: sast-snyk-check
+        workspaces:
+          - name: workspace
+            workspace: workspace
+      - name: sast-unicode-check
+        runAfter:
+          - build-container
+        taskRef:
+          name: sast-unicode-check
         workspaces:
           - name: workspace
             workspace: workspace
--- a/task/sast-unicode-check/0.1/README.md
+++ b/task/sast-unicode-check/0.1/README.md
@@ -1,0 +1,29 @@
+# sast-unicode-check task
+
+## Description:
+
+The sast-unicode-check task uses [find-unicode-control](https://github.com/siddhesh/find-unicode-control.git) tool to perform Static Application Security Testing (SAST) to look for non-printable unicode characters in all text files in a source tree.
+
+## Parameters:
+
+| name                         | description                                                                                                                                   | Default Value                                                                                   | Required |
+|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|----------|
+| FIND_UNICODE_CONTROL_GIT_URL | URL from repository to find unicode control.                                                                                                  | "https://github.com/siddhesh/find-unicode-control.git#c2accbfbba7553a8bc1ebd97089ae08ad8347e58" | No       |
+| FIND_UNICODE_CONTROL_ARGS    | arguments for find-unicode-control command.                                                                                                   | "-p bidi -v -d -t"                                                                              | No       |
+| KFP_GIT_URL                  | Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled.         | ""                                                                                              | No       |
+| PROJECT_NVR                  | Name-Version-Release (NVR) of the scanned project. It is used to find path exclusions (it is optional).                                       | ""                                                                                              | No       |
+| RECORD_EXCLUDED              | Whether to record the excluded findings (defaults to false). If `true`, the the excluded findings will be stored in `excluded-findings.json`. | "false"                                                                                         | No       |
+
+## Results:
+
+| name          | description                              |
+|---------------|------------------------------------------|
+| TEST_OUTPUT   | Tekton task test output.                 |
+
+## Source repository for image:
+
+https://github.com/konflux-ci/konflux-test
+
+## Additional links:
+
+* https://github.com/siddhesh/find-unicode-control.git
--- a/task/sast-unicode-check/0.1/sast-unicode-check.yaml
+++ b/task/sast-unicode-check/0.1/sast-unicode-check.yaml
@@ -1,0 +1,259 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "konflux"
+  name: sast-unicode-check
+spec:
+  description: >-
+    Scans source code for non-printable unicode characters in all text files.
+  results:
+    - description: Tekton task test output.
+      name: TEST_OUTPUT
+  params:
+    - name: image-url
+      type: string
+      description: Image URL.
+      default: ""
+    - name: FIND_UNICODE_CONTROL_GIT_URL
+      type: string
+      description: URL from repository to find unicode control.
+      default: "https://github.com/siddhesh/find-unicode-control.git#c2accbfbba7553a8bc1ebd97089ae08ad8347e58"
+    - name: FIND_UNICODE_CONTROL_ARGS
+      type: string
+      description: arguments for find-unicode-control command.
+      default: "-p bidi -v -d -t"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files.
+      # FIXME: Red Hat internal projects will default to https://gitlab.cee.redhat.com/osh/known-false-positives.git when KONFLUX-4530 is resolved
+      default: ""
+    - name: PROJECT_NVR
+      type: string
+      description: |
+        Name-Version-Release (NVR) of the scanned project.
+        It is used to find path exclusions (it is optional).
+      default: ""
+    - name: RECORD_EXCLUDED
+      type: string
+      description: |
+        Whether to record the excluded findings (defaults to false).
+        If `true`, the the excluded findings will be stored in `excluded-findings.json`.
+      default: "false"
+    - name: caTrustConfigMapName
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: trusted-ca
+    - name: caTrustConfigMapKey
+      type: string
+      description: The name of the key in the ConfigMap that contains the CA bundle data.
+      default: ca-bundle.crt
+  volumes:
+    - name: trusted-ca
+      configMap:
+        name: $(params.caTrustConfigMapName)
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        optional: true
+  steps:
+    - name: sast-unicode-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.6@sha256:5f298d8d990dfa82023e50029b71b08e19c3c9cedb181dfc4bc86c9ecad8700c
+      # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+      # the cluster will set imagePullPolicy to IfNotPresent
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
+        - mountPath: /mnt/trusted-ca
+          name: trusted-ca
+          readOnly: true
+      env:
+        - name: KFP_GIT_URL
+          value: $(params.KFP_GIT_URL)
+        - name: FIND_UNICODE_CONTROL_GIT_URL
+          value: $(params.FIND_UNICODE_CONTROL_GIT_URL)
+        - name: FIND_UNICODE_CONTROL_ARGS
+          value: $(params.FIND_UNICODE_CONTROL_ARGS)
+        - name: PROJECT_NVR
+          value: $(params.PROJECT_NVR)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+        - name: SOURCE_CODE_DIR
+          value: $(workspaces.workspace.path)
+      script: |
+        #!/usr/bin/env bash
+        set -exuo pipefail
+
+        # shellcheck source=/dev/null
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        SCAN_PROP=""
+
+        ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+        if [ -f "$ca_bundle" ]; then
+          echo "INFO: Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+          update-ca-trust
+        fi
+
+        # Clone the source code from upstream repo
+        GIT_URL=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $1}')
+        REV=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $2}')
+
+        # Clone find-unicode-control repository
+        if ! git clone "${GIT_URL}" find-unicode-control; then
+            echo "Failed to clone the repository: ${GIT_URL}" >&2
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 1
+        fi
+
+        if [[ -n "${REV}" ]]; then
+            if ! git -C ./find-unicode-control/ checkout "${REV}"; then
+                echo "Failed to checkout the repository: ${GIT_URL} to ${REV}" >&2
+                note="Task $(context.task.name) failed: For details, check Tekton task log."
+                ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+                echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+                exit 1
+            fi
+            SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}"
+        else
+            git_url_suffix=$(git -C ./find-unicode-control/ rev-parse HEAD)
+            SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}#${git_url_suffix}"
+        fi
+
+        # Find unicode control
+        FUC_EXIT_CODE=0
+        mapfile -t fuc_args <<< "${FIND_UNICODE_CONTROL_ARGS}"
+        LANG=en_US.utf8 ./find-unicode-control/find_unicode_control.py "${fuc_args[@]}" "${SOURCE_CODE_DIR}/source" \
+            >raw_sast_unicode_check_out.txt \
+            2>raw_sast_unicode_check_out.log \
+            || FUC_EXIT_CODE=$?
+        if [[ "${FUC_EXIT_CODE}" -ne 0 ]] && [[ "${FUC_EXIT_CODE}" -ne 1 ]]; then
+            echo "Failed to run find-unicode-control command" >&2
+            cat raw_sast_unicode_check_out.log
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 1
+        fi
+
+        # Translate the output format
+        if ! sed -i raw_sast_unicode_check_out.txt -E -e 's|(.*:[0-9]+)(.*)|\1: warning:\2|' -e 's|^|Error: UNICONTROL_WARNING:\n|'; then
+            echo "Error: failed to translate the unicontrol output format" >&2
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 1
+        fi
+
+        # Process all results as configured with CSGERP_OPTS
+        CSGERP_OPTS=(
+            --mode=json
+            --remove-duplicates
+            --embed-context=3
+            --set-scan-prop="${SCAN_PROP}"
+            --strip-path-prefix="${SOURCE_CODE_DIR}"/source/
+        )
+        # In order to generate csdiff/v1, we need to add the whole path of the source code as
+        # sast-unicode-check only provides an URI to embed the context
+        if ! csgrep "${CSGERP_OPTS[@]}" raw_sast_unicode_check_out.txt > processed_sast_unicode_check_out.json 2> processed_sast_unicode_check_out.err; then
+            echo "Error occurred while running csgrep with CSGERP_OPTS:"
+            cat processed_sast_unicode_check_out.err
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 1
+        fi
+
+        csgrep --mode=evtstat processed_sast_unicode_check_out.json
+
+        # Filter known false positives if KFP_GIT_URL is set
+        if [ -n "${KFP_GIT_URL}" ]; then
+            echo "Filtering false positives in results files using ${KFP_GIT_URL}..." >&2
+
+            # Build initial csfilter-kfp command
+            csfilter_kfp_cmd=(
+                csfilter-kfp
+                --verbose
+                --kfp-git-url="${KFP_GIT_URL}"
+            )
+
+            # Append --project-nvr option if PROJECT_NVR is set
+            if [[ -n "${PROJECT_NVR}" ]]; then
+                csfilter_kfp_cmd+=(--project-nvr="${PROJECT_NVR}")
+            fi
+
+            # Append --record-excluded option if RECORD_EXCLUDED is true
+            if [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                csfilter_kfp_cmd+=(--record-excluded="excluded-findings.json")
+            fi
+
+            if ! "${csfilter_kfp_cmd[@]}" processed_sast_unicode_check_out.json > sast_unicode_check_out.json 2> sast_unicode_check_out.error; then
+                echo "Failed to filter known false positives" >&2
+                cat sast_unicode_check_out.error
+                note="Task $(context.task.name) failed: For details, check Tekton task log."
+                ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+                echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+                exit 1
+            fi
+        else
+            echo "KFP_GIT_URL is not set. Skipping false positive filtering." >&2
+            mv processed_sast_unicode_check_out.json sast_unicode_check_out.json
+        fi
+
+        # Generate sarif report
+        csgrep --mode=sarif sast_unicode_check_out.json > sast_unicode_check_out.sarif
+        if [[ "${FUC_EXIT_CODE}" -eq 0 ]]; then
+            note="Task $(context.task.name) success: No finding was detected"
+            ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
+        elif [[ "${FUC_EXIT_CODE}" -eq 1 ]] && [[ ! -s  sast_unicode_check_out.sarif ]]; then
+            note="Task $(context.task.name) success: Some findings were detected, but filtered by known false positive"
+            ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
+        else
+            echo "sast-unicode-check test failed because of the following issues:"
+            cat sast_unicode_check_out.json
+            TEST_OUTPUT=
+            parse_test_output "$(context.task.name)" sarif sast_unicode_check_out.sarif  || true
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+        fi
+        echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee "$(results.TEST_OUTPUT.path)"
+    - name: upload
+      image: quay.io/konflux-ci/oras:latest@sha256:f4b891ee3038a5f13cd92ff4f473faad5601c2434d1c6b9bccdfc134d9d5f820
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      env:
+        - name: IMAGE_URL
+          value: $(params.image-url)
+      script: |
+        #!/usr/bin/env bash
+
+        if [ -z "${IMAGE_URL}" ]; then
+          echo 'No image-url param provided. Skipping upload.'
+          exit 0;
+        fi
+
+        UPLOAD_FILES="sast_unicode_check_out.sarif excluded-findings.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              continue
+            fi
+
+            if [ "${UPLOAD_FILES}" == "excluded-findings.json" ]; then
+                MEDIA_TYPE=application/json
+            else
+                MEDIA_TYPE=application/sarif+json
+            fi
+
+            echo "Selecting auth"
+            select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
+            echo "Attaching to ${IMAGE_URL}"
+            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        done
+  workspaces:
+  - name: workspace
--- a/task/sast-unicode-check/OWNERS
+++ b/task/sast-unicode-check/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs: https://go.k8s.io/owners
+approvers:
+  - integration-team
+reviewers:
+  - integration-team


.tekton/pull-request.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
-    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/yaml-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml]"
+    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/yaml-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml]"
```
tkdchen, line 8: Missing ','

tkdchen (commented):
Missing `task/task-name/OWNERS` as well.

.tekton/pull-request.yaml
hanchuntao: Updated, thanks.

hanchuntao:
@tkdchen I want to test this task, do you know how to run the CI? thanks!

hanchuntao:
> Missing `task/task-name/OWNERS` as well.

updated, added OWNERS, thanks

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
```diff
@@ -0,0 +1,232 @@
...
+      default: ""
+    - description: Image digest to report findings for.
+      name: image-digest
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - name: FIND_UNICODE_CONTROL_GIT_URL
+      type: string
+      description: URL from repository to find unicode control
+      default: "https://github.com/siddhesh/find-unicode-control.git"
```
kdudka, line 30: If the task runs are expected to be reproducible, we need to pin the scanner to specific revision.  This expectation is unmet with the existing `sast-snyk-check` task because it uses an external scanning service, which can be updated anytime.  Nevertheless, given the fact that we are cloning a personal git repository rather than an official project with a well-defined release procedure,  I would say we should pin the scanner to a specific (verified) revision.

@14rcole What is the preference of the Konflux team on this?

@hanchuntao We should also record the revision of the used scanner in scan properties via `csgrep --set-scan-prop` in a similar way as we record the revision of KFP: https://github.com/csutils/csdiff/pull/207/files

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
```diff
@@ -0,0 +1,232 @@
...
+      description: URL from repository to download known false positives files
+      default: "https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+    - default: ""
+      description: Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional)
+      name: PROJECT_NVR
+      type: string
+    - default: ""
+      description: File to store all excluded findings to (it is optional)
+      name: RECORD_EXCLUDED
+      type: string
```
kdudka, line 42: I think this should be a boolean.  I would prefer to make all the SAST scanning CI tasks use the same file name for this to avoid special-casing in the tools/services that will consume the data from Konflux.

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
```diff
@@ -0,0 +1,232 @@
...
+                    --verbose 
+                    --kfp-git-url="${KFP_GIT_URL}"
+                ）
+
+                if [[ -z "${PROJECT_NVR}" ]]; then
+                    cmd+=(--project-nvr="${PROJECT_NVR}")
+                if
+
+                if [[ -z "${RECORD_EXCLUDED}" ]]; then
+                    cmd+=(--record-excluded="${RECORD_EXCLUDED}")
```
kdudka, line 150: What is the point of writing the file if it is not uploaded?  The file will be removed as soon as the CI pipeline finishes.

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
```diff
@@ -0,0 +1,232 @@
...
+            CSGERP_OPTS=(
+                --mode=json
+                --event='error|warning'
+                --remove-duplicates
+                --embed-context=3
+            )
+            # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+            set -x
+            csgrep --mode=json --prepend-path-prefix="${SOURCE_CODE_DIR}"/ "${SCAN_RESULT}" | csgrep "${CSGERP_OPTS[@]}" | csgrep --mode=json --strip-path-prefix="${SOURCE_CODE_DIR}"/source/  > processed_sast_find_unicode_control_check_out.json
+            set +x
```
kdudka, line 178: I would not disable the XTRACE mode here.  If something fails in the subsequent commands, recording the commands that have been executed is pretty useful for debugging.

kdudka (changes requested):

kdudka:
@hanchuntao Please also include a reference to https://issues.redhat.com/browse/OSH-739 in the commit message.

kdudka (changes requested):
@hanchuntao Please rename the task to `sast-unicode-check` as stated in the summary of [OSH-739](https://issues.redhat.com/browse/OSH-739).

hanchuntao:
> @hanchuntao Please rename the task to `sast-unicode-check` as stated in the summary of [OSH-739](https://issues.redhat.com/browse/OSH-739).

updated

hanchuntao:
> @hanchuntao Please also include a reference to https://issues.redhat.com/browse/OSH-739 in the commit message.

updated

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
hanchuntao: yes, I have updated the code, set -x always

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
hanchuntao: Updated, the final result is:
```
"scan": {
[sast-unicode-check]         "find-unicode-control": "https://github.com/siddhesh/find-unicode-control.git c2accbfbba7553a8bc1ebd97089ae08ad8347e58",
[sast-unicode-check]         "known-false-positives-git-url": "https://gitlab.cee.redhat.com/osh/known-false-positives.git#297df9a6d1a62e0c16a974e8cf1f3770a3b0dc4c"
[sast-unicode-check]     },
```

.tekton/pull-request.yaml
```diff
@@ -59,6 +59,14 @@ spec:
         workspaces:
           - name: workspace
             workspace: workspace
+      - name: sast-unicode-check
+        runAfter:
+          - fetch-repository
```
kdudka, line 64: Will this work actually?  I think the `upload` step takes `IMAGE_URL` and `IMAGE_DIGEST`, which are not known until the `build-container` task finishes.

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,30 @@
+# sast-unicode-check task
+
+## Description:
+
+The sast-unicode-check task uses [find-unicode-control](https://github.com/siddhesh/find-unicode-control.git)  tool to perform Static Application Security Testing (SAST) to look for non-printable unicode characters in all text files in a source tree.
```
kdudka, line 5: Please remove the doubled space after `)`.

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Parameters:
+
+| name                         | description                                                                                      |
+|------------------------------|--------------------------------------------------------------------------------------------------|
+| FIND_UNICODE_CONTROL_GIT_URL | URL from repository to find unicode control                                                      |
+| KFP_GIT_URL                  | Link to the known-false-positives repository. If left blank, results won't be filtered           |
```
kdudka, line 12: Please change the description to `Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled.` to make it consistent with #1411.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,246 @@
...
+      default: ""
+    - description: Image digest to report findings for.
+      name: image-digest
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - name: FIND_UNICODE_CONTROL_GIT_URL
+      type: string
+      description: URL from repository to find unicode control
+      default: "https://github.com/siddhesh/find-unicode-control.git"
```
kdudka, line 30: This needs to be pinned to a specific (verified) revision of the tool for the scans to be reproducible.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,246 @@
...
+          echo "No ${UPLOAD_FILE} exists. Skipping upload."
+          exit 0;
+        fi
+
+        echo "Selecting auth"
+        select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
+        echo "Attaching to ${IMAGE_URL} via the OCI 1.1 Referrers API"
+        oras attach --no-tty --registry-config "$HOME/auth.json" --distribution-spec v1.1-referrers-api --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        echo "Attaching to ${IMAGE_URL} via the OCI 1.1 Referrers Tag"
+        oras attach --no-tty --registry-config "$HOME/auth.json" --distribution-spec v1.1-referrers-tag --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
kdudka, line 244: Please update this as done in #1435.

kdudka (changes requested):
@hanchuntao Please close the review threads that have been resolved.  I do not have permissions to do so.

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
hanchuntao: yes, it should be boolean, but the konflux not support boolean, it only support string, array, so I change the description, here use `true` or `false`

task/sast-find-unicode-control/0.1/sast-find-unicode-control.yaml
hanchuntao: Updated code, if the excluded file exist, also upload this file

.tekton/pull-request.yaml
hanchuntao: Updated, change it to 
```
        runAfter:
          - build-image-index
 ` ``

File: task/sast-unicode-check/0.1/sast-unicode-check.yaml (hunk 10 above)
Comment (hanchuntao) [reply task/sast-unicode-check/0.1/sast-unicode-check.yaml]: Updated

File: task/sast-unicode-check/0.1/sast-unicode-check.yaml (hunk 9 above)
Comment (hanchuntao) [reply task/sast-unicode-check/0.1/sast-unicode-check.yaml]: update, added a specific revision of this repo

File: task/sast-unicode-check/0.1/README.md (hunk 8 above)
Comment (hanchuntao) [reply task/sast-unicode-check/0.1/README.md]: Updated

File: task/sast-unicode-check/0.1/README.md [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+|------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| FIND_UNICODE_CONTROL_GIT_URL | URL from repository to find unicode control                                                                                                                                                            |
+| KFP_GIT_URL                  | Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled.                                                                  |
+| PROJECT_NVR                  | Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional)                                                                                                       |
+| RECORD_EXCLUDED              | Whether to record the excluded findings (it is optional, true or false). <br/>If `true`, the the excluded findings will be stored in `record_excluded.json`. <br/>It should be used with `PROJECT_NVR` |
```
kdudka, line 14: I fail to see why `RECORD_EXCLUDED` should be used with `PROJECT_NVR`.  The flag `RECORD_EXCLUDED` controls whether the excluded findings should be recorded or not whereas `PROJECT_NVR` enables using of `path-exclude.txt` files in KFP.  These are orthogonal options and any combination of them makes sense.

hanchuntao: yes, you are right. I misunderstood it, I have updated it

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,32 @@
...
+| RECORD_EXCLUDED              | Whether to record the excluded findings (it is optional, true or false). If `true`, the the excluded findings will be stored in `record_excluded.json`. |
+
+## Results:
+
+| name          | description                              |
+|---------------|------------------------------------------|
+| IMAGE_DIGEST  | Digest of the artifact just pushed       |
+| IMAGE_URL     | Repository where the artifact was pushed |
+| TEST_OUTPUT   | Tekton task test output.                 |
+| UPLOAD_FILE   | File to upload.                          |
```
kdudka, line 23: This feels wrong.  The only result on the other `sast-*-check` tasks is `TEST_OUTPUT`.  Let's stick to the exising interface unless there is a need to change it.

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Parameters:
+
+| name                         | description                                                                                                                                             |
+|------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------|
+| FIND_UNICODE_CONTROL_GIT_URL | URL from repository to find unicode control                                                                                                             |
+| KFP_GIT_URL                  | Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled.                   |
+| PROJECT_NVR                  | Name-Version-Release (NVR) of the scanned project, used to find path exclusions (it is optional)                                                        |
+| RECORD_EXCLUDED              | Whether to record the excluded findings (it is optional, true or false). If `true`, the the excluded findings will be stored in `record_excluded.json`. |
```
kdudka, line 14: `it is optional, true or false` -> `defaults to false`

Please also change the file name to `excluded-findings.json` for consistency.

task/sast-unicode-check/0.1/README.md
hanchuntao: Updated

task/sast-unicode-check/0.1/README.md
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+      description: URL from repository to find unicode control
+      default: "https://github.com/siddhesh/find-unicode-control.git#c2accbfbba7553a8bc1ebd97089ae08ad8347e58"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files
+      default: "https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+    - default: ""
+      description: |
+        Name-Version-Release (NVR) of the scanned project.
+        It used to find path exclusions (it is optional)
```
kdudka, line 38: `It used to` -> `It is used to`

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                return 1
+            else
+                echo "Message: Succeed to check usage of find-unicode-control" >&2
+            fi
+        }
+
+        # Find unicode control
+        unicontrol_scan()
+        {
+            FUC_EXIT_CODE=0
```
kdudka, line 120: Why do you assign 0 first and then conditionally overwrite it?

The code will be easier to follow if you (unconditionally) assign the value only once after the command is executed.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+            LANG=en_US.utf8 ./cloned/find_unicode_control.py -p bidi -v -d -t "${SOURCE_CODE_DIR}/source" \
+            >raw_sast_unicode_check_out.txt \
+            2>raw_sast_unicode_check_out.log \
```
kdudka, line 123: Please indent the lines to make it obvious that they are continuation of the previous command.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                --mode=json
+                --event='error|warning'
+                --remove-duplicates
+                --embed-context=3
+                --set-scan-prop="${SCAN_PROP}"
+            )
+            # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
+            csgrep --mode=json --prepend-path-prefix="${SOURCE_CODE_DIR}"/ raw_sast_unicode_check_out.txt \
+            | csgrep "${CSGERP_OPTS[@]}" \
+            | csgrep --mode=json --strip-path-prefix="${SOURCE_CODE_DIR}"/source/ \
```
kdudka, line 156: Please indent the lines to make it obvious that they are continuation of the previous command.

kdudka (changes requested):

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: yes, I have remove this setting

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+  params:
+    - description: Image URL.
+      name: image-url
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
+      default: ""
+    - description: Image digest to report findings for.
+      name: image-digest
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
```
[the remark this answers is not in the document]
hanchuntao: Removed this line.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+  description: >-
+    Scans source code for non-printable unicode characters in all text files.
+  results:
+    - description: Tekton task test output.
+      name: TEST_OUTPUT
+  params:
+    - description: Image URL.
+      name: image-url
+      type: string
+      # In a future 0.4 version of the task, drop the default to make this required
```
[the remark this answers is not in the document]
hanchuntao: Removed this line.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: But you should assign `FUC_EXIT_CODE=$?` unconditionally, in the same way as `status=$?` is assigned.  The use of `||` here is redundant and confusing.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
kdudka: If I see correctly, you removed only the comments but not the defaults that the comments were about?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+      type: string
+      default: ""
+    - name: FIND_UNICODE_CONTROL_GIT_URL
+      type: string
+      description: URL from repository to find unicode control
+      default: "https://github.com/siddhesh/find-unicode-control.git#c2accbfbba7553a8bc1ebd97089ae08ad8347e58"
+    - name: KFP_GIT_URL
+      type: string
+      description: URL from repository to download known false positives files
+      default: "https://gitlab.cee.redhat.com/osh/known-false-positives.git"
```
kdudka, line 32: Please change the default value of `KFP_GIT_URL` to an empty string for now with the following comment:
```
```

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                return 1
+            else
+                echo "Message: Succeed to check usage of find-unicode-control" >&2
+            fi
+        }
+
+        # Find unicode control
+        unicontrol_scan()
+        {
+            LANG=en_US.utf8 ./cloned/find_unicode_control.py -p bidi -v -d -t "${SOURCE_CODE_DIR}/source" \
```
kdudka, line 118: According to acceptance criteria of https://issues.redhat.com/browse/OSH-739 a parameter should be provided to override the default arguments passed to the scanner.  Let's introduce the `FIND_UNICODE_CONTROL_ARGS` task parameter for this and `-p bidi -v -d -t` as its default value.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                echo "Message: Succeed to translate the unicontrol output format" >&2
+            fi
+        }
+
+        # Process all results as configured with CSGERP_OPTS
+        process_all_results()
+        {
+            CSGERP_OPTS=(
+                --mode=json
+                --event='error|warning'
```
kdudka, line 148: The argument `--event='error|warning'` seems redundant because the `sed` command produces only `warning:` messages.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                --remove-duplicates
+                --embed-context=3
+                --set-scan-prop="${SCAN_PROP}"
+            )
+            # In order to generate csdiff/v1, we need to add the whole path of the source code as Snyk only provides an URI to embed the context
```
kdudka, line 153: How is this related to Snyk?  Seems like a copy-paste error.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+            csgrep --mode=json --prepend-path-prefix="${SOURCE_CODE_DIR}"/ raw_sast_unicode_check_out.txt \
```
kdudka, line 154: It should not be needed to run `csgrep --prepend-path-prefix=...` before `csgrep --embed-contex=...` if `csgrep` is invoked in the correct directory.  This way we can optimize out an extra JSON serialization/deserialization pass.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+            csgrep --mode=evtstat processed_sast_unicode_check_out.json
+        }
+
+        # Filter known false positive
+        filter_kfp()
+        {
+            # We check if the KFP_GIT_URL variable is set to apply the filters or not
+            if [[ -z "${KFP_GIT_URL}" ]]; then
+                mv processed_sast_unicode_check_out.json sast_unicode_check_out.json
+                echo "Error: KFP_GIT_URL variable not defined. False positives won't be filtered" >&2
```
kdudka, line 168: From end user perspective, `KFP_GIT_URL` is a task parameter.  That is how we document it.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                    echo "Message: Succeed to filter known false positives" >&2
+                fi
+
+                if [ -f "raw-excluded-findings.json" ]; then
+                    csgrep --mode=json --strip-path-prefix="${SOURCE_CODE_DIR}"/source/ raw-excluded-findings.json > excluded-findings.json
+                fi
+            fi
+        }
+
+        show_output()
```
kdudka, line 200: `show_output()` is a misleading name for a function that writes the SARIF file to be uploaded.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+        if [ -z "${IMAGE_URL}" ] || [ -z "${IMAGE_DIGEST}"  ]; then
+          echo 'No image-url or image-digest param provided. Skipping upload.'
+          exit 0;
+        fi
+
+        UPLOAD_FILES="sast_unicode_check_out.sarif excluded-findings.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              exit 0;
```
kdudka, line 250: As mentioned by @rhyw on another pull request.  The `exit 0;` should rather be `continue`.

kdudka (changes requested):

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+        #!/usr/bin/env bash
+        set -exuo pipefail
+        . /utils.sh
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        SCAN_PROP=""
+
+        # Installation of Red Hat certificates for cloning Red Hat internal repositories
+        curl -sS https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem > /etc/pki/ca-trust/source/anchors/2015-RH-IT-Root-CA.crt
+        curl -sS https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem > /etc/pki/ca-trust/source/anchors/2022-IT-Root-CA.pem
```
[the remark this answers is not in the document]
hanchuntao: Yes, for internal networks, this solution can work, for public clusters, cannot work
So I need to remove it, or add FIXME here?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+                return 1
+            else
+                echo "Message: Succeed to clone the repository: ${GIT_URL}" >&2
+            fi
+
+            if [[ -n "${REV}" ]]; then
+                # Get git url suffix
+                git_url_suffix=$(git -C ./cloned/ rev-parse HEAD)
+                if [[ "${git_url_suffix}" != "${REV}" ]]; then
+                    (cd cloned && git checkout --quiet "${REV}")
```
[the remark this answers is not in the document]
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,265 @@
...
+        update-ca-trust
+
+        # Clone the source code from upstream repo
+        clone_unicontrol()
+        {
+            GIT_URL=$(echo ${FIND_UNICODE_CONTROL_GIT_URL} | awk -F'#' '{print $1}')
+            REV=$(echo ${FIND_UNICODE_CONTROL_GIT_URL} | awk -F'#' '{print $2}')
+
+            # Clone find-unicode-control repository
+            git clone --depth=1 "${GIT_URL}" cloned >&2
```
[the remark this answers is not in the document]
hanchuntao: I use `cloned`, because I find some other tasks use it, if it is unusual, I will  use implicit name `find-unicode-control`.

 `git clone --depth=1 "${GIT_URL}" cloned >&2`, `"${GIT_URL}"` is only a git repository url, not container REV

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated, remove the judgment, only use the `KFP_GIT_URL ` exist logic

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Change it to `generate_sarif_report`

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
chmeliik: Other tasks handle this by mounting a CA cert from a configmap

https://github.com/konflux-ci/build-definitions/blob/dc50185eed271866a7fc4f80aaa8285fd86423af/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L178-L184

https://github.com/konflux-ci/build-definitions/blob/dc50185eed271866a7fc4f80aaa8285fd86423af/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L142-L147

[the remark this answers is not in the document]
hanchuntao: @chmeliik thank you for your response. So I can use the code directly? or create a configmap about `the ca-bundle.crt`?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: The semicolon after `continue` is redundant and confusing.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
kdudka: Thanks for the update!  I meant let's use `find-unicode-control` as the implicit name but it needs to be explicitly specified in the `git clone` command (so that it works for arbitrary URLs):
```
git clone "${GIT_URL}" find-unicode-control
```

The problem with `--depth=1` has not been fixed.  Currently this works only because c2accbfbba7553a8bc1ebd97089ae08ad8347e58 is the last commit in the default upstream branch.  Once upstream pushes one more commit to the default branch, the `git checkout` command will start to fail (with the default `FIND_UNICODE_CONTROL_GIT_URL`).  You can try this out by setting `https://github.com/siddhesh/find-unicode-control.git#8599acba673ca4e300c818315bc79edea3509f93` as `FIND_UNICODE_CONTROL_GIT_URL`.  It will break if `--depth=1` is used.

kdudka:
@hanchuntao Please close all review threads that have been resolved.  I do not have privileges to close them myself.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,274 @@
...
+    - name: RECORD_EXCLUDED
+      type: string
+      default: "false"
+      description: |
+        Whether to record the excluded findings (defaults to false).
+        If `true`, the the excluded findings will be stored in `excluded-findings.json`.
+  volumes:
+    - name: trusted-ca
+      configMap:
+        name: $(params.caTrustConfigMapName)
```
jperezdealgaba, line 53: I think these params missing from the `params` section and should be added

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
chmeliik: The existence of the configmap is up to cluster admins. The task just has to take the name of the configmap as a parameter

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: @jperezdealgaba Do we need any secrets for the `sast-unicode-check` actually?  Are they needed to configure the CA certificates?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: @hanchuntao Sorry, my previous suggestion was wrong.  I did not realize that the whole script runs with `set -e`.  So the `|| FUC_EXIT_CODE=$?` construction is actually needed.  Nevertheless, this also means that the `status=$?` lines will not work as expected (because they will never assign a non-zero exit code).

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,274 @@
...
+
+        # Clone the source code from upstream repo
+        clone_unicontrol()
+        {
+            GIT_URL=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $1}')
+            REV=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $2}')
+
+            # Clone find-unicode-control repository
+            git clone --depth=1 "${GIT_URL}" >&2
+            status=$?
```
kdudka, line 101: See my previous comment.  The `status=$?` will not work with `set -e` because non-zero exit code of the previous command will terminate the shell.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,274 @@
...
+                    git -C ./find-unicode-control/ checkout --quiet "${REV}"
+                fi
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}"
+            else
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}#${git_url_suffix}"
+            fi
+
+            # Check usage of find-unicode-control to confirm the tool can work
+            ./find-unicode-control/find_unicode_control.py --help >&2
+            status=$?
```
kdudka, line 122: See my previous comment.  The `status=$?` will not work with `set -e` because non-zero exit code of the previous command will terminate the shell.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,274 @@
...
+                echo "Message: Succeed to run find-unicode-control command" >&2
+            else
+                # Failed to run the command
+                echo "Error: failed to run find unicode control command" >&2
+                return 1
+            fi
+
+            # Translate the output format
+            sed -i raw_sast_unicode_check_out.txt -E -e 's|(.*:[0-9]+)(.*)|\1: warning:\2|' -e 's|^|Error: UNICONTROL_WARNING:\n|'
+            status=$?
```
kdudka, line 153: See my previous comment.  The `status=$?` will not work with `set -e` because non-zero exit code of the previous command will terminate the shell.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,274 @@
...
+
+            if [[ -n "${PROJECT_NVR}" ]] && [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                CMD+=(
+                    --project-nvr="${PROJECT_NVR}"
+                    --record-excluded=raw-excluded-findings.json
+                )
+            fi
+
+            "${CMD[@]}" processed_sast_unicode_check_out.json  > sast_unicode_check_out.json
+            status=$?
```
kdudka, line 196: See my previous comment.  The `status=$?` will not work with `set -e` because non-zero exit code of the previous command will terminate the shell.

kdudka (changes requested):

task/sast-unicode-check/0.1/sast-unicode-check.yaml
jperezdealgaba: Hey! I am not sure if I understand your question. 
They are not being added as secret but as a `configmap`. In the task provided by Konflux team, they are adding it as such: 
https://github.com/konflux-ci/build-definitions/blob/dc50185eed271866a7fc4f80aaa8285fd86423af/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml#L142-L147 . 

This configmap is later mounted as a volume in order to be able to use it and install the certificates.
Secrets can be used for the same purpose, but I wouldn't deviate from the Konflux implementation. 
Does that answer your question?

kdudka: @jperezdealgaba You said:
> The volumeMounts section is also missing to mount the secrets

... which made me wondering which **secrets** are we talking about in the context of `sast-unicode-check`.

As I understand it, no secrets should be needed to run this task.

jperezdealgaba: Right. Sorry for the typo. `secrets` = `configmap`.
Edited initial comment

kdudka: Got it, thank you for clarifying it!

jperezdealgaba:
I did the changes and I am testing them. The pipeline is right now full and the tasks are not being scheduled

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
hanchuntao: ok, thanks @chmeliik

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: @jperezdealgaba yes, you are right, I have added params: `caTrustConfigMapName` and `caTrustConfigMapKey`

task/sast-unicode-check/0.1/sast-unicode-check.yaml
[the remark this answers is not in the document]
hanchuntao: Yes, you are right, after test, need remove `--depth=1`
```
chhan@chhan-mac build-definitions % FIND_UNICODE_CONTROL_GIT_URL="https://github.com/siddhesh/find-unicode-control.git#8599acba673ca4e300c818315bc79edea3509f93"
chhan@chhan-mac build-definitions % GIT_URL=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $1}')
chhan@chhan-mac build-definitions % echo $GIT_URL
https://github.com/siddhesh/find-unicode-control.git
chhan@chhan-mac build-definitions % REV=$(echo "${FIND_UNICODE_CONTROL_GIT_URL}" | awk -F'#' '{print $2}')
chhan@chhan-mac build-definitions % echo $REV
8599acba673ca4e300c818315bc79edea3509f93
chhan@chhan-mac build-definitions % git clone --depth=1 "${GIT_URL}" find-unicode-control    
Cloning into 'find-unicode-control'...
remote: Enumerating objects: 14, done.
remote: Counting objects: 100% (14/14), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 14 (delta 0), reused 9 (delta 0), pack-reused 0 (from 0)
Receiving objects: 100% (14/14), 7.69 KiB | 1.54 MiB/s, done.
chhan@chhan-mac build-definitions % git -C ./find-unicode-control/ checkout --quiet "${REV}"
fatal: reference is not a tree: 8599acba673ca4e300c818315bc79edea3509f93
chhan@chhan-mac build-definitions % 


chhan@chhan-mac build-definitions % git clone "${GIT_URL}" find-unicode-control  
Cloning into 'find-unicode-control'...
remote: Enumerating objects: 64, done.
remote: Counting objects: 100% (64/64), done.
remote: Compressing objects: 100% (43/43), done.
remote: Total 64 (delta 29), reused 41 (delta 15), pack-reused 0 (from 0)
Receiving objects: 100% (64/64), 14.68 KiB | 224.00 KiB/s, done.
Resolving deltas: 100% (29/29), done.
chhan@chhan-mac build-definitions % git -C ./find-unicode-control/ checkout --quiet "${REV}"
chhan@chhan-mac build-definitions % 
```

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,30 @@
...
+| TEST_OUTPUT   | Tekton task test output.                 |
+
+## Source repository for image:
+
+https://github.com/konflux-ci/konflux-test
+
+## Additional links:
+
+* https://github.com/siddhesh/find-unicode-control.git
+* https://gitlab.cee.redhat.com/osh/known-false-positives.git
```
jperezdealgaba, line 30: Do we want to point a public resource to our internal GitLab?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+                    git -C ./find-unicode-control/ checkout --quiet "${REV}"
+                fi
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}"
+            else
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}#${git_url_suffix}"
+            fi
+
+            # Check usage of find-unicode-control to confirm the tool can work
+            USAGE_EXIT_CODE=0
+            ./find-unicode-control/find_unicode_control.py --help >&2 || USAGE_EXIT_CODE=$?
```
jperezdealgaba, line 130: Do we want to print this to the user? It seems there was a misconfiguration in the task logs

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+            csgrep --mode=evtstat processed_sast_unicode_check_out.json
+        }
+
+        # Filter known false positive
+        csfilter_kfp()
+        {
+            echo "Message: Filtering false positives in results files using ${KFP_GIT_URL}..." >&2
+
+            # Build initial csfilter-kfp command
+            csfilter_kfp_cmd="csfilter-kfp --verbose --kfp-git-url=${KFP_GIT_URL}"
```
jperezdealgaba, line 190: I think the reference to the image should change as `csfilter-kfp` the tool is not present

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+                echo "Error: failed to filter known false positives" >&2
+                return 1
+            fi
+
+            if [ -f "raw-excluded-findings.json" ]; then
+                csgrep --mode=json --strip-path-prefix="${SOURCE_CODE_DIR}"/source/ raw-excluded-findings.json > excluded-findings.json
+            fi
+        }
+
+        generate_sarif_report()
```
jperezdealgaba, line 216: The generation of the SARIF file is just one line. Do we want to call this function generate_sraif_report? The rest are parsing for the task output

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+        {
+            csgrep --mode=sarif sast_unicode_check_out.json > sast_unicode_check_out.sarif
+            if [[ "${FUC_EXIT_CODE}" -eq 0 ]]; then
+                note="Task $(context.task.name) success: No finding was detected"
+                ERROR_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
```
jperezdealgaba, line 221: Has this been tested? It seems there is an error in the logs:
`[sast-unicode-check] + note='Unexpected error: Script errored at command: return 1.'`

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: Yep, the first time I saw this in the CI log, I thought there was a bug in the CI task that it invoked the scanner with incorrect options.  Then I saw there is no `--version` option in the scanner and said so be it, let's use `--help` to print something.  The fact that @jperezdealgaba spotted the same issue suggests that the command should rather be removed.  It does not bring any value really.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
kdudka: To be honest, I am not sure if the use of these shell functions really improves anything.  They are all executed one by one in the same order as they are defined.  They do not use any local variables and the few occurrences of `return 1` can be replaced by `exit 1` without any observable change in behavior.  If we inline these functions, the whole script will be shorter with fewer nesting levels, and we can skip the discussion about suitable function names.

task/sast-unicode-check/0.1/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Parameters:
+
+| name                         | description                                                                                                                                   | Default Value                                                                                   | Required |
+|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|----------|
+| FIND_UNICODE_CONTROL_GIT_URL | URL from repository to find unicode control.                                                                                                  | "https://github.com/siddhesh/find-unicode-control.git#c2accbfbba7553a8bc1ebd97089ae08ad8347e58" | Yes      |
+| FIND_UNICODE_CONTROL_ARGS    | arguments for find-unicode-control command.                                                                                                   | "-p bidi -v -d -t"                                                                              | Yes      |
```
kdudka, line 12: I think that `Required` = `Yes` is meaningless for parameters with a default value.  In case the parameter is not provided, the default value will be used.  My understanding of `Required` is that the task will fail to start if the parameter is not provided.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+        if [ -z "${IMAGE_URL}" ] || [ -z "${IMAGE_DIGEST}"  ]; then
+          echo 'No image-url or image-digest param provided. Skipping upload.'
+          exit 0;
+        fi
+
+        UPLOAD_FILES="sast_unicode_check_out.sarif excluded-findings.json"
+        for UPLOAD_FILE in ${UPLOAD_FILES}; do
+            if [ ! -f "${UPLOAD_FILE}" ]; then
+              echo "No ${UPLOAD_FILE} exists. Skipping upload."
+              continue;
```
kdudka, line 266: Please remove the semicolon after `continue`.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}"
+            else
+                SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}#${git_url_suffix}"
+            fi
+
+            # Check usage of find-unicode-control to confirm the tool can work
+            USAGE_EXIT_CODE=0
+            ./find-unicode-control/find_unicode_control.py --help >&2 || USAGE_EXIT_CODE=$?
+            if [[ "${USAGE_EXIT_CODE}" -eq 0 ]]; then
+                echo "Message: Succeed to check usage of find-unicode-control" >&2
```
kdudka, line 132: I would drop **all** the code that prints `Message: Succeed to ...`.  The whole script is executed with `set -x`, so it is clear what is being executed and we expect all these commands to succeed under normal circumstances.  So it is sufficient just to handle the failures, in my opinion.

kdudka: I also find the handling of failures inconsistent.  The exit code of `git checkout --quiet ...` is not explicitly checked despite its success highly depends on the value of the provided `KFP_GIT_URL`.  In contrast to this, the failure of `sed -i` command is very unlikely and its explicit handling just complicates the shell script.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+                csfilter_kfp_cmd+=" --project-nvr="${PROJECT_NVR}""
+            fi
+
+            # Append --record-excluded option if RECORD_EXCLUDED is true
+            if [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                csfilter_kfp_cmd+=" --record-excluded=raw-excluded-findings.json"
+            fi
+
+            KFP_EXIT_CODE=0
+            ${csfilter_kfp_cmd[@]} processed_sast_unicode_check_out.json > sast_unicode_check_out.json || KFP_EXIT_CODE=$?
```
kdudka, line 203: The use of `${csfilter_kfp_cmd[@]}` suggests that `$csfilter_kfp_cmd` should be an array but it is rather constructed as a scalar.  Please have a look at the other open pull requests that use an array for this already.

kdudka (changes requested):

task/sast-unicode-check/0.1/README.md
hanchuntao: Removed it.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Removed it

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated, add some code when `csfilter-kfp` run failed

task/sast-unicode-check/0.1/README.md
hanchuntao: Changed `FIND_UNICODE_CONTROL_GIT_URL` and `FIND_UNICODE_CONTROL_ARGS` to `No`

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: the latest pipeline result: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/chhan/applications/unicode-control-test/pipelineruns/find-unicode-control-on-pull-request-l9xrn
but should add install pacakges and certificated steps
```
        # Only used for testing
        microdnf install dnf -y
        dnf install -y 'dnf-command(copr)'
        dnf copr enable copr.devel.redhat.com/@osh/csdiff -y
        dnf install -y csdiff git python3-file-magic

        # Installation of Red Hat certificates for cloning Red Hat internal repositories
        curl -sS https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem > /etc/pki/ca-trust/source/anchors/2015-RH-IT-Root-CA.crt
        curl -sS https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem > /etc/pki/ca-trust/source/anchors/2022-IT-Root-CA.pem
        update-ca-trust
```

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Updated, using array for this command

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,281 @@
...
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        optional: true
+  steps:
+    - name: sast-unicode-check
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.6@sha256:5f298d8d990dfa82023e50029b71b08e19c3c9cedb181dfc4bc86c9ecad8700c
+      # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+      # the cluster will set imagePullPolicy to IfNotPresent
+      workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      volumeMounts:
```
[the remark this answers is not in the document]
hanchuntao: Updated. should add the `volumeMounts `

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: OK, removed this command

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: I have removed these functions, only using command lines

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,269 @@
...
+    Scans source code for non-printable unicode characters in all text files.
+  results:
+    - description: Tekton task test output.
+      name: TEST_OUTPUT
+  params:
+    - name: image-url
+      type: string
+      description: Image URL.
+      default: ""
+    - name: image-digest
```
kdudka, line 21: Why does the task need the `image-digest` parameter?

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,269 @@
...
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 1
+        fi
+
+        if [[ -n "${REV}" ]]; then
+            # Get git url suffix
+            git_url_suffix=$(git -C ./find-unicode-control/ rev-parse HEAD)
+            if [[ "${git_url_suffix}" != "${REV}" ]]; then
```
kdudka, line 120: This condition is redundant.  If a specific revision is given, we should simply run `git checkout`.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,269 @@
...
+                    echo "Failed to checkout the repository: ${GIT_URL} to ${REV}" >&2
+                    note="Task $(context.task.name) failed: For details, check Tekton task log."
+                    ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+                    echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+                    exit 1
+                fi
+            fi
+            SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}"
+        else
+            SCAN_PROP="find-unicode-control-git-url:${FIND_UNICODE_CONTROL_GIT_URL}#${git_url_suffix}"
```
kdudka, line 131: The `${git_url_suffix}` is now unset in the `else` branch.  So this cannot work.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,269 @@
...
+                echo "Failed to filter known false positives" >&2
+                cat sast_unicode_check_out.error
+                note="Task $(context.task.name) failed: For details, check Tekton task log."
+                ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+                echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+                exit 1
+            fi
+
+            if [ -f "raw-excluded-findings.json" ]; then
+                csgrep --mode=json --strip-path-prefix="${SOURCE_CODE_DIR}"/source/ raw-excluded-findings.json > excluded-findings.json
```
kdudka, line 210: This step is redundant.  If it is really needed, there must be a bug elsewhere.  The path should be canonicalized before processing KFP.

kdudka (changes requested):

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Yes, It should be removed, I don't know why it exists; I think it exists in the original version

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: yes, updated code, add `git_url_suffix` in the else branch

task/sast-unicode-check/0.1/sast-unicode-check.yaml
hanchuntao: Yes, It was caused by changing the order of the steps, not need this step now. updated code

hanchuntao:
> @hanchuntao The code changes look good. I believe you tested the latest revision.

yes, I have tested the latest version and fixed all of the lint issues, now waiting for Konflux team review

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -0,0 +1,259 @@
...
+
+        # Find unicode control
+        FUC_EXIT_CODE=0
+        mapfile -t fuc_args <<< "${FIND_UNICODE_CONTROL_ARGS}"
+        LANG=en_US.utf8 ./find-unicode-control/find_unicode_control.py "${fuc_args[@]}" "${SOURCE_CODE_DIR}/source" \
+            >raw_sast_unicode_check_out.txt \
+            2>raw_sast_unicode_check_out.log \
+            || FUC_EXIT_CODE=$?
+        if [[ "${FUC_EXIT_CODE}" -ne 0 ]] && [[ "${FUC_EXIT_CODE}" -ne 1 ]]; then
+            echo "Failed to run find unicode control command" >&2
```
dheerajodha, line 137:
```suggestion
            echo "Failed to run find-unicode-control command" >&2
```
so it reads better

dheerajodha (commented):
just a small comment, rest LGTM

.tekton/pull-request.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
-    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml]"
+    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml, task/sast-unicode-check.yaml/0.1/sast-unicode-check.yaml]"
```
dheerajodha, line 8:
```suggestion
    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml, task/sast-unicode-check/0.1/sast-unicode-check.yaml]"
```
To fix the "pipelinerun start failure" as reported [here](https://github.com/konflux-ci/build-definitions/pull/1401/checks?check_run_id=31275525824)

.tekton/pull-request.yaml
```diff
@@ -52,6 +52,14 @@ spec:
         workspaces:
           - name: workspace
             workspace: workspace
+      - name: sast-unicode-check
+        runAfter:
+          - build-image-index
```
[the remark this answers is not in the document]
hanchuntao: use `fetch-repository` ?

hanchuntao:
```
Failure snippet:
task check-task-pipeline-repo-existence has the status "Failed":
Please contact Build team - #forum-stonesoup-build that the missing repos should be created in:
- https://quay.io/organization/redhat-appstudio-tekton-catalog
- https://quay.io/organization/konflux-ci
```

@dheerajodha, do you know how to fix this issue in the CI pipeline?

Josh-Everett (approved):

dheerajodha:
> @dheerajodha, do you know how to fix this issue in the CI pipeline?

The Konflux build PLR has passed now, @tkdchen has [opened a PR](https://github.com/konflux-ci/build-definitions/pull/1499) that fixes the 1 final failing PR check. It shouldn't take long for that PR to get merged, and then I'll merge this PR after that if you don't mind.

hanchuntao:
@dheerajodha Because all of the checks passed, what should we do next? Can somebody with write access merge the code?

kdudka:
@hanchuntao Are you sure you tested the changes before they were merged?

I do not think this could work.  See https://github.com/konflux-ci/build-definitions/pull/1524/commits/82046fa03133b5d5f1bdcaf33b0ba56881a18a07 for details.

hanchuntao:
> @hanchuntao Are you sure you tested the changes before they were merged?
> 
> I do not think this could work. See [82046fa](https://github.com/konflux-ci/build-definitions/commit/82046fa03133b5d5f1bdcaf33b0ba56881a18a07) for details.

I tested all of the changes. I don't know what happened, let me check it first, thanks

PR #1404, opened 2024-09-11
Tagging appstudio-utils with latest

joejstuart:
Using a floating tag so applications can use the latest utilities without updating the image tag.

https://issues.redhat.com/browse/EC-865

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -64,6 +64,17 @@ spec:
           - name: source
             workspace: workspace
 
+      - name: apply-additional-image-tags
+        params:
+          - name: IMAGE
+            value: $(tasks.build-container.results.IMAGE_REF)
+          - name: ADDITIONAL_TAGS
+            value: ["latest"]
+        runAfter:
+          - build-container
+        taskRef:
+          name: apply-tags
+
       - name: build-bundles
         params:
           - name: revision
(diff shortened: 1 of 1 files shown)


.tekton/push.yaml
```diff
@@ -64,6 +64,17 @@ spec:
           - name: source
             workspace: workspace
 
+      - name: apply-additional-image-tags
+        params:
+          - name: IMAGE
+            value: quay.io/konflux-ci/appstudio-utils:{{ revision }}
```
lcarva, line 70: It would be nice to use the result from the `build-container` task, something like `$(tasks.build-container.results.IMAGE_REF)`.

joejstuart: Oh, good thinking. Updated.

.tekton/push.yaml
```diff
@@ -64,6 +64,17 @@ spec:
...
+      - name: apply-additional-image-tags
+        params:
+          - name: IMAGE
+            value: $(tasks.build-container.results.IMAGE_REF)
+          - name: ADDITIONAL_TAGS
+            value: ["latest"]
+        runAfter:
+          - build-container
+        taskRef:
+          name: apply-tags
```
lcarva, line 76: Oh! You'll have to update the annotation ` pipelinesascode.tekton.dev/task` so this Task is available.

joejstuart: Good catch! I had no idea.

lcarva (approved):

PR #1420, opened 2024-09-13
CI: don't ignore build-and-push failures

chmeliik:
Fail the pipeline when build-and-push.sh fails, don't proceed to the e2e-tests (which would fail anyway)


chmeliik:
The pipeline failed on the build-bundles task as expected 
image

zregvart (approved):

psturc (approved):

PR #1435, opened 2024-09-17
De-duplicate attachment of OCI artifacts

ralphbean:
In PSSECAUT-605, a user reports that there is some funny behavior that is likely due to us invoking oras attach *twice* here.

There's no good reason to do it twice. Let's instead invoke it once, without specifying the distribution spec, and let oras choose for us.


dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

kasemAlem:
I've rebased and now all checks are passing @ralphbean

ralphbean:
Thanks @kasemAlem !

PR #1437, opened 2024-09-18
add sast-shellcheck task

rhyw:


rhyw:
Marking as draft for now since we'll need to comply with some sast convention first.

task/sast-shell-check/0.1/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+ShellCheck is a static analysis tool, gives warnings and suggestions for bash/sh shell scripts.
+
+## Params:
+
+| Name              | Description                                                                                  | Default Value | Required |
+|-------------------|----------------------------------------------------------------------------------------------|---------------|----------|
+| KFP_GIT_URL       | Git repository to download known false positives files from.                                 | ""            | No       |
+| PROJECT_NVR       | Name-Version-Release (NVR) of the scanned project, used to find path exclusions (optional).  | ""            | No       |
+| RECORD_EXCLUDED   | Whether to record the excluded findings (it is optional, true or false).<br/>If `true`, the the excluded findings will be stored in `record_excluded.json`.  | "false"            | No       |
```
kdudka, line 15: `it is optional, true or false` -> `defaults to false`

Please also change the file name to `excluded-findings.json` for consistency.

kdudka (changes requested):

task/sast-shell-check/0.1/README.md
rhyw: Addressed

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,175 @@
...
+                exit 1
+            else
+                mv "${OUTPUT_FILE}.filtered" "$OUTPUT_FILE"
+                echo "Filtered results saved back to $OUTPUT_FILE"
+            fi
+        else
+            echo "KFP_GIT_URL is not set. Skipping false positive filtering."
+        fi
+
+        csgrep --mode=json --event "$CSGREP_EVENT_FILTER" "$OUTPUT_FILE"
```
kdudka, line 127: This command will use `$OUTPUT_FILE` as an input file and write the filtered list of findings to `stdout`.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,175 @@
...
+          value: $(params.RECORD_EXCLUDED)
+        - name: CSGREP_EVENT_FILTER
+          value: $(params.CSGREP_EVENT_FILTER)
+      script: |
+        #!/usr/bin/env bash
+        set -x
+        . /utils.sh
+
+        # install dependencies
+        microdnf install -y csdiff
```
kdudka, line 64: This command can be removed if you update the reference to the `konflux-test` image, thanks to https://github.com/konflux-ci/konflux-test/pull/292

kdudka (changes requested):

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: addressed in https://github.com/konflux-ci/build-definitions/pull/1437/commits/a09977ba917bdd71155cce6bf670e76a40a23ba6

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: addressed in https://github.com/konflux-ci/build-definitions/pull/1437/commits/cfaec5bd26b5f17880f66a4221df5a0be3499423

rhyw:
@kdudka I added 2 commits for review. I'll squash them into one single commit before it's getting merged.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,177 @@
...
+      type: string
+      default: ""
+    - name: image-digest
+      description: Image digest to report findings for.
+      type: string
+      default: ""
+    - name: KFP_GIT_URL
+      description: git repository to download known false positives files from
+      type: string
+      default: ""
```
kdudka, line 28: Please insert the following comment so that we remember to override the default once Konflux is ready:
```
```

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -121,7 +121,12 @@ spec:
             echo "KFP_GIT_URL is not set. Skipping false positive filtering."
         fi
 
-        csgrep --mode=json --event "$CSGREP_EVENT_FILTER" "$OUTPUT_FILE"
+        if [[ "$IMP_FINDINGS_ONLY" == "true" ]]; then
+            CSGREP_EVENT_FILTER='\[SC(1020|1035|1054|1066|1068|1073|1080|1083|1099|1113|1115|1127|1128|1143|2043|2050|\
+                                      2055|2057|2066|2069|2071|2077|2078|2091|2092|2157|2171|2193|2194|2195|2215|2216|\
+                                      2218|2224|2225|2242|2256|2258|2261)\]$'
```
kdudka, line 127: @rhyw You cannot write it like this because the leading spaces would be included in the regular expression passed to the `--event` option of `csgrep`.

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: Fixed in https://github.com/konflux-ci/build-definitions/pull/1437/commits/1c1285c904b0a45d8a524b47dfabb58f53abeb00

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,179 @@
...
+            fi
+        else
+            echo "KFP_GIT_URL is not set. Skipping false positive filtering."
+        fi
+
+        if [[ "$IMP_FINDINGS_ONLY" == "true" ]]; then
+            CSGREP_EVENT_FILTER='\[SC(1020|1035|1054|1066|1068|1073|1080|1083|1099|1113|1115|1127|1128|1143|2043|2050|'
+            CSGREP_EVENT_FILTER+='2055|2057|2066|2069|2071|2077|2078|2091|2092|2157|2171|2193|2194|2195|2215|2216|'
+            CSGREP_EVENT_FILTER+='2218|2224|2225|2242|2256|2258|2261)\]$'
+            csgrep --mode=json --event "$CSGREP_EVENT_FILTER" "$OUTPUT_FILE" > "$OUTPUT_FILE"
```
kdudka, line 130: This command cannot be safely executed because shell may overwrite `$OUTPUT_FILE` before its original content is read by `csgrep`.  I would suggest to move the whole `if` condition before the first invocation of `csgrep` and simply append the `--event=...` argument to the `CSGREP_OPTS` array in the `then` branch.  The current `--event='error|warning'` argument needs to be moved to the `else` branch.  This will also optimize out unnecessary serialization/deserialization of the JSON data (and speed up KFP processing in case the full list of findings is huge).

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,179 @@
...
+            fi
+
+            # Execute the command and capture any errors
+            if ! "${csfilter_cmd[@]}" "$OUTPUT_FILE" > "${OUTPUT_FILE}.filtered" 2> "${OUTPUT_FILE}.error"; then
+                echo "Error occurred while filtering known false positives:"
+                cat "${OUTPUT_FILE}.error"
+                note="Task $(context.task.name) failed: For details, check Tekton task log."
+                ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+                echo "${ERROR_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+                rm -f "${OUTPUT_FILE}.filtered" "${OUTPUT_FILE}.error"
```
kdudka, line 116: I do not think you need to explicitly remove any files before exiting.  The whole overlay file system is going to be dropped when the container exits.

kdudka (changes requested):

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: addressed in https://github.com/konflux-ci/build-definitions/pull/1437/commits/ac33aad0a7fafa701249357663eda6d4aefda828

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: removed

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+        # generate all shellcheck result JSON files to $SC_RESULTS_DIR, which defaults to ./shellcheck-results/
+        /usr/share/csmock/scripts/run-shellcheck.sh "$SOURCE_CODE_DIR"
+
+        CSGREP_OPTS=(
+            --mode=json
+            --strip-path-prefix="$SOURCE_CODE_DIR"/
+            --remove-duplicates
+            --embed-context=3
+        )
```
hanchuntao, line 81: We should also record the revision of the used scanner in scan properties via csgrep --set-scan-prop in a similar way as we record the revision of KFP: https://github.com/csutils/csdiff/pull/207/files

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+          value: $(params.KFP_GIT_URL)
+        - name: PROJECT_NVR
+          value: $(params.PROJECT_NVR)
+        - name: RECORD_EXCLUDED
+          value: $(params.RECORD_EXCLUDED)
+        - name: IMP_FINDINGS_ONLY
+          value: $(params.IMP_FINDINGS_ONLY)
+      script: |
+        #!/usr/bin/env bash
+        set -x
```
jperezdealgaba, line 62: We are printing all logs coming from every command. Do we want to have this?
In previous comments I received, they stated that we should not print everything to only provide useful comments to the users.

kdudka: I dare to say Yes.  Unless this produces too much noise on glob expansion, loop iteration, etc., the output of `set -x` (a.k.a XTRACE) usually provides the most useful data for debugging because it shows us the real commands that were executed.  For example, `rpmbuild` also enables XTRACE by default in the generated build scripts.  The problem with printing the fancy looking messages such as `executing Snyk Client ... done` instead of XTRACE is that:

- They hide important details.
- They may become incorrect over time because the code that prints these messages needs to by manually kept in sync with other code changes (unlike XTRACE).

I think this is a matter of taste.  Different users/developers will have different opinion on this.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+            if [ -n "${PROJECT_NVR}" ]; then
+                csfilter_cmd+=" --project-nvr=${PROJECT_NVR}"
+            fi
+
+            if [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                csfilter_cmd+=" --record-excluded=excluded-findings.json"
+            fi
+
+            # Execute the command and capture any errors
+            if ! "${csfilter_cmd[@]}" "$OUTPUT_FILE" > "${OUTPUT_FILE}.filtered" 2> "${OUTPUT_FILE}.error"; then
```
kdudka, line 123: The use of `${csfilter_kfp_cmd[@]}` suggests that `$csfilter_kfp_cmd` should be an array but it is rather constructed as a scalar.  Please have a look at #1359 which uses an array for this already.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+        - name: IMP_FINDINGS_ONLY
+          value: $(params.IMP_FINDINGS_ONLY)
+      script: |
+        #!/usr/bin/env bash
+        set -x
+        . /utils.sh
+
+        # FIXME: once https://github.com/konflux-ci/konflux-test/pull/303 is available in latest image,
+        # work around below is no longer needed
+        microdnf install -y ShellCheck
```
hanchuntao, line 67: If this task needs `ShellCheck `,  I think need to install it in https://github.com/konflux-ci/konflux-test
such as: 
https://github.com/konflux-ci/konflux-test/pull/292
https://github.com/konflux-ci/konflux-test/pull/295

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+        microdnf download csmock-plugin-shellcheck && rpm -ivh --nodeps ./csmock-plugin-shellcheck-*.rpm && rm -f ./csmock-plugin-shellcheck-*.rpm
+
+        OUTPUT_FILE="shellcheck-results.json"
+        SOURCE_CODE_DIR=$(workspaces.workspace.path)/source
+
+        # generate all shellcheck result JSON files to $SC_RESULTS_DIR, which defaults to ./shellcheck-results/
+        /usr/share/csmock/scripts/run-shellcheck.sh "$SOURCE_CODE_DIR"
```
hanchuntao, line 74: I think the import item of this task is shellcheck, so should check the exit code of this line

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+            # build initial csfilter-kfp command
+            csfilter_cmd="csfilter-kfp --verbose --kfp-git-url=${KFP_GIT_URL}"
+
+            # Append --project-nvr option if PROJECT_NVR is set
+            if [ -n "${PROJECT_NVR}" ]; then
+                csfilter_cmd+=" --project-nvr=${PROJECT_NVR}"
+            fi
+
+            if [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                csfilter_cmd+=" --record-excluded=excluded-findings.json"
```
hanchuntao, line 119: I think we also need `--strip-path-prefix` for the `excluded-findings.json` file

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,184 @@
...
+            CSGREP_OPTS+=(
+                --event="$CSGREP_EVENT_FILTER"
+            )
+        else
+            CSGREP_OPTS+=(
+                --event="info|warning"
+            )
+        fi
+
+        csgrep "${CSGREP_OPTS[@]}" ./shellcheck-results/*.json > "$OUTPUT_FILE"
```
hanchuntao, line 96: Here we can also redirects errors to standard output, so when it is error, can show the error message

task/sast-shell-check/0.1/sast-shell-check.yaml
kdudka: It is not that easy.  `csmock-plugin-shellcheck` does not check the exit code of this script either.  The latest version does but it prints a warning only: https://github.com/csutils/csmock/commit/b2dda803e6f35174a46f987da602b609b1b26be3

The `shellcheck` processes are wrapped by `timeout`.  So if any of them is terminated, the overall exit code will be non-zero.  The success of the SAST scanning should rather be measured by presence of the produced JSON files.

task/sast-shell-check/0.1/sast-shell-check.yaml
kdudka: I thought that Tekton captures the standard error output, too.

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: Yeah the path is already stripped in the initial json file thus we're already good here.

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: the scan prop in the sarif is like this:
```
  "inlineExternalProperties": [
    {
      "externalizedProperties": {
        "csmock-plugin-shellcheck": "csmock-plugin-shellcheck-3.7.0-1.el9"
      }
    }
  ],
```

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: Addressed in https://github.com/konflux-ci/build-definitions/commit/9431a2fb607cee12a1f132f1756a13c293229c1c . it's a scalar(string) here thus a quote works.

task/sast-shell-check/0.1/sast-shell-check.yaml
kdudka: Once https://github.com/konflux-ci/konflux-test/pull/303 is resolved (and the resulting image used by this task), the version should be given by the base image.  Nevertheless, I think we should rather record the version of `ShellCheck` itself, not the helper script.

rhyw: Sure. I've updated it, now the sarif excerpt looks like this:
{code}
$ cat shellcheck-results.sarif | jq '.inlineExternalProperties'
[
  {
    "externalizedProperties": {
      "ShellCheck": "ShellCheck-0.8.0-2.el9"
    }
  }
]
{code}

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,187 @@
...
+            # predefined list of shellcheck important findings
+            CSGREP_EVENT_FILTER='\[SC(1020|1035|1054|1066|1068|1073|1080|1083|1099|1113|1115|1127|1128|1143|2043|2050|'
+            CSGREP_EVENT_FILTER+='2055|2057|2066|2069|2071|2077|2078|2091|2092|2157|2171|2193|2194|2195|2215|2216|'
+            CSGREP_EVENT_FILTER+='2218|2224|2225|2242|2256|2258|2261)\]$'
+            CSGREP_OPTS+=(
+                --event="$CSGREP_EVENT_FILTER"
+            )
+        else
+            CSGREP_OPTS+=(
+                --event="info|warning"
```
kdudka, line 95: This should have been `error|warning`.  We definitely do **not** want to suppress the findings with `error` events, which denote the highest severity.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,187 @@
...
+            if [ -n "${PROJECT_NVR}" ]; then
+                csfilter_cmd+=" --project-nvr=${PROJECT_NVR}"
+            fi
+
+            if [[ "${RECORD_EXCLUDED}" == "true" ]]; then
+                csfilter_cmd+=" --record-excluded=excluded-findings.json"
+            fi
+
+            # Execute the command and capture any errors
+            if ! "${csfilter_cmd}" "${OUTPUT_FILE}" > "${OUTPUT_FILE}.filtered" 2> "${OUTPUT_FILE}.error"; then
```
kdudka, line 126: I do not think this could work.  `"${csfilter_cmd}"` will be treated by bash as a name of a single command rather than a command followed by a list of arguments.

kdudka (changes requested):

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: indeed, array is a better option here than string. Updated in https://github.com/konflux-ci/build-definitions/pull/1437/commits/5bbfa2a1ca5dcbd83d6d94ca8866e54070c7a0ce

rhyw:
Commits squashed for brevity.

kdudka (approved):
The code changes look good.  Two remarks though:
- The CI is red because of two ShellCheck issues: https://github.com/konflux-ci/build-definitions/actions/runs/11229456532/job/31227287880?pr=1437 - I think we could use a ShellCheck directive and a small code change to make the CI green again.
- Out of OSH-210, this seems to be the only pull request that does not handle trust to the internal Red Hat CA, which is needed to access the internal KFP git repo.  I think we should add the optional `trusted-ca` volume based on config map and optionally install the CA bundle from there, as done in #1359.

rhyw:
It's weird the pipeline got stuck, I'm getting:
`User rhyw is not allowed to trigger CI via pull_request on this repo.`

Can anybody tell how can I re-trigger the CI?

hanchuntao:
> It's weird the pipeline got stuck, I'm getting: `User rhyw is not allowed to trigger CI via pull_request on this repo.`
> 
> Can anybody tell how can I re-trigger the CI?

you can ask the reviewers to approve your request to trigger the CI

hanchuntao:
Hello konflux  integration-team, could you help to review the code, thanks

rhyw:
Hey can anyone from the konflux team please give a review on this? Thanks!

rhyw:
Added the `sast-shell-check-oci-task`

rhyw:
@brunoapimentel @tkdchen @owtaylor @14rcole @Josh-Everett @jsztuka would you mind give a reivew when got time please?

This has been pending for some time.

rhyw:
Rebased / added missing OWNERS file for the oci-ta task.

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,215 @@
...
+        trap 'handle_error $(results.TEST_OUTPUT.path)' EXIT
+
+        ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+        if [ -f "$ca_bundle" ]; then
+          echo "INFO: Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+          update-ca-trust
+        fi
+
+        # FIXME: once https://github.com/konflux-ci/konflux-test/pull/303 is available in latest image,
```
dirgim, line 92: Hi, the PR in question is present in the latest [v1.4.8 release](https://quay.io/repository/redhat-appstudio/konflux-test?tab=tags) of the konflux-test image and it can be found [here](https://quay.io/repository/redhat-appstudio/konflux-test?tab=tags).

rhyw: Thanks. Fixed!

rhyw:
PR rebased. Also renamed param "PROJECT_NVR" to "PROJECT_NAME" to be consistent with other sast tasks, as per suggestion from @jperezdealgaba .

rhyw:
Let me know if a squash is needed. Thanks!

task/sast-shell-check/0.1/sast-shell-check.yaml
```diff
@@ -0,0 +1,211 @@
...
+            # build initial csfilter-kfp command
+            csfilter_kfp_cmd=(
+                csfilter-kfp
+                --verbose
+                --kfp-git-url="${KFP_GIT_URL}"
+            )
+
+            # Append --project-nvr option if PROJECT_NAME is set
+            if [ -n "${PROJECT_NAME}" ]; then
+                csfilter_kfp_cmd+=(--project-nvr="${PROJECT_NAME}")
```
jperezdealgaba, line 142: @rhyw What we did in other SAST tasks is that we have added a new parameter called `COMPONENT`, which is taken from the metadata labels: https://github.com/konflux-ci/build-definitions/blob/main/task/sast-snyk-check/0.3/sast-snyk-check.yaml#L98-L101 . That parameter is always used unless the user adds a `PROJECT_NAME` : https://github.com/konflux-ci/build-definitions/blob/main/task/sast-snyk-check/0.3/sast-snyk-check.yaml#L110-L113 . Would you mind updating the PR?

jperezdealgaba:
You you mind also updating the OWNERS file to add @kdudka as we do here; https://github.com/konflux-ci/build-definitions/pull/1538 ? 

Thanks!

task/sast-shell-check/0.1/sast-shell-check.yaml
rhyw: Updated, thanks!

14rcole (approved):
/lgtm

kdudka:
It seems that `ec-task-checks` failed again with:
```
- result: RPMS_DATA
        task: rpms-signature-scan
success: false
```
I am not able to access the task log myself: https://console.redhat.com/application-pipeline/ns/konflux-ci/pipelinerun/build-definitions-pull-request-6xts4/logs/ec-task-checks

kdudka:
@dirgim The CI is green but the merge has failed.  What should be the next steps?

dirgim:
There are some infra-related issues with e2e tests and don't seem related to this change, rerunning should work as far as I can see.

kdudka:
@dirgim Could you please move this forward?  This merge request has been waiting for merge since November 4th.

rhyw:
Thanks all for your time!

PR #1439, opened 2024-09-18
move ta-generator to task-generator directory

tnevrlka:
ta-generator and task-generator are currently on the same directory level.
Move ta-generator so it's a subdirectory of the task-generator leading to a better file organization in the repository.

- update .gitignore so it ignores vendor directories and diff
files used in trusted-artifacts task generation testing
- reflect the move of the trusted-artifacts
task generator to the task-generator directory
in the README
- add a README into every task-generator subdirectory


chmeliik (approved):
Optionally, we could make the task-generator directory a Go workspace to tie the two submodules together.

But TBH, I'm not entirely sure what problems Go workspaces solve, so we can leave it as is

tnevrlka:
Thank you for the review! 

I would like to avoid using Go workspace if possible. The [Go Modules reference](https://go.dev/ref/mod#go-work-file) mentions `It is generally inadvisable to commit go.work files into version control systems` and I'm not familiar with Go workspaces enough to understand if our use case warrants it.

Issue #1268, opened 2024-08-08
Use commit-timestamp in buildah task in pursuit of reproducible builds


ralphbean:
There are lots of challenges to producing bit-wise reproducible builds. One of them is to get all tools all the way through the build chain to use the same time. See https://reproducible-builds.org/docs/ and in particular https://reproducible-builds.org/docs/source-date-epoch/

* We should expose commit-timestamp as SOURCE_BUILD_EPOCH in all of our tasks. Not all tools respect that, but it is the standard environment variable that they should respect.
* Notably, buildah doesn't respect it, but it does accept a `--timestamp` argument that will help produce the same bits, assuming that nothing inside the containerfile being built is not sensitive to the time.
* We should also provide a `--build-arg SOURCE_BUILD_EPOCH=$SOURCE_BUILD_EPOCH` to each buildah build, so that _if_ a containerfile accepts that as an ARG, it can take advantage of that.

Doing these things won't give us all 100% bit-wise reproducible builds, but it will close some gaps in that direction.

pierDipi:
One issue we noticed that is related to not having reproducible builds is:

We have an operator repository that is aggregating other components images using the custom `latest` tag which eventually turns into a resolved SHA.

when a pipeline runs on push, a task after pushing or apply-tags could fail (for whatever reason), in that case, if we re-run the pipeline, the image revision (and therefore tag) is the same, but the SHA is different which ends up invalidating the "previous SHAs" in the operator repository.

So far, I see only a few solutions:
- run apply-tags task only if all other tasks succeeded
- append a timestamp to the "default revision tag" so that no SHA is "invalidated"

PR #1443, opened 2024-09-18
A new task to generate labels

ralphbean:
For https://issues.redhat.com/browse/KONFLUX-4274. Turns out people really do use those labels (see linked issues).

This can have a future connection with https://github.com/konflux-ci/build-definitions/issues/1268 once we do that. I.e., use the commit timestamp for the release label and build-date, rather than the real world date. But, that's for another day...

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -72,6 +72,11 @@ spec:
         hours, days, and weeks, respectively.
       type: string
       default: ""
+    - name: LABELS
+      description: Additional key=value labels that should be applied to the
+        image
+      type: array
+      default: []
     - name: PREFETCH_INPUT
       description: In case it is not empty, the prefetched content should
         be made available to the build.
@@ -226,7 +231,10 @@ spec:
     - name: build
       image: quay.io/konflux-ci/buildah-task:latest@sha256:860a239c5f25376a435a514ae6d53a5c75b1fa492461d17774e9b7cb32d1e275
       args:
+        - --build-args
         - $(params.BUILD_ARGS[*])
+        - --labels
+        - $(params.LABELS[*])
       workingDir: /var/workdir
       volumeMounts:
         - mountPath: /var/lib/containers
@@ -302,11 +310,34 @@ spec:
             sed -e '/^#/d' -e '/^\s*$/d' "${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}"
           )
         fi
-        # Append BUILD_ARGS
-        # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
-        # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
-        # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
-        build_args+=("$@")
+
+        LABELS=()
+        # Split `args` into two sets of arguments.
+        while [[ $# -gt 0 ]]; do
+            case $1 in
+                --build-args)
+                    shift
+                    # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
+                    # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
+                    # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
+                    while [[ $# -gt 0 && $1 != --* ]]; do
+                                                          build_args+=("$1")
+                                                                              shift
+            done
+                    ;;
+                --labels)
+                    shift
+                    while [[ $# -gt 0 && $1 != --* ]]; do
+                                                          LABELS+=("--label" "$1")
+                                                                                    shift
+            done
+                    ;;
+                *)
+                    echo "unexpected argument: $1" >&2
+                    exit 2
+                    ;;
+          esac
+        done
 
         BUILD_ARG_FLAGS=()
         for build_arg in "${build_args[@]}"; do
@@ -380,13 +411,16 @@ spec:
           VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume ${mount_point}:${YUM_REPOS_D_TARGET}"
         fi
 
-        LABELS=(
+        DEFAULT_LABELS=(
           "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
           "--label" "architecture=$(uname -m)"
           "--label" "vcs-type=git"
         )
-        [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
-        [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+        [ -n "$COMMIT_SHA" ] && DEFAULT_LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
+        [ -n "$IMAGE_EXPIRES_AFTER" ] && DEFAULT_LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+
+        # Concatenate defaults and explicit labels. If a label appears twice, the last one wins.
+        LABELS=("${DEFAULT_LABELS[@]}" "${LABELS[@]}")
 
         ACTIVATION_KEY_PATH="/activation-key"
         ENTITLEMENT_PATH="/entitlement"
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -103,6 +103,10 @@ spec:
     description: Whether to skip stages in Containerfile that seem unused by subsequent stages
     type: string
     default: "true"
+  - name: LABELS
+    description: Additional key=value labels that should be applied to the image
+    type: array
+    default: []
 
   results:
   - description: Digest of the image just built
@@ -179,7 +183,11 @@ spec:
     - name: COMMIT_SHA
       value: $(params.COMMIT_SHA)
     args:
+      - --build-args
       - $(params.BUILD_ARGS[*])
+      - --labels
+      - $(params.LABELS[*])
+
     script: |
       #!/bin/bash
       set -e
@@ -239,11 +247,28 @@ spec:
           sed -e '/^#/d' -e '/^\s*$/d' "${SOURCE_CODE_DIR}/${BUILD_ARGS_FILE}"
         )
       fi
-      # Append BUILD_ARGS
-      # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
-      # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
-      # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
-      build_args+=("$@")
+
+      LABELS=()
+      # Split `args` into two sets of arguments.
+      while [[ $# -gt 0 ]]; do
+          case $1 in
+              --build-args)
+                  shift
+                  # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
+                  # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
+                  # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
+                  while [[ $# -gt 0 && $1 != --* ]]; do build_args+=("$1"); shift; done
+                  ;;
+              --labels)
+                  shift
+                  while [[ $# -gt 0 && $1 != --* ]]; do LABELS+=("--label" "$1"); shift; done
+                  ;;
+              *)
+                  echo "unexpected argument: $1" >&2
+                  exit 2
+                  ;;
+          esac
+      done
 
       BUILD_ARG_FLAGS=()
       for build_arg in "${build_args[@]}"; do
@@ -317,13 +342,16 @@ spec:
         VOLUME_MOUNTS="${VOLUME_MOUNTS} --volume ${mount_point}:${YUM_REPOS_D_TARGET}"
       fi
 
-      LABELS=(
+      DEFAULT_LABELS=(
         "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
         "--label" "architecture=$(uname -m)"
         "--label" "vcs-type=git"
       )
-      [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
-      [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+      [ -n "$COMMIT_SHA" ] && DEFAULT_LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
+      [ -n "$IMAGE_EXPIRES_AFTER" ] && DEFAULT_LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+
+      # Concatenate defaults and explicit labels. If a label appears twice, the last one wins.
+      LABELS=("${DEFAULT_LABELS[@]}" "${LABELS[@]}")
 
       ACTIVATION_KEY_PATH="/activation-key"
       ENTITLEMENT_PATH="/entitlement"
--- a/task/generate-labels/0.1/generate-labels.yaml
+++ b/task/generate-labels/0.1/generate-labels.yaml
@@ -1,0 +1,90 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "konflux"
+  name: generate-labels
+spec:
+  description: |
+    Generate labels based on templates.
+
+    Usage may look like the following.
+
+    > - name: generate-labels
+    >   params:
+    >   - name: label-templates
+    >     value: ["release=$SOURCE_DATE_EPOCH", "build-date=$SOURCE_DATE"]
+
+    The following environment variables are defined for use in label-templates
+
+    * ACTUAL_DATE - a date time string containing the time this task runs, formatted +'%Y-%m-%dT%H:%M:%SZ'
+    * ACTUAL_DATE_EPOCH - the timestamp at the time this task runs
+    * SOURCE_DATE - a date time string containing the provided source timestamp, formatted +'%Y-%m-%dT%H:%M:%SZ'
+    * SOURCE_DATE_EPOCH - the timestamp provided as a param meant to represent the timestamp at which the source was last modified
+
+  params:
+  - name: label-templates
+    description: An array of templates that should be rendered and exposed as an array of labels
+    type: array
+  - name: source-date-epoch
+    description: A standardised environment variable for build tools to consume in order to produce reproducible output.
+    default: ""
+  results:
+    - name: labels
+      description: The rendered labels, rendered from the provided templates
+      type: array
+  steps:
+    - name: render
+      image: quay.io/konflux-ci/yq:latest@sha256:91a7fe73fd28bbfd93ebdcc68cdfd7087d4f2612c627ccd0d5f984190eb7dba6
+      env:
+      - name: SOURCE_DATE_EPOCH
+        value: "$(params.source-date-epoch)"
+      args:
+      - --templates
+      - $(params.label-templates[*])
+      script: |
+          #!/bin/bash
+
+          templates=()
+          while [[ $# -gt 0 ]]; do
+              case $1 in
+                  --templates)
+                      shift
+                      while [[ $# -gt 0 && $1 != --* ]]; do templates+=("$1"); shift; done
+                      ;;
+                  *)
+                      echo "unexpected argument: $1" >&2
+                      exit 2
+                      ;;
+              esac
+          done
+
+          ACTUAL_DATE_EPOCH=$(date -u +'%s')
+          ACTUAL_DATE=$(date -u --date=@"$ACTUAL_DATE_EPOCH" +'%Y-%m-%dT%H:%M:%SZ')
+
+          if [ "$SOURCE_DATE_EPOCH" == "" ]; then
+            SOURCE_DATE_EPOCH="$ACTUAL_DATE_EPOCH"
+          fi
+          SOURCE_DATE=$(date -u --date=@"$SOURCE_DATE_EPOCH" +'%Y-%m-%dT%H:%M:%SZ')
+
+          # Export, so that these are available to the subshell below
+          export ACTUAL_DATE
+          export ACTUAL_DATE_EPOCH
+          export SOURCE_DATE
+          export SOURCE_DATE_EPOCH
+
+          printf "[]" > result.json
+
+          export label
+          for template in "${templates[@]}"; do
+            echo "Processing template $template"
+            label=$(echo "$template" | envsubst)
+            echo "Yielding label $label"
+            yq -oj -i '. += [strenv(label)]' result.json
+          done
+
+          echo "Created the following labels:"
+          tee "$(results.labels.path)" < result.json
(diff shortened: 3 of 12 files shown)


task/buildah/0.2/buildah.yaml
```diff
@@ -312,13 +318,16 @@ spec:
...
+      EPOCH=$(date -u --date="$BUILD_DATE" +'%s')
       LABELS=(
-        "--label" "build-date=$(date -u +'%Y-%m-%dT%H:%M:%S')"
+        "--label" "build-date=$BUILD_DATE"
         "--label" "architecture=$(uname -m)"
         "--label" "vcs-type=git"
       )
       [ -n "$COMMIT_SHA" ] && LABELS+=("--label" "vcs-ref=$COMMIT_SHA")
       [ -n "$IMAGE_EXPIRES_AFTER" ] && LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+      [ "$EPOCH_AS_RELEASE_LABEL" == "true" ] && LABELS+=("--label" "release=$EPOCH")
```
chmeliik, line 330: Not too happy about more redhat-isms in the buildah task. But at least this one is disabled by default.

Have you considered the alternative suggested in slack, where the buildah task would receive a generic `LABELS` param and users would need to pass the result of a generate-labels task as the value of this param?

```yaml
    - name: build-container
      params:
      - name: LABELS
        value:
          - $(tasks.generate-labels.results.LABELS[*])
```

It would require bigger changes on the users' part (adding a task vs. just setting a param for the buildah task) but could avoid yet another hardcoded redhat-ism

chmeliik: As for how to approach a generate-labels task:

a) Hardcode redhat-specific labels, don't add it to the default pipeline, document in downstream docs how to use this task

b) Make it generic, document in downstream docs how to use it redhat-ically. Let's say something like this
```yaml
    - name: generate-labels
      params:
      - name: LABEL_TEMPLATES
        value:
          - "release=$epoch"
          - build-date=$build_date"
```
where the task defines a pre-determined set of placeholders and substitues them using whichever mechanism we'd like, e.g. `envsubst`

c) Let each team write their own

ralphbean: Hm, I think I did consider it but I discarded it quickly in an emotional decision based on frustration that this release label keeps rearing its head; I had hoped we could be done with it. I'm more calm this morning, and so can slow down to think.

Hm, I want to isolate redhat-isms, but I also don't want to put a burden on new redhat users (who already have to work through enabling prefetch, hermetic, source containers, etc.)

Consider this combination:

* Select option (2). Make a generic task `generate-labels` as you described.
* But then, with kustomize, add that task with a redhat-ical configuration, to the default pipelines available in the redhat instances. We already have a need for this downstream-customization creeping up in the ecosystem preflight check task, which probably should not appear in the upstream default pipeline definitions.
* We can still document it as a how-to, but new downstream users in redhat would already start with the release label set as expected.

Is that the lowest-friction path?

chmeliik: > But then, with kustomize, add that task with a redhat-ical configuration, to the default pipelines available in the redhat instances. We already have a need for this downstream-customization creeping up in the ecosystem preflight check task, which probably should not appear in the upstream default pipeline definitions.

As in, the customization would happen on the consumer side, down the stream from build-definitions? That sounds ideal. The consumer would have to build a modified Pipeline bundle, so it's not the most straightforward change if we want it fast, but it sounds like the right solution.

(And until then, we can document which labels to set in the downstream docs)

ralphbean: Heh, side note. I think we're going to need a gettext step image. I don't think we have one already sitting around.

chmeliik: Ah, true. Alternatively, python stdlib [string.Template](https://docs.python.org/3/library/string.html#template-strings) does something similar

```python
import collections
import os
import string
import sys

env = os.environ
# if you want envsubst behavior for nonexistent vars (replace with empty string)
# env = collections.defaultdict(str, env)

for arg in sys.argv[1:]:
    print(string.Template(arg).substitute(env))

```

ralphbean: I added `gettext` to our yq image in https://github.com/konflux-ci/yq-container/pull/48

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -20,6 +20,11 @@ spec:
       description: Name of secret which contains subscription activation key
       type: string
       default: activation-key
+    - name: ADDITIONAL_LABELS
+      description: Additional key=value labels that should be applied to the
+        image
+      type: string
+      default: '[]'
```
ralphbean, line 27: I tried for a while to make this of `type: array`, but I couldn't figure out how to get an array typed param into the task script as an env var. The only other array type param we have gets supplied as `args`, and I don't think we can do that with _two_ params.

Here, I'm treating it as a json string.

task/buildah/0.2/buildah.yaml
ralphbean: OK, I updated the buildah task here to accept a new `ADDITIONAL_LABELS` param, and added a `generate-labels` task.

The idea is that you can use the `commit-timestamp` result from the `git-clone` task as the `SOURCE_DATE_EPOCH` param for the `generate-labels` task.

Eventually, we should drop the `build-date` label that's built into the buildah task and start expecting people to use generate-labels.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
chmeliik: I do think an array of `--label` args would be nicer. This should probably work:

```diff
diff --git a/task/buildah/0.2/buildah.yaml b/task/buildah/0.2/buildah.yaml
index 0c992bbc..b2fe87b0 100644
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -177,10 +177,34 @@ spec:
     - name: COMMIT_SHA
       value: $(params.COMMIT_SHA)
     args:
+      - --build-args
       - $(params.BUILD_ARGS[*])
+      - --labels
+      - $(params.LABELS[*])
     script: |
       #!/bin/bash
       set -e
+
+      build_args=()
+      labels=()
+
+      while [ $# -gt 0 ]; do
+          case $1 in
+              --build-args)
+                  shift
+                  while [[ $# -gt 0 && $1 != --* ]]; do build_args+=("$1"); shift; done
+                  ;;
+              --labels)
+                  shift
+                  while [[ $# -gt 0 && $1 != --* ]]; do labels+=("$1"); shift; done
+                  ;;
+              *)
+                  echo "unexpected argument: $1" >&2
+                  exit 2
+                  ;;
+          esac
+      done
+
       ca_bundle=/mnt/trusted-ca/ca-bundle.crt
       if [ -f "$ca_bundle" ]; then
         echo "INFO: Using mounted CA bundle: $ca_bundle"
```

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+  params:
+  - name: label-templates
+    description: A json array of templates that should be rendered and exposed as a list of labels
+    type: string
+  - name: source-date-epoch
+    description: A standardised environment variable for build tools to consume in order to produce reproducible output.
+    default: ""
+  results:
+    - name: labels
+      description: The rendered labels, rendered from the provided templates
```
chmeliik, line 37: Can we make the `label-templates` param and `labels` result both arrays?

chmeliik: JSON strings are not great for composability, and special values could mess up the JSON

ralphbean: Yes - I didn't see how, but you found a way above. I'll do it.

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - name: LABEL_TEMPLATES
+        value: "$(params.label-templates)"
+      - name: SOURCE_DATE_EPOCH
+        value: "$(params.source-date-epoch)"
+      script: |
+          #!/bin/bash
+
+          ACTUAL_DATE_EPOCH=$(date -u +'%s')
+          # shellcheck disable=SC2034  # Unused variable used implicitly in envsubst
+          ACTUAL_DATE=$(date -u --date=@"$ACTUAL_DATE_EPOCH" +'%Y-%m-%dT%H:%M:%S')
```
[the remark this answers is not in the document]
ralphbean: Fixed in b21b304901599a70dc8d95725cfc83ccbf773ba3

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+    Usage may look like the following.
+
+    > - name: generate-labels
+    >   params:
+    >   - name: LABEL_TEMPLATES
+    >     value: '["release=$SOURCE_DATE_EPOCH", "build-date=$SOURCE_DATE"]'
+
+    The following environment variables are defined for use in LABEL_TEMPLATES
+
+    * ACTUAL_DATE - a date time string containing the time this task runs, formatted +'%Y-%m-%dT%H:%M:%S'
```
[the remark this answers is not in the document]
ralphbean: Done in 27453ec2d648f91f7f7662b165546655f7f339cd.

I noticed that the `build-date` label string (while UTC) does *not* have a trailing Z in the current `buildah` task.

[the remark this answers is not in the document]
chmeliik: D'oh, yeah. Hopefully the exact format doesn't matter that much (FWIW both are valid ISO-8601)

task/generate-labels/0.1/generate-labels.yaml
ralphbean: OK. Update in place now at 5da7b94f6ded2317d3fe1d9f755420900d067b80.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
ralphbean: OK. Update in place now at https://github.com/konflux-ci/build-definitions/commit/5da7b94f6ded2317d3fe1d9f755420900d067b80.

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+      env:
+      - name: LABEL_TEMPLATES
+        value: "$(params.label-templates)"
+      - name: SOURCE_DATE_EPOCH
+        value: "$(params.source-date-epoch)"
+      script: |
+          #!/bin/bash
+
+          ACTUAL_DATE_EPOCH=$(date -u +'%s')
+          ACTUAL_DATE=$(date -u --date=@"$ACTUAL_DATE_EPOCH" +'%Y-%m-%dT%H:%M:%SZ')
```
arewm, line 51: In the example, you set `build-date` so this will result in the label being added twice: https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml#L379

If time-based labels are being set here, they will be inconsistent with the `build-date` label which is added in the buildah task itself. I know you deferred any reproducibility work to the future, but this seems like it would be likely to cause confusion since both are labels being set.

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+    * ACTUAL_DATE_EPOCH - the timestamp at the time this task runs
+    * SOURCE_DATE - a date time string containing the provided source timestamp, formatted +'%Y-%m-%dT%H:%M:%SZ'
+    * SOURCE_DATE_EPOCH - the timestamp provided as a param meant to represent the timestamp at which the source was last modified
+
+  params:
+  - name: label-templates
+    description: A json array of templates that should be rendered and exposed as a list of labels
+    type: string
+  - name: source-date-epoch
+    description: A standardised environment variable for build tools to consume in order to produce reproducible output.
```
arewm, line 33: Is this the date epoch for the sources used or just the source of the epoch to use for calculations?

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+  name: generate-labels
+spec:
+  description: |
+    Generate labels based on templates.
+
+    Usage may look like the following.
+
+    > - name: generate-labels
+    >   params:
+    >   - name: LABEL_TEMPLATES
```
arewm, line 18: the parameter you have defined is `label-templates`.

task/generate-labels/0.1/generate-labels.yaml
ralphbean: Definitely.

Can we just remove `build-date` from the buildah task though? Others may be depending on it.

I guess, so long as enterprise contract doesn't require it - we could remove it and guide people to put this new task in place if and when they're ready.

task/generate-labels/0.1/generate-labels.yaml
ralphbean: Good catch!

ralphbean: Updated!

task/generate-labels/0.1/generate-labels.yaml
ralphbean: Both?

It is the source of the epoch to use for calculations.

It _can_ be the date epoch for the sources used and _should_ be, but nothing requires that it.

task/generate-labels/0.1/generate-labels.yaml
arewm: It can be possible for the removal to only affect the current users of the buildah tasks if we enable `generate-labels` by default and set it to define that build-date.

Or we could just check in the buildah task to see if the `build-date` label is being set externally and then not set it in the task.

We can consider fully removing the functionality in the next breaking change of the tasks.

task/generate-labels/0.1/generate-labels.yaml
arewm: You are saying that this date should be the same as the date of the git commit? Would this pattern only really make sense in a reproducible build scenario (and/or a hermetic build) as we wouldn't be pulling in dynamic content to differentiate from the content committed to git?

task/generate-labels/0.1/generate-labels.yaml
arewm: I think that would be good to do.

arewm:
There has been more interest in this type of functionality from users (i.e. to set/forward the git tags so that they can be used at release time). Should we add this to the default pipelines?

@mmorhun , have there been any criteria defined for when to add tasks to the default pipelines?

samdoran:
I would love the ability to pass in a list of `key=value` labels to the build tasks. That wolud save some steps compared to what I am doing today.

I'm not as clear on the value of the `generate-labels` task. It seems limited to set of four values that can be applied to arbitrary keys.

What would provide more utility and flexibility is a task parameter for an executable in the repository that returns a JSON string of `key=value` pairs.

arewm:
> I'm not as clear on the value of the generate-labels task. It seems limited to set of four values that can be applied to arbitrary keys.

The specific values mentioned are just those that it helps to _interpolate_. I think that it should be possible to modify this task parameters to add any variables that you want. For example, would something like this work?


```yaml
   - name: label-templates
     value: 
        - "release=$SOURCE_DATE_EPOCH"
        - "build-date=$SOURCE_DATE"
        - "short-commit=$(tasks.git-clone.results.short-commit)"
   - name: source-date-epoch
     value: '$(tasks.git-clone.results.commit-timestamp)
```

(well, `commit-timestamp` might actually provide a pretty date, not an epoch ... )

This task can be used to inject content into labels (from other tasks' results, for example). This does provide some indirection ... but if you don't need any of the label templates then you can also use the new parameter in the buildah task to directly add desired labels without having to use an ARG-based indirection.

task/generate-labels/0.1/generate-labels.yaml
ralphbean: Done in 100d2341.

chmeliik:
> What would provide more utility and flexibility is a task parameter for an executable in the repository that returns a JSON list of key=value pairs.

This is tricky, because executing arbitrary code before the build voids the "trust" of the build pipeline.

It would be doable in a trusted-artifacts pipeline - it would have to be a task that takes the source-code artifact as input but doesn't return a source-code artifact as output (any potential modifications would be local-only). And rather than running an executable from the source repo, the task could just execute any arbitrary script passed in.

Maybe should just add copy-and-pasteable snippets to the docs showing how to insert custom tasks into a trusted-artifacts pipeline in the least painful way

ralphbean:
Yeah, perhaps _if_ we ran that arbitrary executable with `unshare` and no net, that would be acceptable. But, in any case - I think it's out of scope of this change.

samdoran:
> This is tricky, because executing arbitrary code before the build voids the "trust" of the build pipeline. It would be doable in a trusted-artifacts pipeline...

I figured there would be some challenges with running arbitrary code. I'm doing this today in a custom task with a trusted artifact and it meets our needs while not compromising the integrity of build artifact.

Perhaps this is something best left to individual pipeline authors to implement as needed.

> Maybe we should just add copy-and-pasteable snippets to the docs showing how to insert custom tasks into a trusted-artifacts pipeline in the least painful way.

That's probably the best solution. It took me way more time than I would like to admit to get something so "simple" working, especially with the trusted artifacts model.

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+          export ACTUAL_DATE_EPOCH
+          export SOURCE_DATE
+          export SOURCE_DATE_EPOCH
+
+          printf "[]" > result.json
+
+          for template in $(echo "${LABEL_TEMPLATES}" | yq -r '.[]'); do
+            echo "Processing template $template"
+            label=$(echo "$template" | envsubst)
+            yq -oj -i '. += ["'"$label"'"]' result.json
```
chmeliik, line 69: I saw a nice trick for passing arguments to yq (which doesn't have `--arg` like jq): use env vars
```suggestion
            yq -oj -i '. += [env(label)]' result.json
```
This avoids the quoting issues (and doesn't break if a label value contains double-quotes)

task/buildah/0.2/buildah.yaml
```diff
@@ -317,13 +343,27 @@ spec:
...
+      [ -n "$COMMIT_SHA" ] && DEFAULT_LABELS+=("vcs-ref=$COMMIT_SHA")
+      [ -n "$IMAGE_EXPIRES_AFTER" ] && DEFAULT_LABELS+=("quay.expires-after=$IMAGE_EXPIRES_AFTER")
+
+      LABELS=()
+      for label in "${ADDITIONAL_LABELS[@]}"; do
+        echo "Adding label from param $label"
+        LABELS+=("--label" "$label")
+      done
+      for label in "${DEFAULT_LABELS[@]}"; do
+        if [[ ${ADDITIONAL_LABELS[*]} =~ ${label//=*} ]]; then
```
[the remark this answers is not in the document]
ralphbean: Done in 1498af6d97691b347e273e7a7e3c7bf68db67eee.

task/buildah/0.2/buildah.yaml
```diff
@@ -239,12 +247,30 @@ spec:
...
+                  echo "unexpected argument: $1" >&2
+                  exit 2
+                  ;;
+          esac
+      done
+
       # Append BUILD_ARGS
       # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
       # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
       # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
```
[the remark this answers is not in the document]
ralphbean: Done in 08ffb807c93b3385f1a11b3e73a459c7c953890f.

task/buildah/0.2/buildah.yaml
```diff
@@ -103,6 +103,10 @@ spec:
     description: Whether to skip stages in Containerfile that seem unused by subsequent stages
     type: string
     default: "true"
+  - name: ADDITIONAL_LABELS
+    description: Additional key=value labels that should be applied to the image
+    type: array
+    default: []
```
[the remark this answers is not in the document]
ralphbean: Done in 1498af6d97691b347e273e7a7e3c7bf68db67eee.

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,73 @@
...
+spec:
+  description: |
+    Generate labels based on templates.
+
+    Usage may look like the following.
+
+    > - name: generate-labels
+    >   params:
+    >   - name: label-templates
+    >     value: '["release=$SOURCE_DATE_EPOCH", "build-date=$SOURCE_DATE"]'
```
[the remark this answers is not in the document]
ralphbean: Oh, because I didn't understand the `args` approach to getting an array value usable inside the script. I see it now and will revise this.

task/generate-labels/0.1/generate-labels.yaml
ralphbean: Done in b702049870eac85d0645fad0899cef9159f56a1e.

task/generate-labels/0.1/generate-labels.yaml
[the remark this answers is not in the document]
ralphbean: Done in da44aaecbb569bf92aeacef02eebae2a2774a8f9.

task/buildah/0.2/buildah.yaml
```diff
@@ -360,6 +385,7 @@ spec:
       unshare -Uf $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w ${SOURCE_CODE_DIR}/$CONTEXT -- buildah build \
         $VOLUME_MOUNTS \
         "${BUILDAH_ARGS[@]}" \
+        "${DEFAULT_LABELS[@]}" \
```
chmeliik, line 388: Should this be the `LABELS` array? (And we should prepend `DEFAULT_LABELS` to it)

task/generate-labels/0.1/generate-labels.yaml
```diff
@@ -0,0 +1,89 @@
...
+          export SOURCE_DATE
+          export SOURCE_DATE_EPOCH
+
+          printf "[]" > result.json
+
+          for template in "${templates[@]}"; do
+            echo "Processing template $template"
+            export label=$(echo "$template" | envsubst)
+            echo "Yielding label $label"
+            yq -oj -i '. += [strenv(label)]' result.json
```
chmeliik, line 85: Good catch on using `strenv` rather than `env` :+1:

task/buildah/0.2/buildah.yaml
chmeliik: Or just pass both `"${DEFAULT_LABELS[@]}" "${LABELS[@]}" \` in this order

ralphbean: Revised in b978eda5, for clarity.

task/buildah/0.2/buildah.yaml
```diff
@@ -239,11 +247,28 @@ spec:
...
         )
       fi
-      # Append BUILD_ARGS
-      # Note: this may result in multiple --build-arg=KEY=value flags with the same KEY being
-      # passed to buildah. In that case, the *last* occurrence takes precedence. This is why
-      # we append BUILD_ARGS after the content of the BUILD_ARGS_FILE - they take precedence.
-      build_args+=("$@")
+
+      LABELS=()
+      # Split `args` into two sets of arguments.
```
mmorhun, line 252: It looks complicated.
Could we have additional labels in a dedicated env var fro example?

ralphbean: I started that way @mmorhun, but if we make the labels available to the script as an env var, then I don't think I can make the `param` an `array`.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -302,11 +310,34 @@ spec:
...
+                    while [[ $# -gt 0 && $1 != --* ]]; do
+                                                          build_args+=("$1")
+                                                                              shift
+            done
+                    ;;
+                --labels)
+                    shift
+                    while [[ $# -gt 0 && $1 != --* ]]; do
+                                                          LABELS+=("--label" "$1")
+                                                                                    shift
```
tkdchen, line 332: Indentation issue.

task/buildah/0.2/buildah.yaml
mmorhun: Will a delimiter in the env work? Like `label1=value1|label2=something else`

chmeliik (approved):
LGTM, please squash into a number of commits that you consider appropriate before merging

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
chmeliik: I think that's just the TA task generation doing some formatting ¯\\\_(ツ)_/¯

ralphbean:
Related: https://github.com/konflux-ci/build-definitions/pull/756

PR #1445, opened 2024-09-18
git-clone: Add configurable `short-commit` result

samdoran:
It's helpful to have a short commit SHA for use in other tasks, such as `apply-tags`. The length of `short-commit` is also configurable, giving greater flexibility to the pipeline author.

### DIFF
--- a/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
+++ b/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
@@ -66,6 +66,10 @@ spec:
       description: Revision to checkout. (branch, tag, sha, ref, etc...)
       type: string
       default: ""
+    - name: shortCommitLength
+      description: Length of short commit SHA
+      type: string
+      default: "7"
     - name: sparseCheckoutDirectories
       description: Define the directory patterns to match or exclude when
         performing a sparse checkout.
@@ -103,6 +107,9 @@ spec:
       description: The precise commit SHA that was fetched by this Task.
     - name: commit-timestamp
       description: The commit timestamp of the checkout
+    - name: short-commit
+      description: The commit SHA that was fetched by this Task limited to
+        params.shortCommitLength number of characters
     - name: url
       description: The precise URL that was fetched by this Task.
   volumes:
@@ -153,6 +160,8 @@ spec:
           value: $(params.submodules)
         - name: PARAM_DEPTH
           value: $(params.depth)
+        - name: PARAM_SHORT_COMMIT_LENGTH
+          value: $(params.shortCommitLength)
         - name: PARAM_SSL_VERIFY
           value: $(params.sslVerify)
         - name: PARAM_HTTP_PROXY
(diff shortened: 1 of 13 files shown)


chmeliik:
@samdoran @jbpratt (#1457) what are your use cases?

IMO returning this type of metadata should not be the clone task's responsibility (nor should `commit-timestamp`, but we got bullied into accepting that one because "it is needed to unblock RHEL AI")

For custom git metadata needs, it would be best if you just ran whatever script you want in your own pipeline. But writing Tekton tasks can be annoying, so I understand why you would want this as a task result...

chmeliik:
Though to be fair, the technical limitation of returning all the git metadata one could possibly want _should be_ fixable by enabling https://tekton.dev/docs/pipelines/tasks/#larger-results-using-sidecar-logs

And I suppose it would be nice for the clone task to provide all the results out of the box. It might not "do one thing well," but it *would* be convenient :smile:

task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
```diff
@@ -237,6 +243,7 @@ spec:
           exit "${EXIT_CODE}"
         fi
         printf "%s" "${RESULT_SHA}" >"$(results.commit.path)"
+        printf "%s" "${RESULT_SHA::$(params.shortCommitLength)}" > "$(results.short-commit.path)"
```
chmeliik, line 246: It would be better to use `git rev-parse --short=<length>` in the unlikely case where a prefix of the configured length is not unique within the repo

jbpratt:
Hey @chmeliik , here is the thread where I discussed my use case with others! https://redhat-internal.slack.com/archives/C031USXS2FJ/p1726817746562859?thread_ts=1726244072.482159&cid=C031USXS2FJ

task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
samdoran: Sounds good. I went back and forth on using `rev-parse` or expansion. That's a good argument for the former.

samdoran:
> what are your use cases?

My use case is similar to what @jbpratt described with a few extra caveats. The short commit hash is available in the release pipeline, so we use that to tag release image. It would be nice to have parity between test and release image tags.

I want to reference the Konflux build images (redhat-user-workloads) to run integration tests. For reasons, we use the image tag as a name for a clowder job invocation. That name is limited to 63 characters (also for reasons?), so `bonfire deploy` fails because the name field is too long.

I would prefer if this was avalible in PAC as a dynamic variable. But I don't have the ability to change that (since it requires a code change and release) whereas I can modify Tekton tasks.

This is kinda gross but works (parameter to `add-tags` task): `pr-{{ body.pull_request.number }}-$(tasks.clone-repository.results.short-commit)`

> For custom git metadata needs, it would be best if you just ran whatever script you want in your own pipeline. But writing Tekton tasks can be annoying, so I understand why you would want this as a task result

For another one of our applications, I wrote a custom task to run a script to get the desired version and that was _quite_ painful.

I don't mind making my own task and keeping it to myself, but it seems like having the short SHA would be beneficial to others. Especially since many of us are coming from app-interface where we have a long history of using the short commit SHA in many places.

samdoran:
I'll address the failing tests.

chmeliik (approved):
Ack, your use cases make sense. The prevailing sentiment seems to be that adding custom code to a Tekton pipeline is a PITA, which I can't really refute :sweat_smile: 

LGTM :+1:

chmeliik:
Just need to re-generate git-clone-oci-ta again

samdoran:
I fixed the failures related to `git-clone-oci-ta.`

chmeliik:
Ah, one more thing. @samdoran if you don't mind, could you squash the commits?

samdoran:
Sure thing!

gbenhaim (approved):

samdoran:
Thank you for merging and all the great reviews!

PR #1447, opened 2024-09-19
feat: Validate individual tasks in GA CI

tisutisu:
This PR adds a Github Actions CI to validate the task tests (if present inside the task directory) and few example tests for `git-clone` task for reference.

Story: https://issues.redhat.com/browse/STONEBLD-2910


task/git-clone/0.1/tests/run.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
lcarva, line 1: Have you considering using a Pipeline to encapsulate the Task execution and the required verification? The release folks took this approach, example: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/check-data-keys/tests/test-check-data-keys-fail-missing-cdn-key.yaml

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+            if [[ -z ${all_status} || -z ${reason} ]];then
+                echo "Could not find a created taskrun or pipelinerun in ${target_ns}"
+            fi
+
+            breakit=True
+            for status in ${all_status};do
+                [[ ${status} == *ERROR || ${reason} == *Fail* || ${reason} == Couldnt* ]] && show_failure ${testname} ${target_ns}
+
+                if [[ ${status} != True ]];then
+                    breakit=
```
mmorhun, line 254: Should we rather have false here?

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+    done
+
+    # I would refactor this to a function but bash limitation is too great, really need a rewrite the sooner
+    # the uglness to pass a hashmap to a function https://stackoverflow.com/a/17557904/145125
+    local cnt=0
+    local all_status=''
+    local reason=''
+    local maxloop=60 # 10 minutes max
+
+    set +x
```
mmorhun, line 213: Why do we need `set +x` here?

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+    for task in ${all_tasks}; do
+        # Skip if it is kustomize related yaml files
+        if [[ ${task} == *"kustomization.yaml" || ${task} == *"patch.yaml" || ${task} == *"recipe.yaml" ]]; then
+                continue
+        fi
+        tektor validate ${task}
+    done
+}
+
+function test_resource_creation() {
```
mmorhun, line 124: The function name and what id does differs

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+}
+
+function detect_new_changed_resources() {
+    # detect for changes in tests dir of the task
+    echo ${CHANGED_FILES} |grep -o 'task/[^\/]*/[^\/]*/tests/[^/]*'|xargs -I {} dirname {}|sed 's/\(tests\).*/\1/g'
+    # detect for changes in the task manifest
+    echo ${CHANGED_FILES} |grep -o 'task/[^\/]*/[^\/]*/*[^/]*.yaml'|xargs -I {} dirname {}|awk '{print $1"/tests"}'
+}
+
+function get_new_changed_tasks() {
```
mmorhun, line 39: function name is not clear

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+  # TODO(adrcunha): Get rid of this workaround.
+  echo -n "$1"> ${TEST_RESULT_FILE}
+}
+
+function detect_changed_e2e_test() {
+    # check if any file from test/ directory is changed
+    echo ${CHANGED_FILES} |grep "test/" || true
+}
+
+function detect_new_changed_resources() {
```
mmorhun, line 32: function name is unclear

test/run-test.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
mmorhun, line 1: Isn't it duplicates `hack/run-test.sh`?

mmorhun (commented):
Can we check task results / logs / created images, etc. ?

README.md
```diff
@@ -132,6 +132,33 @@ Specify the Quay repository using the `QUAY_NAMESPACE` environment variable in t
...
+CI also validates the task yamls using [tektor](https://github.com/lcarva/tektor)
+
+### Running Task tests locally
+
+Prerequisites: Deploy [upstream konflux](https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#bootstrapping-the-cluster) locally.
+
+A helper script called `run-test.sh` is present inside `test` directory to help the developer run the test. Just specify resource type as first argument, resource name as second argument and the version as third argument i.e:
+
+```
+./test/run-test.sh task git-clone 0.1
```
chmeliik, line 157: > type as first argument, resource name as second argument and the version as third argument

This seems unnecessary, passing the actual directory as the argument would work much better with tab completion :)

README.md
```diff
@@ -132,6 +132,33 @@ Specify the Quay repository using the `QUAY_NAMESPACE` environment variable in t
...
+If you have a directory called `tests/` inside the task directory, CI will create a random `Namespace`, apply the task first in the namespace and then all the test yaml files (containing TaskRun or PipelineRun) inside the `tests` directory.
+
+We can use the functionality of running scripts before applying the tested task or the other yaml files. There are two different scripts that are automatically applied if present. These are applied using the `source` bash script:
+
+- `pre-apply-task-hook.sh`: Script to run before applying the task, it can include `kubectl` commands to apply any kubernetes resources in the cluster
+- `pre-apply-taskrun-hook.sh`: Script to run before applying the task run or pipeline run yaml files, it can also include `kubectl` commands to apply any kubernetes resources in the cluster
+
+For example refer: Git Clone task [tests](./task/git-clone/0.1/tests/run.yaml)
+
+CI also validates the task yamls using [tektor](https://github.com/lcarva/tektor)
```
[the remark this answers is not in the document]
lcarva: Filed https://issues.redhat.com/browse/EC-1019

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+function validate_task_yaml_using_tektor() {
+    all_tasks="$*"
+    for task in ${all_tasks}; do
+        # Skip if it is kustomize related yaml files
+        if [[ ${task} == *"kustomization.yaml" || ${task} == *"patch.yaml" || ${task} == *"recipe.yaml" ]]; then
+                continue
+        fi
+        tektor validate ${task}
+    done
+}
```
chmeliik, line 122: Let's not mix linting and tests, this should be a separate GH workflow IMO

.github/workflows/run-task-tests.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Get all changed files in the PR from task or test directories
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          # Avoid using single or double quotes for multiline patterns
+          files: |
+            task/**
+            test/**
+
+      - name: Free Disk Space
```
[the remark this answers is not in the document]
tisutisu: took it from [here](https://github.com/konflux-ci/konflux-ci/blob/f6d9d0fe8f34199eb118febcbf7f7944ae7772a9/.github/workflows/sanity-test.yml#L33), don't aware of the exact reason, but I think might be helping to create more disk space in the github action env. Don't find any issue to keep it our CI as well.

task/git-clone/0.1/tests/run.yaml
tisutisu: @lcarva could you please help with more context around where these tests are running what approach they are following... any doc/link ?

hack/run-test.sh
```diff
@@ -0,0 +1,59 @@
...
+resourcedir=${RESOURCE}/${NAME}/${VERSION}
+
+if [[ ! -d ${resourcedir}/tests ]];then
+    echo "No 'tests' directory is located in ${resourcedir}"
+    exit 1
+fi
+
+tns=${NAME}-${VERSION//./-}
+
+# Delete the tns if already exists
```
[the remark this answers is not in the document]
tisutisu: updated now.

hack/run-test.sh
```diff
@@ -0,0 +1,59 @@
...
+${KUBECTL_CMD} delete ns ${tns} >/dev/null 2>/dev/null || :
```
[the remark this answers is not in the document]
tisutisu: updated now.

hack/run-test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+${KUBECTL_CMD} create namespace ${tns}
+
+#create appstudio-pipeline SA in the test namespace if not already
+if ! ${KUBECTL_CMD} get sa appstudio-pipeline -n ${tns} | grep 'appstudio-pipeline'; then
+   $KUBECTL_CMD create sa appstudio-pipeline -n ${tns}
+fi
+
+test_resource_creation ${RESOURCE}/${NAME}/${VERSION}/tests
```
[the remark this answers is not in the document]
tisutisu: some left overs,  forgot to remove this file, it got moved to `test/run-test.sh`

task/git-clone/0.1/tests/pre-apply-task-hook.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "example of pre-apply-task-hook, this script can be used to run kubernetes commands"
```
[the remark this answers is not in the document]
tisutisu: since we had only one example test for now, I think this might help people to understand how the `pre-apply-task-hook.sh` script might look like, it is kind of placeholder for now, I guess we can remove it in future when we have some tests which will contain `pre-apply-task-hook.sh`. WDYT?

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+  # script, even if the tests pass.
+  # We store the real test result to return it later, ignoring any teardown
+  # failure in kubetest.
+  # TODO(adrcunha): Get rid of this workaround.
+  echo -n "$1"> ${TEST_RESULT_FILE}
+}
+
+function detect_changed_e2e_test() {
+    # check if any file from test/ directory is changed
+    echo ${CHANGED_FILES} |grep "test/" || true
```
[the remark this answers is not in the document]
tisutisu: updated

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+    ${KUBECTL_CMD} get --ignore-not-found=true -n ${tns} pipeline -o yaml
+    echo "--- StepAction Dump"
+    ${KUBECTL_CMD} get --ignore-not-found=true -n ${tns} stepaction -o yaml
+    echo "--- Container Logs"
+    for pod in $(${KUBECTL_CMD} get pod -o name -n ${tns}); do
+        echo "----POD_NAME: ${pod}---"
+        ${KUBECTL_CMD} logs --all-containers -n ${tns} ${pod} || true
+    done
+    exit 1
+
```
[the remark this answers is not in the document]
tisutisu: updated

test/common.sh
```diff
@@ -0,0 +1,276 @@
...
+        local testname=${runtestdir%%/*}
+        runtest=${runtest//tests}
+
+        # in case a task is being removed then it's directory
+        # doesn't exists, so skip the test for YAML
+        [ ! -d "${runtest%%/*}/${testname}" ] && continue
+
+        runtest="${runtest}${testname}.yaml"
+        skipit=
+        for ignore in ${TEST_YAML_IGNORES};do
```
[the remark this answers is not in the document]
tisutisu: updated

test/common.sh
tisutisu: don't see any issue doing that. updated.

test/common.sh
tisutisu: to avoid unnecessary logs being printed in the console, while we looping through to check task run status, reason and results, if something fails `show_failure()` will dump the test resources in the console.

test/common.sh
tisutisu: updated it to `test_resource_creation_and_validation`, hope the name fits with its functionality now

test/common.sh
tisutisu: changed it to `print_changed_tasks_to_console`, hope it better fit its functionality.

test/common.sh
tisutisu: changed it to `print_changed_resources_to_console` now

test/run-test.sh
tisutisu: some leftover, forgot to remove it

.github/workflows/run-task-tests.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+
+      - name: Install tektor
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: |
+          go install github.com/lcarva/tektor@latest
```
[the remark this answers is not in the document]
tisutisu: Pinned to a sha now

README.md
tisutisu: we could also test `stepactions` in similar ways, that functionality is not added fully in the current PR, which we can extend to in the future, so kept the type (`task` or `stepactions`) as first argument in the `run-test.sh` script.

.github/workflows/run-task-tests.yaml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          ref: "${{ github.event.pull_request.head.sha }}"
+          path: build-definitions
+
+      - name: Checkout konflux-ci/konflux-ci Repository
+        if: steps.changed-files.outputs.any_changed == 'true'
+        uses: actions/checkout@v3
+        with:
+          repository: 'konflux-ci/konflux-ci'
+          path: konflux-ci
```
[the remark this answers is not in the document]
tisutisu: updated to use specific commit from `konflux-ci/konflux-ci` while checking out. But for configuring the renovate to update the SHA, does only renovate config is sufficient or do we need to run any script on push at `konflux-ci/konflux-ci` repo?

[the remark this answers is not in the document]
chmeliik: should be doable with a regex manager, roughly similar to https://docs.renovatebot.com/modules/manager/regex/#advanced-capture

README.md
chmeliik: Would `./test/run-test.sh stepactions/eaas-copy-secrets-to-ephemeral-cluster/0.1` not work?

.github/workflows/run-task-tests.yaml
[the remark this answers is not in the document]
chmeliik: How long does this take to run? According to the README, around 3 minutes https://github.com/jlumbroso/free-disk-space?tab=readme-ov-file
Which doesn't seem worth it unless really needed

task/git-clone/0.1/tests/run.yaml
chmeliik: Here we should verify that the task really cloned the repo as expected (e.g. contains the expected files, is really at revision `1.0.0`, ...). And this test should set the example for how to verify that the task works as expected.

That will probably require wrapping the Task in a Pipeline as Lui suggested

test/common.sh
```diff
@@ -0,0 +1,260 @@
...
+  # script, even if the tests pass.
+  # We store the real test result to return it later, ignoring any teardown
+  # failure in kubetest.
+  # TODO(adrcunha): Get rid of this workaround.
+  echo -n "$1"> ${TEST_RESULT_FILE}
+}
+
+function detect_changed_e2e_test() {
+    # check if any file from test/ directory is changed
+    echo ${CHANGED_FILES} | grep "test/" || true
```
chmeliik, line 29: This function currently always returns `0` no? I.e. it's always "true"?

test/common.sh
```diff
@@ -0,0 +1,260 @@
...
+}
+
+function print_changed_resources_to_console() {
+    # detect for changes in tests dir of the task
+    echo ${CHANGED_FILES} | grep -o 'task/[^\/]*/[^\/]*/tests/[^/]*'|xargs -I {} dirname {}|sed 's/\(tests\).*/\1/g'
+    # detect for changes in the task manifest
+    echo ${CHANGED_FILES} | grep -o 'task/[^\/]*/[^\/]*/*[^/]*.yaml'|xargs -I {} dirname {}|awk '{print $1"/tests"}'
```
chmeliik, line 36:
```suggestion
    echo "${CHANGED_FILES}" | grep -o 'task/[^\/]*/[^\/]*/tests/[^/]*'|xargs -I {} dirname {}|sed 's/\(tests\).*/\1/g'
    # detect for changes in the task manifest
    echo "${CHANGED_FILES}" | grep -o 'task/[^\/]*/[^\/]*/*[^/]*.yaml'|xargs -I {} dirname {}|awk '{print $1"/tests"}'
```
without the quotes, `echo` will print all the changed files on a single line and the `grep` probably won't work right

test/common.sh
```diff
@@ -0,0 +1,260 @@
...
+        [ ! -d "${runtest%%/*}/${testname}" ] && continue
+
+        runtest="${runtest}${testname}.yaml"
+        skipit=
+        for ignore in ${TEST_YAML_IGNORES}; do
+            [[ ${ignore} == "${testname}" ]] && skipit=True
+        done
+
+        # don't test the tasks which are deprecated
+        cat ${runtest} | grep 'tekton.dev/deprecated: \"true\"' && skipit=True
```
chmeliik, line 97: We use the `build.appstudio.redhat.com/expires-on` label (but its value can be in the future)

test/common.sh
```diff
@@ -0,0 +1,260 @@
...
+
+# Set the return code that the test script will return.
+# Parameters: $1 - return code (0-255)
+function set_test_return_code() {
+  # kubetest teardown might fail and thus incorrectly report failure of the
+  # script, even if the tests pass.
+  # We store the real test result to return it later, ignoring any teardown
+  # failure in kubetest.
+  # TODO(adrcunha): Get rid of this workaround.
+  echo -n "$1"> ${TEST_RESULT_FILE}
```
chmeliik, line 24: Is this relevant for us?

test/test-tasks.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+source $(dirname $0)/common.sh
+
+TMPF=$(mktemp /tmp/.mm.XXXXXX)
+clean() { rm -f ${TMPF}; }
+trap clean EXIT
+
+# Configure the number of parallel tests running at the same time, start from 0
+MAX_NUMBERS_OF_PARALLEL_TASKS=2 # => 8
```
chmeliik, line 10: What does the `start from 0` and `# => 8` mean here?

test/common.sh
chmeliik: Ah nevermind, we check the stdout output,  not the return code

README.md
tisutisu: yes that should work.... updated it now

test/test-tasks.sh
tisutisu: just a leftover, removed it.

task/git-clone/0.1/tests/run.yaml
tisutisu: makes sense now, this is another approach to validate the tasks (we were not aware of while investigating), the approach taken release-service-catalog repo has more flexibility on what to verify in the task results. With current approach taken in this PR, we are just validating task results are non-empty.  If we want to choose the approach taken in release-service-catalog repo, the current PR needs to be rewritten almost.

ralphbean: As a contributor to the release-service-catalog, I just want to say that I like the approach they're using. Once I got the hang of writing `mocks.sh` for my tests there, it became easy to add new functional tests for my tasks including both positive and negative cases.

They have nice features, like the ability to assert that the task should fail hard and exit with a non-zero status code on certain inputs, or to assert that the task should succeed, but have certain properties in its output.

For local development / inner-loop, I can stand up kind locally, install upstream tekton, and then run `export TEST_ITEMS=path/to/the/task/Im/working/on` and then run the `.github/scripts/test_tekton_tasks.sh` from their repo in a loop while I hack on my task or test.

tisutisu: @ralphbean thanks for the feedback, looks like the approach used in release-service-catalog is better, will move to it.

tisutisu:
Closing in favour of https://github.com/konflux-ci/build-definitions/pull/1644

.github/workflows/run-task-tests.yaml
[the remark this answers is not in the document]
tisutisu: Removed it now. Please check [here](https://github.com/konflux-ci/build-definitions/pull/1644).

task/git-clone/0.1/tests/run.yaml
tisutisu: took this approach in [this PR](https://github.com/konflux-ci/build-definitions/pull/1644)  and closed the current one

PR #1457, opened 2024-09-20
task/git-clone*: add --short SHA to results

jbpratt:
add the short commit SHA of the default length (7 characters) to the results of the tasks and regen the readme


chmeliik:
We can probably close this in favor of #1445 (discussing there how much we need this result)

PR #1458, opened 2024-09-23
ci: migrate yamllint to GitHub Actions

tnevrlka:
[STONEBLD-2791](https://issues.redhat.com/browse/STONEBLD-2791)
- Container image used in yaml-lint task is outdated and using it causes rate-limiting issues with docker.io
- Remove the yaml-lint task to task-lint and remove yamllint part
- Create a GitHub Action that runs yamllint
- Fix yamllint warnings
  - Double quote the `on` keyword
`on` -> `"on"`
  - Fix comment indentation

Tested locally by using `act`. Added random things to random YAMLs in order to make the lint fail and it indeed failed.

### DIFF
--- a/.github/workflows/yaml-lint.yaml
+++ b/.github/workflows/yaml-lint.yaml
@@ -1,0 +1,12 @@
+---
+name: yamllint
+"on":
+  pull_request:
+    branches: [main]
+jobs:
+  yamllint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Lint YAML files
+        run: yamllint .
--- a/.github/workflows/check-buildah-remote.yaml
+++ b/.github/workflows/check-buildah-remote.yaml
@@ -7,9 +7,9 @@ jobs:
     name: Check Buildah Remote
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332  # v4
       - name: Install Go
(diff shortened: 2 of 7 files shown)


.tekton/tasks/yaml-lint.yaml
```diff
@@ -1,50 +0,0 @@
...
-          if yq '.spec?.steps[] | .script' $task | grep -q '\$(params\.'; then
-            FAILED_TASKS="$FAILED_TASKS $task"
-          fi
-        done
-        if [ -n "$FAILED_TASKS" ]; then
-          echo "Tasks contains params in script section (https://tekton.dev/docs/pipelines/tasks/#substituting-in-script-blocks)"
-          echo $FAILED_TASKS | tr ' ' '\n' | sort
-          exit 1
-        fi
-      workingDir: $(workspaces.shared-workspace.path)/source
```
chmeliik, line 50: The yamllint task also did something entirely not yamllint-y (^ this check)

We need to preserve this check

.github/workflows/yaml-lint.yaml
```diff
@@ -0,0 +1,12 @@
...
+"on":
+  pull_request:
+    branches: [main]
+jobs:
+  yamllint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Lint YAML files
+        run: yamllint .
```
chmeliik, line 12: > create a GitHub Action that installs yamllint and then runs it

It looks like it doesn't even install it. Is yamllint just pre-installed on GH runners?

tnevrlka: Yes, it is

chmeliik: Nice. Could you just update the commit message?

chmeliik (approved):

PR #1463, opened 2024-09-25
feat(CLOUDDST-24366) Update description of show-sbom

emilyzheng:
SPDX format will be supported by build-definitions later. Update description of show-sbom tasks so that they are not limited to CycloneDX.

### DIFF
--- a/task/show-sbom-rhdh/0.1/README.md
+++ b/task/show-sbom-rhdh/0.1/README.md
@@ -1,6 +1,6 @@
 # show-sbom-rhdh task
 
-Shows the Software Bill of Materials (SBOM) generated for the built image in CyloneDX JSON format.
+Shows the Software Bill of Materials (SBOM) generated for the built image.
 The 'task.*' annotations are processed by Red Hat Developer Hub (RHDH) so that the log content can be rendered in its UI.
 
 ## Parameters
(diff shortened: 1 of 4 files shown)


emilyzheng:
@rcerven @taylormadore could you please review this?

rcerven:
@arewm should there be mention about spdx, if we aren't even generating sbom in spdx ?

arewm:
The implementation isn't specific to either cyclonedx nor spdx. If we want to change this, I think we should actually remove the wording instead of adding more.

Can SPDX even be attached or does it have to be an attestation? If the latter, then the current task definition won't be able to show the SPDX sbom.

midnightercz:
> The implementation isn't specific to either cyclonedx nor spdx. If we want to change this, I think we should actually remove the wording instead of adding more.
> 
> Can SPDX even be attached or does it have to be an attestation? If the latter, then the current task definition won't be able to show the SPDX sbom.

I thought we already produce sboms in attestations

emilyzheng:
@midnightercz So will spdx sbom be attached like what's done for cyclonedx sbom in upload-sbom? Or buildah tasks will generate it as attestation? I think spdx is supported by cosign attach sbom:

```
$ cosign attach sbom --help
...
    --type='spdx':
	type of sbom (spdx|cyclonedx|syft)
```

midnightercz:
So I checked and most of the tekton tasks use sbom attach --sbom-type=<xxx>, where <xxx> is cyclonedx or spdx based on task paramater

rcerven (approved):

PR #1470, opened 2024-09-25
apply-tags: Tag multi-arch images as expected

samdoran:
When copying an image manifest, `skopeo` by default only copies the image matching the architecture of the system running `skopeo`. It does not copy the image manifest as expected.

In my testing, adding the `--multi-arch all` flag on single architecture images works without error.

### DIFF
--- a/task/apply-tags/0.1/apply-tags.yaml
+++ b/task/apply-tags/0.1/apply-tags.yaml
@@ -47,7 +47,7 @@ spec:
           IMAGE_WITHOUT_TAG=$(echo "$IMAGE" | sed 's/:[^:]*$//')
           for tag in "$@"; do
             echo "Applying tag $tag"
-            skopeo copy docker://$IMAGE docker://$IMAGE_WITHOUT_TAG:$tag
+            skopeo copy --multi-arch all docker://"$IMAGE" docker://"$IMAGE_WITHOUT_TAG:$tag"
           done
         else
           echo "No additional tags parameter specified"
@@ -70,7 +70,7 @@ spec:
           for tag in "${tags_array[@]}"
           do
               echo "Applying tag $tag"
-              skopeo copy docker://$IMAGE docker://$IMAGE_WITHOUT_TAG:$tag
+              skopeo copy --multi-arch all docker://"$IMAGE" docker://"$IMAGE_WITHOUT_TAG:$tag"
           done
         else
           echo "No additional tags specified in the image labels"
--- a/task/apply-tags/0.1/apply-tags.yaml
+++ b/task/apply-tags/0.1/apply-tags.yaml
@@ -47,7 +47,7 @@ spec:
           IMAGE_WITHOUT_TAG=$(echo "$IMAGE" | sed 's/:[^:]*$//')
           for tag in "$@"; do
             echo "Applying tag $tag"
-            skopeo copy --multi-arch all docker://"$IMAGE" docker://"$IMAGE_WITHOUT_TAG:$tag"
+            skopeo copy --multi-arch index-only docker://"$IMAGE" docker://"$IMAGE_WITHOUT_TAG:$tag"
           done
         else
           echo "No additional tags parameter specified"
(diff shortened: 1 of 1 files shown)


task/apply-tags/0.1/apply-tags.yaml
```diff
@@ -47,7 +47,7 @@ spec:
           IMAGE_WITHOUT_TAG=$(echo "$IMAGE" | sed 's/:[^:]*$//')
           for tag in "$@"; do
             echo "Applying tag $tag"
-            skopeo copy docker://$IMAGE docker://$IMAGE_WITHOUT_TAG:$tag
+            skopeo copy --multi-arch all docker://$IMAGE docker://$IMAGE_WITHOUT_TAG:$tag
```
chmeliik, line 50: Does this apply the same tag to the image index and all the arch-specific images? Is that what we want?

samdoran: In my testing, this applies the tags to the image index only not the archi-specific images, which is the goal. I'm building multi-arch images and what currently happens is that the additional tags are applied only to a single image, not the index.

samdoran: This will not apply tags to the arch-specific images. The same tag **cannot** be applied to multiple objects, but an object can have multiple tags.

Looking at the [skopeo man pag](https://www.mankier.com/1/skopeo-copy#--multi-arch), the default behavior when copying a multi-arch image is to only copy the image that matches the system architecture. That lines up with what I'm seeing and seems incorrect.

samdoran: Copying a single architecture image with the `--multi-arch all` flag doesn't change the current behavior.
```
> skopeo copy --multi-arch all docker://quay.io/samdoran/ubi-hive:pr-119-latest-linux-x86-64 docker://quay.io/samdoran/ubi-hive:pr-119-latest-linux-x86-64-2
Copying blob 40049f9dc504 skipped: already exists
Copying blob 2196b3591613 skipped: already exists
Copying blob 2e1c8a5459bc skipped: already exists
Copying config 71cf39a0ad done   |

> skopeo inspect docker://quay.io/samdoran/ubi-hive:pr-119-latest-linux-x86-64-2 | jello _.Architecture
"amd64"
```

samdoran: You can [look at the repo](https://quay.io/repository/samdoran/ubi-hive?tab=tags&tag=latest) to get a better sense of for the result.

samdoran: Compare that to the `pr-119-latest` and `pr-119-fe84dee` images in [redhat-user-worloads](https://quay.io/repository/redhat-user-workloads/cost-mgmt-dev-tenant/ubi-hive?tab=tags) where `pr-119-fe84dee` is not an image index (it's an x86 image).

samdoran:
I fixed the CI failures.

task/apply-tags/0.1/apply-tags.yaml
samdoran: The `index-only` flag only works if the blobs to be copied already exist in the destination. I think that's ok for this task since we expecte the blobs to already be there.

Let me try that real quick.

samdoran: Thas seems to work.
```
> skopeo copy --multi-arch index-only docker://quay.io/samdoran/ubi-hive:pr-119-latest docker://quay.io/samdoran/ubi-hive:pr-119-index-only
Getting image list signatures
Copying 0 images generated from 2 images in list
Writing manifest list to image destination
Storing list signatures
```

Using `all` seems more reliable since it will copy an image if needed, otherwise it just copies the index. I would be fine with `index-only`, though.

chmeliik: This task should really only apply additional tags, so index-only sounds right

samdoran:
Thank you for merging!

PR #1471, opened 2024-09-26
deprecate images using BASE_IMAGES_DIGESTS result

tnevrlka:
[STONEBLD-2747](https://issues.redhat.com/browse/STONEBLD-2747)
- Results with BASE_IMAGES_DIGESTS are too large and cause builds to hit the Tekton result size
limit
- Deprecate all tasks using the result so they will be unsupported after 2024-11-13T00:00:00Z

### DIFF
--- a/task/s2i-java/0.1/s2i-java.yaml
+++ b/task/s2i-java/0.1/s2i-java.yaml
@@ -5,6 +5,7 @@ metadata:
     app.kubernetes.io/version: "0.1"
     build-definition.include: add-sbom-and-push
     build.appstudio.redhat.com/build_type: "java"
+    build.appstudio.redhat.com/expires-on: 2024-11-13T00:00:00Z
   annotations:
     tekton.dev/displayName: s2i java
     tekton.dev/pipelines.minVersion: "0.19"
--- a/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.1/buildah-oci-ta.yaml
@@ -4,6 +4,7 @@ kind: Task
 metadata:
   name: buildah-oci-ta
   annotations:
+    build.appstudio.redhat.com/expires-on: "2024-11-13T00:00:00Z"
     tekton.dev/pipelines.minVersion: 0.12.1
     tekton.dev/tags: image-build, konflux
   labels:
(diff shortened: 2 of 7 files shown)


mmorhun (approved):

task/s2i-java/0.1/s2i-java.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     app.kubernetes.io/version: "0.1"
     build-definition.include: add-sbom-and-push
     build.appstudio.redhat.com/build_type: "java"
+    build.appstudio.redhat.com/expires-on: 2024-11-13T00:00:00Z
```
chmeliik, line 8: The s2i tasks have no 0.2 version, so this deprecates the entire s2i pipeline. But if we're sure that nobody is using them, it is fine

tnevrlka: There's a notice period of 1.5 months. Somebody would surely reach out if they were affected, wouldn't they?

task/buildah-rhtap/0.1/buildah-rhtap.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "containers, rhtap"
+    build.appstudio.redhat.com/expires-on: 2024-11-13T00:00:00Z
```
chmeliik, line 10: buildah-rhtap also has no 0.2 version and we can't deprecate it

I have to finish https://github.com/konflux-ci/build-definitions/pull/1195 first :smile:

task/s2i-java/0.1/s2i-java.yaml
chmeliik: It's a warning, not a failure. Nobody looks at warnings :smile:

chmeliik: (it will start failing after the notice period is over)

task/buildah-rhtap/0.1/buildah-rhtap.yaml
tnevrlka: Okay, I thought it'd be fine since it would get merged before that. But you're right that it makes more sense to deprecate only **after** there's something to migrate to.

task/s2i-java/0.1/s2i-java.yaml
tnevrlka: @chmeliik Are you okay with resolving this? Due to [STONEBLD-2818](https://issues.redhat.com/browse/STONEBLD-2818)

chmeliik: Sounds like we know nobody is using the tasks :+1:

PR #1472, opened 2024-09-26
Delete source to image pipelines and tasks

mmorhun:
Since S2I pipelines (and corresponding tasks) is not used anymore for quite some time, we are removing them.


chmeliik (commented):
Should we wait until the s2i tasks are deprecated according to the expires-on label?

mmorhun:
@chmeliik the tasks are not used anywhere, I think it's safe to delete now.

tkdchen (commented):
These lines are also deprecated:

https://github.com/konflux-ci/build-definitions/blob/main/hack/build-and-push.sh#L278-L279

chmeliik (approved):
:tada:

tkdchen (approved):

PR #1473, opened 2024-09-26
Kustomize support trusted-artifacts Task generator

zregvart:
Adds the ability to have a `kustomization.yaml` file as the base file for generating the OCI-TA Task variant; if it is used as a base the equivalent of `kustomize build` will be run expecting the resulting rendered template to contain the single Task definition.


task-generator/trusted-artifacts/main_test.go
```diff
@@ -36,6 +39,7 @@ func TestGolden(t *testing.T) {
 
 			switch dir.Name() {
 			case "buildah":
```
lcarva, line 41: Should this be?

```
case "buildah":
  image = ""
case "kustomized":
  image = ""
```

zregvart: Missed a `fallthrough` but didn't care as tests were passing. Made it simpler in the next revision

lcarva (approved):

PR #1475, opened 2024-09-26
Deprecate buildah task resource variations

mmorhun:
Since it's possible to set task resources from pipeline run in Konflux ([docs](https://konflux-ci.dev/docs/how-tos/configuring/overriding-compute-resources/)), we are deprecation `buildah-*gb` tasks.

### DIFF
--- a/task/buildah-10gb/0.1/patch.yaml
+++ b/task/buildah-10gb/0.1/patch.yaml
@@ -7,3 +7,9 @@
 - op: replace
   path: /spec/steps/0/computeResources/requests/memory
   value: 8Gi
+- op: add
+  path: /metadata/annotations/build.appstudio.redhat.com~1expires-on
+  value: "2025-01-21T00:00:00Z"
+- op: add
+  path: /metadata/annotations/build.appstudio.redhat.com~1expiry-message
+  value: "Instead of using this task, switch to buildah task (you can find latest bundle reference here: https://quay.io/repository/konflux-ci/tekton-catalog/task-buildah) and configure build step resources as described in the doc: https://konflux-ci.dev/docs/how-tos/configuring/overriding-compute-resources/"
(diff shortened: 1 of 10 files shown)


rcerven (approved):

arewm (approved):

chmeliik (commented):
With https://github.com/enterprise-contract/ec-policies/pull/1156 is in, we can make the expiry message more useful. The default says `Upgrade to a newer version of the task`, which doesn't apply here. We should suggest to switch to the regular buildah task and link the resource configuration docs

mkosiarc (approved):

task/buildah-10gb/0.1/patch.yaml
```diff
@@ -7,3 +7,9 @@
 - op: replace
   path: /spec/steps/0/computeResources/requests/memory
   value: 8Gi
+- op: add
+  path: /metadata/annotations/build.appstudio.redhat.com~1expires-on
+  value: "2025-01-21T00:00:00Z"
+- op: add
+  path: /metadata/annotations/build.appstudio.redhat.com~1expiry-message
+  value: "Instead of using this task, switch to buildah task and configure build step resources as described in the doc: https://konflux-ci.dev/docs/how-tos/configuring/overriding-compute-resources/"
```
chmeliik, line 15: optional: we could also link the quay repo for the buildah task to help users find the right bundle ref https://quay.io/repository/konflux-ci/tekton-catalog/task-buildah

chmeliik (approved):

task/buildah-10gb/0.1/patch.yaml
mmorhun: Makes sense.

tnevrlka:
@mmorhun Can we merge this?

PR #1483, opened 2024-09-30
OCI-attach reports from `clair-scan` 0.2 Task

zregvart:
This attaches the "clair"-formatted output of the `clair-action` command to the scanned image. Contrary to the `sast-snyk-check` Task only a single variant of attachment method is supported, based on the registry support. For quay.io OCI Distribution 1.1. Referrers API will be used.

The `clair-action` needs to be run twice as the Rego rules executed in the `conftest-vulnerabilities` require the "quay" format, which does not include any date information, whereas the (future) EC policy Rego rules require the "clair" format, which does.

Resolves: https://issues.redhat.com/browse/EC-837

Notes for reviewers:
- should this be uploading both OCI 1.1 distribution referrers tag and API variants, analogous to the sast-snyk-check Task?
- is there a better MIME type other than `application/vnd.redhat.clair-report+json` that could be used?
- should the report be transformed into a in-toto enveloped [vuln](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md) predicate attestation and attached as such?

### DIFF
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -214,6 +214,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES_PROCESSED| Images processed in the task.| |
+|REPORTS| Mapping of image digests to report digests| |
 |SCAN_OUTPUT| Clair scan result.| |
 |TEST_OUTPUT| Tekton task test output.| |
 ### clamav-scan:0.1 task results
--- a/task/clair-scan/0.2/README.md
+++ b/task/clair-scan/0.2/README.md
@@ -17,10 +17,11 @@ analyzing the components of a container image and comparing them against Clair's
 
 ## Results:
 
-| name              | description              |
-|-------------------|--------------------------|
-| TEST_OUTPUT | Tekton task test output. |
-| SCAN_OUTPUT | Clair scan result.       |
+| name              | description                              |
+|-------------------|------------------------------------------|
+| TEST_OUTPUT       | Tekton task test output.                 |
+| SCAN_OUTPUT       | Clair scan result.                       |
+| REPORTS           |Mapping of image digests to report digests|
 
 ## Clair-action repository:
 https://github.com/quay/clair-action
--- a/task/clair-scan/0.2/clair-scan.yaml
+++ b/task/clair-scan/0.2/clair-scan.yaml
@@ -34,6 +34,8 @@ spec:
       description: Clair scan result.
     - name: IMAGES_PROCESSED
       description: Images processed in the task.
+    - name: REPORTS
+      description: Mapping of image digests to report digests
   stepTemplate:
     volumeMounts:
       - name: trusted-ca
@@ -92,14 +94,28 @@ spec:
           value: $(params.image-url)
         - name: IMAGE_DIGEST
           value: $(params.image-digest)
+      workingDir: /tekton/home
       script: |
         #!/usr/bin/env bash
+
+        set -o errexit
+        set -o nounset
+        set -o pipefail
 
         imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
         images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
         digests_processed=()
 
-        for sha_file in /tekton/home/image-manifest-*.sha; do
+        # the quay report format used by the Conftest rules in the
+        # conftest-vulnerabilities step doesn't contain the "issued" date which
+        # we require in the policy rules, so we resort to running clair-action
+        # twice to produce both quay and clair formatted output
+        clair_report() {
+          { clair-action report --image-ref="$1" --db-path=/tmp/matcher.db --format=quay | tee "clair-result-$2.json"; } && \
+          { clair-action report --image-ref="$1" --db-path=/tmp/matcher.db --format=clair > "clair-report-$2.json"; }
+        }
+
+        for sha_file in image-manifest-*.sha; do
           if [ -e "$sha_file" ]; then
             arch_sha=$(cat "$sha_file")
             arch=$(basename "$sha_file" | sed 's/image-manifest-//;s/.sha//')
@@ -107,7 +123,7 @@ spec:
 
             echo "Running clair-action on $arch image manifest."
             # run the scan for each image manifest in the image index
-            clair-action report --image-ref=$arch_specific_digest --db-path=/tmp/matcher.db --format=quay | tee /tekton/home/clair-result-$arch.json || true
+            clair_report "${arch_specific_digest}" "${arch}" || true
 
             digests_processed+=("\"$arch_sha\"")
           fi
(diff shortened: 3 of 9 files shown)


lcarva:
> should the report be transformed into a in-toto enveloped [vuln](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md) predicate attestation and attached as such?

Probably, but Konflux is not setup for this now.

task/clair-scan/0.2/clair-scan.yaml
```diff
@@ -92,22 +92,32 @@ spec:
...
+        set -o pipefail
+
         imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
         images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
         digests_processed=()
 
-        for sha_file in /tekton/home/image-manifest-*.sha; do
+        clair_report() {
+          { clair-action report --image-ref="$1" --db-path=/tmp/matcher.db --format=quay | tee "clair-result-$arch.json"; } && \
+          { clair-action report --image-ref="$1" --db-path=/tmp/matcher.db --format=clair > "clair-report-$arch.json"; }
```
lcarva, line 109: What's the performance impact of running `clair-action` twice? Not sure how long each command takes to run.

task/clair-scan/0.2/clair-scan.yaml
```diff
@@ -92,22 +92,32 @@ spec:
...
+        set -o errexit
+        set -o nounset
+        set -o pipefail
+
         imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
         images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
         digests_processed=()
 
-        for sha_file in /tekton/home/image-manifest-*.sha; do
+        clair_report() {
```
lcarva, line 107: Would it be better if `$arch` was provided as the second parameter to this function? `arch=$2`?

task/clair-scan/0.2/clair-scan.yaml
```diff
@@ -117,7 +127,42 @@ spec:
...
+
+        if ! compgen -G "clair-report-*.json" > /dev/null; then
+          echo 'No Clair reports generated. Skipping upload.'
+          exit 0
+        fi
+
+        echo "Selecting auth"
+        select-oci-auth "$IMAGE_URL" > "$HOME/auth.json"
+
+        base_image="${IMAGE_URL/:*/}"
```
lcarva, line 152: Maybe let's call this something like REPOSITORY? "Base image" has a different meaning that may cause confusion.

task/clair-scan/0.2/clair-scan.yaml
```diff
@@ -117,7 +127,42 @@ spec:
...
+          report_file="$1"
+          arch="${report_file/*-}"
+          echo "${arch/.json/}"
+        }
+
+        MEDIA_TYPE='application/vnd.redhat.clair-report+json'
+        for f in clair-report-*.json; do
+          image_ref="${base_image}@$(cat "image-manifest-$(arch "$f").sha")"
+          echo "Attaching $f to ${image_ref}"
+          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${image_ref}" "$f:${MEDIA_TYPE}"
```
lcarva, line 164: On the EC side, we can't trust this information since there's no guarantee that the attached data was actually produced by this Task. Anyone with access to the registry can easily fake the report. This data is informative at best.

This Task needs to emit a digest of the expected data.

If we had identity-based signatures (keyless) enabled in Konflux, we would still need to do this since in that case users could still fake a report via some rogue Pipeline. Slightly harder, but not a risk we should take.

simonbaird (approved):

zregvart:
> > should the report be transformed into a in-toto enveloped [vuln](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md?rgh-link-date=2024-09-30T20%3A06%3A38Z) predicate attestation and attached as such?
> 
> Probably, but Konflux is not setup for this now.

We can choose to use the format, regardless of the attestation being signed or not

task/clair-scan/0.2/clair-scan.yaml
zregvart: In the past 30 days most seem to end within < 30 sec (65%), I would guess that this correlates to the image size. I don't see any caching in claircore or clair-action, so it will be at least double that

task/clair-scan/0.2/clair-scan.yaml
zregvart: I did remember that after pushing, will run up against the result limit though...

task/clair-scan/0.2/clair-scan.yaml
lcarva: Ack. That's probably ok.

task/clair-scan/0.2/clair-scan.yaml
lcarva: Yeah, we may have to create a combined result of some sort. You could, for example, put them altogether in an Image Index.

pipelines/docker-build-multi-platform-oci-ta/README.md
```diff
@@ -214,6 +214,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES_PROCESSED| Images processed in the task.| |
+|REPORTS| Mapping of image digests to report digests| |
```
simonbaird, line 217: Maybe:

```suggestion
|REPORT_MAP| Mapping of image digests to report digests| |
```

or...

```suggestion
|REPORT_REFS| Mapping of image digests to report digests| |
```

task/clair-scan/0.2/README.md
```diff
@@ -17,10 +17,11 @@ analyzing the components of a container image and comparing them against Clair's
...
 
-| name              | description              |
-|-------------------|--------------------------|
-| TEST_OUTPUT | Tekton task test output. |
-| SCAN_OUTPUT | Clair scan result.       |
+| name              | description                              |
+|-------------------|------------------------------------------|
+| TEST_OUTPUT       | Tekton task test output.                 |
+| SCAN_OUTPUT       | Clair scan result.                       |
+| REPORTS           |Mapping of image digests to report digests|
```
simonbaird, line 24: Is it json? Maybe:

```suggestion
| REPORT_REFS           |Mapping of image digests to report digests in JSON format|
```

task/clair-scan/0.2/clair-scan.yaml
```diff
@@ -92,22 +94,32 @@ spec:
...
+        set -o nounset
+        set -o pipefail
+
         imagewithouttag=$(echo -n $IMAGE_URL | sed "s/\(.*\):.*/\1/")
         images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
         digests_processed=()
 
-        for sha_file in /tekton/home/image-manifest-*.sha; do
+        clair_report() {
+          { clair-action report --image-ref="$1" --db-path=/tmp/matcher.db --format=quay | tee "clair-result-$2.json"; } && \
```
simonbaird, line 110: Maybe a comment to acknowledge that we don't want to be running it twice, but currently there's no other way to produce the two types of output.

lcarva (approved):

lcarva:
> > > should the report be transformed into a in-toto enveloped [vuln](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md?rgh-link-date=2024-09-30T20%3A06%3A38Z) predicate attestation and attached as such?
> > 
> > 
> > Probably, but Konflux is not setup for this now.
> 
> We can choose to use the format, regardless of the attestation being signed or not

The benefit the format would bring is that it links the data to a predicate (image). You are doing that already in the `REPORTS` result. Using an in-toto envelope would standardize that. I don't think it covers any gaps though.

lcarva:
> should this be uploading both OCI 1.1 distribution referrers tag and API variants, analogous to the sast-snyk-check Task?

For posterity, in the arch meeting it was mentioned that double uploading may cause some issues. The Tasks that were doing it, no longer do it, see https://github.com/konflux-ci/build-definitions/commit/c746a251af63d1b91bf79b8f759737aaba0227ab.

> is there a better MIME type other than application/vnd.redhat.clair-report+json that could be used?

I think this one is reasonable since it is a completely custom format.

zregvart:
Last call for discussions:
 - the result name (`REPORTS` vs `REPORT_MAP` vs `REPORT_REFS`), the `REPORTS` suits me fine because it is short and I don't need to change anything
 - using in-toto + vuln predicate, would be nice, though doesn't seem to address an issue we have
 - using image index for the reports, would be better to keep the result smaller, a new step would need to be added for this, probably utilizing `quay.io/konflux-ci/buildah-task`

I'm lazy to take on more work here and I'm hovering over the merge button. Let me know if you feel strongly about any of this.

jsztuka:
Im team `REPORTS` only.
Bullet point 3 looks like candidate for follow up story. If it turns out we want it.

jsztuka (approved):
lgtm

jsztuka:
Great work, thanks!

zregvart:
Thanks for the reviews/feedback!

PR #1484, opened 2024-10-01
Use same task sbom processing in buildah 0.1 and 0.2

midnightercz:


tnevrlka:
@midnightercz Thank you for the contribution! What is the reason for this change? 
As far as I know, we want to generally avoid updating 0.1 versions. 

cc @mmorhun

midnightercz:
Hi, I need to do changes in 0.2 to support spdx format. I wasn't sure if 0.1 should be modified as well, so I filed this PR to unify tasks so I can do changes only in build-tasks-dockerfiles in relevant scripts instead of duplicating the functionality to "inline" python code in 0.1. This change might be unneeded if you decide 0.1 shouldn't be touched as it's deprecated and going to be removed.

mmorhun:
I think we should leave 0.1 task as is and make changes to 0.2 only.

PR #1490, opened 2024-10-04
fix: oci-copy task triggers unbounded var error when not using AWS creds

pastequo:
Some build is failing because of this.

I think the problem is that set -u is set after the function definition but *before* the function is called

EDIT: I'm talking about this set: https://github.com/konflux-ci/build-definitions/blob/main/task/oci-copy/0.1/oci-copy.yaml#L186
Build is failing because of 
> /tekton/scripts/script-1-qxwgg: line 11: AWS_ACCESS_KEY_ID: unbound variable

When I'm not providing AWS creds

### DIFF
--- a/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
+++ b/task/oci-copy-oci-ta/0.1/oci-copy-oci-ta.yaml
@@ -135,7 +135,7 @@ spec:
           method="GET"
 
           curl_args=(--fail --silent --show-error)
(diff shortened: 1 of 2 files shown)


pastequo:
I tested my changes in a PR, it fixes my issue. I haven't tested if it introduces regression

chmeliik (approved):
Thanks for the fix!

chmeliik (commented):
After looking at the rest of the task - would you mind making the same fix for the `BEARER_TOKEN` variable? I think it has the same problem

pastequo:
@chmeliik Should be done as well. I haven't checked the rest of the file...

ralphbean (approved):

PR #1492, opened 2024-10-06
fix(KFLUXBUGS-1693): remove fail-unsigned param from task

avi-biton:
The fail-unsigned parameter was removed from the task.
The task will only fail if the scanner fails, otherwise it will end successfully providing the RPMS_DATA result with the signature keys and the number of unsigned RPMs.
EC will process the result and determine whether it should pass or fail.

- Bumping to v0.2
- Update task
- Add MIGRATION.md
- Update template version


avi-biton:
@lcarva Can you please take a look?

lcarva (approved):

PR #1495, opened 2024-10-08
Include image index digest in clair-scan Task

lcarva:
This was missed when the v0.2 version of the clair-scan Task was introduced.

Ref: https://issues.redhat.com/browse/EC-906


robnester-rh (approved):
LGTM

PR #1499, opened 2024-10-10
Install oc if it is not present

tkdchen:


dheerajodha (approved):
Thanks!

.github/workflows/check-readmes.yaml
```diff
@@ -10,6 +10,13 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
 
+      - name: Install oc
+        run: |
+          url=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest-4.17/openshift-client-linux.tar.gz
+          if ! which oc; then
+            sudo bash -c "curl -sL "$url" | tar -xz -C /usr/bin oc"
```
kdudka, line 17: You do not need to run `curl` (neither the `xz` decompressor) through `sudo`.  Only the write to `/usr/bin/` requires root privileges.  You should also run `curl` with `--fail` so that it provides a meaningful error message when the URL starts to return `404` etc.

kdudka (changes requested):

.github/workflows/check-readmes.yaml
tkdchen: PTAL

kdudka: I think it can be further simplified/secured like this:
```sh
curl --fail -sL "$url" | gzip -cd | sudo -- tar -x -C /usr/bin oc
```

kdudka (approved):

.github/workflows/check-readmes.yaml
tkdchen: > I think it can be further simplified/secured like this:
> 
> ```shell
> curl --fail -sL "$url" | gzip -cd | sudo -- tar -x -C /usr/bin oc
> ```

Cool. But, why can't keep `z` option for tar?

kdudka: You can but in that case `gzip` will unnecessarily run with `root` privileges.

tkdchen: Updated

zregvart:
`kubectl` [is present](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools), could we use `kubectl kustomize` [here](https://github.com/konflux-ci/build-definitions/blob/4e3171e5c76e01d81e4acfda29b2b3fd76f142f3/hack/generate-pipelines-readme.py#L54) instead?

tkdchen:
> `kubectl` [is present](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools), could we use `kubectl kustomize` [here](https://github.com/konflux-ci/build-definitions/blob/4e3171e5c76e01d81e4acfda29b2b3fd76f142f3/hack/generate-pipelines-readme.py#L54) instead?

Yes. kustomize is even included in the runner image and just use `kustomize build` in the script. I have no idea why oc was selected to be used in the script originally. I just thought oc should be a CLI installed commonly.

PR #1501, opened 2024-10-11
Annotate task bundles with version

tkdchen:
[STONEBLD-2821](https://issues.redhat.com//browse/STONEBLD-2821)

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -172,6 +172,9 @@ do
       ANNOTATIONS+=("org.opencontainers.image.source=${VCS_URL}")
       ANNOTATIONS+=("org.opencontainers.image.revision=${VCS_REF}")
       ANNOTATIONS+=("org.opencontainers.image.url=${VCS_URL}/tree/${VCS_REF}/${task_dir}")
+      # Ensure an empty string is set rather than string "null" if the version label is not present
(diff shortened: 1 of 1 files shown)


hack/build-and-push.sh
```diff
@@ -172,6 +172,8 @@ do
       ANNOTATIONS+=("org.opencontainers.image.source=${VCS_URL}")
       ANNOTATIONS+=("org.opencontainers.image.revision=${VCS_REF}")
       ANNOTATIONS+=("org.opencontainers.image.url=${VCS_URL}/tree/${VCS_REF}/${task_dir}")
+      concrete_task_version=$(yq '.metadata.labels."app.kubernetes.io/version"' "$prepared_task_file")
```
mmorhun, line 175: Please use just `task_version`

tkdchen: Variable `task_version` is already used in the `read` command to reference the version from task file path.

hack/build-and-push.sh
```diff
@@ -172,6 +172,8 @@ do
...
+      ANNOTATIONS+=("org.opencontainers.image.version=${concrete_task_version}")
```
chmeliik, line 176: Not all tasks have the version label I think. Will this do something sensible for those that don't (probably set the annotation to an empty string?)

tkdchen: Updated. PTAL.

chmeliik (approved):

PR #1502, opened 2024-10-11
Add rpm-ostree OCI-TA Task variant

zregvart:
Generated for version 0.2 only.

Reference: https://issues.redhat.com/browse/EC-924

### DIFF
--- a/task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
+++ b/task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
@@ -116,6 +116,7 @@ spec:
         - name: COMMIT_SHA
           value: $(params.COMMIT_SHA)
       script: |
+        #!/bin/bash
         set -o verbose
         set -eu
         set -o pipefail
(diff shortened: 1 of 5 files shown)


task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
```diff
@@ -0,0 +1,303 @@
...
+            --tls-verify="${TLSVERIFY}" \
+            --digestfile "/var/workdir/source/image-digest" "${IMAGE}" \
+            "docker://${IMAGE}" && break || status=$?
+        done
+        if [ "$status" -ne 0 ]; then
+            echo "Failed to push sbom image to registry after ${max_run} tries"
+            exit 1
+        fi
+
+                                                                                 <"/var/workdir/source"/image-digest tee "$(results.IMAGE_DIGEST.path)"
```
zregvart, line 282: Something in the shell formatter does this, it's a generated file and this is mostly esthetics so I don't think it is worth troubleshooting, but yell at me if you think it matters...

cgwalters:
Note we have https://issues.redhat.com/browse/BIFROST-408 which would allow us to get rid of this task entirely.

What's the timeframe for this? I would say it's possible within a few weeks we'd gain sufficient knowledge to say whether or not we can execute on deprecating and eventually removing this task entirely.

task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
tkdchen: Is shfmt the tool used?

zregvart: We use https://github.com/mvdan/sh

lcarva:
To avoid blocking different efforts, let's go ahead and get this Task in. We can remove/deprecate it if/when we do so for the `rpm-ostree` Task as well.

zregvart:
Taking the comments above as approval

lcarva (approved):

PR #1504, opened 2024-10-14
Always name attached file Dockerfile in push-dockerfile task

mmorhun:
To have consistency in `.dockerfile` oci artifact of the built image, always create it under `Dockerfile` name


tnevrlka (approved):

mkosiarc (approved):

mkosiarc:
LGTM, except that "push-dockerfile-oci-ta.yaml" needs to be updated as well

PR #1507, opened 2024-10-14
Optionally skip SBOM generation

arewm:
Since we can require SBOMs to be present with EC policies, we can enable
users to optionally speed up their builds by not analyzing repositories
to generate build-time SBOMs.

While we may have a partial SBOM from the prefetched data, we should
just not upload an SBOM at all in order to simplify decisions (i.e.
removing the need to decide if the SBOM is full or partial).


mmorhun (dismissed):
LGTM

arewm:
This cannot be merged until the version of Tekton is updated. I think that is planned on Monday with https://github.com/redhat-appstudio/infra-deployments/pull/5201

tnevrlka:
Hello @arewm. FYI version 0.3 of `buildah` tasks has been released. Would you mind rebasing and applying the changes to the newest version, please? Thank you!

tnevrlka (approved):

arewm:
The 0.2 task can be deprecated whenever we determine that it is best to do it. Since it is not currently deprecated, I think it makes most sense to apply this change to both.

PR #1510, opened 2024-10-15
Revert "chore(deps): update ec"

zregvart:
Reverts konflux-ci/build-definitions#1460


tnevrlka (approved):

mmorhun (approved):

chmeliik:
Could you give some more context about why the revert is needed?

zregvart:
> Could you give some more context about why the revert is needed?

The commit/description is a bit terse as I just used the revert option on GitHub

We had report that the EC started failing with:
```
rego_type_error: package annotation redeclared
```

Seems that we pass in the same Rego file twice to the OPA engine and this causes a compile error. Still investigating why has this started to happen and why it wasn't caught by the tests.

PR #1511, opened 2024-10-15
feat: Add support for specific GCP format for build-vm-image task

pastequo:
Add supporty for specific GCP disk image format in build-vm-image task

Original support from BIB was made in this PR: https://github.com/osbuild/bootc-image-builder/pull/646

Edit: I also checked the release pipeline, it should be ok to have a .tgz: https://github.com/hacbs-release/app-interface-deployments/blob/main/internal-services/catalog/pulp-push-disk-images-task.yaml#L180-L182

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -21,7 +21,7 @@ spec:
       type: string
     - name: IMAGE_TYPE
       type: string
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

pastequo:
I'm anticipating this change in BIB: https://github.com/osbuild/images/pull/987 and use right away the `.tar.gz` file suffix. (As of today it's `.tgz`)

PR #1513, opened 2024-10-16
[ISV-5221][ISV-5222] Add a new step into the “build-index-image” to build and push index sbom.

haripate:
This PR is related with two JIRA Issues:
https://issues.redhat.com/browse/ISV-5221
https://issues.redhat.com/browse/ISV-5222

### DIFF
--- a/task/build-image-index/0.1/README.md
+++ b/task/build-image-index/0.1/README.md
@@ -20,4 +20,5 @@ This takes existing Image Manifests and combines them in an Image Index.
 |IMAGE_URL|Image repository and tag where the built image was pushed|
 |IMAGES|List of all referenced image manifests|
 |IMAGE_REF|Image reference of the built image containing both the repository and the digest|
+|SBOM_BLOB_URL|Reference of SBOM blob digest to enable digest-based verification from provenance|
 
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -47,6 +47,13 @@ spec:
     name: IMAGES
   - description: Image reference of the built image containing both the repository and the digest
     name: IMAGE_REF
+  - name: SBOM_BLOB_URL
+    description: Reference of SBOM blob digest to enable digest-based verification from provenance
+    type: string
+  volumes:
+    - name: shared-dir
+      emptyDir: {}
+
   stepTemplate:
     env:
     - name: BUILDAH_FORMAT
@@ -61,6 +68,9 @@ spec:
       value: $(params.ALWAYS_BUILD_INDEX)
     - name: STORAGE_DRIVER
       value: $(params.STORAGE_DRIVER)
+    volumeMounts:
+      - name: shared-dir
+        mountPath: /index-build-data
   steps:
   - image: quay.io/konflux-ci/buildah-task:latest@sha256:5cbd487022fb7ac476cbfdea25513b810f7e343ec48f89dc6a4e8c3c39fa37a2
     # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
@@ -143,7 +153,67 @@ spec:
         cat "image-digest"
       } > "$(results.IMAGE_REF.path)"
       echo -n "${image_manifests:1:-1}" > "$(results.IMAGES.path)"
+
+      IMAGE_DIGEST=$(cat image-digest)
+
+      INDEX_IMAGE_PULLSPEC="${IMAGE}@${IMAGE_DIGEST}"
+      buildah manifest inspect "$INDEX_IMAGE_PULLSPEC" > /index-build-data/manifest_data.json
     securityContext:
       capabilities:
         add:
           - SETFCAP
+
+  - image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:3b219e0610c06401bb5bd355a4bdfeb7f6700f2bef66f89316739d4aae96c89d
+    name: create-sbom
+    computeResources:
+      limits:
+        memory: 512Mi
+        cpu: 200m
+      requests:
+        memory: 256Mi
+        cpu: 100m
+    script: |
+      #!/bin/bash
+      set -e
+
+      MANIFEST_DATA_FILE="/index-build-data/manifest_data.json"
+      if [ ! -f "$MANIFEST_DATA_FILE" ]; then
+        echo "The manifest_data.json file does not exist. Skipping the SBOM creation..."
+        exit 0
+      fi
+
+      IMAGE_URL="$(cat "$(results.IMAGE_URL.path)")"
+      IMAGE_DIGEST="$(cat "$(results.IMAGE_DIGEST.path)")"
+      echo "Creating SBOM result file..."
+      python3 index_image_sbom_script.py \
+        --image-index-url "$IMAGE_URL" \
+        --image-index-digest "$IMAGE_DIGEST" \
+        --inspect-input-file "$MANIFEST_DATA_FILE" \
+        --output-path /index-build-data/sbom-results.json
+
+  - name: upload-sbom
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+    script: |
+      #!/bin/bash
+      set -e
+
+      SBOM_RESULT_FILE="/index-build-data/sbom-results.json"
+      if [ ! -f "$SBOM_RESULT_FILE" ]; then
+        echo "The sbom_results.json file does not exists. Skipping the SBOM upload..."
+        exit 0
+      fi
+
+      cosign attach sbom --sbom "$SBOM_RESULT_FILE" --type spdx "$(cat "$(results.IMAGE_REF.path)")"
+
+      # Remove tag from IMAGE while allowing registry to contain a port number.
+      sbom_repo="${IMAGE%:*}"
+      sbom_digest="$(sha256sum "$SBOM_RESULT_FILE" | cut -d' ' -f1)"
+      # The SBOM_BLOB_URL is created by `cosign attach sbom`.
+      echo -n "${sbom_repo}@sha256:${sbom_digest}" | tee "$(results.SBOM_BLOB_URL.path)"
+    computeResources:
+      limits:
+        memory: 512Mi
+        cpu: 200m
+      requests:
+        memory: 256Mi
+        cpu: 100m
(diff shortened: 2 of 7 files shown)


task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +162,140 @@ spec:
...
+      python3 /scripts/index-image-sbom-script.py \
+        --image-index-url "$IMAGE_URL" \
+        --image-index-digest "$IMAGE_DIGEST" \
+        --inspect-input-file manifest_data.json
+        --output-path sbom-results.json
+    workingDir: $(workspaces.source.path)
+    securityContext:
+      runAsUser: 0
+
+  - name: inject-sbom-and-push
```
Allda, line 192: What is this step used for in the context of the index image? Do we need it here?

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +162,140 @@ spec:
...
+      requests:
+        memory: 256Mi
+        cpu: 100m
+    env:
+    - name: IMAGE_URL
+      value: $(results.IMAGE_URL.path)
+    - name: IMAGE_DIGEST
+      value: $(results.IMAGE_DIGEST.path)
+    script: |
+      TOADD="${IMAGE_URL}@${IMAGE_DIGEST}"
```
Allda, line 181: `TOADD` variable name doesn't make much sense to me in the context of this step. `INDEX_IMAGE_PULLSPEC` or similar makes more sense.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +162,140 @@ spec:
...
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+    script: |
+      ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+      if [ -f "$ca_bundle" ]; then
+        echo "INFO: Using mounted CA bundle: $ca_bundle"
+        cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+        update-ca-trust
+      fi
+
+      cosign attach sbom --sbom sbom-cyclonedx.json --type cyclonedx "$(cat "$(results.IMAGE_REF.path)")"
```
Allda, line 278: The format of the sbom generated by the script is spdx.

Allda: The `sbom-cyclonedx.json` is not defined anywhere in this task. I assume you mean `sbom-results.json` instead?

task/build-image-index/0.1/build-image-index.yaml
Allda: I thought about this step a little bit more and I think it is not relevant for the index image. This step seems to add sbom into the actual image a push it to the registry. The index image is not an actual image (it doesn't have any layer). It is just a manifest that contains references to a child's images. I think you can remove it completely and also get rid of additional params that you added to the task.

Allda (approved):

task/build-image-index/0.1/README.md
```diff
@@ -20,4 +20,5 @@ This takes existing Image Manifests and combines them in an Image Index.
 |IMAGE_URL|Image repository and tag where the built image was pushed|
 |IMAGES|List of all referenced image manifests|
 |IMAGE_REF|Image reference of the built image containing both the repository and the digest|
+|SBOM_BLOB_URL|Reference, including digest to the SBOM blob|
```
chmeliik, line 23: Could you update this to match the description in the task?

You can do so by running `hack/generate-readme.sh task/build-image-index/0.1/build-image-index.yaml > task/build-image-index/0.1/README.md`

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +150,59 @@ spec:
       capabilities:
         add:
           - SETFCAP
+
+  - image: quay.io/redhat-appstudio/sbom-utility-scripts-image:d4935b35eb0849841412d5d1e481a812e939f2bf
```
chmeliik, line 154: Please use the same image reference as other tasks: `quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:53a3041dff341b7fd1765b9cc2c324625d19e804b2eaff10a6e6d9dcdbde3a91`

If it doesn't include the code you need, you can use a newer version but please pin the image to a digest

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +150,59 @@ spec:
...
+      requests:
+        memory: 256Mi
+        cpu: 100m
+    env:
+    - name: IMAGE_URL
+      value: $(results.IMAGE_URL.path)
+    - name: IMAGE_DIGEST
+      value: $(results.IMAGE_DIGEST.path)
+    script: |
+      INDEX_IMAGE_PULLSPEC="${IMAGE_URL}@${IMAGE_DIGEST}"
```
chmeliik, line 169: Have you tested this? I think this concatenates the result paths, not the result values

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +150,59 @@ spec:
...
+      buildah manifest inspect "$INDEX_IMAGE_PULLSPEC" > manifest_data.json
```
chmeliik, line 170: Does this have to pull the manifest, or is the manifest stored locally? I think it may have to pull it, because `/var/lib/containers` isn't shared between the steps. It would be better to get the inspect data in the previous step, which still has the manifest on disk

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -147,3 +150,59 @@ spec:
...
+      python3 /scripts/index-image-sbom-script.py \
+        --image-index-url "$IMAGE_URL" \
+        --image-index-digest "$IMAGE_DIGEST" \
+        --inspect-input-file manifest_data.json
+        --output-path sbom-results.json
+    workingDir: $(workspaces.source.path)
```
chmeliik, line 176: Let's not write the file into the source workspace, it shouldn't "leak" out of this task

lcarva: This Task does not use any workspaces so this won't work. Let's please also not add workspaces to this Task unless it's absolutely necessary.

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -143,7 +153,60 @@ spec:
...
+      IMAGE_URL="$(cat $(results.IMAGE_URL.path))"
+      IMAGE_DIGEST="$(cat $(results.IMAGE_DIGEST.path))"
+      echo "Creating SBOM result file..."
+      python3 index_image_sbom_script.py \
+        --image-index-url "$IMAGE_URL" \
+        --image-index-digest "$IMAGE_DIGEST" \
+        --inspect-input-file /index-build-data/manifest_data.json \
+        --output-path /index-build-data/sbom-results.json
+    securityContext:
+      runAsUser: 0
```
chmeliik, line 188: I don't think the `runAsUser: 0` would be needed here, the step just writes a file into a pod-local volume

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -72,6 +79,9 @@ spec:
       requests:
         memory: 512Mi
         cpu: 250m
+    volumeMounts:
+      - name: shared-dir
+        mountPath: /index-build-data
```
chmeliik, line 84: Nitpick: you can put this in the `stepTemplate` instead to avoid having to copy it to every step

chmeliik (commented):
Looks good now, tested in https://github.com/chmeliik/multi-arch-konflux-sample/pull/4/files and does what it should :+1: 

Could you address the failing checks?

chmeliik (approved):
Thanks! :tada:

task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -143,7 +153,67 @@ spec:
...
+
+  - name: upload-sbom
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+    script: |
+      #!/bin/bash
+      set -e
+
+      SBOM_RESULT_FILE="/index-build-data/sbom-results.json"
+      if [ ! -f "$SBOM_RESULT_FILE" ]; then
+        echo "The sbom_results.json file does not exists. Skipping the SBOM upload..."
```
lcarva, line 202: Do we need something like `touch "$(results.SBOM_BLOB_URL.path)"` here?

chmeliik: Ah good catch. But I don't think it's actually possible to reach this line, @haripate maybe we can remove the `if [ ! -f "$MANIFEST_DATA_FILE" ]` and `if [ ! -f "$SBOM_RESULT_FILE" ]` conditions?

lcarva: @chmeliik, could it reach here if `ALWAYS_BUILD_INDEX` is true and only one Image Manifest was provided?

chmeliik: Yup, good point. I forgot this task is allowed not to create an image index

chmeliik: I think that without the `touch "$(results.SBOM_BLOB_URL.path)"`, the SBOM_BLOB_URL result will be missing entirely, as opposed to empty if we add the `touch`. Considering no other tasks reference the SBOM_BLOB_URL, that seems fine at the moment.

lcarva: Yeah, seem reasonable.

We may need to revisit this in the future. If this Task does not produce an Image Index, then the SBOM_BLOB_URL should probably point to the SBOM of the Image Manifest.

TBH, I don't really understand why this Task has the `ALWAYS_BUILD_INDEX` parameter. If you don't want to build an Image Index, then maybe just don't include the Task? We're probably making things more difficult to use and manage with such parameters.

chmeliik: I think the idea was that we would have just one build pipeline able to handle single-arch and multi-arch builds without any customization on the user's part. I don't fully remember, @arewm was that it?

arewm: There were at least two reasons why I added this: 
* Single arch pipelines should still be able to easily generate an Image Index if it becomes a requirement. This option can provide consistency with the internal build pipeline where every image generated is behind a manifest list. It also better supports operator-based installations where images generally shouldn't be referenced by image manifest.
* Adding the image index generation to the template improved the kustomization for the multi-arch builds (i.e. the image index and image manifest generating tasks can be next to each other).

Now that we support a matrix of size 1, we could potentially change the pipeline default to the multi-arch pipeline as these will build single-arch by default. We can then change the single-arch pipeline to not have the image index and use that only for builds requiring an Image Manifest.

arewm:
FYI, the reversion of this PR is being proposed in https://github.com/konflux-ci/build-definitions/pull/1565.

PR #1514, opened 2024-10-16
Add show error option to curl

mkosiarc:
This work was triggered by KFLUXBUGS-1285.
User was trying to use dockerfile from an url, but it was failing for a
uknown reason and there was not any information as to why it might be
failing. It only showed the last successful log:

"Fetch Dockerfile from <URL>".

The true reason behind that failure is that the user was trying to fetch
the Dockerfile from an internal gitlab. That instance is accessible only
from internal network and with a certificate issued by Red Hat is needed.
Dockerfile failed "violently", without returning any error msg and did not return any http code.
Also we are using "set -e" which means, we exit on failure.

The reason we did not have any error message from the curl is the "-s"
(silent) option. However, adding -S (show error) means that we will be
outputting errors on failure without showing the curl progress bar

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -268,12 +268,12 @@ spec:
         elif echo "$DOCKERFILE" | grep -q "^https\?://"; then
           echo "Fetch Dockerfile from $DOCKERFILE"
           dockerfile_path=$(mktemp --suffix=-Dockerfile)
-          http_code=$(curl -s -L -w "%{http_code}" --output "$dockerfile_path" "$DOCKERFILE")
+          http_code=$(curl -s -S -L -w "%{http_code}" --output "$dockerfile_path" "$DOCKERFILE")
           if [ $http_code != 200 ]; then
             echo "No Dockerfile is fetched. Server responds $http_code"
             exit 1
           fi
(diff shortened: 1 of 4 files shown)


mkosiarc:
> > In that particular case it was not provided, so the curl command that was fetching the Dockerfile failed "violently" instead of returning any http code.
> 
> Failed violently and didn't print any error message? Or was it just hidden further up the logs so the user didn't notice it?

It did not print anything at all. This is easily replicable by using any Dockerfile from the internal gitlab instance

mkosiarc:
I believe because the curl command is wrapped in http_code=$().
When I was testing it without this, it behaved differently.

chmeliik:
> I believe it is because the curl command is wrapped in http_code=$(). When I was testing it without this, it behaved differently.

Ha, could be. In which case IMO we should print the curl output on failure

mkosiarc:
The way it is setup right now

> > I believe it is because the curl command is wrapped in http_code=$(). When I was testing it without this, it behaved differently.
> 
> Ha, could be. In which case IMO we should print the curl output on failure

That is actually saner approach. The reason was also the -s option. However, it seems that -s can be used with -S, that actually outputs the curl error.

chmeliik (approved):

mmorhun (approved):

tkdchen (approved):

PR #1518, opened 2024-10-18
Change the default build-platforms to only build locallly

arewm:
Now that we can support a matrix of size one out of the box, changing the default configuraiton for multi-platform builds to only run in-cluster.


tkdchen (approved):

chmeliik (approved):

arewm:
Merging this PR without a resolution to konflux-ci/konflux-ci#567 as the multi-platform pipeline cannot be used by default in deployments as it requires the additional configuration of the multi-platform controller.

PR #1522, opened 2024-10-21
Rename CI task that builds appstudio-utils image

tkdchen:
The original name build-container does not reflect what the task does.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -54,13 +54,13 @@ spec:
             workspace: workspace
       - name: sast-unicode-check
         runAfter:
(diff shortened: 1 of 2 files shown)


chmeliik:
```
admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: couldn't add link between build-bundles and build-container: task build-bundles depends on build-container but build-container wasn't present in Pipeline: spec.pipelineSpec.tasks
```

It looks line the `build-container` name was referenced elsewhere in the pipeline

chmeliik (approved):

PR #1523, opened 2024-10-21
CI check for missing TA Task variants

zregvart:
A script and a GitHub workflow change to check if a new Task was added that uses a PVC-backed workspace without a Trusted Artifacts Task variant.

Reference: https://issues.redhat.com/browse/EC-934

### DIFF
--- a/hack/missing-ta-tasks.sh
+++ b/hack/missing-ta-tasks.sh
@@ -1,0 +1,115 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+shopt -s globstar
+
+git_root=$(git rev-parse --show-toplevel)
+policy_file="${git_root}/policies/all-tasks.yaml"
+
+tmp_files=()
+trap 'rm "${tmp_files[@]}" > /dev/null 2>&1' EXIT
+
+# Tasks that are currently missing Trusted Artifact variant
+todo=(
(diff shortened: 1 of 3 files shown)


hack/missing-ta-tasks.sh
```diff
@@ -0,0 +1,106 @@
...
+set -o pipefail
+shopt -s globstar
+
+git_root=$(git rev-parse --show-toplevel)
+
+tmp_files=()
+trap 'rm "${tmp_files[@]}" > /dev/null 2>&1' EXIT
+
+# names of workspaces that will not be mounted via PVC, as inline array
+non_pvc_workspaces='["ssh-directory", "basic-auth", "gitops-auth", "git-basic-auth", "netrc"]'
```
lcarva, line 14: Not sure if it's worth it, but it would be nice if this list was generated from https://github.com/konflux-ci/build-definitions/blob/7fe56f111b0ea0621efa1e3e530003d547091e28/policies/all-tasks.yaml#L10-L14.

lcarva (approved):

hack/missing-ta-tasks.sh
zregvart: I think this is a good idea, incorporated in [`324b27e` (#1523)](https://github.com/konflux-ci/build-definitions/pull/1523/commits/324b27e6d1c737730d051efa1c5763f0da47a670)

PR #1526, opened 2024-10-22
Document trusted-artifacts Task generator

zregvart:
Adds the documentation for the trusted-artifacts Task generator.

Reference: https://issues.redhat.com/browse/EC-933


lcarva (approved):
Nice!

PR #1528, opened 2024-10-23
subscription-manager pre-registration

brianwcook:
This add support for pre-registering with subscription-manager in the buildah tasks (0.2 only). When activation keys are provided the buildah-task container will register and then provide certs to the buildah env via volume mounts so that users do not need to add subscription-manager register commands to the Containerfiles. This makes downstreaming somewhat easier and addresses negative user feedback.

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -229,7 +229,7 @@ spec:
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
         - $(params.CACHI2_ARTIFACT)=/var/workdir/cachi2
     - name: build
-      image: quay.io/konflux-ci/buildah-task:latest@sha256:5cbd487022fb7ac476cbfdea25513b810f7e343ec48f89dc6a4e8c3c39fa37a2
+      image: quay.io/konflux-ci/buildah-task:latest@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c
       args:
         - --build-args
         - $(params.BUILD_ARGS[*])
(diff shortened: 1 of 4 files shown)


skateman:
Tested with the buildah task, works well 🎉

brianwcook:
opened https://github.com/konflux-ci/buildah-container/pull/98 to get subscription-manager included in buildah-task

ezr-ondrej:
Works on ubi9 as well :)

brianwcook:
rebased and removed dnf install.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -429,16 +429,33 @@ spec:
...
-        # when activation keys are used an empty directory on shared emptydir volume to "/etc/pki/entitlement" to prevent certificates from being included in the produced container
-        # To use activation key file 'org' must exist, which means the key 'org' must exist in the key/value secret
-
+        # 1. do not enable activation key and entitlement at same time. If both vars are provided, prefer activation key.
+        # 2. Activation-keys will be used when the key 'org' exists in the activation key secret.
+        # 3. try to pre-register and mount files to the correct location so that users do no need to modify Dockerfiles.
+        # 3. If the Dockerfile contains the string "subcription-manager register", add the activation-keys volume
+        #    to buildah but don't pre-register for backwards compatibility. In this case mount an empty directory on
+        #    shared emptydir volume to "/etc/pki/entitlement" to prevent certificates from being included in the produced
+        #    container.
```
chmeliik, line 438: Without this check, would something we mount break the ability to `subcription-manager register`?

The string check is very flimsy, I would rather drop it if possible

(In fact, currently the check doesn't work at all, it's always `grep: register: No such file or directory`)

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -429,16 +429,33 @@ spec:
...
           cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
           mkdir /shared/rhsm-tmp
           VOLUME_MOUNTS+=(--volume /tmp/activation-key:/activation-key -v /shared/rhsm-tmp:/etc/pki/entitlement:Z)
           echo "Adding activation key to the build"
 
+          if ! grep subscription-manager register "$dockerfile_path"; then
+            # user is not running registration in the Containerfile: pre-register.
+            echo "Pre-registering with subscription manager."
+            subscription-manager register --org "$(cat /tmp/activation-key/org)" --activationkey "$(cat /tmp/activation-key/activationkey)"
+            REGISTERED=$?
```
chmeliik, line 451:
```suggestion
            trap 'subscription-manager unregister' EXIT
```
This will make us unregister always, even if the build fails

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -429,16 +429,33 @@ spec:
...
+            # copy generated certificates to /shared/rhsm-tmp
+            cp /etc/pki/entitlement/*.pem /shared/rhsm-tmp
+
+            # and then mount get /etc/rhsm/ca/redhat-uep.pem into /run/secrets/rhsm/ca
+            VOLUME_MOUNTS+=(--volume /etc/rhsm/ca/redhat-uep.pem:/run/secrets/rhsm/ca/redhat-uep.pem)
+          fi
```
chmeliik, line 457: So if I understand correctly:
* we are generating entitlement certificates in a Fedora container image (the one that runs this script)
* the Fedora image installs `subscription-manager`, which comes with the `redhat-uep.pem` file
* we're mounting those into the build, which could be based on ubi8, ubi9, ubiX-minimal, ...

@ptoscano could you check if we're doing something sensible here?

brianwcook:
i'm doing some add'l testing, don't merge for now.

brianwcook:
this is ready to go. tested with:
* entitlement certificates as key value secrets
* activation keys secret + registration in Dockerfile
* activation keys as secrets and auto-pre-registration

arewm (approved):
LGTM

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
brianwcook: I made it more robust and fixed the bug before merging.

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
brianwcook: did not get this in but I will file a follow up PR. testing all cases manually was getting complicated

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
brianwcook: I _think_ that the redhat-uep.pem is actually generated during the registration, but the rest of what you say is accurate

ptoscano: > I _think_ that the redhat-uep.pem is actually generated during the registration,

```bash
$ rpm -ql subscription-manager-rhsm-certificates 
/etc/rhsm
/etc/rhsm/ca
/etc/rhsm/ca/redhat-entitlement-authority.pem
/etc/rhsm/ca/redhat-uep.pem
/usr/share/licenses/subscription-manager-rhsm-certificates
/usr/share/licenses/subscription-manager-rhsm-certificates/COPYING
```

ptoscano: @chmeliik other than what I wrote above about "where redhat-uep.pem comes from", what you said seems correct

chmeliik: And can it cause some problems? Mounting `redhat-uep.pem` from the Fedora version of the package? Or does it generally not matter?

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -429,16 +429,34 @@ spec:
...
+        REGISTERED="false"
         if [ -e /activation-key/org ]; then
           cp -r --preserve=mode "$ACTIVATION_KEY_PATH" /tmp/activation-key
           mkdir /shared/rhsm-tmp
           VOLUME_MOUNTS+=(--volume /tmp/activation-key:/activation-key -v /shared/rhsm-tmp:/etc/pki/entitlement:Z)
           echo "Adding activation key to the build"
 
+          if ! grep subscription-manager "$dockerfile_path" | grep -q register; then
+            # user is not running registration in the Containerfile: pre-register.
+            echo "Pre-registering with subscription manager."
```
ptoscano, line 449: this ought to exclude comments, otherwise a line such as
```
# install packages with no need to register using subscription-manager
dnf -y install hello
```
will trigger this detection to think the dockerfile is registering on its own

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
ptoscano: It should not be an issue, since `redhat-uep.pem` is shipped by `subscription-manager-rhsm-certificates`, which would need to be upgraded in case any of the certificates it ships needs to be changed. In that case, the update of the package ought to be done enough time in advance to ensure all the users (RHEL and not only, Fedora included) get the new package installed.

PR #1529, opened 2024-10-23
Added Nvidia GPU support to the buildah-remote task

syedriko:
Added Nvidia GPU support to the buildah-remote task

### DIFF
--- a/task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
+++ b/task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
@@ -445,6 +445,11 @@ spec:
       REMOTESSHEOF
       chmod +x scripts/script-build.sh
 
+      PODMAN_NVIDIA_ARGS=()
+      if [[ "$PLATFORM" == "linux-g"* ]]; then
+          PODMAN_NVIDIA_ARGS+=("--device=nvidia.com/gpu=all" "--security-opt=label=disable")
+      fi
+
       if ! [[ $IS_LOCALHOST ]]; then
         rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
         ssh $SSH_ARGS "$SSH_HOST" $PORT_FORWARD podman  run $PODMAN_PORT_FORWARD \
@@ -477,7 +482,7 @@ spec:
           -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
           -v "$BUILD_DIR/results/:/tekton/results:Z" \
           -v "$BUILD_DIR/scripts:/scripts:Z" \
-          --user=0  --rm  "$BUILDER_IMAGE" /scripts/script-build.sh "$@"
+          --user=0 "${PODMAN_NVIDIA_ARGS[@]}" --rm "$BUILDER_IMAGE" /scripts/script-build.sh "$@"
         rsync -ra "$SSH_HOST:$BUILD_DIR/volumes/shared/" /shared/
         rsync -ra "$SSH_HOST:$BUILD_DIR/volumes/workdir/" /var/workdir/
         rsync -ra "$SSH_HOST:$BUILD_DIR/results/" "/tekton/results/"
--- a/task/buildah-remote/0.2/buildah-remote.yaml
+++ b/task/buildah-remote/0.2/buildah-remote.yaml
@@ -552,6 +552,11 @@ spec:
       REMOTESSHEOF
       chmod +x scripts/script-build.sh
 
+      PODMAN_NVIDIA_ARGS=()
+      if [[ "$PLATFORM" == "linux-g"* ]]; then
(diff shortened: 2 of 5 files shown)


brianwcook:
I left you a couple reviews. The buildah-remote tasks are generated by running /hack/generate-buildah-remote.sh in this repo.  This ensures they stay consistent with the normal buildah task. You need to modify generate-buildah-remote.sh so that when you run it, it produces the same diff this PR has. Once you run it, the PR should have 3 changed files: the generate script, buildah-remote/0.1/buildah-remote.yaml and buildah-remote/0.2/buildah-remote.yaml.
 
 After that, you also need to run the /hack/generate-ta-tasks.sh. which will update 2 more files (trusted artifacts versions of the 2 buildah remote tasks). 

Summary: you will modify 1 file (main.go), running two generate commands and adding all those changes here.

syedriko:
Thanks, @brianwcook! PTAL

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -445,6 +445,11 @@ spec:
       REMOTESSHEOF
       chmod +x scripts/script-build.sh
 
+      PODMAN_NVIDIA_ARGS=()
+      if [[ "$PLATFORM" == "linux-g"* ]]; then
```
arewm, line 449: I have tried in the past to _not_ depend on the semantics of the PLATFORM parameter. @ifireball @mshaposhnik, what do you think?

The use of the PLATFORM parameter like this would fall in line with the functionality requested in https://issues.redhat.com/browse/KONFLUX-4073.

chmeliik:
Could you describe what these changes do, how and why? The code change on its own doesn't give me much to work with

task/buildah-remote-oci-ta/0.1/buildah-remote-oci-ta.yaml
```diff
@@ -445,6 +445,11 @@ spec:
...
+          PODMAN_NVIDIA_ARGS+=("--device nvidia.com/gpu=all" "--security-opt=label=disable")
```
chmeliik, line 450: What are the implications of `--security-opt=label=disable`?

chmeliik:
I would also consider dropping this PR in favor of https://github.com/konflux-ci/build-definitions/pull/1530, although that one seems like it potentially gives the user _too much_ control

syedriko:
> Could you describe what these changes do, how and why? The code change on its own doesn't give me much to work with

The goal here is to allow Konflux builds to access Nvidia GPUs on machines so equipped. An example is running PyTorch during container build - https://github.com/openshift/lightspeed-rag-content/blob/main/Containerfile.

This PR is a building block towards support of this scenario. The others are AWS instance type(s) in multi-platform controller https://github.com/redhat-appstudio/infra-deployments/blob/0b936310854c7b4031b967eda33ad8399f12da60/components/multi-platform-controller/production/common/host-config.yaml#L528 and an AMI with Nvidia drivers.

This PR, for platforms that start with "linux-g", tells podman to pass though Nvidia GPU devices to the containers it runs.

>What are the implications of --security-opt=label=disable?
Not too sure about it beyond the obvious, but this same from Nvidia docs https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.14.2/cdi-support.html

syedriko:
@brianwcook @chmeliik @arewm It looks to me like https://github.com/konflux-ci/build-definitions/pull/1530 has morphed into something we can't use instead of this PR. Could you PTAL and see if we can revive this PR.

brianwcook:
I am +1 to merging this. We can revisit the enablement logic (vm's with instance types starting with 'g') at a later date if / when MPC changes how to label platforms, probably with no impact to end users.  In addition the impact from disabling security labeling looks minimal since we 1) do not reuse build VMs 2) we do not run concurrent builds on a VM and 3) this only applies to remote builders, not OpenShift based builds.

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -551,6 +551,11 @@ spec:
       REMOTESSHEOF
       chmod +x scripts/script-build.sh
 
+      PODMAN_NVIDIA_ARGS=()
+      if [[ "$PLATFORM" == "linux-g"* ]]; then
+          PODMAN_NVIDIA_ARGS+=("--device nvidia.com/gpu=all" "--security-opt=label=disable")
```
chmeliik, line 556: Won't this expand to `podman run '--device nvidia.com/gpu=all'` ?

```
Error: unknown flag: --device nvidia.com/gpu
See 'podman run --help'
```

```suggestion
          PODMAN_NVIDIA_ARGS+=("--device=nvidia.com/gpu=all" "--security-opt=label=disable")
```

syedriko: Of course. Thank you! Fixed.

chmeliik (approved):
LGTM :+1:

PR #1530, opened 2024-10-24
Add PRIVILEGED_NESTED to the buildah tasks

shi2wei3:
Resolve https://issues.redhat.com/browse/BIFROST-475

To build bootc case image with buildah, we need additional options to be supported.

xref: https://gitlab.com/fedora/bootc/base-images/-/blob/main/Containerfile?ref_type=heads#L6

### DIFF
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -61,6 +61,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |LABELS| Additional key=value labels that should be applied to the image| []| |
 |PLATFORM| The platform to build on| None| |
 |PREFETCH_INPUT| In case it is not empty, the prefetched content should be made available to the build.| | '$(params.prefetch-input)'|
+|PRIVILEGED_NESTED| Whether to enable privileged mode| false| |
 |SKIP_UNUSED_STAGES| Whether to skip stages in Containerfile that seem unused by subsequent stages| true| |
 |SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| None| '$(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)'|
 |SQUASH| Squash all new and previous layers added as a part of this build, as per --squash| false| |
--- a/task/buildah-remote/0.2/buildah-remote.yaml
+++ b/task/buildah-remote/0.2/buildah-remote.yaml
@@ -119,6 +119,10 @@ spec:
     description: Additional key=value labels that should be applied to the image
     name: LABELS
     type: array
+  - default: "false"
+    description: Whether to enable privileged mode
+    name: PRIVILEGED_NESTED
+    type: string
   - description: The platform to build on
     name: PLATFORM
     type: string
@@ -185,6 +189,8 @@ spec:
       value: $(params.SQUASH)
     - name: SKIP_UNUSED_STAGES
       value: $(params.SKIP_UNUSED_STAGES)
+    - name: PRIVILEGED_NESTED
+      value: $(params.PRIVILEGED_NESTED)
     - name: BUILDER_IMAGE
       value: quay.io/konflux-ci/buildah-task:latest@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c
     - name: PLATFORM
@@ -243,6 +249,7 @@ spec:
         export BUILD_DIR=$(cat /ssh/user-dir)
         export SSH_ARGS="-o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10"
         echo "$BUILD_DIR"
+        # shellcheck disable=SC2086
         ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/volumes"
 
         PORT_FORWARD=""
@@ -381,6 +388,12 @@ spec:
       fi
 
       BUILDAH_ARGS+=("${BUILD_ARG_FLAGS[@]}")
+
+      if [ "${PRIVILEGED_NESTED}" == "true" ]; then
+        BUILDAH_ARGS+=("--security-opt=label=disable")
+        BUILDAH_ARGS+=("--cap-add=all")
+        BUILDAH_ARGS+=("--device=/dev/fuse")
+      fi
 
       if [ -n "${ADD_CAPABILITIES}" ]; then
         BUILDAH_ARGS+=("--cap-add=${ADD_CAPABILITIES}")
@@ -552,7 +565,16 @@ spec:
       chmod +x scripts/script-build.sh
 
       if ! [[ $IS_LOCALHOST ]]; then
+        PRIVILEGED_NESTED_FLAGS=()
+        if [[ "${PRIVILEGED_NESTED}" == "true" ]]; then
+          # This is a workaround for building bootc images because the cache filesystem (/var/tmp/ on the host) must be a real filesystem that supports setting SELinux security attributes.
+          # https://github.com/coreos/rpm-ostree/discussions/4648
+          # shellcheck disable=SC2086
+          ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/var/tmp"
+          PRIVILEGED_NESTED_FLAGS=(--privileged --mount "type=bind,source=$BUILD_DIR/var/tmp,target=/var/tmp,relabel=shared")
+        fi
         rsync -ra scripts "$SSH_HOST:$BUILD_DIR"
+        # shellcheck disable=SC2086
         ssh $SSH_ARGS "$SSH_HOST" $PORT_FORWARD podman  run $PODMAN_PORT_FORWARD \
           --tmpfs /run/secrets \
           -e BUILDAH_FORMAT="$BUILDAH_FORMAT" \
@@ -574,6 +596,7 @@ spec:
           -e ADD_CAPABILITIES="$ADD_CAPABILITIES" \
           -e SQUASH="$SQUASH" \
           -e SKIP_UNUSED_STAGES="$SKIP_UNUSED_STAGES" \
+          -e PRIVILEGED_NESTED="$PRIVILEGED_NESTED" \
           -e COMMIT_SHA="$COMMIT_SHA" \
           -e DOCKERFILE="$DOCKERFILE" \
           -v "$BUILD_DIR/workspaces/source:$(workspaces.source.path):Z" \
@@ -585,6 +608,7 @@ spec:
           -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
           -v "$BUILD_DIR/results/:/tekton/results:Z" \
           -v "$BUILD_DIR/scripts:/scripts:Z" \
+          "${PRIVILEGED_NESTED_FLAGS[@]}" \
           --user=0  --rm  "$BUILDER_IMAGE" /scripts/script-build.sh "$@"
         rsync -ra "$SSH_HOST:$BUILD_DIR/workspaces/source/" "$(workspaces.source.path)/"
         rsync -ra "$SSH_HOST:$BUILD_DIR/volumes/shared/" /shared/
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -107,6 +107,10 @@ spec:
     description: Additional key=value labels that should be applied to the image
     type: array
     default: []
+  - name: PRIVILEGED_NESTED
+    description: Whether to enable privileged mode
+    type: string
+    default: "false"
 
   results:
   - description: Digest of the image just built
@@ -166,6 +170,8 @@ spec:
       value: $(params.SQUASH)
     - name: SKIP_UNUSED_STAGES
       value: $(params.SKIP_UNUSED_STAGES)
+    - name: PRIVILEGED_NESTED
+      value: $(params.PRIVILEGED_NESTED)
 
   steps:
   - image: quay.io/konflux-ci/buildah-task:latest@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c
@@ -302,6 +308,12 @@ spec:
       fi
 
       BUILDAH_ARGS+=("${BUILD_ARG_FLAGS[@]}")
+
+      if [ "${PRIVILEGED_NESTED}" == "true" ]; then
+        BUILDAH_ARGS+=("--security-opt=label=disable")
+        BUILDAH_ARGS+=("--cap-add=all")
+        BUILDAH_ARGS+=("--device=/dev/fuse")
+      fi
 
       if [ -n "${ADD_CAPABILITIES}" ]; then
         BUILDAH_ARGS+=("--cap-add=${ADD_CAPABILITIES}")
--- a/pipelines/docker-build-oci-ta/README.md
+++ b/pipelines/docker-build-oci-ta/README.md
@@ -58,6 +58,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE_EXPIRES_AFTER| Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | '$(params.image-expires-after)'|
 |LABELS| Additional key=value labels that should be applied to the image| []| |
 |PREFETCH_INPUT| In case it is not empty, the prefetched content should be made available to the build.| | '$(params.prefetch-input)'|
+|PRIVILEGED_NESTED| Whether to enable privileged mode| false| |
 |SKIP_UNUSED_STAGES| Whether to skip stages in Containerfile that seem unused by subsequent stages| true| |
 |SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| None| '$(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)'|
 |SQUASH| Squash all new and previous layers added as a part of this build, as per --squash| false| |
--- a/pipelines/docker-build/README.md
+++ b/pipelines/docker-build/README.md
@@ -57,6 +57,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE_EXPIRES_AFTER| Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | '$(params.image-expires-after)'|
 |LABELS| Additional key=value labels that should be applied to the image| []| |
 |PREFETCH_INPUT| In case it is not empty, the prefetched content should be made available to the build.| | '$(params.prefetch-input)'|
+|PRIVILEGED_NESTED| Whether to enable privileged mode| false| |
 |SKIP_UNUSED_STAGES| Whether to skip stages in Containerfile that seem unused by subsequent stages| true| |
 |SQUASH| Squash all new and previous layers added as a part of this build, as per --squash| false| |
 |STORAGE_DRIVER| Storage driver to configure for buildah| vfs| |
--- a/pipelines/fbc-builder/README.md
+++ b/pipelines/fbc-builder/README.md
@@ -55,6 +55,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE_EXPIRES_AFTER| Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | '$(params.image-expires-after)'|
 |LABELS| Additional key=value labels that should be applied to the image| []| |
 |PREFETCH_INPUT| In case it is not empty, the prefetched content should be made available to the build.| | |
+|PRIVILEGED_NESTED| Whether to enable privileged mode| false| |
 |SKIP_UNUSED_STAGES| Whether to skip stages in Containerfile that seem unused by subsequent stages| true| |
 |SQUASH| Squash all new and previous layers added as a part of this build, as per --squash| false| |
 |STORAGE_DRIVER| Storage driver to configure for buildah| vfs| |
--- a/task-generator/remote/main.go
+++ b/task-generator/remote/main.go
@@ -164,6 +164,7 @@ if ! [[ $IS_LOCALHOST ]]; then
   export BUILD_DIR=$(cat /ssh/user-dir)
   export SSH_ARGS="-o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10"
   echo "$BUILD_DIR"
+  # shellcheck disable=SC2086
   ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/workspaces" "$BUILD_DIR/scripts" "$BUILD_DIR/volumes"
 
   PORT_FORWARD=""
@@ -234,13 +235,23 @@ if ! [[ $IS_LOCALHOST ]]; then
 			}
 		}
 		ret += "\nif ! [[ $IS_LOCALHOST ]]; then"
+		ret += "\n"
+		ret += `  PRIVILEGED_NESTED_FLAGS=()
+  if [[ "${PRIVILEGED_NESTED}" == "true" ]]; then
+    # This is a workaround for building bootc images because the cache filesystem (/var/tmp/ on the host) must be a real filesystem that supports setting SELinux security attributes.
+    # https://github.com/coreos/rpm-ostree/discussions/4648
+    # shellcheck disable=SC2086
+    ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/var/tmp"
+    PRIVILEGED_NESTED_FLAGS=(--privileged --mount "type=bind,source=$BUILD_DIR/var/tmp,target=/var/tmp,relabel=shared")
+  fi`
 		ret += "\n  rsync -ra scripts \"$SSH_HOST:$BUILD_DIR\""
 		containerScript := "scripts/script-" + step.Name + ".sh"
 		for _, e := range step.Env {
 			env += "    -e " + e.Name + "=\"$" + e.Name + "\" \\\n"
 		}
 		podmanArgs += "    -v \"$BUILD_DIR/scripts:/scripts:Z\" \\\n"
-		ret += "\n  ssh $SSH_ARGS \"$SSH_HOST\" $PORT_FORWARD podman  run " + env + "" + podmanArgs + "    --user=0  --rm  \"$BUILDER_IMAGE\" /" + containerScript + ` "$@"`
+		podmanArgs += "    \"${PRIVILEGED_NESTED_FLAGS[@]}\" \\\n"
+		ret += "\n  # shellcheck disable=SC2086\n  ssh $SSH_ARGS \"$SSH_HOST\" $PORT_FORWARD podman  run " + env + "" + podmanArgs + "    --user=0  --rm  \"$BUILDER_IMAGE\" /" + containerScript + ` "$@"`
 
 		// Sync the contents of the workspaces back so subsequent tasks can use them
 		for _, workspace := range task.Spec.Workspaces {
--- a/task/buildah-oci-ta/0.2/README.md
+++ b/task/buildah-oci-ta/0.2/README.md
@@ -23,6 +23,7 @@ When prefetch-dependencies task was activated it is using its artifacts to run b
 |IMAGE_EXPIRES_AFTER|Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.|""|false|
 |LABELS|Additional key=value labels that should be applied to the image|[]|false|
 |PREFETCH_INPUT|In case it is not empty, the prefetched content should be made available to the build.|""|false|
+|PRIVILEGED_NESTED|Whether to enable privileged mode|false|false|
 |SKIP_UNUSED_STAGES|Whether to skip stages in Containerfile that seem unused by subsequent stages|true|false|
 |SOURCE_ARTIFACT|The Trusted Artifact URI pointing to the artifact with the application source code.||true|
 |SQUASH|Squash all new and previous layers added as a part of this build, as per --squash|false|false|
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -82,6 +82,10 @@ spec:
         be made available to the build.
       type: string
       default: ""
+    - name: PRIVILEGED_NESTED
+      description: Whether to enable privileged mode
+      type: string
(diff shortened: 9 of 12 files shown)


chmeliik:
> I have no idea about how to test my changes by myself because I don't know how to build a buildah task container image, it uses the released one in the task definition.

You can use the [tkn](https://github.com/tektoncd/cli/releases) CLI tool:

```
tkn bundle push quay.io/acmiel-test/tekton-catalog/task-buildah:test -f task/buildah/0.2/buildah.yaml
```

Push it to your own quay.io namespace, make the repo public, use the task in your pipeline

chmeliik:
Would this change potentially allow users to do insecure things / something that would bypass the hermeticity of the build / break SBOM completeness?

shi2wei3:
Those are all the options we need for bootc image right now

`buildah build --security-opt=label=disable --cap-add=all --device /dev/fuse ...`

chmeliik:
> > Would this change potentially allow users to do insecure things / something that would bypass the hermeticity of the build / break SBOM completeness?
> 
> well...will there be enough information about what options were used that we can control it via enterprise contract policies?

I believe so. Should we have the policies ready before we allow production use?

brianwcook:
> > > Would this change potentially allow users to do insecure things / something that would bypass the hermeticity of the build / break SBOM completeness?
> > 
> > 
> > well...will there be enough information about what options were used that we can control it via enterprise contract policies?
> 
> I believe so. Should we have the policies ready before we allow production use?

Yes, or else this will probably weaken the security of all other pipelines. Meaning, I think we should ensure the existing policies require sane values for new options before releasing the ability to configure those same new options.

brianwcook:
@shi2wei3 Is this going to eventually replace the ostree builder task we have now?

shi2wei3:
> @shi2wei3 Is this going to eventually replace the ostree builder task we have now?

For bootc image, I think we plan to use buildah task to replace rpm-ostree task, but I don't know if there are other projects whick still using it.

xref: https://issues.redhat.com/browse/BIFROST-408

Cc: @cgwalters , please correct me if I'm wrong.

cgwalters:
Yes the goal is to drop the rpm-ostree task

cgwalters:
> Would this change potentially allow users to do insecure things 

Yeah but I think we established that's true already in
https://issues.redhat.com/browse/KFLUXBUGS-1499

It'd make sense for this to be opt-in...and perhaps the high level flag should just be --enable-nested-containers (which would map to these underlying options). For example, `--device /dev/fuse` is quite safe to add, but *other* devices less so.

brianwcook:
@Sergey Yedrikov ***@***.***> wdyt, could this feature work in
instead of the specific The GPU one?

syedriko:
> @syedriko wdyt, could this feature work in instead of the specific The GPU one?
It absolutely could.
And while we're adding knobs to the buildah task, can we expose TMPDIR, as well? With hermetic build, I'm running out of space under /var/tmp and I'm not seeing a way to point TMPDIR someplace else.

shi2wei3:
I need to re-work on this PR since it not works as expected during my test.

shi2wei3:
@cgwalters , sorry to bother, but the workflow in this konflux buildah task is different than directly run `buildah build --security-opt=label=disable --cap-add=all --device /dev/fuse ...` locally, I don't think I have enough experience to sort it out for now, could you help me to take a look?

What I did in this PR will add `--security-opt=label=disable` and `--device /dev/fuse` to the `podman run` [cmd which runs remotely](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml#L576), and also pass them to the generated `scripts/script-build.sh`, and finally add those two options to the `buildah_cmd` in the generated build script and it will be triggered in [a new unshare ns](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-remote-oci-ta/0.2/buildah-remote-oci-ta.yaml#L544).


With this PR, our bootc image build will return the following error message:

```
[build] Adding the entitlement to the build
[build] [1/2] STEP 1/3: FROM registry.stage.redhat.io/rhel9/bootc-image-builder:9.6 AS builder
[build] [1/2] STEP 2/3: ARG MANIFEST=rhel-bootc.yaml
[build] [1/2] STEP 3/3: RUN --mount=type=cache,target=/workdir --mount=type=bind,rw=true,src=.,dst=/buildcontext,bind-propagation=shared . /cachi2/cachi2.env &&     rpm-ostree compose image --cachedir=/workdir --format=ociarchive --initialize /buildcontext/${MANIFEST} /buildcontext/out.ociarchive
[build] bwrap: Can't mount proc on /newroot/proc: Operation not permitted
[build] error: bwrap test failed, see <https://github.com/coreos/rpm-ostree/pull/429>: bwrap(true): Child process killed by signal 1
[build] error: compose tree failed: ExitStatus(unix_wait_status(256))
[build] subprocess exited with status 1
[build] subprocess exited with status 1
[build] Error: building at STEP "RUN --mount=type=cache,target=/workdir --mount=type=bind,rw=true,src=.,dst=/buildcontext,bind-propagation=shared . /cachi2/cachi2.env &&     rpm-ostree compose image --cachedir=/workdir --format=ociarchive --initialize /buildcontext/${MANIFEST} /buildcontext/out.ociarchive": exit status 1
```

I checked from the log, these options has been passed to the podman run command which runs over ssh:
```
[build] + ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10 <user>@<ip> podman run --cap-add=all --security-opt=label=disable --device=/dev/fuse ...
```

cgwalters:
That `podman run` invocation is just going to end up invoking `buildah` right? Are the options being passed to `buildah` too?

Remember there's actually *three* levels of containerization going on AIUI:

- podman
- buildah
- rpm-ostree invoking bwrap

And it's the 3rd level we're trying to enable.

Isn't the outer `podman run` invocation already passing some of these caps?

shi2wei3:
> That `podman run` invocation is just going to end up invoking `buildah` right? Are the options being passed to `buildah` too?

Right, and it invokes `buildah` inside of `unshare`, see
```
unshare -Uf "${UNSHARE_ARGS[@]}" --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w "${SOURCE_CODE_DIR}/$CONTEXT" -- sh -c "$command"
```
And I've already passed those options to `buildah` cmd, from the debug log, I can find those options and please help me to check whether the caps in buildah debug log is sufficient for bootc image building.
```
# unshare -Uf --net --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w source/. -- sh -c ip link set lo up && buildah build --volume /tmp/cachi2:/cachi2 --volume /var/workdir/fetched.repos.d:/etc/yum.repos.d --volume /tmp/entitlement:/etc/pki/entitlement --log-level=debug --pull=never --security-opt=label=disable --device=/dev/fuse --cap-add=all --label build-date=2024-11-04T07:51:51 --label architecture=aarch64 --label vcs-type=git --label vcs-ref=37fb332eaf93837afb28dd0dffa01fa456b1cc4f --label quay.expires-after=5d --tls-verify=true --no-cache --ulimit nofile=4096:4096 -f /tmp/Containerfile.BhWzeP -t <repo:tag> .

[build] time="2024-11-04T07:51:52Z" level=debug msg="effective capabilities: [audit_control=true audit_read=true audit_write=true block_suspend=true bpf=true checkpoint_restore=true chown=true dac_override=true dac_read_search=true fowner=true fsetid=true ipc_lock=true ipc_owner=true kill=true lease=true linux_immutable=true mac_admin=true mac_override=true mknod=true net_admin=true net_bind_service=true net_broadcast=true net_raw=true perfmon=true setfcap=true setgid=true setpcap=true setuid=true sys_admin=true sys_boot=true sys_chroot=true sys_module=true sys_nice=true sys_pacct=true sys_ptrace=true sys_rawio=true sys_resource=true sys_time=true sys_tty_config=true syslog=true wake_alarm=true]"
[build] time="2024-11-04T07:51:52Z" level=debug msg="Pull Policy for pull [never]"
...

```

> 
> Remember there's actually _three_ levels of containerization going on AIUI:
> 
> * podman
> * buildah
> * rpm-ostree invoking bwrap
> 
> And it's the 3rd level we're trying to enable.

Sorry, I didn't get what you mean here, is the development work is not completed yet to support building bootc image from buildah with Containerfile?

> 
> Isn't the outer `podman run` invocation already passing some of these caps?

The outer `podman run` runs over ssh with a non-root user, this remote user will be delete after the task is completed.  I'm not sure what caps it has, but I'm sure I've added these options to the outer `podman run` command, I can find them in the debug output:
```
ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10 <non-root-user>@<ip> podman run --cap-add=all --security-opt=label=disable --device=/dev/fuse <other options>  --user=0 --rm quay.io/konflux-ci/buildah-task:latest@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c /scripts/script-build.sh
```



I tried to reproduce this failure by simulate what konflux buildah task did in my local env, but without any luck, I got a different error, maybe that's because there are still some differences I didn't find out between my local environment and the konflux environment.

```
# cat /scripts/script-build.sh
#!/bin/bash

# ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=60 -o ServerAliveCountMax=10 <user>@<local_vm> \
#   podman run --rm --user=0 \
#   --cap-add=all --security-opt=label=disable --device=/dev/fuse \
#   -v "/rhel-bootc:/rhel-bootc:Z" \
#   -v "/scripts:/scripts:Z" \
#   quay.io/konflux-ci/buildah-task:latest /scripts/script-build.sh

chown root:root /var/lib/containers
echo 'root:1:4294967294' | tee -a /etc/subuid >>/etc/subgid
COMMAND="buildah build --security-opt=label=disable --cap-add=all --device /dev/fuse -t localhost/rhel-bootc -f Containerfile"
unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah --log-level=debug pull <bib stage image>
unshare -Uf --net --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w /rhel-bootc -- sh -c "$COMMAND"


# podman run --rm --user=0 --cap-add=all --security-opt=label=disable --device=/dev/fuse -v "/rhel-bootc:/rhel-bootc:Z" -v "/scripts:/scripts:Z" quay.io/konflux-ci/buildah-task:latest /scripts/script-build.sh
Trying to pull quay.io/konflux-ci/buildah-task:latest...
Copying blob 3a8929253bff done   |
Copying blob 885f6bb378fc done   |
Copying blob 01f40b4c926e done   |
Copying blob 17b076046ff4 done   |
Copying blob 3a2b607ffaba done   |
Copying config f4c834bdec done   |
time="2024-11-04T08:55:19Z" level=debug msg="effective capabilities: [audit_control=true audit_read=true audit_write=true block_suspend=true bpf=true checkpoint_restore=true chown=true dac_override=true dac_read_search=true fowner=true fsetid=true ipc_lock=true ipc_owner=true kill=true lease=true linux_immutable=true mac_admin=true mac_override=true mknod=true net_admin=true net_bind_service=true net_broadcast=true net_raw=true perfmon=true setfcap=true setgid=true setpcap=true setuid=true sys_admin=true sys_boot=true sys_chroot=true sys_module=true sys_nice=true sys_pacct=true sys_ptrace=true sys_rawio=true sys_resource=true sys_time=true sys_tty_config=true syslog=true wake_alarm=true]"
time="2024-11-04T08:55:19Z" level=debug msg="[graphdriver] trying provided driver \"overlay\""
time="2024-11-04T08:55:19Z" level=debug msg="overlay: imagestore=/var/lib/shared"
time="2024-11-04T08:55:19Z" level=debug msg="overlay: imagestore=/usr/lib/containers/storage"
time="2024-11-04T08:55:19Z" level=debug msg="overlay: mount_program=/usr/bin/fuse-overlayfs"
Error: mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted
time="2024-11-04T08:55:19Z" level=debug msg="[graphdriver] trying provided driver \"overlay\""
time="2024-11-04T08:55:19Z" level=debug msg="overlay: imagestore=/var/lib/shared"
time="2024-11-04T08:55:19Z" level=debug msg="overlay: imagestore=/usr/lib/containers/storage"
time="2024-11-04T08:55:19Z" level=debug msg="overlay: mount_program=/usr/bin/fuse-overlayfs"
time="2024-11-04T08:55:19Z" level=warning msg="failed to shutdown storage: \"mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted\""
Error: mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted
time="2024-11-04T08:55:19Z" level=warning msg="failed to shutdown storage: \"mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted\""


# unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah info
Error: mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted
WARN[0000] failed to shutdown storage: "mount /var/lib/containers/storage/overlay:/var/lib/containers/storage/overlay, flags: 0x1000: operation not permitted"
# cat /etc/redhat-release
Fedora release 40 (Forty)
# rpm -q buildah
buildah-1.37.3-1.fc40.x86_64
```


If you want to check the full debug output of my failed test, please find the link in the last pipeline job on MR 1446 in the downsteam rhel-bootc repo.

cgwalters:
> Sorry, I didn't get what you mean here, is the development work is not completed yet to support building bootc image from buildah with Containerfile?

This one definitely works, I use it often (usually with podman, not buildah but it should be the same in general). We can add a CI check that it works but hopefully of course this gets finished and every PR is our CI.

cgwalters:
> time="2024-11-04T08:55:19Z" level=debug msg="[graphdriver] trying provided driver \"overlay\""

I thought konflux was forcing on vfs? That may be a difference.

shi2wei3:
@cgwalters I can reproduce the failure locally now with the following one-line command:
```
podman run --rm --user=0 --cap-add=all --security-opt=label=disable --device=/dev/fuse -v "/rhel-bootc:/rhel-bootc:Z" -v "/home/wshi/.docker/:/root/.docker:Z" -e STORAGE_DRIVER=vfs quay.io/konflux-ci/buildah-task:latest unshare -Uf --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w /rhel-bootc -- sh -c "buildah build --security-opt=label=disable --cap-add=all --device /dev/fuse -t localhost/rhel-bootc -f Containerfile"
```
I tried to add `--privileged` to `podman run`, then the bwrap error is gone, but I hit a new error now...
```
Downloading from 'appstream'...done
Downloading from 'baseos'...done
Importing packages...done: 441
Skipping file /usr/lib/systemd/system/sysinit.target.wants/systemd-firstboot.service from checkout
Skipping file /usr/lib/systemd/system-generators/systemd-gpt-auto-generator from checkout
Skipping file /usr/etc/grub.d/08_fallback_counting from checkout
Skipping file /usr/etc/grub.d/10_reset_boot_success from checkout
Skipping file /usr/etc/grub.d/12_menu_auto_hide from checkout
Skipping file /usr/lib/systemd/ from checkout
Checking out packages...done
Running pre scripts...20 done
Running post scripts...done
error: Installing packages: While applying overrides for pkg shadow-utils: Copyup usr/bin/chage: Setting xattrs: fsetxattr(security.selinux): Operation not supported
error: compose tree failed: ExitStatus(unix_wait_status(256))
subprocess exited with status 1
subprocess exited with status 1
Error: building at STEP "RUN --mount=type=cache,target=/workdir --mount=type=bind,rw=true,src=.,dst=/buildcontext,bind-propagation=shared rpm-ostree compose image --cachedir=/workdir --format=ociarchive --initialize /buildcontext/${MANIFEST} /buildcontext/out.ociarchive": exit status 1
```

And I'm not sure if adding `--privileged` to this rootless container can get approved, I saw we use it in rpm-ostree task.

cgwalters:
> And I'm not sure if adding --privileged to this rootless container can get approved, 

I think this is not really different from `--cap-add=all`. In the end 

> error: Installing packages: While applying overrides for pkg shadow-utils: Copyup usr/bin/chage: Setting xattrs: fsetxattr(security.selinux): Operation not supported

The way this is working today with a default `podman build` is because of the `RUN --mount=type=cache,target=/workdir` - that cache directory will be a temporary dir that's not on overlayfs so we can set `security.selinux` on it. Is that somehow being lost here maybe? Or maybe when using VFS it has different semantics?

We may also be able to fix this by changing that inner containerfile to use `RUN --mount=type=tmpfs,target=/workdir` or so?

cgwalters:
> Anything that needs more permissions than the current OpenShift security context allows will need to be scheduled to a VM but bootc builds already are anyway.

Yes. This relates to a larger picture point around why we haven't prioritized trying to lessen the privileges that the build requires - it's basically because anyone who is making bootc container images is going to want to *boot* it for testing - and that means a VM or bare metal.

So in the end, that infrastructure (containers and VMs/metal) needs to be coupled and available. The reality is of course today the build and test infrastructure is wildly distinct; to the best of my knowledge the Konflux hardware provisioning and management is distinct from Testing Farm for example which is where a lot of the bootc tests run.

shi2wei3:
> We may also be able to fix this by changing that inner containerfile to use `RUN --mount=type=tmpfs,target=/workdir` or so?

I google it and found you answered it when Brian hit the similar issue, and it works when I test it locally, I've update the buildah task and test it in MR1446.

https://github.com/coreos/rpm-ostree/discussions/4648

shi2wei3:
> Just FYI, the Checkton errors should go away if you rebase (it diffs the errors between main and your branch and tries to report only the new ones, which doesn't seem to work correctly if your branch is behind)

Thanks for the remind, I'll rebase and update the changes when my test is pass.

cgwalters:
> I google it and found you answered it when Brian hit the similar issue, 

Heh yeah, totally forgot about that...

shi2wei3:
Test result is PASS for building bootc image with this PR on konflux, there are other issues like supporting hermetic build environment and missing some labels in generated images, I'll track them in bootc repo.

Please help me to review this PR. @chmeliik @brianwcook

pipelines/docker-build-multi-platform-oci-ta/README.md
```diff
@@ -61,6 +63,8 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |LABELS| Additional key=value labels that should be applied to the image| []| |
 |PLATFORM| The platform to build on| None| |
 |PREFETCH_INPUT| In case it is not empty, the prefetched content should be made available to the build.| | '$(params.prefetch-input)'|
+|PRIVILEGED| Whether to add --privileged to podman run which runs remotely| false| |
```
cgwalters, line 66: Here's my proposal: let's consolidate all `PRIVILEGED`, `SECURITY_OPTS` and `DEVICES` to just: `PRIVILEGED_NESTED` or something.

It's totally fine if e.g. `--privileged` automatically implices `--devices /dev/fuse` - there is no loss of generality here; privileged code can use fuse anyways.

I think it's OK if `PRIVILEGED_NESTED` implies `--security-opts=label=disable` too.

shi2wei3: Yes, that's a good idea, but I saw in previous comments,  @brianwcook suggest to use this PR to replace https://github.com/konflux-ci/build-definitions/pull/1529/files , where they use `--security-opts` and `--devices` separately and they don't use `--privileged`.

I add `--devices` and `--security-opts` along with `--privileged` is because the outermost `podman run` could just use `--privileged` but the inner `buildah build` doesn't accept `--privileged` and I need to pass  `--devices` and `--security-opts` to it, do you have any suggests for this?

brianwcook:
We don't _have_ to consolidate it - I will leave it to you since I think
you have a solid handle on all the different options. If it's going to be
super confusing to try to actual get all the knobs turned the right way for
these tasks, then multiple tasks statically configured the right way may be
the path with the better user experience.

Brian

pipelines/docker-build-multi-platform-oci-ta/README.md
shi2wei3: Merged into a single `PRIVILEGED_NESTED` now.

shi2wei3:
@brianwcook @chmeliik @cgwalters Please help to review this PR, I have tested I can build a bootc image successfully.

task/buildah/0.2/buildah.yaml
```diff
@@ -107,6 +107,10 @@ spec:
     description: Additional key=value labels that should be applied to the image
     type: array
     default: []
+  - default: "false"
+    description: Whether to use privileged nested containers when run remotely, this will add --privileged to podman run and --security-opt=label=disable --cap-add=all --device=/dev/fuse to buildah build
+    name: PRIVILEGED_NESTED
```
chmeliik, line 112: > when run remotely, this will add --privileged to podman

Does this option only work for the remote variants of the task? If so, we should either call that out more clearly or only add the option to the remote tasks

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -552,6 +566,11 @@ spec:
       chmod +x scripts/script-build.sh
 
       if ! [[ $IS_LOCALHOST ]]; then
+        PRIVILEGED_NESTED_FLAG=""
+        if [[ "${PRIVILEGED_NESTED}" == "true" ]]; then
+          ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/var/tmp"
+          PRIVILEGED_NESTED_FLAG="--privileged --mount type=bind,source=$BUILD_DIR/var/tmp,target=/var/tmp,relabel=shared"
```
chmeliik, line 572: Why does specifically `/var/tmp` need to be a mount? Could you add a comment for that here?

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -585,6 +605,7 @@ spec:
           -v "$BUILD_DIR/.docker/:/root/.docker:Z" \
           -v "$BUILD_DIR/results/:/tekton/results:Z" \
           -v "$BUILD_DIR/scripts:/scripts:Z" \
+          "$PRIVILEGED_NESTED_FLAG" \
```
chmeliik, line 608: I think this variable expansion won't work as expected, it turns multiple flags into a single argument

An array would work better

```bash
PRIVILEGED_NESTED_FLAGS=(--privileged --mount --type=bind,...)
```

chmeliik (commented):
@shi2wei3 we will need Enterprise Contract policies in place before merging this PR (the usage of PRIVILEGED_NESTED should be disallowed by default). Would you be willing to contribute those?

Otherwise, we can file a story for the EC team, cc @lcarva @zregvart

shi2wei3:
> @shi2wei3 we will need Enterprise Contract policies in place before merging this PR (the usage of PRIVILEGED_NESTED should be disallowed by default). Would you be willing to contribute those?
> 
> Otherwise, we can file a story for the EC team, cc @lcarva @zregvart

I'm not familiar with how EC policies work, could EC team help to work on that? Where should I file a story?

chmeliik:
> > @shi2wei3 we will need Enterprise Contract policies in place before merging this PR (the usage of PRIVILEGED_NESTED should be disallowed by default). Would you be willing to contribute those?
> > Otherwise, we can file a story for the EC team, cc @lcarva @zregvart
> 
> I'm not familiar with how EC policies work, could EC team help to work on that? Where should I file a story?

They're fairly straightforward, conceptually this would be very similar to https://github.com/enterprise-contract/ec-policies/blob/a1a28ad2a42298463004f1df59dbf3fc7a03fe30/policy/release/buildah_build_task/buildah_build_task.rego#L36
The story, if needed, should go to https://issues.redhat.com/projects/EC

shi2wei3:
> They're fairly straightforward, conceptually this would be very similar to https://github.com/enterprise-contract/ec-policies/blob/a1a28ad2a42298463004f1df59dbf3fc7a03fe30/policy/release/buildah_build_task/buildah_build_task.rego#L36 The story, if needed, should go to https://issues.redhat.com/projects/EC

Thanks for the guidance. I'll give it a try.

task/buildah/0.2/buildah.yaml
shi2wei3: Updated the description

task/buildah-remote/0.2/buildah-remote.yaml
shi2wei3: Added a comment for explanation.

task/buildah-remote/0.2/buildah-remote.yaml
shi2wei3: Changed to array.

task/buildah-remote/0.2/buildah-remote.yaml
```diff
@@ -552,6 +566,11 @@ spec:
       chmod +x scripts/script-build.sh
 
       if ! [[ $IS_LOCALHOST ]]; then
+        PRIVILEGED_NESTED_FLAG=""
+        if [[ "${PRIVILEGED_NESTED}" == "true" ]]; then
+          ssh $SSH_ARGS "$SSH_HOST"  mkdir -p "$BUILD_DIR/var/tmp"
```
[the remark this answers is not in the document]
shi2wei3: Used the directive to disable it.

task/buildah/0.2/buildah.yaml
```diff
@@ -107,6 +107,10 @@ spec:
     description: Additional key=value labels that should be applied to the image
     type: array
     default: []
+  - default: "false"
+    description: This option is only available for remote tasks, it will add --privileged to podman run and --security-opt=label=disable --cap-add=all --device=/dev/fuse to buildah build
```
chmeliik, line 111: I'll pretend I'm a Konflux user

What is a "remote" task? If this task is not one of those, why does it have this option?

zregvart:
Replying on behalf of the EC team, we can put in place a policy rule similar to the one already mentioned that checks if the `PRIVILEGED_NESTED` parameter is set to `true` and emit a violation if it is. This is fairly trivial for us to do.

I've filed https://issues.redhat.com/browse/EC-1011 to tackle this. Not sure if it is blocking, the merge of this change, we can prioritize and get this in very soon if requested.

task/buildah/0.2/buildah.yaml
shi2wei3: How could I remove this option in buildah task but leave it in buildah-remote tasks, after running ./hack/generate-buildah-remote.sh, this option will be removed from buildah-remote task either.

shi2wei3:
> I've filed https://issues.redhat.com/browse/EC-1011 to tackle this. Not sure if it is blocking, the merge of this change, we can prioritize and get this in very soon if requested.

Thanks for the help, per the comment from @chmeliik the EC policy is blocking this PR to be merged.

task/buildah/0.2/buildah.yaml
shi2wei3: Ah, it seems there is a recipe.yaml in each tasks might be used for that, let me try.

shi2wei3: Okay, I should append the remote param in task-generator/remote/main.go, updated.

task/buildah/0.2/buildah.yaml
```diff
@@ -303,6 +303,12 @@ spec:
 
       BUILDAH_ARGS+=("${BUILD_ARG_FLAGS[@]}")
 
+      if [ "${PRIVILEGED_NESTED}" == "true" ]; then
+        BUILDAH_ARGS+=("--security-opt=label=disable")
+        BUILDAH_ARGS+=("--cap-add=all")
+        BUILDAH_ARGS+=("--device=/dev/fuse")
+      fi
```
chmeliik, line 310: The scripts sets `set -u`, so this will always fail because of the undefined variable.

Could you also add a comment that this code is only here because it's needed in the buildah-remote task?

shi2wei3: And there is another related issue, when I dropped PRIVILEGED_NESTED param in buildah.yaml, task-generator/remote/main.go won't add `-v PRIVILEGED_NESTED="$PRIVILEGED_NESTED"` to the `podman run` command runs over ssh, then the inner scripts/script-build.sh won't have $PRIVILEGED_NESTED set.

How about we add param PRIVILEGED_NESTED back to all buildah tasks to address all these problems, literally it have privilege mode enabled for local buildah.yaml because we add option `--security-opt=label=disable --cap-add=all` to `buildah build`, I just don't have a real world use case to verify it works for local build, but for bootc image we only use remote task for multi-arch image building,

shi2wei3: Added `PRIVILEGED_NESTED` back and removed "remote" keyword in description.
Bootc image building test is PASS.

chmeliik: So we just don't know if it works in the non-remote task. That's fine ¯\\\_(ツ)\_/¯

task/buildah/0.2/buildah.yaml
```diff
@@ -107,6 +107,12 @@ spec:
     description: Additional key=value labels that should be applied to the image
     type: array
     default: []
+  - name: PRIVILEGED_NESTED
+    description: Whether to enable privileged mode, it will add --privileged to podman
+      run and --security-opt=label=disable --cap-add=all --device=/dev/fuse to buildah
```
chmeliik, line 112: Nitpick: let's remove the part about adding `--privileged` to podman. It's an implementation detail that users don't care about, and it's not true for the non-remote task

chmeliik (approved):
LGTM apart from the param description

task/buildah/0.2/buildah.yaml
shi2wei3: Updated.

shi2wei3:
> LGTM apart from the param description

Thanks for the all the suggestions, and we are pending on the EC policy.

zregvart:
> Thanks for the all the suggestions, and we are pending on the EC policy.

I don't think you should be. There will be a delay in adoption of this change and by the time any project reaches the release pipeline we should have the EC policy in effect.

If this is something that you want to push through very soon, I suggest considering a contribution to [ec-policies](https://github.com/enterprise-contract/ec-policies/).

chmeliik:
> > Thanks for the all the suggestions, and we are pending on the EC policy.
> 
> I don't think you should be. There will be a delay in adoption of this change and by the time any project reaches the release pipeline we should have the EC policy in effect.
> 
> If this is something that you want to push through very soon, I suggest considering a contribution to [ec-policies](https://github.com/enterprise-contract/ec-policies/).

I think the adoption by the bootc team may be quite quick, no? But I won't block this if you both think that's OK

shi2wei3:
e2e-tests are failed, but I don't have the access to it. @chmeliik Could you help to take a look?

chmeliik:
> e2e-tests are failed, but I don't have the access to it. @chmeliik Could you help to take a look?

You may be able to access them as follows:
* install https://github.com/sayan-biswas/kubectl-tekton
* login to https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com
* `kubectl tekton -n konflux-ci logs pipelinerun build-definitions-pull-request-p6m2s`
  * you can see the pipelinerun name in the `Details` view for the build-definitions-pull-request check

chmeliik:
But in this case, I'm getting `No logs found`. Happens sometimes.

/retest

shi2wei3:
Thanks @chmeliik . Could we get this PR merged, the corresponding [ec-policy ](https://github.com/enterprise-contract/ec-policies/pull/1221)is also ready for merge, I don't have write access to do that.

PR #1534, opened 2024-10-24
feat(STONEINTG-1078): Combine all FBC validation tasks into one

arewm:
This change combines the inspect-image, fbc-validate, and fbc-related-image-check into a single task. ~It depends on functionality that needs to be added to EC to parse the produced trusted artifact in order to assess whether the related images are valid.~

All of the old tasks will be deprecated and the required tasks need to be updated to require only

```
- [fbc-related-image-check, validate-fbc]
```

This will ensure that users can still be guaranteed to have a valid FBC fragment and appropriate related images.

EDIT: In order to merge this PR quicker, the functionality for moving the related image check was backed out. In a future PR, we can update the task to output the list of related images and have EC check them for validity.

EDIT: This PR has sat around long enough so that I was able to [add a helper script to oras](https://github.com/konflux-ci/oras-container/pull/133) to attach artifacts. I have included this functionality in the task so that we can later remove the related image checking after EC supports the attached file (without a version bump).

### DIFF
--- a/task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
+++ b/task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
@@ -6,6 +6,7 @@ metadata:
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "konflux"
+    build.appstudio.redhat.com/expires-on: "2025-01-31T00:00:00Z"
   name: fbc-related-image-check
 spec:
   description: >-
--- a/task/fbc-related-image-check/0.2/MIGRATION.md
+++ b/task/fbc-related-image-check/0.2/MIGRATION.md
@@ -1,0 +1,48 @@
+## Deprecation notice
+
+This task is deprecated, please remove it from your pipeline.
+Deprecation date: 2025-01-31
+
+# Migration from 0.1 to 0.2
+
+Version 0.2:
+
+No changes within this version, its only purpose is to provide information on how to remove this task from your pipeline.
+
+## Action from users
+
+To remove this task from your pipeline please follow these steps:
+
+1. Remove the fbc-related-image-check task definition from your FBC pipelines similar to this change:
+
+```diff
+--- a/.tekton/original-pipelinerun.yaml
++++ b/.tekton/new-pipelinerun.yaml
+@@ -323,26 +323,6 @@ spec:
+       workspaces:
+       - name: workspace
+         workspace: workspace
+-    - name: fbc-related-image-check
+-      runAfter:
+-      - fbc-validate
+-      taskRef:
+-        params:
+-        - name: name
+-          value: fbc-related-image-check
+-        - name: bundle
+-          value: quay.io/konflux-ci/tekton-catalog/task-fbc-related-image-check:0.1@sha256:0fae84cc832d21c250334ab1d285db92e7e22e916ea342d044e46136c502d2f8
+-        - name: kind
+-          value: task
+-        resolver: bundles
+-      when:
+-      - input: $(params.skip-checks)
+-        operator: in
+-        values:
+-        - "false"
+-      workspaces:
+-      - name: workspace
+-        workspace: workspace
+     workspaces:
+     - name: workspace
+     - name: git-auth
+```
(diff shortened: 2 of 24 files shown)


arewm:
@yashvardhannanavati @grokspawn , I cannot assign you as reviewers, but can you double check this PR? It is a big PR but the content should be largely unchanged from when it was separated into three tasks.

grokspawn:
> @yashvardhannanavati @grokspawn , I cannot assign you as reviewers, but can you double check this PR? It is a big PR but the content should be largely unchanged from when it was separated into three tasks.

Just ran across the gh notification now.  I'll take a look.  I also couldn't assign me. 😞

task/fbc-related-image-check/0.2/MIGRATION.md
```diff
@@ -0,0 +1,48 @@
+## Deprecation notice
+
+This task is deprecated, please remove it from your pipeline.
+Deprecation date: 2024-12-31
```
dirgim, line 4: This seems like a bit of an aggressive deadline considering there's not a lot of working days until that date?

dirgim: Same applies to the other tasks.

lcarva: (I'd avoid using a date that falls on a common holiday.)

arewm: I can put it into January. I had this as a placeholder initially. What would be a better date, January 15th?

dirgim: Based on recent user feedback on the [verify-signed-rpm](https://github.com/konflux-ci/build-definitions/tree/main/task/verify-signed-rpms) task being required in EC, I'd suggest Jan 31.

dirgim (approved):
LGTM

task/fbc-related-image-check/0.1/fbc-related-image-check.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "konflux"
+    build.appstudio.redhat.com/expires-on: "2025-01-31T00:00:00Z"
```
jsztuka, line 9: Do we need to keep track of this? Or we update this value in next 2 months?

jsztuka, line 9: Will it be possible to announce this also on users channel for general awareness?

jsztuka (approved):
lgtm

grokspawn (approved):
Sorry it took so long.  With the deprecation and consolidations there's a bunch of room to cover.  Oh, and we also debuted OLMv1, v1.0. :)

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -0,0 +1,463 @@
...
+            echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 0
+        fi
+
+        BASE_IMAGE_REPOSITORY="$(jq -r '.Name | sub("[^/]+/"; "") | sub("[:@].*"; "")' "$BASE_IMAGE_INSPECT")"
+        echo "Detected base image repository: $BASE_IMAGE_REPOSITORY"
+        echo -n "$BASE_IMAGE_REPOSITORY" > /shared/BASE_IMAGE_REPOSITORY
+
+        note="Step inspect-image completed: Check inspected JSON files under /var/workdir/inspect-image."
+    - name: extract-and-validate
```
yashvardhannanavati, line 181: Having these steps as separate stepActions might be a cleaner approach? But we can do that later too

arewm: I put them in one step to simplify the results reporting. We can certainly change that to produce intermediate results in the task steps and then have a single step at the end to set the result value.

yashvardhannanavati: yeah, it works as is too. Just an idea :)

yashvardhannanavati (approved):

PR #1543, opened 2024-10-30
push to a unique tag in buildah tasks

tnevrlka:
[KFLUXBUGS-1539](https://issues.redhat.com/browse/KFLUXBUGS-1539)
- Built image in buildah task is pushed to a location specified in the `IMAGE` param
- This location does not have to be unique when running multiple PipelineRuns on the
same revision
- A race condition can occur resulting in a failure
- Besides pushing to `IMAGE`, push to a tag named after the TaskRun name which
should be unique

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -662,6 +662,15 @@ spec:
 
         buildah commit "${BUILDAH_ARGS[@]}" $container $IMAGE
 
+        echo "Pushing to ${IMAGE%:*}:${TASKRUN_NAME}"
+        if ! retry buildah push \
+          --tls-verify="$TLSVERIFY" \
+          "$IMAGE" \
+          "docker://${IMAGE%:*}:$(context.taskRun.name)"; then
+          echo "Failed to push sbom image to ${IMAGE%:*}:$(context.taskRun.name) after ${max_run} tries"
+          exit 1
+        fi
+
         echo "Pushing to ${IMAGE}"
         if ! retry buildah push \
           --tls-verify="$TLSVERIFY" \
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -629,6 +629,19 @@ spec:
         #!/bin/bash
         set -e
 
+        retry() {
+          status=-1
+          max_run=5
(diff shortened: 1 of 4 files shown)


task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -656,10 +659,18 @@ spec:
           status=0
           [ "$run" -gt 1 ] && sleep $sleep_sec
           echo "Pushing sbom image to registry"
+
+          echo "Pushing to ${IMAGE%:*}:${TASKRUN_NAME}"
```
MartinBasti, line 663: This should have retry as well IMHO

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -649,20 +662,21 @@ spec:
...
-        if [ "$status" -ne 0 ]; then
-          echo "Failed to push sbom image to registry after ${max_run} tries"
+        echo "Pushing to ${IMAGE%:*}:${TASKRUN_NAME}"
+        if ! retry buildah push \
+          --tls-verify="$TLSVERIFY" \
+          "$IMAGE" \
+          "docker://${IMAGE%:*}:$(context.taskRun.name)"; then
+          echo "Failed to push sbom image to ${IMAGE%:*}:$(context.taskRun.name) after ${max_run} tries"
+          exit 1
+        fi
```
MartinBasti, line 672: IMO this makes readability better; but just nitpick and personal preference
```suggestion
        echo "Pushing to ${IMAGE%:*}:${TASKRUN_NAME}"
        if ! retry buildah push \
          --tls-verify="$TLSVERIFY" \
          "$IMAGE" \
          "docker://${IMAGE%:*}:$(context.taskRun.name)";
         then
          echo "Failed to push sbom image to ${IMAGE%:*}:$(context.taskRun.name) after ${max_run} tries"
          exit 1
        fi
```

MartinBasti: I meant:
```
if [cond]
then  # on newline
  code
fi
```

tnevrlka:
The webhook error should be fixed... /retest

task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
MartinBasti: work for me

MartinBasti (approved):

PR #1558, opened 2024-10-31
Guard Task execution via changed files

zregvart:
The new Task in `.tekton/tasks/changed-dirs.yaml` produces a list of changed directories for a particular pull request as an array result. Given this result guards can be done, e.g. using CEL expressions to limit when a Task on the Pipeline needs to run.
This way we can skip expensive Tasks that are unrelated to the change done in the pull request.

Similar to work in #1188 and #524, with the distinction that the PipelineRun is executed, only potentially not in full.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
-    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml, task/sast-unicode-check/0.1/sast-unicode-check.yaml]"
+    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, .tekton/tasks/buildah.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.2/sast-snyk-check.yaml, task/sast-unicode-check/0.1/sast-unicode-check.yaml, .tekton/tasks/task-switchboard.yaml]"
     pipelinesascode.tekton.dev/max-keep-runs: "5"
 spec:
   params:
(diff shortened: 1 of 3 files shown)


zregvart:
Marked as draft, not sure if this approach will work, any further ideas welcome!

.tekton/tasks/changed-dirs.yaml
```diff
@@ -0,0 +1,47 @@
...
+        for path in $(gh pr view "https://github.com/konflux-ci/build-definitions/pull/$(params.pr_number)" --json files --jq '.files.[].path'); do
+          dirs+=("${path%%/*}")
+        done
+        # deduplicate dirs
+        readarray -t dirs < <(printf '%s\n' "${dirs[@]}" | sort -u)
+        # join $dirs with '","', e.g. (a b c) -> 'a","b","c","'
+        escaped_json="$(printf '%s\",\"' "${dirs[@]}")"
+        # add the square brackets and remove the trailing ',"'
+        escaped_json="[\"${escaped_json::-2}]"
+        echo -n "${escaped_json}" > "$(results.changed_directories.path)"
```
lcarva, line 47: The number of directories in this repo will blow past the result limit.

```
$ find . -type d | wc -c | numfmt --to=iec
11K
```

zregvart: It's only creating an array of top-level directories, so the current size is:

```
$ ls -d */|jq -cRn '[inputs]'|wc -c
107
```

lcarva: Great!

chmeliik:
It does seem too limiting without `enable-cel-in-whenexpression` For many of the steps, guarding them on one specific directory is not enough :(

zregvart:
> It does seem too limiting without `enable-cel-in-whenexpression` For many of the steps, guarding them on one specific directory is not enough :(

I agree, the alternative, as you pointed out is to not build a generic solution but a specific one, which would mean moving the logic into the `changed-dirs` Task (and in the process probably renaming it to something like `task-switchboard`). Tedious to maintain and not nearly as nice.

I do think the first iteration could be to look at the top level directories, that might be good enough.

There could be a middle ground, e.g. provide a set of expressions (in CEL or something else) and when those evaluate as `true` emit one item in the result array of `changed-dirs`, then use `when` based on that value. If that makes sense. I can try to prototype this idea on this PR.

> I was going to create a script that gets the list of changed files _and_ evaluates some gitignore-style conditions in one step to enable skipping e2e-tests. It would avoid the result size limit problems and be more flexible than the default `when` (but less flexible than `when` with CEL)

I think that kind of logic should go to the e2e harness, e.g. using [`Skip`](https://onsi.github.io/ginkgo/#skipping-specs) based on the Tasks/Pipelines changed. Might be easier to maintain this in golang, with a bit of effort this could payoff greatly.

zregvart:
This version evaluates expressions over all files in a PR using OPA/Rego. The [`expressions` parameter](https://github.com/konflux-ci/build-definitions/pull/1558/files#diff-e1eb482b2093660ed8e19db808321a572bbfb86758055c532eadd30f039c0ddeR47-R52) determine by evaluating to `true` what result is omitted, the function [`strings.any_prefix_match`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-strings-stringsany_prefix_match) comes in quite handy.

.tekton/pull-request.yaml
```diff
@@ -36,7 +36,27 @@ spec:
...
+          - name: pr_number
+            value: "{{ pull_request_number }}"
+          - name: utils_image
+            value: quay.io/konflux-ci/pull-request-builds:appstudio-utils-{{revision}}
+          - name: expressions
+            value:
+              - tasks := strings.any_prefix_match(input, "task/")
+              - tasks_pipelines := strings.any_prefix_match(input, ["task/", "pipelines/"])
+              - e2e_tests := strings.any_prefix_match(input, ["task/", "pipelines/"])
+              - check_partner_tasks := strings.any_prefix_match(input, "partners/")
```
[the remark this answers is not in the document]
chmeliik: (could also get more specific with it, but `.tekton/` and `hack/` would be good enough for me)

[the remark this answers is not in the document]
zregvart: Very true, added...

chmeliik (approved):
LGTM :taco:

PR #1564, opened 2024-11-01
always unregister system from red hat subscription manager

brianwcook:


arewm (approved):

c-kruse (commented):
I was able to confirm that with this changeset I was able to install subscription content with the buildah-oci-ta task and an activation-key secret.

PR #1565, opened 2024-11-01
Revert "[ISV-5221] Add new step to inject and push SBOMs in build-ima…

rcerven:
…ge-index task."

This reverts commit 05cb6766a2d851b4473d1a4a7913eb8d3ff5e969.

reverting it because it is causing problems with results limit, since it added more steps


arewm (approved):
These changes are causing the Tekton results to overflow since new steps results in a smaller space for results in each task step.

chmeliik (changes requested):
I would like stronger proof that the problem was the result limit. Have you tested with the affected user that this resolves the issue?

chmeliik:
> I would like stronger proof that the problem was the result limit. Have you tested with the affected user that this resolves the issue?

Probably no need to bother. The affected user was using an older version of the task, this change wasn't even included. We can probably close this.

PR #1567, opened 2024-11-01
Add link between e2e timeout and resource pruning

arewm:
We moved the stale application pruning to closer to the timeout of the PLR. In order to reduce unexpected errors when increasing the timeout, we should add a comment for the behavior.


MartinBasti (approved):

arewm:
Merging. No need to make this comment-only change use compute to enable the merge.

PR #1570, opened 2024-11-03
feat: added new task export-image-tar

Jared-Sprague:
This pull request is to add a new simple task for exporting images as oci-archive tar file and pushing as an oci artifact.  This is needed in some cases where images must be shared without the use of an image registry.

### DIFF
--- a/task/export-image-tar/0.1/export-image-tar.yaml
+++ b/task/export-image-tar/0.1/export-image-tar.yaml
@@ -1,0 +1,79 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: export-image-tar
+spec:
+  params:
+    - name: input-image
+      type: string
+      description: The input image that is to be saved as an oci-archive tar file
+    - name: tar-filename
+      default: "image.tar"
+      description: The tar filename of the saved input image
+      type: string
+    - name: dest-image-tag
+      default: ""
+      description: The tag of the destination image when the tar file is loaded. Use this if you want to hide the source repository name, or if you have issues with digest missmatch which can happen sometimes with `podman save` and `podman load`
+      type: string
+    - name: output-artifact
+      type: string
+      description: The output url that points to the OCI artifact of the tar image
+  results:
+    - description: Digest of the tar artifact just pushed
+      name: ARTIFACT_DIGEST
+    - description: Image repository where the tar artifact was pushed
+      name: ARTIFACT_URL
+  steps:
+    - name: save-tar
+      image: registry.access.redhat.com/ubi9/podman:latest
+      script: |
+        podman --version
+
+        podman pull $(params.input-image)
+
+        echo "saving: $(params.tar-filename)"
+
+        # Check if dest-image-tag is not an empty string
+        if [[ -n "$(params.dest-image-tag)" ]]; then
+            # Tag the base image with the destination image tag
+            echo "tagging image: $(params.dest-image-tag)"
+            podman tag "$(params.input-image)" "$(params.dest-image-tag)"
+            
+            # Save the tagged image
+            podman save --format oci-archive -o "/mnt/artifacts/$(params.tar-filename)" "$(params.dest-image-tag)"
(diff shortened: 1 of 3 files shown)


task/export-image-tar/0.1/export-image-tar.yaml
```diff
@@ -0,0 +1,79 @@
...
+      name: ARTIFACT_DIGEST
+    - description: Image repository where the tar artifact was pushed
+      name: ARTIFACT_URL
+  steps:
+    - name: save-tar
+      image: registry.access.redhat.com/ubi9/podman:latest
+      script: |
+        podman --version
+
+        podman pull $(params.input-image)
```
arewm, line 32: How should this work when the image is multi-arch? With this command, it will only pull the _one_ image matching the system's architecture (i.e. x86). In the future, if we are going to change over to arm node pools by default, this would effectively be a breaking change as it would pull a different image (if that is even possible).

Jared-Sprague: @arewm Oh that is a really good point. I could add the an `arch` parameter to pass through to `podman pull --arch ARCH`

Jared-Sprague: @arewm Ok I added an `input-arch` param to specify a specific arch to pull if it's different from the host arch.

chmeliik (commented):
What does this PR allow you to do that you can't already achieve with `skopeo copy docker://$IMAGE oci-archive:archive.tar`?

chmeliik:
> @chmeliik This makes it possible to do this as part of a pipeline task for a release. There are many tasks you can do manually that konflux automates for you. Like I can clone a repo with `git clone` but there still is a `clone-repository` task. Our release requirements are such that we need to give customers the ability to download an image tar file directly. Since there is no task currently in Konflux for saving and uploading an image as a tar file. That is what this is for. And just because you can do it manually with skopeo doesn't help with CI/CD.

Is the oras-uploaded artifact pullable directly? I thought you pretty much needed `oras pull` to download something uploaded by `oras push` (which is why it seems no different from telling your customer to use skopeo)

arewm:
At some level, these operations can be done at release time, but that comes down to an implementation detail. I feel like operations that modify artifacts should be done before release time to enable them to be more fully tested. Build-time activities are those like producing container images, binary artifacts, tarfiles, etc. Release-time activities are those like retrieving, signing, pushing, copying content. This separation can also ensure that we have appropriate provenance available for the produced artifacts.

In this case, the release pipeline wouldn't know that it is processing a container image necessarily. It would just be passed a path to retrieve a tarball and a destination for it.

I think that this task could be changed to do something similar to what I am exploring here: https://github.com/konflux-ci/build-definitions/pull/1534/files#diff-ac0937f0ae23d79101fafbbe1d7b9325c299712e27bcb8a781e0ec34ef8a4643R441-R447

These tarballs are tightly coupled with the container image(s) itself. Instead of producing separate artifacts, it would make sense to create artifacts and attach them to the original container using the referrer's API. Based on feedback from the contract team, however, it has been requested to _not_ reuse the Trusted Artifact flow for this and instead to create a separate utility to simplify the artifact pushing. This approach is, however, less supported than it is to generate a tarball on its own. It would require the feature requested in https://issues.redhat.com/browse/KONFLUX-5222.

chmeliik:
> At some level, these operations can be done at release time, but that comes down to an implementation detail. I feel like operations that modify artifacts should be done before release time to enable them to be more fully tested. Build-time activities are those like producing container images, binary artifacts, tarfiles, etc. Release-time activities are those like retrieving, signing, pushing, copying content. This separation can also ensure that we have appropriate provenance available for the produced artifacts.

Agreed. And by this definition, this task is a release task. It doesn't produce anything new, it copies the built image into the same registry but in a weird wrapper

arewm:
Right. It could be handled on either side of build or release. I added this topic to the Konflux community architecture call for tomorrow.

PR #1573, opened 2024-11-04
set strict mode for buildah build step

tnevrlka:
[KFLUXBUGS-1681](https://issues.redhat.com/browse/KFLUXBUGS-1681)

- Failure of dockerfile-json parsing is not handled and buildah attempts to build
the image, leading to confusing errors
- Fail the build immediately if an error in a pipeline occurs

Previously
image

After the change
image

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -256,7 +256,7 @@ spec:
           value: $(params.DOCKERFILE)
       script: |
         #!/bin/bash
-        set -e
+        set -euo pipefail
         ca_bundle=/mnt/trusted-ca/ca-bundle.crt
         if [ -f "$ca_bundle" ]; then
           echo "INFO: Using mounted CA bundle: $ca_bundle"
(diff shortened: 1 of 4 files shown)


task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -348,10 +348,18 @@ spec:
           BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
         done
 
-        BASE_IMAGES=$(
-          dockerfile-json "${BUILD_ARG_FLAGS[@]}" "$dockerfile_copy" |
-            jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName | select(test("^oci-archive:") | not)'
-        )
+        if ! dockerfile_json_output=$(
```
MartinBasti, line 351: How about using

```
set -euo pipefail
```

and have it safe for everyhthing in script. I assume that reason why it doesn't fail is missing `-o pipefail` and thus error is hidded.


https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

MartinBasti (approved):

PR #1590, opened 2024-11-08
Fix rhsm ca issue

brianwcook:
this fixes an issue affecting ubi-minimal (microdnf) for ubi8 only. Has been tested against
ubi8, ubi9, ubi8-minimal, ubi9-minimal


brianwcook:
merging to fix a user issue, e2e tests are temporarily broken per @flacatus

PR #1594, opened 2024-11-11
Add imageContentSources to eaas-create-ephemeral-cluster-hypershift-aws

amisstea:
The user may choose to pass the value as a quoted string with explicit newlines ("\n") or by using the "|" operator, among other options. A tool like `yq` is therefore necessary since basic variable expansion within a heredoc isn't adequate at preventing issues with malformed yaml.

As a result. the container image was changed to one which provides both the `yq` and `kubectl` binaries. This is a more upstream friendly image since it's publicly pullable.

### DIFF
--- a/stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
+++ b/stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
@@ -6,7 +6,7 @@ spec:
   description: >-
     This StepAction provisions an ephemeral cluster using Hypershift with 3 worker nodes in AWS.
     It does so by creating a ClusterTemplateInstance in a space on an EaaS cluster.
-  image: registry.redhat.io/openshift4/ose-cli@sha256:15da03b04318bcc842060b71e9dd6d6c2595edb4e8fdd11b0c6781eeb03ca182
+  image: quay.io/konflux-ci/appstudio-utils:5d8df82bdad43c3a71c9eb1a525a25e946777dc4@sha256:0901c58b85d6a05f75782bd8dd64b849a5b7d44f9f36c78a6ee33161278c96a4
   params:
     - name: eaasSpaceSecretRef
       type: string
@@ -32,6 +32,12 @@ spec:
       type: string
       default: 30m
       description: How long to wait for cluster provisioning to complete.
+    - name: imageContentSources
+      type: string
+      default: ""
+      description: >-
+        Alternate registry information containing a list of sources and their mirrors in yaml format.
+        See: https://hypershift-docs.netlify.app/how-to/disconnected/image-content-sources
   results:
     - name: clusterName
       description: The name of the generated ClusterTemplateInstance resource.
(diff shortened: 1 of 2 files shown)


amisstea:
@hmariset please review

hmariset (approved):
/lgtm

stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
```diff
@@ -6,7 +6,8 @@ spec:
   description: >-
     This StepAction provisions an ephemeral cluster using Hypershift with 3 worker nodes in AWS.
     It does so by creating a ClusterTemplateInstance in a space on an EaaS cluster.
-  image: registry.redhat.io/openshift4/ose-cli@sha256:15da03b04318bcc842060b71e9dd6d6c2595edb4e8fdd11b0c6781eeb03ca182
+  image: public.ecr.aws/bitnami/kubectl:1.31.2@sha256:df433ab6de9adfb0223df79167719099726f946a68c1784d7739a8b6b63cd137
```
chmeliik, line 9: There is a list of allowed registries for step images: https://github.com/release-engineering/rhtap-ec-policy/blob/ba7c0a249386ac15df6495e31140de7fb038187f/data/rule_data.yml#L10

Side note, we'll need to extend the CI check to take StepActions into account as well

chmeliik: > Side note, we'll need to extend the CI check to take StepActions into account as well

Filed https://issues.redhat.com/browse/EC-1010

amisstea: Ah, good point. I'll have a look for an alternate image from those allowed registries/orgs.

chmeliik: `quay.io/konflux-ci/appstudio-utils` might work, it contains both `yq` and `kubectl` and other tasks already use it

stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
```diff
@@ -51,40 +57,45 @@ spec:
...
-          value: $VERSION
-        - name: timeout
-          value: $TIMEOUT
+      parameters: []
     EOF
 
+    yq -i '.spec.parameters += {"name": "instanceType", "value": strenv(INSTANCE_TYPE)}' cti.yaml
+    yq -i '.spec.parameters += {"name": "version", "value": strenv(VERSION)}' cti.yaml
+    yq -i '.spec.parameters += {"name": "timeout", "value": strenv(TIMEOUT)}' cti.yaml
+    yq -i '.spec.parameters += {"name": "imageContentSources", "value": strenv(IMAGE_CONTENT_SOURCES)}' cti.yaml
```
chmeliik, line 80: Just to double-check: the `value` should be a yaml string, not the object parsed from that string? (I think you would want `env()` rather than `strenv()` for the latter)

chmeliik (approved):
LGTM assuming the value should be a string

stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
amisstea: Yes, the `value` should always be parsed as a string. The yq [docs](https://mikefarah.gitbook.io/yq/operators/env-variable-operators) indicate `strenv()` should be used over `env()` and that was my experience when testing with both.

```
env which takes a single environment variable name and parse the variable as a yaml node (be it a map, array, string, number of boolean)

strenv which also takes a single environment variable name, and always parses the variable as a string.
```

amisstea:
@chmeliik can we add this one to the merge queue now?

PR #1596, opened 2024-11-12
Add OCI/TA variant of the tkn-bundle Task

zregvart:
Resolves: https://issues.redhat.com/browse/EC-979

### DIFF
--- a/task/tkn-bundle-oci-ta/0.1/recipe.yaml
+++ b/task/tkn-bundle-oci-ta/0.1/recipe.yaml
@@ -1,0 +1,9 @@
+---
+base: ../../tkn-bundle/0.1/tkn-bundle.yaml
+add:
+  - use-source
+removeWorkspaces:
+  - source
+preferStepTemplate: true
+replacements:
+  workspaces.source.path: /var/workdir/source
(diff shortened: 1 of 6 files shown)


task/tkn-bundle-oci-ta/0.1/recipe.yaml
```diff
@@ -0,0 +1,9 @@
+---
+base: ../../tkn-bundle/0.1/tkn-bundle.yaml
+add:
+  - use-source
+removeWorkspaces:
+  - source
+preferStepTemplate: true
+replacements:
+  workspaces.source.path: /var/workdir/source
```
lcarva, line 9: Should this be `/var/workdir`?

zregvart: I don't think so, the restored source is in [`/var/workdir/source`](https://github.com/konflux-ci/build-definitions/pull/1596/files#diff-d36a1c9b71c7827d73359c092cf21a3169f0bb7a876a4454272132fcbc140e32R60), then the Task reads [from there (example)](https://github.com/konflux-ci/build-definitions/pull/1596/files#diff-d36a1c9b71c7827d73359c092cf21a3169f0bb7a876a4454272132fcbc140e32R90-R95)

lcarva (approved):
/lgtm

PR #1599, opened 2024-11-12
update synk upstream docs link

ashwindasr:
Replaces outdated synk documentation link


arewm (approved):
Thanks for the updated documentation links!

PR #1605, opened 2024-11-13
feat: Update prefetch-dependencies to 0.2 to support sub-paths

ligangty:
We're developing a new build pipeline to support build maven zip oci-artifacts, which needs cachi2 support to read multiple lockfiles of "generic" pkg type. This PR added that subpaths support.

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.2/README.md
+++ b/task/prefetch-dependencies-oci-ta/0.2/README.md
@@ -1,0 +1,50 @@
+# prefetch-dependencies-oci-ta task
+
+Task that uses Cachi2 to prefetch build dependencies. The fetched dependencies and the
+application source code are stored as a trusted artifact in the provided OCI repository.
+For additional info on Cachi2, see docs at
+https://github.com/containerbuildsystem/cachi2#basic-usage.
+
+## Configuration
+
+Config file must be passed as a YAML string. For all available config options please check
+[available configuration parameters] page.
+
+Example of setting timeouts:
+
+```yaml
+params:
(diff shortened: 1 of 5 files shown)


chmeliik:
@ligangty could you give the PR a more complete description? What is it for? What does cachi2 not support that you want it to support?

ligangty:
@chmeliik In our new pipeline, we will use the recent feature of cachi2 to download "generic" type of artifacts(for maven) from cachi2 specific lockfile (should be covered in here: https://github.com/containerbuildsystem/cachi2/pull/692), and also need to specify multiple paths of lockfiles with json flag like:
`
{
     "packages": [
         {"type": "generic", "path": "path1"}, 
         {"type": "generic", "path": "path2"} 
     ] 
}
`
Seems current prefetch task does not support this requirement, that's why I gave this PR. I'm not sure if I need to give this as "0.2" or just change it directly in "0.1". Please correct me if I'm in wrong way.

chmeliik:
> I think that should work fine in version 0.1 (without the changes in this PR) 🤔

Have you tried setting `prefetch-input` to `{ "packages": [ {"type": "generic", "path": "path1"}, {"type": "generic", "path": "path2"} ] }`?

ligangty:
> > I think that should work fine in version 0.1 (without the changes in this PR) 🤔
> 
> Have you tried setting `prefetch-input` to `{ "packages": [ {"type": "generic", "path": "path1"}, {"type": "generic", "path": "path2"} ] }`?

@chmeliik this is working! Thanks for the info. I'll close this one.

PR #1623, opened 2024-11-19
use regex to ignore commented  "subscription-manager register"

### LINKED ISSUE #1620: exclude comments when detecting if a task is registering itself with rhsm
this ought to exclude comments, otherwise a line such as
```
# install packages with no need to register using subscription-manager
dnf -y install hello
```
will trigger this detection to think the dockerfile is registering on its own

_Originally posted by @ptoscano in https://github.com/konflux-ci/build-definitions/pull/1528#discussion_r1816773041_

brianwcook:
Closes  #1620


chmeliik (approved):
LGTM, just need to re-generate buildah-remote-oci-ta as well

Thanks for the fix :+1:

PR #1630, opened 2024-11-21
Make EC strict mode configurable

Zokormazo:
Allow modifying `STRICT`. This allows to mark an EC as informative and not gate changes on it.

This is useful for us while we iterate to fix all the violations without gating yet the pull requests

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -71,6 +71,15 @@ spec:
       type: string
       description: PipelineRun ID
       default: "pr/$(context.pipelineRun.name)"
+    - name: STRICT
+      type: string
+      description: |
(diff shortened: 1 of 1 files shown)


zregvart (approved):
LGTM. cc @konflux-ci/ec

lcarva:
I see the value of doing this, just please be wary that if the check is green no one usually digs through the check to see if there are any issues. If setting strict to false, there is a high chance that you won't find out about issues until release time which is usually not a great user experience.

Could you update the description of the parameter to provide some sort of warning to future users thinking about setting struct to false?

Zokormazo:
> Could you update the description of the parameter to provide some sort of warning to future users thinking about setting struct to false?

Sure

PR #1635, opened 2024-11-21
fix: extract region from aws s3 urls

ralphbean:
It turns out that IBM Cloud and AWS s3 bucket urls are constructed differently, and the region is in a different place in the string.

Add a conditional here to catch AWS-style urls.


ralphbean:
Going to force merge this one.

PR #1643, opened 2024-11-25
feat(MMENG-4237): add build-maven-zip task

ligangty:
This task will be used in the new Konflux MRRC maven build pipeline. It is responsible for these work:

* Consume cachi2 prefetched aritfacts of "generic-fetcher" package type
* Generate checksums(md5, sha1, sha256) for all artifacts
* Bundle the artifacts and checksums into maven zip 
* Upload the maven zip to quay.io as OCI-artifact
* Attach sbom generated by cachi2 to the OCI-artifact
* Store these results into IMAGE_DIGEST, IMAGE_URL and SBOM_BLOB_URL

It also includes the oci-ta version.

BTW, we've added a new owner alias to put our team members in, and also mark the owners of this task as our team for future update.

### DIFF
--- a/task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
+++ b/task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
@@ -1,0 +1,209 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: build-maven-zip-oci-ta
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: maven-build, konflux
+  labels:
+    app.kubernetes.io/version: "0.1"
+    build.appstudio.redhat.com/build_type: maven-zip
+spec:
+  description: |-
+    Build-maven-zip task builds prefetched maven artifacts into a OCI-artifact with zip bundle  and pushes the OCI-artifact into container registry.
+    In addition it will use the SBOM file in prefetch-task, pushes the SBOM file to same registry of zip oci-artifact using cosign tool.
+    Note that this task needs the output of prefetch-dependencies task. If it is not activated, there will not be any output from this task.
+  params:
+    - name: CACHI2_ARTIFACT
+      description: The Trusted Artifact URI pointing to the artifact with
+        the prefetched dependencies.
+      type: string
+      default: ""
+    - name: FILE_NAME
+      description: The zip bundle file name of archived artifacts
+      type: string
+      default: maven-repository
+    - name: IMAGE
+      description: Reference of the OCI-Artifact this build task will produce.
+      type: string
+    - name: IMAGE_EXPIRES_AFTER
+      description: Delete image tag after specified time. Empty means to keep
+        the image tag. Time values could be something like 1h, 2d, 3w for
+        hours, days, and weeks, respectively.
+      type: string
+      default: ""
+    - name: PREFETCH_ROOT
+      description: The root directory of the artifacts under the prefetched
+        directory. Will be kept in the maven zip as the top directory for
+        all artifacts.
+      type: string
+      default: maven-repository
+    - name: caTrustConfigMapKey
+      description: The name of the key in the ConfigMap that contains the
+        CA bundle data.
+      type: string
+      default: ca-bundle.crt
+    - name: caTrustConfigMapName
+      description: The name of the ConfigMap to read CA bundle data from.
+      type: string
+      default: trusted-ca
+  results:
+    - name: IMAGE_DIGEST
+      description: Digest of the OCI-Artifact just built
+    - name: IMAGE_REF
+      description: OCI-Artifact reference of the built OCI-Artifact
+    - name: IMAGE_URL
+      description: OCI-Artifact repository and tag where the built OCI-Artifact
+        was pushed
+    - name: SBOM_BLOB_URL
+      description: Reference of SBOM blob digest to enable digest-based verification
+        from provenance
+      type: string
+  volumes:
+    - name: shared
+      emptyDir: {}
+    - name: trusted-ca
+      configMap:
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        name: $(params.caTrustConfigMapName)
+        optional: true
+    - name: workdir
+      emptyDir: {}
+  stepTemplate:
+    env:
+      - name: FILE_NAME
+        value: $(params.FILE_NAME)
+      - name: IMAGE
+        value: $(params.IMAGE)
+      - name: IMAGE_EXPIRES_AFTER
+        value: $(params.IMAGE_EXPIRES_AFTER)
+      - name: PKG_ROOT
+        value: $(params.PREFETCH_ROOT)
+    volumeMounts:
+      - mountPath: /shared
+        name: shared
+      - mountPath: /var/workdir
+        name: workdir
+  steps:
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:81c4864dae6bb11595f657be887e205262e70086a05ed16ada827fd6391926ac
+      args:
+        - use
+        - $(params.CACHI2_ARTIFACT)=/var/workdir/cachi2
+    - name: prepare
+      image: quay.io/konflux-ci/appstudio-utils@sha256:d0fecb67dd98d874704f2e3d20686363de7cbb42c15ad2b1d8e8c38118c50078
+      workingDir: /var/workdir
+      script: |
+        #!/bin/bash
+        set -euo pipefail
+        # Generate checksums for all maven artifact files. It will ignore the checksum files
+        # and signature files if they existed there
+        pkgpath="/var/workdir/cachi2/output/deps/generic"
+        if [ -d "$pkgpath/${PKG_ROOT}" ]; then
+          echo "Generating checksums for artifacts"
+          while IFS= read -r -d '' f; do
+            md5sum "$f" | awk '{print $1}' >"$f.md5"
+            sha1sum "$f" | awk '{print $1}' >"$f.sha1"
+            sha256sum "$f" | awk '{print $1}' >"$f.sha256"
+          done < <(find "$pkgpath/${PKG_ROOT}" -type f ! -name "*.md5" \
+            ! -name "*.sha1" ! -name "*.sha128" ! -name "*.sha256" \
+            ! -name "*.sha512" ! -name "*.asc" -print0)
+          # Bundle the artifacts and checksums together into a zip file
+          cd "$pkgpath"
+          echo "create maven zip to /var/workdir/cachi2/output/${FILE_NAME}.zip"
+          zip -rq "${FILE_NAME}.zip" "${PKG_ROOT}"
+          mv "${FILE_NAME}.zip" "/shared/${FILE_NAME}.zip"
+        else
+          echo "No ${PKG_ROOT} dir found, can not generate maven zip!"
+          exit 1
+        fi
+      computeResources:
+        limits:
+          cpu: "4"
+          memory: 8Gi
+        requests:
+          cpu: "1"
+          memory: 2Gi
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
+    - name: build
+      image: quay.io/konflux-ci/oras:latest@sha256:9d6db5840c70e65fefe041201cc7ffe2d1661bd0582b590b54787213ccfd76e9
+      workingDir: /var/workdir
+      volumeMounts:
+        - mountPath: /mnt/trusted-ca
+          name: trusted-ca
+          readOnly: true
+      script: |
+        #!/bin/bash
+        set -euo pipefail
+        if [ -f "/shared/${FILE_NAME}.zip" ]; then
+          mv "/shared/${FILE_NAME}.zip" "./${FILE_NAME}.zip"
+          select-oci-auth "$IMAGE" >auth.json
+          [ -n "$IMAGE_EXPIRES_AFTER" ] && EXPIRE_LABEL=("--annotation" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+          oras push "$IMAGE" \
+            --registry-config auth.json \
+            "${EXPIRE_LABEL[@]}" \
+            --artifact-type application/vnd.maven+zip "${FILE_NAME}.zip"
+          RESULTING_DIGEST=$(oras resolve --registry-config auth.json "${IMAGE}")
+          echo -n "$IMAGE" | tee "$(results.IMAGE_URL.path)"
+          echo -n "$RESULTING_DIGEST" | tee "$(results.IMAGE_DIGEST.path)"
+          echo -n "${IMAGE}@${RESULTING_DIGEST}" | tee "$(results.IMAGE_REF.path)"
+        else
+          echo "The maven zip file is not found!"
+          exit 1
+        fi
+
+        # Save the SBOM produced by Cachi2 so it can be merged into the final SBOM later
+        if [ -f "/var/workdir/cachi2/output/bom.json" ]; then
+          cp -vf "/var/workdir/cachi2/output/bom.json" ./sbom-cyclonedx.json
+        else
+          echo "The SBOM file for fetched artifacts is not found!"
+          exit 1
+        fi
+      computeResources:
+        limits:
+          cpu: "4"
+          memory: 8Gi
+        requests:
+          cpu: "1"
+          memory: 2Gi
+      securityContext:
+        capabilities:
+          add:
+            - SETFCAP
+    - name: upload-sbom
+      image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+      workingDir: /var/workdir
+      volumeMounts:
+        - mountPath: /mnt/trusted-ca
+          name: trusted-ca
+          readOnly: true
+      script: |
+        #!/bin/bash
+        set -euo pipefail
+        ca_bundle=/mnt/trusted-ca/ca-bundle.crt
+        if [ -f "$ca_bundle" ]; then
+          echo "INFO: Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+          update-ca-trust
+        fi
+
+        cosign attach sbom --sbom sbom-cyclonedx.json --type cyclonedx "$IMAGE"
+
+        # Remove tag from IMAGE while allowing registry to contain a port number.
+        sbom_repo="${IMAGE%:*}"
+        sbom_tag="sha256-$(                                <"$(results.IMAGE_DIGEST.path)" cut -d: -f2).sbom"
+        # The SBOM_BLOB_URL is created by `cosign attach sbom`.
+        echo -n "${sbom_repo}:${sbom_tag}" | tee "$(results.SBOM_BLOB_URL.path)"
+      computeResources:
+        limits:
+          cpu: 200m
+          memory: 512Mi
+        requests:
+          cpu: 100m
+          memory: 256Mi
--- a/task/build-maven-zip/0.1/README.md
+++ b/task/build-maven-zip/0.1/README.md
@@ -1,0 +1,28 @@
+# build-maven-zip task
+
+Build-maven-zip task builds prefetched maven artifacts into a OCI-artifact with zip bundle  and pushes the OCI-artifact into container registry.
+In addition it will use the SBOM file in prefetch-task, pushes the SBOM file to same registry of zip oci-artifact using cosign tool.
+Note that this task needs the output of prefetch-dependencies task. If it is not activated, there will not be any output from this task.
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|IMAGE|Reference of the OCI-Artifact this build task will produce.||true|
+|PREFETCH_ROOT|The root directory of the artifacts under the prefetched directory. Will be kept in the maven zip as the top directory for all artifacts.|maven-repository|false|
+|FILE_NAME|The zip bundle file name of archived artifacts|maven-repository|false|
+|IMAGE_EXPIRES_AFTER|Delete image tag after specified time. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.|""|false|
+|caTrustConfigMapName|The name of the ConfigMap to read CA bundle data from.|trusted-ca|false|
+|caTrustConfigMapKey|The name of the key in the ConfigMap that contains the CA bundle data.|ca-bundle.crt|false|
+
+## Results
+|name|description|
+|---|---|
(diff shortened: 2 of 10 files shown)


task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,225 @@
...
+        cpu: '4'
+      requests:
+        memory: 2Gi
+        cpu: '1'
+    script: |
+      #!/bin/bash
+      set -e
+      # Generate checksums for all maven artifact files. It will ignore the checksum files
+      # and signature files if they existed there
+      BYPASS_FILE_TYPES=("md5" "sha1" "sha128" "sha256" "sha512" "asc")
```
tkdchen: Can we use Bash Associative arrays to implement the bypass? It's more like operating a dict in other programming language, which would make the code more readable than using flag+loop.

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,225 @@
...
+    computeResources:
+      limits:
+        memory: 8Gi
+        cpu: '4'
+      requests:
+        memory: 2Gi
+        cpu: '1'
+    script: |
+      #!/bin/bash
+      set -e
```
tkdchen, line 79: Let's add `set -o pipefail`

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,212 @@
...
+      if [ -f "/shared/${BUNDLE_NAME}.zip" ]; then
+        mv "/shared/${BUNDLE_NAME}.zip" "./${BUNDLE_NAME}.zip"
+        select-oci-auth "$IMAGE" > auth.json
+        [ -n "$IMAGE_EXPIRES_AFTER" ] && EXPIRE_LABEL=("--annotation" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+        oras push "$IMAGE" \
+        --registry-config auth.json \
+        "${EXPIRE_LABEL[@]}" \
+        --artifact-type application/vnd.maven+zip "${BUNDLE_NAME}.zip"
```
tkdchen, line 126: > Upload the maven zip to quay.io as OCI-artifact

Upload the zip package as a standalone artifact or attach it to the image built by the build-container task?

Env var `IMAGE` is described in the `spec.params` as following:

> Reference of the image buildah will produce.

But, this `oras push` pushes the zip package to the registry as image `$IMAGE`.

I guess you probably need `oras attach` to attach the zip package to the image `$IMAGE`.

Or, does the param description need to be changed?

By the way, for easier-to-read, please indent the continuous lines with at least 2 spaces. 2 is preferred personally.

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,212 @@
...
+        cpu: 200m
+      requests:
+        memory: 256Mi
+        cpu: 100m
+    script: |
+      if [ -f sbom-cachi2.json ]; then
+        mv sbom-cachi2.json sbom-cyclonedx.json
+        echo "Creating sbom-purl.json"
+        python3 /scripts/create_purl_sbom.py
+      fi
```
tkdchen, line 165: The previous step https://github.com/konflux-ci/build-definitions/pull/1643/files#diff-6199ec974f25eac323954ab607a9160cb1fe2a523abf0e255775588e1e2bda01R136 has already ensured the existence of file `./sbom-cachi2.json`.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: This one is removed by using "find" to bypass them.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: This task will directly push the zip file as oci-artifact, so there will not be any images generated. This IMAGE param is used here because the pipeline will use IMAGE to pass the registry name which we can not change, so I'd prefer to leave it here, but I'd like to change the description.

ligangty: The indent changed in new commit.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: Fixed.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: This BYPASS has been removed in commit. Instead the logic is implemented through "find" directly.

task/build-maven-zip/0.1/build-maven-zip.yaml
tkdchen: But, the logic looks ambiguous. In previous step, if the bom.json does not exist, it is an error and will fail the task, then no cachi2-sbom.json is created, then perhaps fail the pipeline, but in this step, the `sbom-cachi2.json` becomes optional for this build task.

ligangty: Ok. I've removed it in new commit.

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
```diff
@@ -0,0 +1,220 @@
...
+        while IFS= read -r -d '' f
+        do
+          md5sum "$f" | awk '{print $1}'> "$f.md5"
+          sha1sum "$f" | awk '{print $1}'> "$f.sha1"
+          sha256sum "$f" | awk '{print $1}'> "$f.sha256"
+        done <   <(find "$pkgpath/${PKG_ROOT}" -type f ! -name "*.md5" \
+        ! -name "*.sha1" ! -name "*.sha128" ! -name "*.sha256" \
+        ! -name "*.sha512" ! -name "*.asc" -print0)
+        # Bundle the artifacts and checksums together into a zip file
+        cd "$pkgpath" || exit
```
MartinBasti, line 106: What's purpose of that exit on this line?

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,210 @@
...
+        cpu: '4'
+      requests:
+        memory: 2Gi
+        cpu: '1'
+    script: |
+      #!/bin/bash
+      set -o pipefail
+      # Generate checksums for all maven artifact files. It will ignore the checksum files
+      # and signature files if they existed there
+      pkgpath="$(workspaces.source.path)/cachi2/output/deps/$INPUT"
```
chmeliik, line 82: possible forms of `INPUT` at the moment:
1. just a package manager name, e.g. `generic`
2. `{"type": "<package manager name>", <more attributes>}`
3. an array of the objects from above, e.g. `[{"type": "generic", ...}, {"type": "rpm", ...}, ...]`
4. `{"packages": [<an array like above>], "flags": [<supported flags>]}`

This only supports the simplest one

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,210 @@
...
+          md5sum "$f" | awk '{print $1}'> "$f.md5"
+          sha1sum "$f" | awk '{print $1}'> "$f.sha1"
+          sha256sum "$f" | awk '{print $1}'> "$f.sha256"
+        done <   <(find "$pkgpath/${PKG_ROOT}" -type f ! -name "*.md5" \
+        ! -name "*.sha1" ! -name "*.sha128" ! -name "*.sha256" \
+        ! -name "*.sha512" ! -name "*.asc" -print0)
+        # Bundle the artifacts and checksums together into a zip file
+        cd "$pkgpath" || exit
+        echo "create maven zip to $(workspaces.source.path)/cachi2/output/${BUNDLE_NAME}.zip"
+        zip -rq "${BUNDLE_NAME}.zip" "${PKG_ROOT}"
```
chmeliik, line 96: The `generic` fetcher isn't just for maven artifacts. Is it ok to bundle all kinds of arbitrary artifacts into the "Maven zip"?

task/build-maven-zip/0.1/build-maven-zip.yaml
chmeliik: TBH, this task probably shouldn't take `PREFETCH_INPUT` into account at all, it should only handle `cachi2/output/deps/generic`

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
ligangty: This is a checkton reported error. It recommends to use this exit for cd.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: Sounds ok. I'll use it directly.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: The design for this generic-fetcher can support all kinds of "artifacts" here, not maven. But actually maven does not have a strict rule to limit what's maven artifacts, it's also support all kinds of files as "artifacts", so we don't have a rule to distinguish the non-maven artifacts in these arbitrary ones. So this should be only controlled by the user's lockfiles.

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
chmeliik: You might want `set -e` at the beginning of every script instead, unless this is really the only command that needs error handling

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: Fixed in new commit

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
MartinBasti: +1 for `set -euo pipefail`

ligangty: Oh sorry forget this one. Just added and fixed that exit.

task/build-maven-zip/0.1/README.md
```diff
@@ -0,0 +1,29 @@
...
+Build-maven-zip task builds prefetched maven artifacts into a OCI-artifact with zip bundle  and pushes the OCI-artifact into container registry.
+In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
+Note that this task needs the output of prefetch-dependencies task. If it is not activated, there will not be any output from this task.
+
+## Parameters
+
+| name                 | description                                                            | default value    | required |
+| -------------------- | ---------------------------------------------------------------------- | ---------------- | -------- |
+| IMAGE                | Reference of the OCI-Artifact this build task will produce.            |                  | true     |
+| PREFETCH_ROOT        | The root directory of the artifacts in the prefetched directory.       | maven-repository | false    |
```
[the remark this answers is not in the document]
ligangty: This is something else. It's under that deps/generic. We had an agreement with the pilot user to make sure this root is defined in the lockfile which default as "maven-repository". So just leave a param here if user want to change it.

task/build-maven-zip/0.1/README.md
```diff
@@ -0,0 +1,29 @@
+# Build-maven-zip task
+
+Build-maven-zip task builds prefetched maven artifacts into a OCI-artifact with zip bundle  and pushes the OCI-artifact into container registry.
+In addition it generates a SBOM file, injects the SBOM file into final container image and pushes the SBOM file as separate image using cosign tool.
```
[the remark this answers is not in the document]
ligangty: It will attach the sbom to the oci-artifact by using cosign tool. But maybe it's not a image. I think I should change the description here.

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,204 @@
...
+      limits:
+        memory: 512Mi
+        cpu: 200m
+      requests:
+        memory: 256Mi
+        cpu: 100m
+    script: |
+      mv sbom-cachi2.json sbom-cyclonedx.json
+      echo "Creating sbom-purl.json"
+      python3 /scripts/create_purl_sbom.py
```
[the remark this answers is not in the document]
ligangty: I just copied this one from buildah task but I don't make sure how this create_prul_sbom.py is working. From its name I'm guessing it will do some changes for the purl on the sbom, so I just leave it here. If someone can tell that this is not usable, I'll remove it.

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
chmeliik: Ack, could you update the description to make it clearer?

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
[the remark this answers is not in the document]
ligangty: Yes it's created manually. After created I've run the generate-ta-tash.sh but seems nothing changed. Should I remove the oci-ta one and run that script again?

task/build-maven-zip/0.1/README.md
```diff
@@ -0,0 +1,29 @@
...
+Note that this task needs the output of prefetch-dependencies task. If it is not activated, there will not be any output from this task.
+
+## Parameters
+
+| name                 | description                                                            | default value    | required |
+| -------------------- | ---------------------------------------------------------------------- | ---------------- | -------- |
+| IMAGE                | Reference of the OCI-Artifact this build task will produce.            |                  | true     |
+| PREFETCH_ROOT        | The root directory of the artifacts in the prefetched directory.       | maven-repository | false    |
+| BUNDLE_NAME          | The zip bundle name of archived artifacts.                             | maven-repository | false    |
```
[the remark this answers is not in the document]
ligangty: Not any prefer from my side. Ok to change.

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
chmeliik: Oh, and strictly speaking, this task doesn't _generate_ an SBOM file. It just pushes the one that cachi2 generated

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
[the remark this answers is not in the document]
chmeliik: Yeah. Remove the oci-ta variant, create the `recipe.yaml` for it and run the generator script

task/build-maven-zip/0.1/build-maven-zip.yaml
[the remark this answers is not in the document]
chmeliik: It doesn't do anything to `sbom-cyclonedx.json`, it creates `sbom-purl.json`. Not applicable

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
```diff
@@ -0,0 +1,214 @@
...
+      mv sbom-cachi2.json sbom-cyclonedx.json
+      echo "Creating sbom-purl.json"
+      python3 /scripts/create_purl_sbom.py
+    securityContext:
+      runAsUser: 0
+  - name: upload-sbom
+    image: quay.io/konflux-ci/appstudio-utils:ab6b0b8e40e440158e7288c73aff1cf83a2cc8a9@sha256:24179f0efd06c65d16868c2d7eb82573cce8e43533de6cea14fec3b7446e0b14
+    workingDir: /var/workdir
+    script: |
+      #!/bin/bash
```
MartinBasti, line 176: `set -euo pipefail` also here?

task/build-maven-zip/0.1/build-maven-zip.yaml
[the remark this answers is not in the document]
ligangty: Ok. I'll remove it.

task/build-maven-zip/0.1/README.md
```diff
@@ -0,0 +1,29 @@
...
+| caTrustConfigMapName | The name of the ConfigMap to read CA bundle data from.                 | trusted-ca       | false    |
+| caTrustConfigMapKey  | The name of the key in the ConfigMap that contains the CA bundle data. | ca-bundle.crt    | false    |
+
+## Results
+
+| name          | description                                                                       |
+| ------------- | --------------------------------------------------------------------------------- |
+| IMAGE_DIGEST  | Digest of the OCI-Artifact just built                                             |
+| IMAGE_URL     | Image repository and tag where the built OCI-Artifact was pushed                  |
+| SBOM_BLOB_URL | Reference of SBOM blob digest to enable digest-based verification from provenance |
```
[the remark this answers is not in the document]
ligangty: Fixed in new commit

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
ligangty: Changed in new commit

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
ligangty: Updated in new commit.

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
ligangty: Added in new commit.

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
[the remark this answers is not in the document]
ligangty: Ok. Done in new commit.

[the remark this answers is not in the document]
ligangty: Weird. Seems this always report yamllint error in CI. Do you know why this happens? @chmeliik

task/build-maven-zip/0.1/build-maven-zip.yaml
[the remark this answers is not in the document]
ligangty: Removed in new commit.

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
ligangty: I've updated the description here in new commit.

task/build-maven-zip-oci-ta/0.1/build-maven-zip-oci-ta.yaml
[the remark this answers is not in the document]
chmeliik: >    Error: 14:141 [new-line-at-end-of-file] no new line character at the end of file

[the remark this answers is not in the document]
ligangty: @chmeliik but this is generated by the generate-ta-task.sh, if I did change to this file, the "Validate PR - Trusted Artifact variants / Check Trusted Artifact variants" will report failure of ta not updated error.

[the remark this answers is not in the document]
ligangty: Oh sorry, checked at wrong file. It's recipe.yaml not the build-maven-zip-oci-ta.yaml. Fixed.

task/build-maven-zip/0.1/README.md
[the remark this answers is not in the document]
chmeliik: > But the default name should not include .zip there because it is handled in the task logic.

Ah, that's what I meant - that the parameter should be the whole filename, not just the part before the task adds `.zip`. It's just clearer what the parameter means that way

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,194 @@
...
+  steps:
+  - image: quay.io/konflux-ci/appstudio-utils@sha256:d0fecb67dd98d874704f2e3d20686363de7cbb42c15ad2b1d8e8c38118c50078
+    name: prepare
+    computeResources:
+      limits:
+        memory: 8Gi
+        cpu: '4'
+      requests:
+        memory: 2Gi
+        cpu: '1'
```
chmeliik, line 74: Does it need that much memory? It's just creating one zip file

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,194 @@
...
+    workingDir: $(workspaces.source.path)
+  - image: quay.io/konflux-ci/oras:latest@sha256:9d6db5840c70e65fefe041201cc7ffe2d1661bd0582b590b54787213ccfd76e9
+    name: build
+    computeResources:
+      limits:
+        memory: 8Gi
+        cpu: '4'
+      requests:
+        memory: 2Gi
+        cpu: '1'
```
chmeliik, line 113: Same here, that's way too much just to run oras

task/build-maven-zip/0.1/build-maven-zip.yaml
```diff
@@ -0,0 +1,194 @@
...
+    script: |
+      #!/bin/bash
+      set -euo pipefail
+      if [ -f "/shared/${FILE_NAME}.zip" ]; then
+        mv "/shared/${FILE_NAME}.zip" "./${FILE_NAME}.zip"
+        select-oci-auth "$IMAGE" > auth.json
+        [ -n "$IMAGE_EXPIRES_AFTER" ] && EXPIRE_LABEL=("--annotation" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+        oras push "$IMAGE" \
+          --registry-config auth.json \
+          "${EXPIRE_LABEL[@]}" \
```
chmeliik, line 123: Won't the task always fail when  `IMAGE_EXPIRES_AFTER` is empty? (Because `EXPIRE_LABEL` will be unset here)

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: I heard from user that the prefetched artifacts may contain more than 10k+ files, and the whole zip file size will be 1GB+ or even great than 2GB. So maybe 2Gi is better?

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: Not sure how much memory will oras cost to upload a 2GB+ file. If we can make sure the file size will bring big impact on oras upload, we can reduce it to 1Gi.

task/build-maven-zip/0.1/build-maven-zip.yaml
ligangty: Not sure. I will have a test with my user workspace.

ligangty: Just had a quick test which set the "image-expires-after" to "" in my test pipeline, seems it does not break anything. The oci-artifact is pushed to quay.io without any expire label.

task/build-maven-zip/0.1/build-maven-zip.yaml
chmeliik: ok

tkdchen (approved):

PR #1646, opened 2024-11-25
fix: delete propagated kubeconfig

yftacherzog:
when getting ephemeral cluster credentials the kubeconfig file was assigned to the KUBECONFIG env var which is deleted before the script ends.

This change removes that assignment and instead explicitly mentions the kubeconfig path when it's used.


amisstea:
LGTM :+1:. Just need to wait for @JoaoFula to confirm it works before merging.

tnevrlka (approved):

PR #1659, opened 2024-11-28
Improve generation of base images SBOMs

mkosiarc:
most functional changes are in the related PR that updates the
base_images_sbom_script.py
https://github.com/konflux-ci/build-tasks-dockerfiles/pull/191

Here, we are just updating on how we generate the inputs for this
script.
We are now passing the whole parsed Dockerfile in json format to that
script, which allows us to better parse/detect base images.

Also, the format of the /shared/base_images_digests file was changed.
Previously we could rely on the order of the image references with the
digests in the file. Now we need to provide a mapping from an image
reference as it was used in the Dockerfile to the full image reference
with digests that was used during build and generated by buildah.

The mapping is done as:
<image-reference-used-in-dockerfile> <full-image-reference-with-digest>

Also, the sbom utility image has to be updated together in the same
PR/commit, otherwise it would break konflux temporarily

[KFLUXBUGS-1718](https://issues.redhat.com//browse/KFLUXBUGS-1718)

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -163,7 +163,7 @@ spec:
         add:
           - SETFCAP
 
-  - image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:f1c07f273536a00fa6539e2aa41b3fddab3bb282a157c9676572858bfd19d7e4
+  - image: quay.io/redhat-appstudio/sbom-utility-scripts-image@sha256:e1347023ef1e83d52813c26384f551e3a03e482539d17a647955603e7ea6b579
     name: create-sbom
     computeResources:
       limits:
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -348,9 +348,9 @@ spec:
           BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
         done
 
+        dockerfile-json "${BUILD_ARG_FLAGS[@]}" "$dockerfile_copy" >/shared/parsed_dockerfile.json
         BASE_IMAGES=$(
-          dockerfile-json "${BUILD_ARG_FLAGS[@]}" "$dockerfile_copy" |
-            jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName | select(test("^oci-archive:") | not)'
+          jq -r '.Stages[] | select(.From | .Stage or .Scratch | not) | .BaseName | select(test("^oci-archive:") | not)' /shared/parsed_dockerfile.json
         )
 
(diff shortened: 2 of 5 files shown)


mkosiarc:
has to be merged only after https://github.com/konflux-ci/build-tasks-dockerfiles/pull/191

chmeliik (commented):
> `<image-reference-used-in-dockerfile>|<full-image-reference-with-digest>`
> The character that creates the mapping "|" was chosen arbitrarily, just
because it cannot be a valid part of image reference.

It would be a bit more natural to just separate them by whitespace, but this is OK too

mkosiarc:
> > `<image-reference-used-in-dockerfile>|<full-image-reference-with-digest>`
> > The character that creates the mapping "|" was chosen arbitrarily, just
> > because it cannot be a valid part of image reference.
> 
> It would be a bit more natural to just separate them by whitespace, but this is OK too

I guess I don't mind using space instead of "|". I just used it for more visibility and because of an (possibly irrational) fear that additional empty space might find its way to the file, causing problems with the parsing.

chmeliik:
> > > `<image-reference-used-in-dockerfile>|<full-image-reference-with-digest>`
> > > The character that creates the mapping "|" was chosen arbitrarily, just
> > > because it cannot be a valid part of image reference.
> > 
> > 
> > It would be a bit more natural to just separate them by whitespace, but this is OK too
> 
> I guess I don't mind using space instead of "|". I just used it for more visibility and because of an (possibly irrational) fear that additional empty space might find its way to the file, causing problems with the parsing.

Python handles additional spaces without issues

```
In [1]: "   image1\t \t image2     \n".split()
Out[1]: ['image1', 'image2']
```

mkosiarc:
> > > > `<image-reference-used-in-dockerfile>|<full-image-reference-with-digest>`
> > > > The character that creates the mapping "|" was chosen arbitrarily, just
> > > > because it cannot be a valid part of image reference.
> > > 
> > > 
> > > It would be a bit more natural to just separate them by whitespace, but this is OK too
> > 
> > 
> > I guess I don't mind using space instead of "|". I just used it for more visibility and because of an (possibly irrational) fear that additional empty space might find its way to the file, causing problems with the parsing.
> 
> Python handles additional spaces without issues
> 
> ```python
> In [1]: "   image1\t \t image2     \n".split()
> Out[1]: ['image1', 'image2']
> ```
> 
> (and so does Bash, in fact)
> 
> ```shell
> $ read -r image1 image2 <<< $'   image1\t \t image2     \n'
> $ echo "$image1; $image2;"
> image1; image2;
> ```

I am aware of that :D I meant accidentally at the start or end of the file, on top of being a white space in the middle (resulting in more splits) - but we have control over that in the tasks - that's why I implied that the fear is probably just irrational and we can make the switch to using space.

mkosiarc:
Yeah, like I said, it was more irrational. I can make the switch to using space.

chmeliik (approved):

PR #1658, opened 2024-11-27
Move prefetch devs out of the larger build team

chmeliik:
They generally don't maintain anything in the build pipeline except for
the prefetch task.

Also: Don't run e2e-tests just for OWNERS changes (note that the e2e-tests _will_ trigger on this PR because it changes a file in `.tekton`)

PR #1669, opened 2024-11-29
Setup CODEOWNERS

chmeliik:
CODEOWNERS will allow us to make the ownership of individual tasks
actually mean something.

We will have to give every owner write permissions, but we will be able
to set up branch protection rules to block PRs from getting merged if
they don't have approvals from the owners.

This will allow the CODEOWNERS of tasks and stepactions to fully
self-service their PRs without giving them the ability to mess with
other parts of the repository.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,96 @@
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+*   @konflux-ci/build-maintainers
+
+/task/acs-deploy-check                      @konflux-ci/build-maintainers
+/task/acs-image-check                       @konflux-ci/build-maintainers
+/task/acs-image-scan                        @konflux-ci/build-maintainers
+/task/apply-tags                            @konflux-ci/build-maintainers
+/task/build-image-index                     @konflux-ci/build-maintainers
+/task/build-image-manifest                  @konflux-ci/build-maintainers
+/task/buildah                               @konflux-ci/build-maintainers
+/task/buildah-10gb                          @konflux-ci/build-maintainers
+/task/buildah-20gb                          @konflux-ci/build-maintainers
+/task/buildah-24gb                          @konflux-ci/build-maintainers
+/task/buildah-6gb                           @konflux-ci/build-maintainers
+/task/buildah-8gb                           @konflux-ci/build-maintainers
+/task/buildah-min                           @konflux-ci/build-maintainers
+/task/buildah-oci-ta                        @konflux-ci/build-maintainers
+/task/buildah-remote                        @konflux-ci/build-maintainers
+/task/buildah-remote-oci-ta                 @konflux-ci/build-maintainers
+/task/buildah-rhtap                         @konflux-ci/build-maintainers
+/task/download-sbom-from-url-in-attestation @konflux-ci/build-maintainers
+/task/gather-deploy-images                  @konflux-ci/build-maintainers
+/task/git-clone                             @konflux-ci/build-maintainers
+/task/git-clone-oci-ta                      @konflux-ci/build-maintainers
+/task/init                                  @konflux-ci/build-maintainers
+/task/push-dockerfile                       @konflux-ci/build-maintainers
+/task/push-dockerfile-oci-ta                @konflux-ci/build-maintainers
+/task/show-sbom                             @konflux-ci/build-maintainers
+/task/show-sbom-rhdh                        @konflux-ci/build-maintainers
+/task/slack-webhook-notification            @konflux-ci/build-maintainers
+/task/source-build                          @konflux-ci/build-maintainers
+/task/source-build-oci-ta                   @konflux-ci/build-maintainers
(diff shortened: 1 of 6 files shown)


chmeliik:
Changed the CODEOWNERS based on https://github.com/konflux-ci/build-definitions/pull/1658

Addressed the OWNERS_ALIASES vs. GH team mismatches
* `build-maintainers` matches the previous `build-team` now, minus the prefetch-only maintainers
* `ec` matches `ec-team` now
* `integration-service-maintainers` is `integration-team` + `chipspeak, jencull` (and those are integration team members, so that seems fine)

chmeliik:
Need to add one more thing - Renovate groupings based on CODEOWNERS. Otherwise Renovate PRs will need approvals from many different sets of owners

chmeliik:
CC @acornett21 @amisstea @arewm @avi-biton @bcrochet @brianwcook @brunoapimentel @cgwalters @eskultety @gbenhaim @gurnben @hmariset @jbpratt @kdudka @komish @ligangty @oamsalem @omeramsc @ralphbean @scoheb @skattoju @taylormadore @yftacherzog @yma96

This PR adds you as a CODEOWNER (because you were already in an OWNERS file), which means you'll now have the power to self-service changes to your tasks. On the other hand, it means your approval will be required in order for those changes to be merged. Let me know if you have any concerns about that.

chmeliik:
Btw, the groupings based on filenames seem to work well. See for example https://github.com/chmeliik/build-definitions/pull/45, https://github.com/chmeliik/build-definitions/pull/46, which both update the oras image, but in two separate PRs according to the matchFileNames groupings

chmeliik:
Added `/task/validate-fbc` and fixed the formatting (my auto-aligner script had a bug)

CODEOWNERS
```diff
@@ -0,0 +1,115 @@
...
+/task/sast-shell-check                      @konflux-ci/integration-service-maintainers @kdudka
+/task/sast-shell-check-oci-ta               @konflux-ci/integration-service-maintainers @kdudka
+/task/sast-snyk-check                       @konflux-ci/integration-service-maintainers @kdudka
+/task/sast-unicode-check                    @konflux-ci/integration-service-maintainers @kdudka
+
+# renovate groupName=preflight
+/task/ecosystem-cert-preflight-checks   @acornett21 @bcrochet @komish @skattoju
+
+# renovate groupName=eaas
+/task/provision-env-with-ephemeral-namespace    @amisstea @avi-biton @gbenhaim @oamsalem @yftacherzog
```
yftacherzog, line 78:
```suggestion
/task/provision-env-with-ephemeral-namespace    @amisstea @avi-biton @gbenhaim @omeramsc @yftacherzog
```

hack/update_renovate_json_based_on_codeowners.py
```diff
@@ -0,0 +1,112 @@
...
+        if not line.startswith("#"):
+            pattern, *_ = line.split(maxsplit=1)
+            patterns.append(pattern)
+        elif m := renovate_directive_pat.match(line):
+            groupname = m.group(1)
+
+    if groupname:
+        patterns = list(map(_codeowners_pattern_to_glob_pattern, patterns))
+        return {"groupName": groupname, "matchFileNames": patterns}
+    else:
```
tnevrlka, line 53: nit: the `else` is unnecessary afaik

tnevrlka (approved):
🌮

CODEOWNERS
chmeliik: :scream: good catch. @arewm we may have invited the wrong Omar into konflux-ci

hack/update_renovate_json_based_on_codeowners.py
chmeliik: Yup. Also reordered to avoid making the `return None` look like the happy path :smile:

CODEOWNERS
chmeliik: Fixed. Went through the rest of the list, I think the rest are legit

CODEOWNERS
```diff
@@ -0,0 +1,115 @@
...
+
+# renovate groupName=integration
+/task/clair-scan                @konflux-ci/integration-service-maintainers
+/task/clamav-scan               @konflux-ci/integration-service-maintainers
+/task/deprecated-image-check    @konflux-ci/integration-service-maintainers
+/task/fbc-related-image-check   @konflux-ci/integration-service-maintainers
+/task/fbc-validation            @konflux-ci/integration-service-maintainers
+/task/inspect-image             @konflux-ci/integration-service-maintainers
+/task/sast-snyk-check-oci-ta    @konflux-ci/integration-service-maintainers
+/task/sast-unicode-check-oci-ta @konflux-ci/integration-service-maintainers
```
14rcole, line 60: Both of these tasks should also have @kdudka as a maintainer

chmeliik: Makes sense, added

chmeliik:
There hasn't been pushback to this approach (or the pushback has been resolved). The CODEOWNERS file is now fully valid. Merging

PR #1670, opened 2024-11-29
Use revision when pull_request_number is missing

zregvart:
It seems that the `pull_request_number` might not be defined when running checks on a PR in merge queue. So if the value is empty we can try looking up the pull request number by searching for the commit that contained it.


chmeliik (approved):
Did trigger the expected tasks, LGTM

PR #1673, opened 2024-12-02
feat(MMENG-4247): Add maven-zip-build pipeline

ligangty:
This new build pipeline will do:
* Accept the cachi2 lockfiles for generic-fetcher as source code in git repository
* Prefetch the artifacts in the cachi2 lockfiles in prefetch task
* Build the OCI-artifact of maven zip bundle which is generated by the artifacts fetched through cachi2 in build-maven-zip task

It only uses these four tasks:
* init 0.2
* git-clone 0.1
* prefetch-dependencies 0.1
* build-maven-zip 0.1

The ta version will use the oci-ta versions of tasks instead of above.

### DIFF
--- a/pipelines/maven-zip-build-oci-ta/kustomization.yaml
+++ b/pipelines/maven-zip-build-oci-ta/kustomization.yaml
@@ -1,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../maven-zip-build
+
+patches:
+- path: patch.yaml
+  target:
+    kind: Pipeline
--- a/pipelines/maven-zip-build-oci-ta/patch.yaml
+++ b/pipelines/maven-zip-build-oci-ta/patch.yaml
@@ -1,0 +1,94 @@
+---
+- op: add
+  path: /spec/description
+  value: |
+    This pipeline will build the maven zip to oci-artifact while maintaining trust after pipeline customization.
+
+    _Uses `prefetch-dependencies` to fetch all artifacts which will be the content of the maven zip, and then uses `build-maven-zip-oci-ta` to create zip and push it to quay.io as oci-artifact. Information is shared between tasks using OCI artifacts instead of PVCs.
+    This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/repository/konflux-ci/tekton-catalog/pipeline-maven-zip-build-oci-ta?tab=tags)_
+- op: replace
+  path: /metadata/name
+  value: maven-zip-build-oci-ta
+- op: replace
+  path: /metadata/labels
+  value:
+    "pipelines.openshift.io/used-by": "build-cloud"
+    "pipelines.openshift.io/runtime": "generic"
+    "pipelines.openshift.io/strategy": "maven-zip"
+# clone-repository Task
+- op: replace
+  path: /spec/tasks/1/taskRef/name
+  value: git-clone-oci-ta
+- op: add
+  path: /spec/tasks/1/params/-
+  value:
+    name: ociStorage
+    value: $(params.output-image).git
+- op: add
+  path: /spec/tasks/1/params/-
+  value:
+    name: ociArtifactExpiresAfter
+    value: $(params.image-expires-after)
+- op: remove
+  path: /spec/tasks/1/workspaces/0
+# prefetch-dependencies Task
+- op: replace
+  path: /spec/tasks/2/taskRef/name
+  value: prefetch-dependencies-oci-ta
+- op: add
+  path: /spec/tasks/2/params/-
+  value:
+    name: SOURCE_ARTIFACT
+    value: $(tasks.clone-repository.results.SOURCE_ARTIFACT)
+- op: add
+  path: /spec/tasks/2/params/-
+  value:
+    name: ociStorage
+    value: $(params.output-image).prefetch
+- op: add
+  path: /spec/tasks/2/params/-
+  value:
+    name: ociArtifactExpiresAfter
+    value: $(params.image-expires-after)
+- op: remove
+  path: /spec/tasks/2/workspaces/0
+- op: remove
+  path: /spec/tasks/2/when
+# Replace build-maven-zip task by build-maven-zip-oci-ta task
+- op: replace
+  path: /spec/tasks/3/taskRef
+  value:
+    name: build-maven-zip-oci-ta
+    version: "0.1"
+- op: add
+  path: /spec/tasks/3/params/-
+  value:
+    name: CACHI2_ARTIFACT
+    value: $(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)
+- op: remove
+  path: /spec/tasks/3/workspaces/0
+# sast-snyk-check
+- op: replace
+  path: /spec/tasks/4/taskRef/name
+  value: sast-snyk-check-oci-ta
+- op: add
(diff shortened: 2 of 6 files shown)


pipelines/maven-zip-build-ta/README.md
```diff
@@ -0,0 +1,148 @@
+# "maven-zip-build-ta pipeline"
```
MartinBasti, line 1: should be named with suffix `-oci-ta`

ligangty: Fixed in new commit.

pipelines/maven-zip-build-oci-ta/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../template-build
```
chmeliik, line 4: Maybe using `../maven-zip-build` as the base would be easier?

pipelines/maven-zip-build/patch.yaml
```diff
@@ -0,0 +1,125 @@
...
+- op: remove
+  path: /spec/params/7  # hermetic
+- op: remove
+  path: /spec/params/4  # dockerfile
+- op: remove
+  path: /spec/params/3  # path-context
+# Change default value of skip-checks
+- op: replace
+  path: /spec/params/4/default
+  value: true
```
chmeliik, line 46: Have you tried running the pipeline? I think this is invalid

chmeliik:
I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?

ligangty:
> I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?

@chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.

pipelines/maven-zip-build/patch.yaml
ligangty: I've tested the pipeline in my local workspace with corresponding changes and it can pass.

pipelines/maven-zip-build-oci-ta/kustomization.yaml
ligangty: Not sure. Let me have a check.

chmeliik:
> > I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?
> 
> @chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.

Hmm, I see. We can enable snyk at least, because it doesn't can the oci-artifact, it scans the source repo and the prefetched content

ligangty:
> > > I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?
> > 
> > 
> > @chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.
> 
> Hmm, I see. We can enable snyk at least, because it doesn't scan the oci-artifact, it scans the source repo and the prefetched content

Ok, I'll added it and have a try in my workspace.

MartinBasti:
> > > I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?
> > 
> > 
> > @chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.
> 
> Hmm, I see. We can enable snyk at least, because it doesn't scan the oci-artifact, it scans the source repo and the prefetched content

Is there a Enterprise Contract policy that will allow maven artifacts to pass without scans?

pipelines/maven-zip-build/patch.yaml
chmeliik: That's weird :thinking: 

```
$ oc kustomize pipelines/maven-zip-build | oc create -f -
Error from server (InternalError): error when creating "STDIN": Internal error occurred: Internal error occurred: conversion webhook for tekton.dev/v1, Kind=Pipeline failed: unable to unmarshal input: json: cannot unmarshal array into Go struct field PipelineSpec.spec.tasks of type v1.PipelineTask
```

(which is probably about something other than this boolean, but it doesn't work)

ligangty:
> > > > I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?
> > > 
> > > 
> > > @chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.
> > 
> > 
> > Hmm, I see. We can enable snyk at least, because it doesn't scan the oci-artifact, it scans the source repo and the prefetched content
> 
> Is there a Enterprise Contract policy that will allow maven artifacts to pass without scans?

I think we should create a new one later in release phase to disable the clair-scan.

pipelines/maven-zip-build/patch.yaml
ligangty: Ok let me have a check.

pipelines/maven-zip-build-oci-ta/kustomization.yaml
ligangty: Used ../maven-zip-build in new commit.

pipelines/maven-zip-build/patch.yaml
ligangty: Ok, I think this is cause by a typo in build-maven-zip task adding. An extra dash is added there. I've fixed it in new commit.

ligangty:
> > > > I noticed the pipeline has 0 scans, no `snyk`, `clamav`, nothing. Is that OK?
> > > 
> > > 
> > > @chmeliik I confirmed with @ralphbean that all current scan tasks can not work with oci-artifact, that's why I removed all these tasks there. When these tasks have ability to scan the binary content of the oci-artifact we may consider to add them back.
> > 
> > 
> > Hmm, I see. We can enable snyk at least, because it doesn't scan the oci-artifact, it scans the source repo and the prefetched content
> 
> Ok, I'll added it and have a try in my workspace.

Added in new commit.

pipelines/maven-zip-build-oci-ta/patch.yaml
```diff
@@ -0,0 +1,94 @@
...
+    value: $(params.output-image).prefetch
+- op: add
+  path: /spec/tasks/2/params/-
+  value:
+    name: ociArtifactExpiresAfter
+    value: $(params.image-expires-after)
+- op: remove
+  path: /spec/tasks/2/workspaces/0
+- op: remove
+  path: /spec/tasks/2/when
```
chmeliik, line 56: Why remove the `when` for prefetch-dependencies? The regular build-maven-zip pipeline has it, `-oci-ta` does not

chmeliik: Ah, found 51d7cf9ee96eae89520e8632bf013975dbdbdd65. Makes sense

Long live useful commit messages! :tada:

chmeliik (approved):

MartinBasti (approved):

PR #1684, opened 2024-12-03
feat: supply created annotation on bundles

ralphbean:
This will be helpful when debugging, to quickly identify how old a bundle is.

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -5,6 +5,10 @@ set -e -o pipefail
 VCS_URL=https://github.com/konflux-ci/build-definitions
 VCS_REF=$(git rev-parse HEAD)
 
+# https://reproducible-builds.org/docs/source-date-epoch/
+SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)
+BUILD_DATE="$(date --utc --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" +%Y-%m-%d)"
+
 function is_official_repo() {
     # match e.g.
     #   redhat-appstudio-tekton-catalog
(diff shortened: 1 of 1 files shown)


tnevrlka (approved):

hack/build-and-push.sh
```diff
@@ -5,6 +5,10 @@ set -e -o pipefail
 VCS_URL=https://github.com/konflux-ci/build-definitions
 VCS_REF=$(git rev-parse HEAD)
 
+# https://reproducible-builds.org/docs/source-date-epoch/
+SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)
+BUILD_DATE="$(date --utc --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" +%Y-%m-%d)"
```
lcarva, line 10: Unused?

hack/build-and-push.sh
```diff
@@ -171,6 +175,7 @@ do
       ANNOTATIONS=()
       ANNOTATIONS+=("org.opencontainers.image.source=${VCS_URL}")
       ANNOTATIONS+=("org.opencontainers.image.revision=${VCS_REF}")
+      ANNOTATIONS+=("org.opencontainers.image.created=${SOURCE_DATE_EPOCH}")
```
lcarva, line 178: From [spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys)

> org.opencontainers.image.created date and time on which the image was built, conforming to [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6).

I think it would be confusing to use the date of the git commit here because that is not actually the build date. They may be very different.

FYI, there is a build date already: 
```
🐚 skopeo inspect --config docker://quay.io/konflux-ci/tekton-catalog/task-clair-scan:0.2 | jq .created
"2024-11-19T14:34:42.596868488Z"
```

Not sure if the git commit date is that useful. Which use case do you have in mind?

tnevrlka: Seems to me like the [reproducible-builds](https://reproducible-builds.org/docs/source-date-epoch/) and the spec you linked are saying different things

Taken from the Scope part of [reproducible-builds](https://reproducible-builds.org/docs/source-date-epoch/) (can't send a link to the exact chapter)
```
Regardless to whether the SOURCE_DATE_EPOCH argument value is captured into Dockerfile with ARG SOURCE_DATE_EPOCH, the argument value is also used for:

    the created timestamp in the [OCI Image Config](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)
 ```

lcarva: Given that the reproducible-builds docs links to the spec I linked above, I'm gonna go with the spec on this one.

ralphbean:
Thanks for the review guys!

hack/build-and-push.sh
ralphbean: TIL about `skopeo inspect --config`. Knowing that I could have solved the problem that caused me to submit this PR.

I'm going to drop this.

PR #1687, opened 2024-12-04
oci-copy produced SBOM has a better name

kosciCZ:
This enhances the name for the oci-copy produced SBOM, to be `registry/repository@digest `

JIRA: [ISV-5323](https://issues.redhat.com//browse/ISV-5323)


ralphbean:
Awesome - thanks @kosciCZ ! Sorry it sat so long.

PR #1688, opened 2024-12-04
feat(KONFLUX-5480): Convert FBC pipeline to use trusted artifacts

arewm:
This also changes the FBC pipeline to be multi-arch enabled by default.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -36,6 +36,27 @@ spec:
             value: $(params.revision)
           - name: depth
             value: "0"
+      - name: build-appstudio-utils
+        runAfter:
+          - fetch-repository
+        params:
+          - name: IMAGE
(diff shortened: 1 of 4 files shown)


chmeliik (commented):
Why does FBC need to be multi-arch? Isn't it basically just a container of text files?

arewm:
The image contains an OPM binary which is used to serve the graph metadata for OLM. The indexes will be made multi-arch upon release as IIB requests do the conversion. If a user wants to test the FBC fragments before release, however, they will need to build the component as multi-arch to enable testing on all platforms.

This change still builds x86 images by default but lets users optionally add platforms. (this change also will have a side effect of adding the in-cluster remote task to e2e tests).

chmeliik (approved):
Not sure if the FBC pipeline needs prefetch-dependencies, otherwise LGTM

PR #1704, opened 2024-12-06
Remove OWNERS files

chmeliik:
We've switched to CODEOWNERS


chmeliik:
oops, let me split to more PRs by CODEOWNERS

PR #1707, opened 2024-12-06
fix: update maven-zip-build pipelines with recent template-build change

ligangty:
The template-build has added several sast tasks, which breaks the maven-zip-build pipelines. This PR has fixed the pipelines  accordingly. It kept the following tasks which are added recently in template-build:

* sast-coverity-check
* coverity-availability-check
* sast-shell-check
* sast-unicode-check


chmeliik (approved):
Btw, if you want to add yourself and @yma96 as codeowners of `/pipelines/maven-zip-build*` I wouldn't mind

PR #1709, opened 2024-12-06
Remove build-vm-image tasks' OWNERS files

chmeliik:
We use CODEOWNERS instead now.


arewm (approved):
LGTM. I see https://github.com/konflux-ci/build-definitions/blob/main/CODEOWNERS#L94 instead.

PR #1712, opened 2024-12-06
Remove integration tasks' OWNERS files

chmeliik:
We use CODEOWNERS instead now.


dirgim (approved):
LGTM - since the integration team's content in the new [CODEOWNERS](https://github.com/konflux-ci/build-definitions/blob/main/CODEOWNERS) file looks good, I'm fine with removing the old OWNERS files.

PR #1737, opened 2024-12-09
fix(KONFLUX-6028): Stop fetching tags with skopeo inspect

arewm:
By default, skopeo inspect will try to fetch all of the tags for a remote repository. This can take a lot of time if there are many present. Therefore, unless we need tags, we should use one of

* `skopeo inspect --raw`
* `skopeo inspect --no-tags`

as both of these will not try to list all tags.


arewm:
I suspect that the e2e tests will time out due to cluster slowness this morning.

rcerven (approved):

PR #1739, opened 2024-12-09
Set useful SBOM name for rpm-ostree generated images

johnbe11:
Set SBOM name for rpm-ostree generated images to `registry/repository@digest`of the pushed image.

JIRA: [ISV-5322](https://issues.redhat.com//browse/ISV-5322)

### DIFF
--- a/task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
+++ b/task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
@@ -234,7 +234,7 @@ spec:
         - mountPath: /var/lib/containers
           name: varlibcontainers
       script: |
-        syft oci-dir:/var/lib/containers/rhtap-final-image --output cyclonedx-json=/var/workdir/sbom-cyclonedx.json
+        syft oci-dir:/var/lib/containers/rhtap-final-image --output cyclonedx-json="/var/workdir/sbom-cyclonedx.json" --source-name "${IMAGE%:*}@$(cat "$(results.IMAGE_DIGEST.path)")"
       computeResources:
         limits:
           memory: 6Gi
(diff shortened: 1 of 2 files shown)


cgwalters:
CI failure looks unrelated (looks like this repo could use https://github.com/redhat-plumbers-in-action/differential-shellcheck or honestly just less shell script)

I see the task is cloned for a lot of repos, typically changes like this should mirror the buildah task, was that already done?

johnbe11:
@cgwalters I've updated and fixed the previous issue, PTAL

task/rpm-ostree-oci-ta/0.2/rpm-ostree-oci-ta.yaml
```diff
@@ -234,7 +234,7 @@ spec:
         - mountPath: /var/lib/containers
           name: varlibcontainers
       script: |
-        syft oci-dir:/var/lib/containers/rhtap-final-image --output cyclonedx-json=/var/workdir/sbom-cyclonedx.json
+        syft oci-dir:/var/lib/containers/rhtap-final-image --output cyclonedx-json="/var/workdir/sbom-cyclonedx.json" --source-name "${IMAGE%:*}@$(cat "$(results.IMAGE_DIGEST.path)")"
```
cgwalters, line 237: This `"${IMAGE%:*}@$(cat "$(results.IMAGE_DIGEST.path)")"` is changing the image to a digested form? It's surprising this isn't a property gathered earlier in the pipeline.
I had to go back to the bash reference manual to figure out that `%:*` is stripping the tag (I had thought `%:` was special syntax, but it's not).

cgwalters (approved):
Looks sane to me.
Worth noting here we're actively trying to deprecate (and eventually remove) this task, xref https://issues.redhat.com/browse/BIFROST-408

cgwalters:
On a procedural note, I doubt anyone else here cares but personally I would avoid merging from main into the branch for trivial patches like this, I'd rebase instead, it'd be a single commit with a single line change.

PR #1742, opened 2024-12-10
task/opm-render-bundles: add v to entry name

jbpratt:
the entry should have a v in the name according to the skiprange docs

https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange


gurnben (approved):
LGTM!

PR #1744, opened 2024-12-10
feat(KONFLUX-3935) Add pruning check to FBC pipeline

nmars:
Adds a new task to the fbc-builder pipeline to check if the incoming FBC fragment would remove channels or channel entries already present in the target index.

### DIFF
--- a/pipelines/fbc-builder/README.md
+++ b/pipelines/fbc-builder/README.md
@@ -85,6 +85,13 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE_URL| Fully qualified image name.| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
 |POLICY_DIR| Path to directory containing Conftest policies.| /project/repository/| |
 |POLICY_NAMESPACE| Namespace for Conftest policy.| required_checks| |
+### fbc-target-index-pruning-check:0.1 task parameters
+|name|description|default value|already set by|
+|---|---|---|---|
+|IMAGE_DIGEST| Image digest.| None| '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
+|IMAGE_URL| Fully qualified image name.| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
+|RENDERED_CATALOG_DIGEST| Digest for attached json file containing the FBC fragment's opm rendered catalog.| None| '$(tasks.validate-fbc.results.RENDERED_CATALOG_DIGEST)'|
+|TARGET_INDEX| Image name of target index, minus tag.| registry.redhat.io/redhat/redhat-operator-index| 'registry.redhat.io/redhat/redhat-operator-index'|
 ### git-clone-oci-ta:0.1 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
@@ -152,9 +159,9 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES| List of all referenced image manifests| |
-|IMAGE_DIGEST| Digest of the image just built| deprecated-base-image-check:0.4:IMAGE_DIGEST ; validate-fbc:0.1:IMAGE_DIGEST|
+|IMAGE_DIGEST| Digest of the image just built| deprecated-base-image-check:0.4:IMAGE_DIGEST ; validate-fbc:0.1:IMAGE_DIGEST ; fbc-target-index-pruning-check:0.1:IMAGE_DIGEST|
 |IMAGE_REF| Image reference of the built image containing both the repository and the digest| |
-|IMAGE_URL| Image repository and tag where the built image was pushed| show-sbom:0.1:IMAGE_URL ; deprecated-base-image-check:0.4:IMAGE_URL ; apply-tags:0.1:IMAGE ; validate-fbc:0.1:IMAGE_URL|
+|IMAGE_URL| Image repository and tag where the built image was pushed| show-sbom:0.1:IMAGE_URL ; deprecated-base-image-check:0.4:IMAGE_URL ; apply-tags:0.1:IMAGE ; validate-fbc:0.1:IMAGE_URL ; fbc-target-index-pruning-check:0.1:IMAGE_URL|
 |SBOM_BLOB_URL| Reference of SBOM blob digest to enable digest-based verification from provenance| |
 ### buildah-remote-oci-ta:0.3 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
@@ -164,6 +171,11 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE_URL| Image repository and tag where the built image was pushed| |
 |SBOM_BLOB_URL| Reference of SBOM blob digest to enable digest-based verification from provenance| |
 ### deprecated-image-check:0.4 task results
+|name|description|used in params (taskname:taskrefversion:taskparam)
+|---|---|---|
+|IMAGES_PROCESSED| Images processed in the task.| |
+|TEST_OUTPUT| Tekton task test output.| |
+### fbc-target-index-pruning-check:0.1 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES_PROCESSED| Images processed in the task.| |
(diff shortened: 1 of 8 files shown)


MartinBasti (approved):
Approving CODEOWNERS change only

dirgim (dismissed):
LGTM code-wise but needs rebase and resolving the `CODEOWNERS` conflict

chmeliik (dismissed):
build-owned parts LGTM

dirgim (dismissed):
LGTM code-wise

hongweiliu17 (approved):
lgtm

arewm (changes requested):
Is there specific benefit to creating an additional task for this check instead of just including it in the current `validate-fbc` task? 

By creating multiple tasks, we have to perform similar actions twice like fetching/rendering the FBC catalog. I had talked with @yashvardhannanavati about how it would be nice if the functions in the konflux-test image could reuse a common shared cache of the catalog. Separating the FBC verification checks into multiple tasks would invalidate benefits achieved with this approach.

nmars:
> Is there specific benefit to creating an additional task for this check instead of just including it in the current `validate-fbc` task?
>
> By creating multiple tasks, we have to perform similar actions twice like fetching/rendering the FBC catalog. I had talked with @yashvardhannanavati about how it would be nice if the functions in the konflux-test image could reuse a common shared cache of the catalog. Separating the FBC verification checks into multiple tasks would invalidate benefits achieved with this approach.

@arewm The benefit of having a separate check is because we want to be able to configure it for one-off exceptions. This is the proposed way to deal with the rare instances where a team needs to remove a version or channel from the target index. But they would want to ensure the other checks in `validate-fbc` would pass.

pipelines/fbc-builder/README.md
```diff
@@ -189,8 +201,10 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES_PROCESSED| Images processed in the task.| |
+|OCP_VERSION| OCP version derived from base image.| fbc-target-index-pruning-check:0.1:OCP_VERSION|
 |RELATED_IMAGES_DIGEST| Digest for attached json file containing related images| |
 |RELATED_IMAGE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the related images for the FBC fragment.| |
```
nmars, line 206: @arewm How does `RELATED_IMAGE_ARTIFACT` get created?

nmars: Latest change: the validate-fbc task was modified to create a trusted artifact from the `opm render` output. It will be pulled and used in the new check in lieu of running `opm render` again.

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
```diff
@@ -0,0 +1,230 @@
...
+      description: Image digest.
+    - name: TARGET_INDEX
+      description: Image name of target index, minus tag.
+      default: registry.redhat.io/redhat/redhat-operator-index
+    - name: OCP_VERSION
+      description: OCP version.
+    - name: RENDERED_CATALOG_DIGEST
+      description: Digest for attached json file containing the FBC fragment's opm rendered catalog.
+  results:
+    - name: TEST_OUTPUT
```
yashvardhannanavati, line 26: should IMAGES_PROCESSED also be a part of the result?

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
```diff
@@ -0,0 +1,230 @@
...
+    Ensures file-based catalog (FBC) components do not remove versions of operators already added to a released catalog.
+  params:
+    - name: IMAGE_URL
+      description: Fully qualified image name.
+    - name: IMAGE_DIGEST
+      description: Image digest.
+    - name: TARGET_INDEX
+      description: Image name of target index, minus tag.
+      default: registry.redhat.io/redhat/redhat-operator-index
+    - name: OCP_VERSION
```
yashvardhannanavati, line 21: I assume this task will be run after `build-image-index` in the pipeline. May I ask from where are we getting the OCP_VERSION param? I thought you will have to extract it from the fragment inside the task. Am I missing something?

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
```diff
@@ -0,0 +1,230 @@
...
+        fi
+
+        failure_num=0
+        TESTPASSED=true
+
+        fbc_channels=/tmp/olm-channels-fbc-image.json
+        ndx_channels=/tmp/olm-channels-target-index.json
+
+        ### Filter out channels and channel entries from FBC fragment render
+        jq -s 'map(select(.schema == "olm.channel")) | reduce .[] as $obj ([]; . += [{package: $obj.package, channel: $obj.name, entries: [$obj.entries[].name]}])' "${rendered_fbc_image}" > "${fbc_channels}"
```
yashvardhannanavati, line 146: this could be a good addition to konflux-test image as a function. I am ok if you want to make this change in a follow-up PR

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+        echo ""
+        echo "Channels defined in FBC fragment:"
+        jq '.' "${fbc_channels}"
+        echo ""
+
+        ### Filter out channels and channel entries from target index render
+        jq -s 'map(select(.schema == "olm.channel")) | reduce .[] as $obj ([]; . += [{package: $obj.package, channel: $obj.name, entries: [$obj.entries[].name]}])' "${rendered_target_index}" > "${ndx_channels}"
```
yashvardhannanavati, line 154: Line 146 is duplicated here. Another reason to convert this into a function instead.

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
arewm: I had previously suggested to not redo logic that we were doing in the `validate-fbc` task (including the catalog rendering). It might be intended for this result to come from _that_ task. 

After the work that I started doing in https://github.com/konflux-ci/konflux-test/pull/356, my previous comments may have been over-aggressive as the primary motivation was to not duplicate scripts. We can achieve this same result of reducing duplication by instead leveraging the utility scripts.

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
arewm: (and part of the script, of course)

nmars: I don't think that applies in this check, which compares channel entries between the FBC image and the latest target index.

task/fbc-target-index-pruning-check/0.1/fbc-target-index-pruning-check.yaml
yashvardhannanavati: `validate-fbc` isn't emitting OCP_VERSION either : https://github.com/konflux-ci/build-definitions/blob/ec9936c4226e3c6e57f57a1c0e5368d4b433c925/task/validate-fbc/0.1/validate-fbc.yaml#L26 

Also, OCP_VERSION seems to be a required param. @nmars Are we confident this will work in the pipeline?

yashvardhannanavati: Ah, I see you're adding it. Please ignore my previous comment

task/validate-fbc/0.1/README.md
```diff
@@ -16,4 +16,5 @@ Ensures file-based catalog (FBC) components are uniquely linted for proper const
 |TEST_OUTPUT|Tekton task test output.|
 |RELATED_IMAGES_DIGEST|Digest for attached json file containing related images|
 |IMAGES_PROCESSED|Images processed in the task.|
+|OCP_VERSION|OCP version of FBC image.|
```
yashvardhannanavati, line 19: Should `RENDERED_CATALOG_DIGEST ` be added here too?

Also, could you have the changes to validate-fbc in a separate commit?

An afterthought. We should be using konflux-test helper functions to get OCP_VERSION and rendered catalog. Tekton results have a size limit and this approach wouldn't be scalable.

yashvardhannanavati (approved):
one non-blocking comment. looks good

dirgim (approved):
LGTM

arewm (approved):
no more changes requested.

dirgim (approved):
/ok-to-test

chmeliik (approved):

MartinBasti (approved):

PR #1745, opened 2024-12-11
Attach migration file to task bundle

tkdchen:
[STONEBLD-3026](https://issues.redhat.com//browse/STONEBLD-3026)

* Introduce concepts of normal task (e.g. buildah) and extended task
  (e.g. buildah-20gb, buildah-min).
* Refactor the code that processes task build and push.
* For testing purpose, add a simple local cache in order to reduce the
  number of skopeo-inpsect runs for checking existence of task bundles.
  This could also be helpful to reduce the whole script execution during
  testing.
* For testing purpose, script can be configured with environment
  variable TEST_TASKS to run and check the result for a small number of
  tasks.

To test the attach:

* Open a new branch.
* Bump version of a task in the YAML definition, e.g. task/buildah/0.2
* Create task/buildah/0.2/migrations/<the new version>.sh
* Commit the changes
* Run hack/build-and-push.sh by configuring it to push images to your
  own test image repository.
* Find out the buildah task bundle.
* Run `oras discover \<bundle reference>'. The attached migration file
  should be listed.


chmeliik (commented):
:heart: amazing commit message and "doc" in the script itself, thanks!

(will review soon, just wanted to highlight this first)

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+
+# Build and push a task as a Tekton bundle, that is tagged a floating tag and a fixed tag.
+# The task bundle reference with digest is output to stdout in the last line,
+# that is extracted from tkn-bundle-push.
+build_push_task() {
+    local -r task_dir=$1
+    local -r prepared_task_file=$2
+    local -r task_bundle=$3
+    local -r task_file_sha=$4
+    # local -r extra_annotations=${5:-""}
```
chmeliik, line 230: If we don't need the `extra_annotations`, you can drop this commented code :slightly_smiling_face:

hack/build-and-push.sh
```diff
@@ -1,10 +1,56 @@
...
 
+# There are two major types of tasks in build-definitions:
+#
+# - Normal tasks, which are written as Tekton Task resource.
+#
+# - kustomized tasks, which are customized by kustomization based on normal
+#   tasks. kustomized tasks can be customized either based on another normal
+#   task, e.g. task buildah-24gb is based on task buildah. This type of
+#   kustomized task inherits the interface without change. Or based on itself,
+#   e.g. task inspect image.
```
chmeliik, line 11: "Based on itself" means there's a `$task_dir/$task_name.yaml` file _as well as_ a `$task_dir/kustomization.yaml`?

Technically, the code treats those as normal tasks I think

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+    CACHE_FILE="/tmp/build-definitions-build-and-push.cache"
+    declare -r CACHE_FILE
+fi
+
+cache_get() {
+    local -r key=$1
+    local value=
+    if [ -n "$ENABLE_CACHE" ]; then
+        # No match should not fail the script. It means no cached data.
+        read -r _ value < <(grep "^${key}" "$CACHE_FILE" 2>/dev/null) || :
```
chmeliik, line 348: Would this work? The `grep` seems potentially unreliable
```bash
value=$(awk -v key="$key" '$1 == key { print $2 }' < "$CACHE_FILE")
```

Or, to find the last occurrence of the key (since `cache_set` just appends keys):

```bash
value=$(awk -v key="$key" '$1 == key { value = $2 } END { print value }' < "$CACHE_FILE")

```

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+}
+
+# Attach migration file to given task bundle.
+# Arguments: task_dir, concrete_task_version, task_bundle
+attach_migration_file() {
+    local -r task_dir=$1
+    local -r concrete_task_version=$2
+    local -r task_bundle=$3
+
+    local -r task_file="${task_dir}/${task_name}.yaml"
```
chmeliik, line 383: What is `${task_name}` here? It doesn't seem to be one of the arguments :thinking:

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
 
-        output=$(tkn_bundle_push "${ANNOTATION_FLAGS[@]}" -f "$prepared_task_file" "$task_bundle" | save_ref "$task_bundle" "$OUTPUT_TASK_BUNDLE_LIST")
-        echo "$output"
-        task_bundle_with_digest="${output##*$'\n'}"
+    (
+        cd "${migration_file%/*}"
+        retry oras attach \
+            --registry-config "$AUTH_JSON" \
+            --artifact-type "$ARTIFACT_TYPE_TEXT_XSHELLSCRIPT" \
+            --distribution-spec v1.1-referrers-tag \
```
chmeliik, line 436: Do we need to specifically use the referrers-tag rather than the referrers-api?

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+            echo "NOTE: not pushing task-$task_name:$task_version to $QUAY_NAMESPACE; the repo does not exist and $QUAY_NAMESPACE is deprecated"
+            continue
+        fi
+
+        if is_normal_task "$task_dir" "$task_name";  then
+            build_data=$(generate_normal_task_build_data "$task_dir" "$task_name" "$task_version")
+        elif is_kustomized_task "$task_dir" "$task_name";  then
+            build_data=$(generate_kustomized_task_build_data "$task_dir" "$task_name" "$task_version")
+        else
+            continue
```
chmeliik, line 472: We should generally never reach the `continue`, right? Maybe echo a warning if this happens? (It would likely mean we're accidentally skipping some task)

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+        ANNOTATIONS+=("org.opencontainers.image.description=${task_description}")
+    fi
+    if [ -f "${task_dir}/README.md" ]; then
+        ANNOTATIONS+=("org.opencontainers.image.documentation=${VCS_URL}/tree/${VCS_REF}/${task_dir}README.md")
+    fi
+    if [ -f "${task_dir}/TROUBLESHOOTING.md" ]; then
+        ANNOTATIONS+=("dev.tekton.docs.troubleshooting=${VCS_URL}/tree/${VCS_REF}/${task_dir}TROUBLESHOOTING.md")
+    fi
+    if [ -f "${task_dir}/USAGE.md" ]; then
+        ANNOTATIONS+=("dev.tekton.docs.usage=${VCS_URL}/tree/${VCS_REF}/${task_dir}USAGE.md")
```
chmeliik, line 250: `${task_dir}README.md` etc.: all of these look like they assume `${task_dir}` always ends with `/`. But I don't think it does

hack/build-and-push.sh
```diff
@@ -194,19 +194,20 @@ inject_bundle_ref_to_pipelines() {
...
+            {\"name\": \"bundle\", \"value\": \"${task_bundle_with_digest}\"},
+            {\"name\": \"kind\", \"value\": \"task\"}
+        ]
+    }"
+    local -r task_selector="select(.name == \"${task_name}\" and .version == \"${task_version}\")"
+    find "$GENERATED_PIPELINES_DIR" "$CORE_SERVICES_PIPELINES_DIR" -maxdepth 1 -type f -name '*.yaml' | \
+        while read -r pipeline_file; do
+            yq e "(.spec.tasks[].taskRef | ${task_selector}) |= ${bundle_ref}" -i "${pipeline_file}"
+            yq e "(.spec.finally[].taskRef | ${task_selector}) |= ${bundle_ref}" -i "${pipeline_file}"
+        done
```
chmeliik, line 210: Nice - I finally understand what this part of the code even does :taco:

chmeliik (commented):
Nice refactoring! Minor comments, mostly looks good

hack/build-and-push.sh
```diff
@@ -133,95 +180,324 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+
+    # Check if task bundle has an attached migration file.
+    local filename
+    local found=
+    local artifact_refs
+
+    # List attached artifacts, that have specific artifact type and annotation.
+    # Then, find out the migration artifact.
+    artifact_refs=$(
+        retry oras discover "$task_bundle" --artifact-type "$ARTIFACT_TYPE_TEXT_XSHELLSCRIPT" --format json | \
```
chmeliik, line 398: The `--format` flag is new in https://github.com/oras-project/oras/releases/tag/v1.2.0

It would be good to document this or even add an automated check that the version is new enough. And we need to make sure that the appstudio-utils image has a new enough oras installed.

chmeliik (commented):
> Create task/buildah/0.2/migrations/<the new version>.sh

Is this right? Should the `migrations/` directory be under the version-specific directory, or should it be just `task/buildah/migrations/`?

While trying it out, I noticed we lost a lot of output:

<details>

```bash
QUAY_NAMESPACE=acmiel-test/tekton-catalog BUILD_TAG=$(date --iso-8601=date) SKIP_BUILD=1 SKIP_INSTALL=1 OUTPUT_TASK_BUNDLE_LIST=/tmp/task-bundle-list-konflux-ci OUTPUT_PIPELINE_BUNDLE_LIST=pipeline-bundle-list-konflux-ci hack/build-and-push.sh
```

```
✓ Pulled      0.2.2.sh                                                                                                                              71/71  B 100.00%  527µs
  └─ sha256:29f475854dcabc7419fd2d4b2da77b6c2f1fe98e78da8dab88f561b3fda90c1b
✓ Pulled      application/vnd.oci.image.manifest.v1+json                                                                                          783/783  B 100.00%     0s
  └─ sha256:8fcab734213322ad000455ebaf4384fb25af98523e5bc9643cdc5a20dddb2c56
```

</details>

The script no longers says what tasks it's pushing to quay.io

hack/build-and-push.sh
```diff
@@ -133,95 +180,325 @@ if [ "$SKIP_BUILD" == "" ]; then
...
+    local prepared_task_file
+    local task_file_sha
+    local task_bundle
+
+    find task/*/* -maxdepth 0 -type d | awk -F '/' '{ print $0, $2, $3 }' | \
+    while read -r task_dir task_name task_version
     do
-        yq e "$sub_expr_1" -i "${filename}"
-        yq e "$sub_expr_2" -i "${filename}"
+        if [ -n "$TEST_TASKS" ] && echo "$TEST_TASKS" | grep -v "$task_name" 2>/dev/null; then
```
chmeliik, line 458:
```suggestion
        if [ -n "$TEST_TASKS" ] && echo "$TEST_TASKS" | grep -qv "$task_name" 2>/dev/null; then
```
When I set `TEST_TASKS=buildah`, the script was just printing `buildah` over and over again :smile:

hack/build-and-push.sh
tkdchen: No. As the example task `inspect-image` shows, version 0.2 is based on 0.1.

hack/build-and-push.sh
tkdchen: :thinking:  Is referrers-api preferred?

hack/build-and-push.sh
chmeliik: Ooh, interesting. makes sense.

hack/build-and-push.sh
chmeliik: I would say so - if I understand correctly, referrers-tag should be the fallback for registries that don't support referrers-api

Can we leave it unspecified and let oras decide which distribution spec to use?

tkdchen: Can `oras attach` determine that by itself?

`--distribution-spec` is specified in some tasks explicitly.

tkdchen:
@chmeliik Comments are addressed. PTAL.

hack/build-and-push.sh
chmeliik: According to c746a251af63d1b91bf79b8f759737aaba0227ab, it can

chmeliik:
Nice, output is much more readable now. Just not sure about this part:

>  Create task/buildah/0.2/migrations/<the new version>.sh

Is this right? Should the `migrations/` directory be under the version-specific directory, or should it be just `task/buildah/migrations/`?

chmeliik:
> Nice, output is much more readable now. Just not sure about this part:
> 
> > Create task/buildah/0.2/migrations/.sh
> 
> Is this right? Should the `migrations/` directory be under the version-specific directory, or should it be just `task/buildah/migrations/`?

(it would make sense to me under the version-specific directory as well, just checking if that's the intention)

tkdchen:
> Is this right? Should the migrations/ directory be under the version-specific directory, or should it be just task/buildah/migrations/?

@chmeliik Sorry. I forgot to reply to this topic. Yes, the former is intended. That means a version-specific task has its own migrations and not mixed with other versions'.

hack/build-and-push.sh
tkdchen: It's removed from the script.

chmeliik (approved):
LGTM :+1:

PR #1756, opened 2024-12-13
fix(prefetch): fail step when error happen

MartinBasti:
Task were failing silently causing harder to debug errors or misleding users that different step failed. Ensure that steps aren't failing silently and we don't run unnecessary steps after failure.


task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -398,6 +411,9 @@ spec:
       image: quay.io/redhat-appstudio/cachi2:0.15.0@sha256:b141cb5cf4d98e6c5f668f1fe172e1d68f2a44ac1027403fbcff94ce1e68185d
       script: |
         #!/bin/bash
+
+        set -euo pipefail
```
chmeliik, line 415: Oof, I just saw the `unregister-rhsm` step at the end. This means the prefetch task will only unregister itself if it succeeds, otherwise it will leave a registration hanging around in the system.

But exiting on failure properly probably doesn't make that problem much worse than it already was

chmeliik (approved):

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
MartinBasti: questionable is, how much of hanging are we talking about in ephemeral tasks

MartinBasti: I can rework it to hard way using an internal state

MartinBasti: by hard way I mean to `trap` errror, record it and skip all tasks except unregister

MartinBasti:
e2e Fails at this
```
container step-unregister-rhsm: \ncp: cannot stat '/shared/rhsm/consumer/*': No such file or directory\n\
```

chmeliik:
@MartinBasti maybe give it a few /retest ? :smile: 

Users are sometimes confused by the failing-not-failing prefetch task

MartinBasti:
This task is being reworked, I'll just close this PR

MartinBasti:
https://github.com/konflux-ci/build-definitions/pull/1850

PR #1760, opened 2024-12-15
chore(KFLUXVNGD-140): renovate to ignore tekton-tools tasks

yftacherzog:
The tasks that are maintained in tekton-tools are tracked by mintmaker there and have additional specific test cases there.

We should let mintmaker update them rather than have renovate update them here. The updated tasks are then automatically mirrored to this repo.


chmeliik (approved):
Makes sense :+1:

PR #1762, opened 2024-12-16
fix: typo on sast-sny-check-oci-ta

avisiedo:
If statement to select the media type had a typo, so the condition would be always false. This change would fix potential wrong media types.


zregvart:
Thanks, this change needs to go to, ideally first, to https://github.com/konflux-ci/build-definitions/blob/aa56da2399f37c4466a882e02f1bf566dd567631/task/sast-snyk-check/0.3/sast-snyk-check.yaml#L251 as well. Then run `hack/generate-ta-tasks.sh` to generate this file as it is auto-generated from it.

zregvart (approved):

PR #1763, opened 2024-12-17
feat(KONFLUX-5670): Define computeResources for prefetch-dependencies and create-trusted-artifact steps

jhutar:
OK, I was trying to reproduce the issue [KONFLUX-5951](https://issues.redhat.com//browse/KONFLUX-5951) and was able to do so with Gomod Cachi2 configured. I do not know what caused that issue, but noticed that for me all of 50 concurrent prefetch-dependencies pods were running on the same node!

I have picked 1 CPU and 3 GiB of memory as a `request` (and in an effort to slowly migrate to model where we have requests == limits), used the same for `limits`. This is why I used these numbers (graphs from _stone-prd-rh01_):

image
... looks like steps `step-prefetch-dependencies` and `step-create-trusted-artifact` can take about 2 GB (including cache and so)

image
... RSS memory is of-course smaller

image
... regarding CPU it takes about 1 CPU across all containers

image
... again, only containers `step-prefetch-dependencies` and `step-create-trusted-artifact` seems to be bigger players here

### DIFF
--- a/task-generator/trusted-artifacts/ta.go
+++ b/task-generator/trusted-artifacts/ta.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	pipeline "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	core "k8s.io/api/core/v1"
+	resource "k8s.io/apimachinery/pkg/api/resource"
 )
 
 var (
@@ -334,6 +335,16 @@ func perform(task *pipeline.Task, recipe *Recipe) error {
 				},
 			},
 			Args: args,
+			ComputeResources: core.ResourceRequirements{
+				Requests: core.ResourceList{
+					core.ResourceCPU:    resource.MustParse("1"),
+					core.ResourceMemory: resource.MustParse("3Gi"),
+				},
+				Limits: core.ResourceList{
+					core.ResourceCPU:    resource.MustParse("1"),
+					core.ResourceMemory: resource.MustParse("3Gi"),
+				},
+			},
 		}
 
 		if task.Spec.StepTemplate == nil && !recipe.PreferStepTemplate {
--- a/task-generator/trusted-artifacts/go.mod
+++ b/task-generator/trusted-artifacts/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/tektoncd/pipeline v0.66.0
 	github.com/zregvart/tkn-fmt v0.0.0-20240614122620-a2995427266c
 	k8s.io/api v0.30.1
+	k8s.io/apimachinery v0.30.1
 	mvdan.cc/sh/v3 v3.10.0
 	sigs.k8s.io/kustomize/api v0.18.0
 	sigs.k8s.io/kustomize/kyaml v0.18.1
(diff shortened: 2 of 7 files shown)


hugares:
the failing check is saying you need to generate some files:
```
Error: File is out of date, run `hack/generate-ta-tasks.sh` and include the updated file with your changes
```

hugares:
> the failing check is saying you need to generate some files:
> 
> ```
> Error: File is out of date, run `hack/generate-ta-tasks.sh` and include the updated file with your changes
> ```

I think this is due to the fact that -ta version does not have same values as non -ta one

jhutar:
For record, this is a failure rate of `prefetch-dependencies.*` tasks in `stone-prd-rh01` in last week and day:
image
image
And average over time:
image
image

jhutar:
> Hello @zregvart. Is this correct approach please?

OK, ignore me Zoran. Looks like yaml in `task-generator/trusted-artifacts/golden/prefetch-dependencies/ta.yaml` is only used in tests, so changing it will not help with generating yamls, but will help with `go test` run in CI checks.

hugares (approved):
/lgtm

chmeliik (commented):
Is this based based on tests with one specific repo, or do you have data across varied uses of git-clone and prefetch-dependencies?

The memory and cpu usage will vary wildly depending on the size of the repo, the configured package managers, the number of dependencies etc.

task-generator/trusted-artifacts/ta.go
```diff
@@ -334,6 +335,16 @@ func perform(task *pipeline.Task, recipe *Recipe) error {
...
+			ComputeResources: core.ResourceRequirements{
+				Requests: core.ResourceList{
+					core.ResourceCPU:    resource.MustParse("1"),
+					core.ResourceMemory: resource.MustParse("3Gi"),
+				},
+				Limits: core.ResourceList{
+					core.ResourceCPU:    resource.MustParse("1"),
+					core.ResourceMemory: resource.MustParse("3Gi"),
+				},
+			},
```
zregvart, line 347: Ideally the values here would be part of the recipe, nothing that would block this change, this can be added if/when there is a need to modify this in the future

zregvart (approved):

jhutar:
Hello @chmeliik ! Graphs come from real world builds on `stone-prd-rh01`.

If the concern is limits are too small, I think they should be OK until some new huge repo appears (I have pasted a graph for whole week of data).

If the concern is the requests (and limits) are too high for small repos without specific requirements, I assume that should be fine, because in that case the pod/task will be running just briefly, so resources will not be allocated for too long.

task-generator/trusted-artifacts/ta.go
jhutar: I agree (as far as I understand the architecture here) but that would require bigger code change and I did not wanted to go too deep to it for now. But, let me know and I will happily work on it in different PR.

chmeliik:
> Hello @chmeliik ! Graphs come from real world builds on `stone-prd-rh01`.

Ack :+1: 

> If the concern is limits are too small, I think they should be OK until some new huge repo appears (I have pasted a graph for whole week of data).

Yeah this was it. Ok, let's see if the current requests and limits are enough

chmeliik (commented):
LGTM, could you just update the commit message or split into two commits? This PR updates the computeResources for every `create-trusted-artifact` step and for one separate step in the prefetch-dependencies task

jhutar:
Hello @chmeliik ! Updated PR title and commit message. Hope it looks OK. Also happy new year :)

chmeliik (approved):

PR #1772, opened 2024-12-20
Validate added migration file

tkdchen:
[STONEBLD-2827](https://issues.redhat.com//browse/STONEBLD-2827)

### DIFF
--- a/hack/validate-migration.sh
+++ b/hack/validate-migration.sh
@@ -1,0 +1,356 @@
+#!/usr/bin/env bash
+
+# Validate migration file introduced by a branch.
+#
+# This script can be run in the CI against a PR or in local from a topic branch.
+# Before run, all local changes have to be committed.
+#
+# Network is required to execute this script.
+#
+# Checks are implemented as functions whose name has prefix `check_`. Each of
+# them exits with status code 0 to indicate pass, otherwise exits
+# script execution immediately with non-zero code.
+#
+# Test with a local cluster:
+#
+# - Before starting test with local cluster, launch one by `kind create
+#   cluster'.  Then, install tekton pipelines
+#   https://tekton.dev/docs/pipelines/install/#installation
+#
+# - Set IN_CLUSTER to true.
+
+set -euo pipefail
+
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+cd "$SCRIPTDIR/.." || exit 1
+unset SCRIPTDIR
+
+declare -r BUILD_PIPELINE_CONFIG=https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/refs/heads/main/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
+declare -r LABEL_TASK_VERSION=app.kubernetes.io/version
+
+WORK_DIR=$(mktemp -d --suffix=-validate-migrations)
+declare -r WORK_DIR
+
+: "${DEFAULT_BRANCH:=main}"
+declare -r DEFAULT_BRANCH
+
+# By default, do not run with a real kubernetes cluster.
+: "${IN_CLUSTER:=""}"
+declare -r IN_CLUSTER
+
+prepare_pipelines() {
+    local -a pl_names_in_config
+    local pushed_pipelines
+
+    mkdir -p "${WORK_DIR}/pipelines/pushed"
+
+    # Download pushed pipeline bundles from ConfigMap
+    pushed_pipelines=$(
+        curl --fail -sL "$BUILD_PIPELINE_CONFIG" | yq '.data."config.yaml"' | yq '.pipelines[] | .name + " " + .bundle'
+    )
+
+    while read -r pl_name pl_bundle; do
+        pl_names_in_config+=("$pl_name")
+        tkn bundle list "$pl_bundle" pipeline "$pl_name" -o yaml \
+            >"${WORK_DIR}/pipelines/pushed/${pl_name}.yaml"
+    done <<<"$pushed_pipelines"
+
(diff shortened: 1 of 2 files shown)


hack/validate-migration.sh
```diff
@@ -0,0 +1,332 @@
...
+#
+# - Set IN_CLUSTER to true.
+
+set -euo pipefail
+
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+cd "$SCRIPTDIR/.." || exit 1
+unset SCRIPTDIR
+
+declare -r BUILD_PIPELINE_CONFIG=https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/refs/heads/main/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
```
chmeliik, line 28: In the commit message, could you describe more about what the check does and why? In particular, I don't quite understand why we need the pipeline bundles from redhat-appstudio/infra-deployments

hack/validate-migration.sh
```diff
@@ -0,0 +1,332 @@
...
+        while read -r status origin_path; do
+            if ! is_migration_file "$origin_path"; then
+                continue
+            fi
+            case "$status" in
+                A)  # file is added
+                    echo "$origin_path"
+                    ;;
+                D | M)
+                    echo "It is not allowed to delete or modify existing migration file: $origin_path" >&2
```
chmeliik, line 254: It would be nice to suggest what to do instead. I think the solution would to be bump the patch version and add a new migration that fixes the broken migration?

hack/validate-migration.sh
```diff
@@ -0,0 +1,332 @@
...
+
+    if [[ ${#migrations_files[@]} -eq 0 ]]; then
+        echo "No migration."
+        exit 0
+    fi
+
+    local -r n=${#migrations_files[@]}
+    if [[ $n -gt 1 ]]; then
+        echo "error: found $n migration files. Please ensure to include a single migration file per time." >&2
+        exit 1
```
chmeliik, line 319: I think it will often be necessary to include multiple migration files in a single PR. Especially for the buildah tasks, where you have to update 4+ tasks at the same time (buildah, buildah-oci-ta, buildah-remote, buildah-remote-oci-ta, buildah-*gb)

hack/validate-migration.sh
```diff
@@ -0,0 +1,332 @@
...
+    local -r migration_file=$1
+    local -r dir_path=${migration_file%/*}  # remove file name
+    echo "${dir_path%/*}"  # remove path component migrations/
+}
+
+check_migrations_is_in_task_version_specific_dir() {
+    local -r migration_file=$1
+    local -r parent_dir=$(resolve_migrations_parent_dir "$migration_file")
+    local -r result=$(find task/*/* -type d -regex "$parent_dir" -print -quit)
+    if [ -z "$result" ]; then
```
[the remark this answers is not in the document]
tkdchen: > I'm not sure I fully get the `find` call. Would this be equivalent?
> 
> ```shell
> if [[ "$parent_dir" != task/*/* ]]; then
> ```

What kind of match of this expression? This command outputs nothing:

```bash
[[ "task/summary/abc" != task/*/* ]] && echo yes
```

This `find` commands tries to find out a matched task version directory, e.g. `task/buildah/0.2`. The `-regex` does a full regex match.

hack/validate-migration.sh
tkdchen: Fixed in the new commit. PTAL.

hack/validate-migration.sh
```diff
@@ -249,25 +249,35 @@ is_migration_file() {
...
+    file_list=$(git diff --name-status "$(git merge-base HEAD "$DEFAULT_BRANCH")")
+    local seen=
+    while read -r status origin_path; do
+        if ! is_migration_file "$origin_path"; then
+            continue
+        fi
+        case "$status" in
+            A)  # file is added
+                task_dir_path=$(awk -F '/' '{ OFS = "/"; print $1, $2 }' <<<"$origin_path")
+                if grep -q "$task_dir_path" <<<"$seen"; then
```
chmeliik, line 262: I think this will cause problems

```bash
$ grep -q "task/buildah" <<< "task/buildah-remote task/some-other-task" && echo already seen                                                       
already seen
```

Maybe change the `seen` variable to be multiline and check like this?

```bash
grep -q "^${task_dir_path}$" <<< "$seen"
```

hack/validate-migration.sh
```diff
@@ -314,28 +324,23 @@ main() {
...
-    check_migrations_is_in_task_version_specific_dir "$file_path"
-    check_version_match "$file_path"
-    check_apply_on_pipelines "$file_path"
-    check_apply_in_real_cluster
+    for migration_file in $output; do
+        check_pass_shellcheck "$migration_file"
+        check_migrations_is_in_task_version_specific_dir "$migration_file"
+        check_version_match "$migration_file"
+        check_apply_on_pipelines "$migration_file"
+        check_apply_in_real_cluster
```
chmeliik, line 341: Do these work correctly when run multiple times? Do we maybe need to clean something in the `WORK_DIR` before running them again?

hack/validate-migration.sh
tkdchen: Good catch. Updated.

hack/validate-migration.sh
tkdchen: Updated.

tkdchen:
Rebased this PR in order to base on the fix https://github.com/konflux-ci/build-definitions/pull/1788

chmeliik (commented):
LGTM, could you just squash the commits?

tkdchen:
Squash is done.

chmeliik (approved):

PR #1774, opened 2024-12-20
`--json` not supported in version of curl in use

zregvart:
The version of `curl` (7.76.1) we currently use in quay.io/konflux-ci/appstudio-utils does not support the `--json` parameter, so this reverts to using lengthier `--data-binary` and setting the Content-Type and Accept headers.


chmeliik (approved):
LGTM, good catch

PR #1781, opened 2024-12-31
Remove the GitHub Action to check the remote builder task length

arewm:
With the merging of
https://github.com/redhat-appstudio/infra-deployments/pull/5201 the deployment of Konflux now properly supports larger sized of tasks.

Even without this fix merged, the FBC pipeline is now based on the multi-platform pipeline and therefore will test the default pipeline which includes the buildah-remote task. We can fall back to the signal with that test to ensure that the task continues to work properly in the local cluster.


.github/workflows/check-buildah-remote.yaml
chmeliik, line 1: We still want this one. The one with the size check is `.github/workflows/temp-block-buildah.yaml`

arewm: Of course. Good catch.

chmeliik (approved):

PR #1784, opened 2025-01-03
Fix buildah remote

chmeliik:
The new 'icm' step was broken because it didn't take into account the
IMAGE modifications done in every other step. To fix it, switch the step
from "command-style" to "script-style" to make the task-generator tool
notice that it has to do its thing.

Also fix the task-generator because it wasn't handling things very well


tnevrlka (approved):
LGTM, nice commit message

arewm (approved):
Thanks for this fix!

PR #1785, opened 2025-01-03
fix(KONFLUX-6229) fix ICM injection for multiarch images

ralphbean:
This ICM injection step accepts the IMAGE param as the image to operate on, which needs to account for whether or not the IMAGE_APPEND_PLATFORM param is also set to true.

Incorporates: https://github.com/konflux-ci/build-tasks-dockerfiles/pull/214
Fixes: https://issues.redhat.com/browse/KONFLUX-6229


ralphbean:
Oh, his is better! Let's drop this.

PR #1796, opened 2025-01-08
chore: replace multi-platform-runner image with buildah-task

tisutisu:
Part of the story: https://issues.redhat.com/browse/STONEBLD-2716

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -113,7 +113,7 @@ spec:
         echo "declare TAGGED_AS=${TAGGED_AS}" >> /var/workdir/vars
 
     - name: build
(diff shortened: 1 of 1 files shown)


arewm (changes requested):
I had to revert this change the last time it was made because of a change in behavior of buildah between the versions: https://github.com/konflux-ci/build-definitions/pull/1348

tisutisu:
> I had to revert this change the last time it was made because of a change in behavior of buildah between the versions: #1348

@arewm IIRC `rpm-ostree` task change was causing the issue last time, so with this PR I have only updated the `build-vm-image` task. Do you see any problem replacing image for the `build-vm-image` task?

arewm (approved):
@tisutisu , thanks. I didn't realize that there were multiple instances that still needed to be changed. I don't see the same issue that we hit in the rpm-ostree task because we are just pushing the images directly from the remote location instead of trying to rsync the data back to the cluster and then push it. 

I suspect that this change should work fine.

PR #1801, opened 2025-01-08
Improve the FBC related image validation

arewm:
* Change the logic to only report unreleased related images
* Remove newlines from the related image result


task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -72,140 +79,133 @@ spec:
...
+        # is a no-op for now. We can come back later and enforce that the content is identical between
+        # all architectures if desired.
+        images_processed_template='{"image": {"pullspec": "'"$IMAGE_URL"'", "digests": [%s]}}'
+        if [ -n "$manifest_digests" ]; then
+          while read -r _arch arch_sha; do
+            digests_processed+=("\"$arch_sha\"")
+          done < <(echo "$manifest_digests" | jq -r 'to_entries[] | "\(.key) \(.value)"')
         fi
 
+        # If the image is an Image Index, also add the Image Index digest to the list.
```
yashvardhannanavati, line 121: Can I suggest moving lines 121 - 125 towards the end of the task? I think we should only emit IMAGES_PROCESSED result once we have successfully finished processing the image. Current approach isn't incorrect either. I just feel if the task fails on line 134, we didn't really process the image. I'll leave the final call on you.

P.S: For error/failure use cases, we can skip emitting the IMAGES_PROCESSED result. EC only verifies the presence of digests if the task actually emits the result.

yashvardhannanavati (commented):
Minor comments. Looks great!

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -72,140 +79,133 @@ spec:
...
+            echo "No amd64 image manifest found for ${image_with_digest}"
+            note="Step inspect-image failed: Encountered errors while inspecting image. For details, check Tekton task log."
+            TEST_OUTPUT=$(make_result_json -r ERROR -t "$note")
+            echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+            exit 0
+        fi
+        image_with_digest="$(get_image_registry_repository_digest "${IMAGE_URL}@${reference_manifest_digest}")"
+        echo -n "${image_with_digest}" > /shared/image_with_digest
+
+        # FBC content should be identical across different architectures so the IMAGES_PROCESSED result
```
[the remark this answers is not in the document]
arewm: This code is consistent with the prior behavior. We need an architecture matching the running architecture because we have to extract and run `opm` from its base image. We can potentially remove this caveat with a binaryless flow once that is supported. We can also add in a validation to ensure that the catalog is identical between architectures (maybe that would be a warning if it wasn't true).

Does IIB take the content from all architectures or does it only use one architecture when generating the resulting catalog images?

Regardless, I would vote for deferring this.

task/validate-fbc/0.1/validate-fbc.yaml
arewm: If you want to wait until we process the image then we should actually generate this result in the next step. Since we were already processing the manifests here, I thought that it was best to just do all of the manifest operations in one place (I didn't want to duplicate the manifest operations in two steps).

Since the IMAGES_PROCESSED result doesn't adversely affect policy evaluation, I thought it was better to just get the result generation over with (especially since we are not _really_ processing all of the digests).

task/validate-fbc/0.1/validate-fbc.yaml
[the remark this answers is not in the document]
yashvardhannanavati: IIB uses content only from amd64 arch because of the same reason you mentioned. Also, I believe sanity check of the content should be done by CVP or its equivalent which would be the validate-fbc task in Konflux because IIB is only invoked at release time.

But, I agree with you. Let's revisit this once we have the binaryless flow.

task/validate-fbc/0.1/validate-fbc.yaml
yashvardhannanavati: Well, you could just process everything here and store everything in a file. Then, at the end you could simply emit the result.

 But, like I said, the current approach is correct too. I am fine with keeping it as is.

yashvardhannanavati (dismissed):

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -216,18 +216,26 @@ spec:
             - SETFCAP
       computeResources:
         limits:
-          memory: 4Gi
+          memory: 8Gi
         requests:
-          memory: 512Mi
+          memory: 8Gi
```
arewm, line 221: This is a large memory increase, likely because of the use of `jq --slurp` in the konflux-test image. I think that we should look to use something other than `jq` to determine the unique related images between the fragment and the index. We already have the content in files: 
https://github.com/konflux-ci/konflux-test/blob/d28ed88a75b49132cf734ba3ed4e1013b8696492/test/utils.sh#L637-L640

So we should be able to do the set subtraction in bash by sorting the files and then using `comm`: https://stackoverflow.com/a/11964477

dirgim (approved):
LGTM, nice work

PR #1803, opened 2025-01-09
sast-snyk-check: Added functionatlity to ignore directories and folders

jperezdealgaba:
Resolves: https://issues.redhat.com/browse/OSH-795

Making use of the snyk ignore functionality, users are now able to specify the IGNORE parameter with a list of comma-separated directories and folders and they will be ignored from the scans.


task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -57,6 +57,10 @@ spec:
       type: string
       description: Write excluded records in file. Useful for auditing (defaults to false).
       default: "false"
+    - name: IGNORE
```
kdudka, line 60: I would name the parameter `IGNORE_FILE_PATHS` to make it more specific.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -135,13 +141,25 @@ spec:
...
         if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
           SEVERITY_THRESHOLD="low"
         fi
+
+        # We ignore files using snyk ignore if the user set up the IGNORE variable
+        if [[ -n "${IGNORE}" ]]; then
+          paths="(${IGNORE//,/ })" # Split by comma into an array
+          command=""
+          for path in "${paths[@]}"; do
+            command+="snyk ignore --file-path=source/$path && "
```
kdudka, line 156: What is the advantage of chaining the `snyk ignore` commands with `&&` in a string, which is expanded by `eval` later on?  Cannot we just run the `snyk ignore` commands directly?  Note that the script is executed with `set -e`, so the chaining with `&&` should not be needed.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -195,8 +213,8 @@ spec:
 
           # Generation of scan stats
 
-          total_files=$(jq '[.runs[0].properties.coverage[].files] | add' "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json)
-          supported_files=$(jq '[.runs[0].properties.coverage[] | select(.type == "SUPPORTED") | .files] | add' "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json)
+          total_files=$(jq '[.runs[0].properties.coverage[].files // 0] | add' "${SOURCE_CODE_DIR}"/sast_snyk_check_out.json)
```
kdudka, line 216: It is unclear to me how this case is specific to the ignored paths feature.  Cannot the same happen also when the files that we ignore simply do not exist (without specifying the parameter)?  How do these cases differ from each other?

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -135,13 +141,25 @@ spec:
           echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
           exit 0
         fi
-
```
kdudka, line 138: This line seems to be removed by mistake.

kdudka (changes requested):

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: There is no special advantage of doing that. I was just following Snyk docs: https://docs.snyk.io/snyk-cli/scan-and-maintain-projects-using-the-cli/snyk-cli-for-snyk-code/exclude-directories-and-files-from-snyk-code-cli-tests but we execute the commands one by one.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: While testing this, if the scan had results and they were ignored with the .snyk file policy, the resulted `null` value was never converted into `0` in the following line:
https://github.com/konflux-ci/build-definitions/blob/f1fa9480c40f63aae17974aa11d1b9d23c20f582/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml#L243C10-L243C40

With this change, the null value was converted into `0`

jperezdealgaba:
Moving to draft while testing new changes...

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: I would expect the ignored files to be excluded from scanning rather than being excluded from scan results.

jperezdealgaba: And they are, the files are ignored from the scan by looking at the .snyk file and ignoring them.
Will find the root cause of this

jperezdealgaba: I found the root cause of this. This code path was never reached because if no supported projects were found, hence the `total_files` was null, `snyk code` returned a different status code and it skipped this part: https://github.com/konflux-ci/build-definitions/blob/main/task/sast-snyk-check/0.3/sast-snyk-check.yaml#L220
Now, there are supported files but they are ignored and that's how this was found. 
`jq` returns a string set to "null" and that's why the default setting to 0 wasn't working as expected. I have modified the code to make it clearer. LMKWYT

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Modified to call the commands one by one

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: Done

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -142,6 +148,15 @@ spec:
         if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
           SEVERITY_THRESHOLD="low"
         fi
+
+        # We ignore files using snyk ignore if the user set up the IGNORE variable
+        if [[ -n "${IGNORE_FILE_PATHS}" ]]; then
+          paths=("${IGNORE_FILE_PATHS//,/ }") # Split by comma into an array
```
kdudka, line 154: Does this shell construction actually work?

I think you would need to use `eval` to make it work as expected and that would be problematic from security point of view.  Moreover, it would not work for file paths containing spaces anyway.

Cannot we simplify it like this?
```sh
(
  cd "${SOURCE_CODE_DIR}"
  IFS="," 
  for path in $IGNORE_FILE_PATHS; do
    snyk ignore --file-path="source/$path"
  done
)
```

Note that the `if` condition can be removed, too, if you use the code snippet above.

kdudka (changes requested):

task/sast-snyk-check/0.3/sast-snyk-check.yaml
kdudka: Makes sense.  Thanks for explanation!

task/sast-snyk-check/0.3/sast-snyk-check.yaml
jperezdealgaba: I tested it and it worked fine and findings were ignored (i.e. debug run: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-5bxx4), but yours construction is much more simpler, let me test it and I will integrate it

kdudka: I think that `paths=("${IGNORE_FILE_PATHS//,/ }")` would always create an array with exactly one item.

kdudka: > Update: The` /hack/generate-ta-tasks.sh` script doesn't like the subshell `( )` declaration in separate lines, so we will need to tweak this a bit.

They do not need to be on separate lines for the script to work.  It can be formatted e.g. like this:
```sh
(cd "${SOURCE_CODE_DIR}"
IFS="," 
for path in $IGNORE_FILE_PATHS; do
  snyk ignore --file-path="source/$path"
done)
```

jperezdealgaba: I have been trying different possibilities and having the parenthesis in different lines is incorrectly parsed by the script and the output is handled like this:

```
[...]
script: "#!/usr/bin/env bash\n\nset -euo pipefail\n# shellcheck source=/dev/null\n.
        /utils.sh\ntrap 'handle_error $(results.TEST_OUTPUT.path)' EXIT\n\nif
        [[ -z \"${PROJECT_NAME}\" ]]; then\n    PROJECT_NAME=${COMPONENT_LABEL}\nfi\n\necho
        \"The PROJECT_NAME used is: ${PROJECT_NAME}\"\n\n# Installation of
        Red Hat certificates for cloning Red Hat internal repositories\nca_bundle=/mnt/trusted-ca/ca-bundle.crt\nif
        [ -f \"$ca_bundle\" ]; then\n  echo \"INFO: Using mounted CA bundle:
        $ca_bundle\"\n  cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors\n
        \ update-ca-trust\nfi\n\nSNYK_TOKEN_PATH=\"/etc/secrets/snyk_token\"\nif
        [ -f \"${SNYK_TOKEN_PATH}\" ] && [ -s \"${SNYK_TOKEN_PATH}\" ]; then\n
        \ # SNYK token is provided\n  SNYK_TOKEN=\"$(cat ${SNYK_TOKEN_PATH})\"\n
        \ export SNYK_TOKEN\nelse\n  # According to shellcheck documentation,
        the following error can be ignored as it is ignored through indirection:

[...]
```
The cleanest way I found to make it was to store it in a variable:

https://github.com/konflux-ci/build-definitions/blob/d8e9c7d3ab93deeaaf29455c829f262e75b42b4f/task/sast-snyk-check/0.3/sast-snyk-check.yaml#L152-L161

I have also tested it and works as expected:

  - Variable not set: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-nrntb
  - Variable set: https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-d8bf6

kdudka: I can see it now.  The transformation script looks really broken.  Nevertheless, I do not like the artificial variable that would consume anything produced by `snyk ignore` on `stdout`.  What if we formatted the code like this?
```sh
        (cd "${SOURCE_CODE_DIR}" && IFS="," && for path in $IGNORE_FILE_PATHS; do
          snyk ignore --file-path="source/$path"
        done)
```
That does not trigger the bug in my local env.

jperezdealgaba: Sure. It also doesn't trigger in mine. I have also tested it and works as expected:
- https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/jperezde/applications/test-coverity/pipelineruns/osh-cli-container-konflux-test-2-on-pull-request-bkvbr
I have done needed changes

Let me know if we can resolve this comment.

kdudka: Yes, thanks for the update!

kdudka (approved):

dirgim (dismissed):
LGTM

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -142,6 +148,12 @@ spec:
         if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
           SEVERITY_THRESHOLD="low"
         fi
+
+        # We ignore files using snyk ignore if the user set up the IGNORE_FILE_PATHS variable.
+        (cd "${SOURCE_CODE_DIR}" && IFS="," && for path in $IGNORE_FILE_PATHS; do
+          snyk ignore --file-path="source/$path"
```
hongweiliu17, line 154:
```suggestion
          snyk ignore --file-path="source/${path}"
```

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -164,6 +171,12 @@ spec:
         if [ "${IMP_FINDINGS_ONLY}" == "false" ]; then
           SEVERITY_THRESHOLD="low"
         fi
+
+        # We ignore files using snyk ignore if the user set up the IGNORE_FILE_PATHS variable.
+        (cd "${SOURCE_CODE_DIR}" && IFS="," && for path in $IGNORE_FILE_PATHS; do
+          snyk ignore --file-path="source/$path"
```
hongweiliu17, line 177:
```suggestion
          snyk ignore --file-path="source/${path}"
```

hongweiliu17 (approved):
It is better to use `${path}`. Others look good to me

hongweiliu17 (approved):
LGTM

jsztuka (approved):
lgtm

jperezdealgaba:
@jsztuka Where did it fail? I am not able to see the logs

jsztuka:
E2e test part, I need to pull the log from cluster probably, looks like its no longer available.

arewm:
CI has passed upon being rerun.

dirgim (approved):
LGTM

PR #1813, opened 2025-01-13
improve buildah 0.3 migration files

tnevrlka:
- Some users have encountered problems when migrating buildah tasks from 0.2 to 0.3 when using older PipelineRun definitions which still include the results that are being removed in buildah 0.3
- Mention in the Migration files that users should check if their PipelineRun contains the removed results and that they should be removed


chmeliik (approved):

arewm:
Force merging the PR as it is just a documentation update. This will also help the acceptable bundles task be rebuilt to come back into sync.

PR #1814, opened 2025-01-13
chore(KFLUXVNGD-155): Add custom certificate support for git clone and buildah task

hmariset:
Add param to support the custom certificate support for git-clone-oci-ta and
buildah-oci-ta tasks to connect to internal registry. For this we have updated
the use-trusted-artifact/create-trusted-artifact steps, which download/upload
the artifact from/to a registry.

Jira-Url: https://issues.redhat.com/browse/KFLUXVNGD-155

### DIFF
--- a/task-generator/trusted-artifacts/README.md
+++ b/task-generator/trusted-artifacts/README.md
@@ -55,10 +55,12 @@ The following is the list of supported options:
 | `addResult`          | sequence of Tekton [TaskResult]s                 | Additional Tekton Task results to add to the Task |
 | `addVolume`          | sequence of [Volume]s                            | Additional Volumes to add to the Task |
 | `addVolumeMount`     | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Task |
+| `addTAVolumeMount`   | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the `use-trusted-artifact` and `create-trusted-artifact` steps in the Trusted Artifact Task |
 | `base`               | string                                           | Relative path from `recipe.yaml` to the Task definition of the non-Trusted Artifacts Task |
 | `description`        | string                                           | Description of the Trusted Artifacts Task |
 | `displaySuffix`      | string                                           | Additional text to place to the value of `tekton.dev/displayName` annotation from the non-Trusted Artifacts Task to the Trusted Artifacts Task (default: `" oci trusted artifacts"`) |
 | `preferStepTemplate` | boolean                                          | When `true` preference is set to configure common configuration on the `Task.spec.stepTemplate` rather than on each Task Step |
+| `useTAVolumeMount`   | boolean                                          | When `true` VolumeMounts are added to the `use-trusted-artifact` and `create-trusted-artifact` steps. If `addTAVolumeMount` is provided then those VolumeMounts are added. Otherwise, default VolumeMounts are added |
 | `regexReplacements`  | map of strings keys and string values            | Perform regular expression-based replacement with keys being the regular expression and the values being the replacement, see [Replacements](#replacements) |
 | `removeParams`       | sequence of strings                              | Names of Task parameters to remove |
 | `removeVolumes`      | sequence of strings                              | Names of Task Volumes to remove |
--- a/task-generator/trusted-artifacts/recipe.go
+++ b/task-generator/trusted-artifacts/recipe.go
@@ -24,10 +24,12 @@ type Recipe struct {
 	AddResult          []pipeline.TaskResult `json:"addResult"`
 	AddVolume          []core.Volume         `json:"addVolume"`
 	AddVolumeMount     []core.VolumeMount    `json:"addVolumeMount"`
+	AddTAVolumeMount   []core.VolumeMount    `json:"addTAVolumeMount"`
 	Base               string                `json:"base"`
 	Description        string                `json:"description"`
 	DisplaySuffix      string                `json:"displaySuffix"`
 	PreferStepTemplate bool                  `json:"preferStepTemplate"`
+	UseTAVolumeMount   bool                  `json:"useTAVolumeMount"`
 	RegexReplacements  map[string]string     `json:"regexReplacements"`
 	RemoveParams       []string              `json:"removeParams"`
 	RemoveVolumes      []string              `json:"removeVolumes"`
--- a/task-generator/trusted-artifacts/ta.go
+++ b/task-generator/trusted-artifacts/ta.go
@@ -159,8 +159,21 @@ func perform(task *pipeline.Task, recipe *Recipe) error {
 		Name:      "workdir",
 		MountPath: "/var/workdir",
 	}
+	trustedVolumeMount := core.VolumeMount{
+		Name:      "trusted-ca",
+		MountPath: "/etc/pki/tls/certs/ca-custom-bundle.crt",
+		SubPath:   "ca-bundle.crt",
+		ReadOnly:  true,
+	}
+
 	if len(recipe.AddVolumeMount) == 0 {
 		recipe.AddVolumeMount = []core.VolumeMount{workdirVolumeMount}
+	}
+	if len(recipe.AddTAVolumeMount) == 0 {
+		recipe.AddTAVolumeMount = []core.VolumeMount{trustedVolumeMount}
+	}
+	if !recipe.UseTAVolumeMount {
+		recipe.AddTAVolumeMount = []core.VolumeMount{}
 	}
 
 	removeEnv := func(env *[]string) func(core.EnvVar) bool {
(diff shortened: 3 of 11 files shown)


task-generator/trusted-artifacts/ta.go
```diff
@@ -159,6 +159,12 @@ func perform(task *pipeline.Task, recipe *Recipe) error {
 		Name:      "workdir",
 		MountPath: "/var/workdir",
 	}
+	trustedVolumeMount := core.VolumeMount{
+		Name:      "trusted-ca",
```
yftacherzog, line 163: I'm not sure we should hard code the name of the volume. Maybe we can have `trusted-ca` as the default, and include an optional recipe field for specifying a different volume name.

Same goes for the sub-path. The volume might have the certificate stored under a different path.

task-generator/trusted-artifacts/recipe.go
```diff
@@ -24,6 +24,7 @@ type Recipe struct {
 	AddResult          []pipeline.TaskResult `json:"addResult"`
 	AddVolume          []core.Volume         `json:"addVolume"`
 	AddVolumeMount     []core.VolumeMount    `json:"addVolumeMount"`
+	AddTAVolumeMount   []core.VolumeMount      `json:"addTAVolumeMount"`
```
yftacherzog, line 27: We should have it documented [here](https://github.com/konflux-ci/build-definitions/blob/main/task-generator/trusted-artifacts/README.md).

task-generator/trusted-artifacts/ta.go
```diff
@@ -308,6 +318,7 @@ func perform(task *pipeline.Task, recipe *Recipe) error {
 			Name:  "use-trusted-artifact",
 			Image: image,
 			Args:  args,
+			VolumeMounts: recipe.AddTAVolumeMount,
```
yftacherzog, line 321: I have some reservations regarding adding the volume mount to all tasks by default.
This is mainly because it means we're modifying all tasks at once.

I'd prefer if there's a way to make adding the mounts optional. E.g. add a boolean field for that and have it disabled by default and then we enable it for each task we need it throughout the process, and verify we don't break it while we do it.

It will also make this PR smaller.

task-generator/trusted-artifacts/ta.go
hmariset: Updated the code to add the volumemount if specified else default value is used.

task-generator/trusted-artifacts/ta.go
hmariset: Updated to use boolean value useTAVolumeMount

task-generator/trusted-artifacts/README.md
```diff
@@ -55,10 +55,12 @@ The following is the list of supported options:
...
 | `addVolumeMount`     | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Task |
+| `addTAVolumeMount`   | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Trusted Artifact Task |
 | `base`               | string                                           | Relative path from `recipe.yaml` to the Task definition of the non-Trusted Artifacts Task |
 | `description`        | string                                           | Description of the Trusted Artifacts Task |
 | `displaySuffix`      | string                                           | Additional text to place to the value of `tekton.dev/displayName` annotation from the non-Trusted Artifacts Task to the Trusted Artifacts Task (default: `" oci trusted artifacts"`) |
 | `preferStepTemplate` | boolean                                          | When `true` preference is set to configure common configuration on the `Task.spec.stepTemplate` rather than on each Task Step |
+| `useTAVolumeMount`   | boolean                                          | When `true` Volume Mount is added to the Trusted Artifact |
```
yftacherzog, line 63: We need to mention somehow that if `addTAVolumeMount` is not provided and `useTAVolumeMount` is `true` then a default mount is applied.

task-generator/trusted-artifacts/recipe.go
hmariset: Added

task-generator/trusted-artifacts/README.md
hmariset: Updated

task-generator/trusted-artifacts/README.md
```diff
@@ -55,10 +55,12 @@ The following is the list of supported options:
...
+| `addTAVolumeMount`   | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Trusted Artifact Task |
 | `base`               | string                                           | Relative path from `recipe.yaml` to the Task definition of the non-Trusted Artifacts Task |
 | `description`        | string                                           | Description of the Trusted Artifacts Task |
 | `displaySuffix`      | string                                           | Additional text to place to the value of `tekton.dev/displayName` annotation from the non-Trusted Artifacts Task to the Trusted Artifacts Task (default: `" oci trusted artifacts"`) |
 | `preferStepTemplate` | boolean                                          | When `true` preference is set to configure common configuration on the `Task.spec.stepTemplate` rather than on each Task Step |
+| `useTAVolumeMount`   | boolean                                          | When `true` and if `addTAVolumeMount` is provided then this value is added to the Trusted Artifact step, if not then default volume mount is added |
```
yftacherzog, line 63:
```suggestion
| `useTAVolumeMount`   | boolean                                          | When `true` VolumeMounts are added to the Trusted Artifact step. If `addTAVolumeMount` is provided then those VolumeMounts are added. Otherwise,  default VolumeMounts are added |
```

task-generator/trusted-artifacts/README.md
```diff
@@ -55,10 +55,12 @@ The following is the list of supported options:
...
+| `addTAVolumeMount`   | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Trusted Artifact Task |
 | `base`               | string                                           | Relative path from `recipe.yaml` to the Task definition of the non-Trusted Artifacts Task |
 | `description`        | string                                           | Description of the Trusted Artifacts Task |
 | `displaySuffix`      | string                                           | Additional text to place to the value of `tekton.dev/displayName` annotation from the non-Trusted Artifacts Task to the Trusted Artifacts Task (default: `" oci trusted artifacts"`) |
 | `preferStepTemplate` | boolean                                          | When `true` preference is set to configure common configuration on the `Task.spec.stepTemplate` rather than on each Task Step |
+| `useTAVolumeMount`   | boolean                                          | When `true` VolumeMounts are added to the Trusted Artifact step. If `addTAVolumeMount` is provided then those VolumeMounts are added. Otherwise,  default VolumeMounts are added |
```
yftacherzog, line 63:
```suggestion
| `useTAVolumeMount`   | boolean                                          | When `true` VolumeMounts are added to the Trusted Artifact step. If `addTAVolumeMount` is provided then those VolumeMounts are added. Otherwise, default VolumeMounts are added |
```

yftacherzog (dismissed):
lgtm

hmariset:
@konflux-ci/build-maintainers Please review this PR. Thank you!

chmeliik (commented):
@hmariset could you describe the changes a bit better in the commit message?

My first thoughts when reviewing were:

> Add custom certificate support for git clone task

Wait, it supports those already

> Add param to support the custom certificate support for git-clone-oci-ta task to connect to internal registry.

Wait, why would the git-clone task connect to a _registry_? That was probably meant to say Gitlab?


Only  later did I realize that this is for the use-trusted-artifact/create-trusted-artifact steps, which download/upload the artifact from/to a registry

task-generator/trusted-artifacts/README.md
```diff
@@ -55,10 +55,12 @@ The following is the list of supported options:
 | `addResult`          | sequence of Tekton [TaskResult]s                 | Additional Tekton Task results to add to the Task |
 | `addVolume`          | sequence of [Volume]s                            | Additional Volumes to add to the Task |
 | `addVolumeMount`     | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Task |
+| `addTAVolumeMount`   | sequence of [VolumeMount]s                       | Additional VolumeMount to add to the Trusted Artifact Task |
```
chmeliik, line 58: Here as well, it would be good to say that this is specifically for the `use-trusted-artifact` step rather than the whole task

task/buildah-oci-ta/0.3/recipe.yaml
chmeliik, line 1: Could you make this change for `0.2` as well? They're not deprecated yet

chmeliik:
Failed with `"verify-enterprise-contract-run-krkfj" already exists, the server was not able to generate a unique name for the object`

That's a rare one :gem: 

/retest

chmeliik (approved):

chmeliik:
Thanks for the fix!

dirgim (approved):
+1

PR #1815, opened 2025-01-14
Config indy generic proxy as a sidecar of buildah task

sswguo:
Middleware build pipeline needs to access indy for generic proxy requests (like non-java & non-npm artifacts), and it requires the indy generic proxy deployed together with the builder in the same cluster to handle the proxy http format, which openshift routers don't allow.  

In MP+, we deployed this as a standalone service, and in Konflux, the sidecar way maybe a better option, but we hope to start discussing based on this PR, to see which one is better. Or if we need a dedicated task to avoid introducing this to all of the users even there are feature flag.

### DIFF
--- a/task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
@@ -139,6 +139,18 @@ spec:
       description: The name of the ConfigMap to read CA bundle data from.
       type: string
       default: trusted-ca
+    - name: ENABLE_INDY_PROXY
+      type: string
+      description: Enable the indy generic proxy (true/false)
+      default: "false"
+    - name: INDY_PROXY_CLIENT_ID
+      type: string
+      description: The client ID used by indy proxy to communicate with Indy.
+      default: ""
+    - name: INDY_PROXY_CLIENT_CREDENTIAL
+      type: string
+      description: The client credential used by indy proxy to communicate with Indy.
+      default: ""
   results:
     - name: IMAGE_DIGEST
       description: Digest of the image just built
@@ -176,6 +188,17 @@ spec:
       emptyDir: {}
     - name: workdir
       emptyDir: {}
+    - name: indy-generic-proxy-secrets
+      secret:
+        optional: true
+        secretName: indy-generic-proxy-secrets
+    - name: indy-generic-proxy-config
+      configMap:
+        items:
+          - key: application.yaml
+            path: application.yaml
+        name: indy-generic-proxy-config
+        optional: true
   stepTemplate:
     computeResources:
       limits:
@@ -502,6 +525,12 @@ spec:
           done < <(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;)
         fi
 
+        if [ -f "$ca_bundle" ]; then
+          cp -r --preserve=mode /mnt/trusted-ca /tmp/trusted-ca
+          VOLUME_MOUNTS+=(--volume /tmp/trusted-ca:/etc/pki/ca-trust/source/anchors)
+          echo "Adding the trusted-ca to the build"
+        fi
+
         # Prevent ShellCheck from giving a warning because 'image' is defined and 'IMAGE' is not.
         declare IMAGE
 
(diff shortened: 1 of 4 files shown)


task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
```diff
@@ -725,3 +754,26 @@ spec:
...
+          memory: 2Gi
+        requests:
+          cpu: 200m
+          memory: 512Mi
+      script: |
+        if [ "$(params.ENABLE_INDY_PROXY)" == "true" ]; then
+          export QUARKUS_OIDC_CLIENT_CLIENT_ID="$(params.INDY_PROXY_CLIENT_ID)"
+          export QUARKUS_OIDC_CLIENT_CREDENTIALS_SECRET="$(params.INDY_PROXY_CLIENT_CREDENTIAL)"
+          /deployment/start-service.sh
+        fi
```
arewm, line 779: If we are to implement a sidecar like this, is there any reason why it needs to be specific to Quarkus/INDY? Would this have value as a generalization?

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
```diff
@@ -725,3 +754,26 @@ spec:
         requests:
           cpu: 100m
           memory: 256Mi
+  sidecars:
+    - name: indy-generic-proxy
+      image: quay.io/factory2/indy-generic-proxy-service:latest-prod
```
arewm, line 759: What is the provenance of this image?

arewm (commented):
What are the implications if we add this sidecar to every single buildah task that is created? Will this result in a drastic increase in cluster utilization?

sswguo:
> What are the implications if we add this sidecar to every single buildah task that is created? Will this result in a drastic increase in cluster utilization?

Seems there is no feature flag in tekton to control the sidecar, so yeah it will result in downloading the image for every single buildah task.

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
sswguo: Both `QUARKUS_OIDC_CLIENT_CLIENT_ID ` and `QUARKUS_OIDC_CLIENT_CREDENTIALS_SECRET ` are supported in quarkus that we can declare the variable as env, which can be declared in the config file as following: 

```
 quarkus
    oidc-client:
       auth-server-url: ""
       client-id: 
       credentials:
           secret: 
```

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
sswguo: I believe we need to follow konflux pattern to build the image if we agree the sidecar proposal.

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
matejonnet: QUARKUS_* are internal, I believe it doesn't really matter how they are called.
*_INDY_* should these be renamed to more generic name like HTTP_PROXY_* ?

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
```diff
@@ -725,3 +754,36 @@ spec:
         requests:
           cpu: 100m
           memory: 256Mi
+  sidecars:
```
mmorhun, line 757: Despite the functionality is behind a feature flag, it's still specific to PNC only.
I think we should keep the buildah task generic and create a dedicated pipeline for specific purposes (like FBC, Java, etc.)

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
```diff
@@ -139,6 +139,18 @@ spec:
       description: The name of the ConfigMap to read CA bundle data from.
       type: string
       default: trusted-ca
+    - name: ENABLE_HTTP_PROXY
```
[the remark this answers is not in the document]
matejonnet: The description could be updated, but it's a generic http proxy, which in our case provides persistent cache per build.

task/buildah-oci-ta/0.3/buildah-oci-ta.yaml
```diff
@@ -139,6 +139,18 @@ spec:
...
+      type: string
+      description: Enable the http proxy (true/false)
+      default: "false"
+    - name: HTTP_PROXY_CLIENT_ID
```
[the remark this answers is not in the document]
matejonnet: Should we change it to HTTP_PROXY_SSO_CLIENT_ID ?

PR #1817, opened 2025-01-14
git-clone task: Support fetching tags of submodules

andreasgerstmayr:
Add a new `fetchSubmoduleTags` param to the git-clone task to fetch tags from all submodules.

The change should be straightforward, but I didn't test it so far. Is there a _simple_ way to build a container for this task, push it to a public registry and flip this line: https://github.com/os-observability/konflux-tempo/blob/8090ec116d33d2a2bb1bcdc3967159f18a258b3f/.tekton/multi-arch-build-pipeline.yaml#L42 to test this change?

### DIFF
--- a/task/git-clone-oci-ta/0.1/README.md
+++ b/task/git-clone-oci-ta/0.1/README.md
@@ -9,6 +9,7 @@ The git-clone-oci-ta Task will clone a repo from the provided url and store it a
 |caTrustConfigMapName|The name of the ConfigMap to read CA bundle data from.|trusted-ca|false|
 |depth|Perform a shallow clone, fetching only the most recent N commits.|1|false|
 |enableSymlinkCheck|Check symlinks in the repo. If they're pointing outside of the repo, the build will fail. |true|false|
+|fetchSubmoduleTags|Fetch all tags from all submodules.|false|false|
 |fetchTags|Fetch all tags for the repo.|false|false|
 |httpProxy|HTTP proxy server for non-SSL requests.|""|false|
 |httpsProxy|HTTPS proxy server for SSL requests.|""|false|
--- a/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
+++ b/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
@@ -34,6 +34,10 @@ spec:
         Check symlinks in the repo. If they're pointing outside of the repo, the build will fail.
       type: string
       default: "true"
+    - name: fetchSubmoduleTags
+      description: Fetch all tags from all submodules.
+      type: string
+      default: "false"
     - name: fetchTags
       description: Fetch all tags for the repo.
       type: string
@@ -184,6 +188,8 @@ spec:
           value: $(params.userHome)
         - name: PARAM_FETCH_TAGS
           value: $(params.fetchTags)
+        - name: PARAM_FETCH_SUBMODULE_TAGS
+          value: $(params.fetchSubmoduleTags)
         - name: WORKSPACE_SSH_DIRECTORY_BOUND
           value: $(workspaces.ssh-directory.bound)
         - name: WORKSPACE_SSH_DIRECTORY_PATH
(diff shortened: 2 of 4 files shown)


pavolloffay (approved):

chmeliik:
Is there a _simple_ way to build a container for this task, push it to a public registry and flip this line: https://github.com/os-observability/konflux-tempo/blob/8090ec116d33d2a2bb1bcdc3967159f18a258b3f/.tekton/multi-arch-build-pipeline.yaml#L42 to test this change?

```bash
tkn bundle push quay.io/<your-namespace>/tekton-catalog/task-git-clone:some-tag -f task/git-clone/0.1/git-clone.yaml
```

Then use `quay.io/<your-namespace>/tekton-catalog/task-git-clone:some-tag` in your pipeline

chmeliik:
> ```shell
> tkn bundle push quay.io/<your-namespace>/tekton-catalog/task-git-clone:some-tag -f task/git-clone/0.1/git-clone.yaml
> ```
> 
> Then use `quay.io/<your-namespace>/tekton-catalog/task-git-clone:some-tag` in your pipeline
> 
> tkn: https://tekton.dev/docs/cli/

Oh, and you'll have to go to https://quay.io/<your-namespace>/tekton-catalog/task-git-clone and make the repository public

andreasgerstmayr:
Thank you for the instructions! I tested it successfully.

Now we can set `fetchSubmoduleTags=true` and `depth=10`, and `git describe --tags --abbrev=0` in a Dockerfile returns the correct version of a component whose sources are in a submodule.

Fun fact, I tried to run `git describe` without my changes and suddenly it also worked when using `depth=10`. That was quite unexpected, and after a lot of debugging I figured out that cachi2 already fetches the tags (afaics here: https://github.com/containerbuildsystem/cachi2/blob/7557f2c6ab01523de5a544e2ab34e4134b976069/cachi2/core/package_managers/gomod.py#L1255). But that's a side effect of using the prefetch-dependencies task, imho it's much cleaner to explicitly configure the git-clone task to fetch the tags.

chmeliik:
> Fun fact, I tried to run `git describe` without my changes and suddenly it also worked when using `depth=10`. That was quite unexpected, and after a lot of debugging I figured out that cachi2 already fetches the tags (afaics here: https://github.com/containerbuildsystem/cachi2/blob/7557f2c6ab01523de5a544e2ab34e4134b976069/cachi2/core/package_managers/gomod.py#L1255). But that's a side effect of using the prefetch-dependencies task, imho it's much cleaner to explicitly configure the git-clone task to fetch the tags.

Hmm, interesting. I think cachi2 does more or less an equivalent of `git fetch`, with no explicit submodule handling. Maybe `git fetch` is already submodule-aware to some extent?

andreasgerstmayr:
> > Fun fact, I tried to run `git describe` without my changes and suddenly it also worked when using `depth=10`. That was quite unexpected, and after a lot of debugging I figured out that cachi2 already fetches the tags (afaics here: https://github.com/containerbuildsystem/cachi2/blob/7557f2c6ab01523de5a544e2ab34e4134b976069/cachi2/core/package_managers/gomod.py#L1255). But that's a side effect of using the prefetch-dependencies task, imho it's much cleaner to explicitly configure the git-clone task to fetch the tags.
> 
> Hmm, interesting. I think cachi2 does more or less an equivalent of `git fetch --tags`, with no explicit submodule handling. Maybe `git fetch` is already submodule-aware to some extent?

I think cachi2 simply runs this command in the subfolder. This is the config for the prefetch-dependencies task: `[{"type": "gomod", "path": "./tempo"}, {"type": "rpm"}]`.

When run outside the folder, `git fetch --tags` doesn't recursively fetch tags of subfolders (unless you have [`submodule.recurse`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse) enabled in the git config).

chmeliik:
> I think cachi2 simply runs this command in the subfolder. This is the config for the prefetch-dependencies task: `[{"type": "gomod", "path": "./tempo"}, {"type": "rpm"}]`.
> 
> When run outside the folder, `git fetch --tags` doesn't recursively fetch tags of subfolders (unless you have [`submodule.recurse`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse) enabled in the git config).

I believe cachi2 runs `git fetch` in the repo root. But I'm fine with that staying a mystery, this change seems reasonable :smile:

chmeliik (approved):

PR #1819, opened 2025-01-14
Add PNC prebuild task based upon git-clone-oci-ta

rnc:
As middleware (PNC) need to preprocess the source in order to prepare it for building within a container an extra step is required for the PNC custom pipeline

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -117,6 +117,9 @@
 /task/build-maven-zip           @ligangty @yma96
 /task/build-maven-zip-oci-ta    @ligangty @yma96
 
+# renovate groupName=pnc
+/task/pnc-prebuild-git-clone-oci-ta           @rnc
+
 # renovate groupName=oci-copy
 /task/oci-copy          @ralphbean
 /task/oci-copy-oci-ta   @ralphbean
--- a/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
+++ b/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
@@ -1,0 +1,76 @@
+- op: replace
+  path: /metadata/name
+  value: pnc-prebuild-git-clone-oci-ta
+- op: replace
+  path: /metadata/annotations/tekton.dev~1displayName
+  value: pnc prebuild git clone oci trusted artifacts
+- op: replace
+  path: /spec/description
+  value: The pnc-prebuild-git-clone-oci-ta Task will clone a repo from the provided url, apply
+    PNC prebuild modifications and store it as a trusted artifact in the provided OCI repository.
+
+- op: add
+  path: /spec/params/-
+  value:
+    name: BUILD_SCRIPT
+    description: The build script to embed with the Containerfile
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: BUILD_TOOL
+    description: The build tool to use (ant, gradle, maven, sbt).
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: BUILD_TOOL_VERSION
+    description: The build tool version to use (e.g. 3.9.5)
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: JAVA_VERSION
+    description: Java version to use (7, 8, 9, 11, 17, 21, 22, 23)
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: PNC_KONFLUX_TOOLING_IMAGE
+    description: Name of the tooling image.
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: RECIPE_IMAGE
+    description: The image from the build recipe to use
+    type: string
+
+- op: add
+  path: /spec/steps/2
+  value:
+    name: preprocessor
+    image: $(params.PNC_KONFLUX_TOOLING_IMAGE)
+    securityContext:
+      runAsUser: 0
+    computeResources:
+      limits:
+        cpu: 300m
+        memory: 512Mi
+      requests:
+        cpu: 10m
+        memory: 512Mi
+    args:
+      - prepare
+      - --build-tool-version=$(params.BUILD_TOOL_VERSION)
+      - --java-version=$(params.JAVA_VERSION)
+      - --recipe-image=$(params.RECIPE_IMAGE)
+      - --tooling-image=$(params.PNC_KONFLUX_TOOLING_IMAGE)
+      - --type=$(params.BUILD_TOOL)
+      - /var/workdir/source
+    volumeMounts:
+      - mountPath: /var/workdir
+        name: workdir
+    env:
+      - name: BUILD_SCRIPT
+        value: $(params.BUILD_SCRIPT)
--- a/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
+++ b/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
@@ -6,15 +6,17 @@
   value: pnc prebuild git clone oci trusted artifacts
 - op: replace
   path: /spec/description
-  value: The pnc-prebuild-git-clone-oci-ta Task will clone a repo from the provided url, apply
-    PNC prebuild modifications from https://github.com/project-ncl/konflux-tooling and store
-    it as a trusted artifact in the provided OCI repository.
+  value: The pnc-prebuild-git-clone-oci-ta task will clone a repo from the provided url, apply
+    PNC prebuild modifications (from https://github.com/project-ncl/konflux-tooling) and store
+    it as a trusted artifact in the provided OCI repository. The prebuild modifications create
+    a Containerfile and suitable build script in order for the Java based project to be built
+    within a container given build parameters configured from PNC.
 
 - op: add
   path: /spec/params/-
   value:
     name: BUILD_SCRIPT
-    description: The build script to embed with the Containerfile
+    description: Middleware (Maven/Gradle/Ant/SBT) build script to build the project to embed with the Containerfile
     type: string
 - op: add
   path: /spec/params/-
--- a/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
+++ b/task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
@@ -47,7 +47,7 @@
   path: /spec/steps/2
   value:
     name: preprocessor
-    image: quay.io/redhat-user-workloads/konflux-jbs-pnc-tenant/konflux-tooling:latest
+    image: quay.io/konflux-ci/pnc-konflux-tooling@sha256:5c11e8168d13267e12edf7329cd67d993f3ed708e5c57aa1319e3354dd8b2bcf
     securityContext:
       runAsUser: 0
     computeResources:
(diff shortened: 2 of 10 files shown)


task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,76 @@
...
+  value:
+    name: JAVA_VERSION
+    description: Java version to use (7, 8, 9, 11, 17, 21, 22, 23)
+    type: string
+- op: add
+  path: /spec/params/-
+  value:
+    name: PNC_KONFLUX_TOOLING_IMAGE
+    description: Name of the tooling image.
+    type: string
```
arewm, line 41: How often is this going to change? How generic is the image to satisfy a majority of use cases? It seems like it would be better to bake this type of image reference in to a task itself instead of requiring users to define it for each of their onboarded components' pipeline runs.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,76 @@
...
+        cpu: 10m
+        memory: 512Mi
+    args:
+      - prepare
+      - --build-tool-version=$(params.BUILD_TOOL_VERSION)
+      - --java-version=$(params.JAVA_VERSION)
+      - --recipe-image=$(params.RECIPE_IMAGE)
+      - --tooling-image=$(params.PNC_KONFLUX_TOOLING_IMAGE)
+      - --type=$(params.BUILD_TOOL)
+      - /var/workdir/source
```
arewm, line 70: This is a very opaque command. It isn'y clear what this actually does to the source especially since it is running with full access to the internet. I don't know how we would be able to qualify/verify our trust in the changes made here.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: Actually thats a good point ; now we have extracted the code from JBS into a onboarded repository that parameter/value could probably be removed and hardcoded to something like e.g. `quay.io/redhat-user-workloads/konflux-jbs-pnc-tenant/konflux-tooling:latest`

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: Admittedly this command does not need network access and is similar to what we would need if KONFLUX-4903 was available. It runs java code to preprocess the source to ensure its possible to build it inside a container from https://github.com/project-ncl/konflux-tooling and is only meant to be used by middleware/pnc.

matejonnet: If the concern is full internet access we can probably run it within "unshare" but I believe it's a matter of trusting the application, that it rally does what it should?

mmorhun: If it doesn't need internet access, I'd restrict it.

rnc: @mmorhun Are you saying this is your only objection before approval? That tooling image does not currently have unshare installed so we would have to modify the image I suppose.

rnc: I'll look at installing unshare in the tooling image and running this command with `unshare -n /opt/jboss/container/java/run/run-java.sh...`

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,71 @@
+- op: replace
+  path: /metadata/name
+  value: pnc-prebuild-git-clone-oci-ta
+- op: replace
+  path: /metadata/annotations/tekton.dev~1displayName
+  value: pnc prebuild git clone oci trusted artifacts
+- op: replace
+  path: /spec/description
+  value: The pnc-prebuild-git-clone-oci-ta Task will clone a repo from the provided url, apply
+    PNC prebuild modifications from https://github.com/project-ncl/konflux-tooling and store
```
mmorhun, line 10: `konflux-tooling` sounds too generic for me, maybe `pnc-konflux-tooling` would be better?

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,71 @@
...
+    it as a trusted artifact in the provided OCI repository.
+
+- op: add
+  path: /spec/params/-
+  value:
+    name: BUILD_SCRIPT
+    description: The build script to embed with the Containerfile
```
mmorhun, line 17: nitpick: I don't really understand which kind of build script you mean from the description.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,71 @@
...
+      runAsUser: 0
+    computeResources:
+      limits:
+        cpu: 300m
+        memory: 512Mi
+      requests:
+        cpu: 10m
+        memory: 512Mi
+    args:
+      - prepare
```
mmorhun, line 59: nitpick: maybe some comment or extended description would make it more clear what is actually done here.

mmorhun (commented):
Please update the [CODEOWNERS](https://github.com/konflux-ci/build-definitions/blob/main/CODEOWNERS) file.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: Unfortunately I get `unshare: unshare failed: Operation not permitted` when trying to run the above command

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: I am not sure how we can change the image name as isn't it derived from the repository name?

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: Rather than unshare, would it be acceptable to simply define `-DsocksProxyHost=does.not.exist` which according to https://docs.oracle.com/javase/7/docs/technotes/guides/net/proxies.html should force all outgoing TCP connections to go via the proxy (which will then fail) ?

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - --build-tool-version=$(params.BUILD_TOOL_VERSION)
+      - --java-version=$(params.JAVA_VERSION)
+      - --recipe-image=$(params.RECIPE_IMAGE)
+      - --tooling-image=quay.io/redhat-user-workloads/konflux-jbs-pnc-tenant/konflux-tooling:latest
```
arewm, line 65: I don't think that we should reference an image from redhat-user-workloads as this can be accidentally deleted of the component is deleted. Can we release this to be an image in `quay.io/konflux-ci` instead?

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
mmorhun: You can release it via Konflux

mmorhun (approved):
Approving with the only condition: task image is released via Konflux.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: Agreed but we'll have to discuss that on how best to release that single component to the konflux owned namespace.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
rnc: I have not yet added anything to isolate the network. *If* network isolation is mandatory then I could add
```
  +    script: |
  +      #!/bin/bash
  +      /opt/jboss/container/java/run/run-java.sh -DsocksProxyHost=does.not.exist $@
```
to the task to isolate it.

mmorhun (approved):
We need to make CODEOWNERS check pass before the merge.

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
mmorhun: I'd defer it to EC team and prodsec

rnc: Can they be linked in to give their opinion please?

simonbaird: I don't have a strong opinion on the security implications. Re the opaqueness concern, maybe a comment with a link to the source code for the command being run would help address that. Re the "it should run without internet access", I agree it's a good idea, but consider filing a separate Jira for tracking that work and implementing it in a future PR, rather than have it block this one. No need to do everything in one PR, right?

rnc: Re `link to the source code for the command` ; its now linked in the readme

p-rog: From the ProdSec side I don't have any strong opinion.  I will ask some of my team mates if they see any blockers here. My only concern is related to the fact how we can register what was cloned from the provided url? Do we want to verify if the url is trusted somehow? Can we precisely register the source url and manifest the content?

rnc: @p-rog I'm not entirely sure what you mean; I've adapted the git-clone task as KONFLUX-4903 is not yet available.

mmorhun (approved):

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
p-rog: The purpose of this rebuild steps is to collect the java source code to ensure its possible to build it inside the container. My concerns are related to the sources manifesting and potential chance of losing information about the source code and artifacts used to build the particular component, which is crucial from the later vulnerability management and supply chain security risk assessment perspective. But maybe I misunderstand something here and these concerns are unnecessary.

rnc: I think this is diverging off network isolation and perhaps should be entered as a JIRA? Or we can discuss in slack with @matejonnet.  If you're talking generating sboms that would be handled by sbomer I guess?

CODEOWNERS
```diff
@@ -115,6 +115,9 @@
 /task/build-maven-zip           @ligangty @yma96
 /task/build-maven-zip-oci-ta    @ligangty @yma96
 
+# renovate groupName=pnc
+/task/pnc-prebuild-git-clone-oci-ta           @rnc
```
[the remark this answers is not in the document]
arewm: Done

task/pnc-prebuild-git-clone-oci-ta/0.1/patch.yaml
arewm: > unshare: unshare failed: Operation not permitted

The processing here is just calling a script in your image. Did you try to use unshare within the command in your image that was called?

What are the possible impacts of the user-provided script parameter? Are we sufficiently protecting against maliciously crafted content?

rnc: The processing here is running a java program and so unshare couldn't be used within it though I did experiment with unshare running the program but it seemed  using `-DsocksProxyHost` would provide the same effect.
The user provided script is not run in this stage but embedded into the build script. It comes from PNC and is used in the build stage (e.g. mvn -D.... install or gradle -P.....assemble).

PR #1821, opened 2025-01-14
Remove cpu limits from tasks

enkeefe00:
Closes: [KFLUXINFRA-767](https://issues.redhat.com/browse/KFLUXINFRA-767)


task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
```diff
@@ -282,7 +282,6 @@ spec:
         echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee "$(results.TEST_OUTPUT.path)"
       computeResources:
         limits:
-          cpu: "16"
```
kdudka, line 285: We specified the limit on purpose to get more resources for the Coverity scanner.  How is this supposed to work after the change is applied?

cmoulliard:
By removing such CPU limits you will then accept that pods get more CPU (if available on the node) as documented here https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ ?

chmeliik:
@enkeefe00 could you explain the motivation for this change? I don't think anyone here understands it, and it feels like the opposite direction of what @jhutar has been doing (adding resource requests and limits to tasks)

kdudka:
@cmoulliard Thanks for the link!  The question is whether this is the only place where the limit can be set.  Are you sure it cannot happen that a lower limit is set elsewhere and we need to override the number in the `sast-coverity-check` task?

jhutar:
Hello. Generally I'm more concerned about memory request and limit. @hugares , do you have a preference?

Regarding if the CPU limit will come from somewhere else: https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit

> If you do not specify a CPU limit for a Container, then one of these situations applies:
> 
> * The Container has no upper bound on the CPU resources it can use. The Container could use all of the CPU resources available on the Node where it is running.
> * The Container is running in a namespace that has a default CPU limit, and the Container is automatically assigned the default limit. Cluster administrators can use a [LimitRange](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#limitrange-v1-core/) to specify a default value for the CPU limit.

So looking at `stone-stg-rh01`, I see:

```
$ oc -n jhutar-1-tenant get LimitRange/resource-limits -o yaml | yq .spec
limits:
  - default:
      memory: 2Gi
    defaultRequest:
      cpu: 100m
      memory: 256Mi
    type: Container
```

Looks like we only have default CPU request.

Also Tekton plays a role when assigning resources to containers, but looks like that will not be relevant here: https://tekton.dev/docs/pipelines/compute-resources/

chmeliik (commented):
@enkeefe00 makes sense, thanks for linking the relevant resources. 

The failing check is saying that the "trusted-artifacts" task generator wants to put some limits back in place. You can fix that in `task-generator/trusted-artifacts/ta.go` I think

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
kdudka: @enkeefe00 Makes sense.  Thanks for explanation!

chmeliik (approved):

kdudka (approved):

hugares:
@chmeliik Can we merge this PR? I am not familiar with review/merge process in build-definitions

chmeliik:
> @chmeliik Can we merge this PR? I am not familiar with review/merge process in build-definitions

We still need reviews from:
* @konflux-ci/integration-service-maintainers 
* @ligangty or @yma96 
* @cmoulliard

enkeefe00:
@cmoulliard and @yma96 or @ligangty could you review this PR so we can try to merge it soon?

cmoulliard (approved):

chmeliik:
We're only missing reviews from @yma96 or @ligangty and I don't think they would be opposed. Let's force-merge this before we merge https://github.com/konflux-ci/build-definitions/pull/1865 and causing a rebase to be needed

mmorhun (approved):

PR #1798, opened 2025-01-08
change sbom-utility-scripts refs to new location

tnevrlka:
- `quay.io/redhat-appstudio/sbom-utility-scripts-image` got renamed and moved, so it's now `konflux-ci/sbom-utility-scripts`
- This is a part of the effort to decomission redhat-appstudio Quay org
- Change the image references to the new location

PR #1825, opened 2025-01-15
buildah/0.3: support generating SPDX SBOMs

chmeliik:
See commit messages for more details

Depends on https://github.com/konflux-ci/build-definitions/pull/1798 and a subsequent Renovate update (the current revision of the sbom-utility-scripts image doesn't include the SPDX support yet)

Tested in https://github.com/redhat-appstudio/rh-syft/pull/105 (using a test build of the sbom-utility-scripts image)

PR #1826, opened 2025-01-15
rpm-ostree: support generating SPDX SBOMs

chmeliik:
Add an SBOM_TYPE parameter, defaulting to 'cyclonedx' for now. This
parameter determines the format of the Syft-generated SBOM. This must
match the format of the SBOM coming from the prefetch task (if any),
otherwise the SBOM merging will fail.

Depends on https://github.com/konflux-ci/build-definitions/pull/1798 and a subsequent Renovate update (the current revision of the sbom-utility-scripts image doesn't include the SPDX support yet)


cgwalters:
Presumably this is matching a corresponding change to the buildah task; can you link that?

Is this being driven by a specific requirement to match that or is it more a proactive thing?

cgwalters:
Also tangential to this https://github.com/konflux-ci/build-definitions/pull/1856

And again just to reiterate https://issues.redhat.com/browse/BIFROST-408 is close to completion and I think we can drop this task soon, so if this isn't *urgent* I would probably say let's just push to complete https://issues.redhat.com/browse/BIFROST-408

chmeliik:
> Presumably this is matching a corresponding change to the buildah task; can you link that?

Yes, my bad. The buildah change was done in https://github.com/konflux-ci/build-definitions/pull/1825

> Is this being driven by a specific requirement to match that or is it more a proactive thing?

There isn't really a need to match the buildah task, but if rpm-ostree doesn't get SPDX support soon, it may result in an inconvenience to this task's users. I'll try to explain:

* The rpm-ostree task merges a Syft-generated SBOM with an SBOM coming from the prefetch task: https://github.com/konflux-ci/build-definitions/blob/69fdc802b3627999efaf3ca09392ae0dc926953b/task/rpm-ostree/0.2/rpm-ostree.yaml#L230
* The prefetch task (and other tasks) recently gained SPDX support: https://github.com/konflux-ci/build-definitions/pull/1852
* We will soon be releasing a `v0.2` of the prefetch task which will flip the default to SPDX
* The PR proposing the update of prefetch 0.1 -> 0.2 will break the build pipeline unless other SBOM-handling tasks are SPDX ready
* So we're planning to make all the tasks SPDX ready, then do a mass version bump and flip the default to SPDX. That way, users will get a PR which updates the prefetch task, buildah/rpm-ostree task and other SBOM-related tasks at the same time and things will work seamlessly (that's the hope, anyway)

TLDR: if we don't do this, the Mintmaker PR that updates the prefetch task to `0.2` will not be merge-able for rpm-ostree users. If you're OK with that, there's no need for this

chmeliik:
> TLDR: if we don't do this, the Mintmaker PR that updates the prefetch task to 0.2 will not be merge-able for rpm-ostree users. If you're OK with that, there's no need for this

Oh and at some point, ProdSec will start requiring that Konflux releases use SPDX I think. But I'm not sure when that will be

chmeliik:
Dropping since the rpm-ostree task is not maintained anymore

PR #1613, opened 2024-11-15
Make EC verify task timeout infinite

ralphbean:
Before this change, users could use the TIMEOUT param to increase their timeout up from the 5m default, but they would hit a ceiling. If they increased their timeout past 2h, then a tekton default on timeout for the task itself would kick in.

PR #1830, opened 2025-01-17
fix: add 4h task timeout in the ec pipeline

dirgim:
Since the default timeout for tasks is hardcoded to 1h/2h, this change sets the timeout to a higher value to
support users who run bigger sets of images through the verify task.

It is still possible to limit the duration of the check on the pipeline or pipelineRun level, depending on usage.

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -119,3 +119,4 @@ spec:
             value: verify-enterprise-contract
           - name: kind
             value: task
+      timeout: 4h


jsztuka (approved):
Lgtm

dirgim:
> FYI there's also #1613

Ack, but it seems to have stalled, hence I'm proposing this solution in order to unblock users who are running into the hardcoded 2h timeout. If that PR does get unblocked quickly, I'd be happy to close this one.

chmeliik:
Yeah, apparently there was a problem with it, Martin even tried to resolve it but didn't work https://github.com/konflux-ci/build-definitions/pull/1834

joejstuart (approved):

PR #1833, opened 2025-01-17
Add deprecation notice to provision-env-with-ephemeral-namespace

amisstea:


yftacherzog (approved):
/lgtm

Omeramsc (approved):
LGTM

yftacherzog:
@amisstea the change we did here will be overwritten once updates are merged from tekton-tools. I opened a PR there with the same content.

PR #1788, opened 2025-01-06
chore: fix ec-cli version to v0.6.150

tnevrlka:
The latest release of ec-cli is missing the executables. 
This makes it fail build of the appstudio-utils image. 
Fix the version of ec-cli to the latest working release

PR #1842, opened 2025-01-20
Using rolling release of EC

joejstuart:
Since there's not an easy way to update ec use the rolling release.

https://issues.redhat.com/browse/EC-1066

### DIFF
--- a/appstudio-utils/Dockerfile
+++ b/appstudio-utils/Dockerfile
@@ -1,4 +1,8 @@
+FROM quay.io/enterprise-contract/ec-cli:snapshot AS ec-image
 FROM registry.access.redhat.com/ubi9/ubi
+
+COPY --from=ec-image /usr/local/bin/ec_linux_amd64.gz /usr/bin/ec.gz
+RUN gunzip /usr/bin/ec.gz && chmod +x /usr/bin/ec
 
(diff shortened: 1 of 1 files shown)


chmeliik:
We pinned to a specific release in https://github.com/konflux-ci/build-definitions/pull/1788, because the `stable` release has been broken at least twice that I can remember (missing the executable assets)

joejstuart:
> We pinned to a specific release in #1788, because the `stable` release has been broken at least twice that I can remember (missing the executable assets)

We got it fixed here. https://github.com/enterprise-contract/ec-cli/pull/2244. Without regular updates of the cli, it can fall behind the policies which can cause issues.

joejstuart:
@chmeliik I'm going to add an additional verification step to help prevent this also. https://issues.redhat.com/browse/EC-1082

chmeliik (dismissed):
Assuming more robust releases of ec-cli this LGTM

joejstuart:
@chmeliik I misunderstood the issue. It was with the Github release process and not with the ec-cli image. This latest change copies the binary from the ec-cli image bypassing the download from Github. The verification step after the image is built is also added. https://github.com/enterprise-contract/ec-cli/pull/2280/files

chmeliik (approved):
Nice, this seems better :+1:

PR #1848, opened 2025-01-22
feat(STONEINTG-1098): setting default KFP_GIT_URL for sast-snyk-check

hongweiliu17:
feat([STONEINTG-1098](https://issues.redhat.com//browse/STONEINTG-1098)): setting default KFP_GIT_URL for snyk check
    
    * setting default KFP_GIT_URL to
      https://gitlab.cee.redhat.com/osh/known-false-positives.git
      for task running in internal Konflux instance
    * setting default KFP_GIT_URL to empty string for task running
      in external Konflux instance
    
    Signed-off-by: Kamil Dudka <kdudka@redhat.com>
    Signed-off-by: Hongwei Liu <hongliu@redhat.com>

### DIFF
--- a/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
+++ b/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
@@ -38,9 +38,13 @@ spec:
       type: string
       default: "true"
     - name: KFP_GIT_URL
-      description: URL from repository to download known false positives files
+      description: Known False Positives (KFP) git URL (optionally taking
+        a revision delimited by \#). Defaults to "SITE_DEFAULT", which means
+        the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+        for internal Konflux instance and empty string for external Konflux
+        instance. If set to an empty string, the KFP filtering is disabled.
       type: string
-      default: ""
+      default: SITE_DEFAULT
     - name: PROJECT_NAME
       description: Name of the scanned project, used to find path exclusions.
         By default, the Konflux component name will be used.
@@ -128,6 +132,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.labels['appstudio.openshift.io/component']
+        - name: BUILD_PLR_LOG_URL
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['pipelinesascode.tekton.dev/log-url']
       script: |
         #!/usr/bin/env bash
 
@@ -195,6 +203,21 @@ spec:
 
           echo "Results:"
           (set -x && csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
+
+          # Set KFP_GIT_URL to https://gitlab.cee.redhat.com/osh/known-false-positives.git for internal Konflux instance
+          # p01 will be decommissionned due to lack of IBM support
+          if [[ "${KFP_GIT_URL}" == "SITE_DEFAULT" ]]; then
+            case "${BUILD_PLR_LOG_URL}" in
+            *konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com* | *konflux.apps.stone-prod-p01.wcfb.p1.openshiftapps.com*)
+              echo "the task is running within Red Hat network, set KFP_GIT_URL to https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+              KFP_GIT_URL="https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+              ;;
+            *)
+              echo "the task is not running within Red Hat network, set KFP_GIT_URL to empty string"
+              KFP_GIT_URL=
+              ;;
+            esac
+          fi
 
           # We check if the KFP_GIT_URL variable is set to apply the filters or not
           if [[ -z "${KFP_GIT_URL}" ]]; then
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -222,7 +222,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |CACHI2_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the prefetched dependencies.| | '$(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)'|
 |IGNORE_FILE_PATHS| Directories or files to be excluded from Snyk scan (Comma-separated). Useful to split the directories of a git repo across multiple components.| | |
 |IMP_FINDINGS_ONLY| Report only important findings. Default is true. To report all findings, specify "false"| true| |
-|KFP_GIT_URL| URL from repository to download known false positives files| | |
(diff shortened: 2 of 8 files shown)


kdudka:
@hongweiliu17 Thank you for working on this!  I think this approach makes sense.  Nevertheless, I would like to preserve the behavior where `KFP_GIT_URL` set to an empty string means "KFP disabled" so that users get the unfiltered list of findings in this case.

Could we please introduce a documented constant such as `SITE_DEFAULT`, which would be set as a default value for the `KFP_GIT_URL` parameter and handled this way in the task script?

Users would be able to override the default value either to an empty string to disable KFP, or to a custom URL to use their own KFP.

hongweiliu17:
> @hongweiliu17 Thank you for working on this! I think this approach makes sense. Nevertheless, I would like to preserve the behavior where `KFP_GIT_URL` set to an empty string means "KFP disabled" so that users get the unfiltered list of findings in this case.
> 
> Could we please introduce a documented constant such as `SITE_DEFAULT`, which would be set as a default value for the `KFP_GIT_URL` parameter and handled this way in the task script?
> 
> Users would be able to override the default value either to an empty string to disable KFP, or to a custom URL to use their own KFP.

I am trying to understand `SITE_DEFAULT`, that is a new parameter

> @hongweiliu17 Thank you for working on this! I think this approach makes sense. Nevertheless, I would like to preserve the behavior where `KFP_GIT_URL` set to an empty string means "KFP disabled" so that users get the unfiltered list of findings in this case.
> 
> Could we please introduce a documented constant such as `SITE_DEFAULT`, which would be set as a default value for the `KFP_GIT_URL` parameter and handled this way in the task script?
> 
> Users would be able to override the default value either to an empty string to disable KFP, or to a custom URL to use their own KFP.

@kdudka I am trying to understand your meaning and make a a change but name `SITE_DEFAULT` to `KFP_SITE_DEFAULT` to make it unconfusing, can you review it?

kdudka:
@hongweiliu17 Sorry for being unclear!  I did not mean to introduce a new parameter, just a new default value of the existing parameter with a special semantics.  I have drafter my idea as #1849.  What do you think about it?

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -196,6 +204,19 @@ spec:
           echo "Results:"
           (set -x && csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
 
+          if [[ "${KFP_GIT_URL}" == "SITE_DEFAULT" ]]; then
+            case "${BUILD_PLR_LOG_URL}" in
+            *console.redhat.com* | *console.dev.redhat.com*)
```
kdudka, line 209: Sorry for the confusion in #1849.  I thought that `console.redhat.com` was the internal Konflux cluster.  Nevertheless, I would invert the logic to make it more community-friendly.

Could we please match the internal Konflux instances and fallback to empty string for everything else?

If a new Konflux cluster appears in the future, it is safer to assume that it does not have access to RH internal network.

kdudka (changes requested):

hongweiliu17:
> @hongweiliu17 Sorry for being unclear! I did not mean to introduce a new parameter, just a new default value of the existing parameter with a special semantics. I have drafted my idea as #1849. What do you think about it?

Updated. Thanks!

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
hongweiliu17: Updated.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
```diff
@@ -196,6 +204,21 @@ spec:
           echo "Results:"
           (set -x && csgrep --mode=evtstat sast_snyk_check_out_all_findings.json)
 
+          # Set KFP_GIT_URL to https://gitlab.cee.redhat.com/osh/known-false-positives.git for internal Konflux instance
+          # p01 will be decommissionned due to lack of IBM support
+          if [[ "${KFP_GIT_URL}" == "SITE_DEFAULT" ]]; then
+            case "${BUILD_PLR_LOG_URL}" in
+            *konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com* | *konflux.apps.stone-prod-p01.wcfb.p1.openshiftapps.com*)
+              echo "the task is not running within Red Hat network, set KFP_GIT_URL to empty string"
```
kdudka, line 212: These log messages are swapped with each other in the `-oci-ta` copy of the task, which is also the reason why the `Validate PR - Trusted Artifact variants / Check Trusted Artifact variants (pull_request)` CI test is failing.  Either you need to execute `hack/generate-ta-tasks.sh` locally and commit its result, or apply the following patch:
```diff
--- a/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
+++ b/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
@@ -209,11 +209,11 @@ spec:
           if [[ "${KFP_GIT_URL}" == "SITE_DEFAULT" ]]; then
             case "${BUILD_PLR_LOG_URL}" in
             *konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com* | *konflux.apps.stone-prod-p01.wcfb.p1.openshiftapps.com*)
-              echo "the task is not running within Red Hat network, set KFP_GIT_URL to empty string"
+              echo "the task is running within Red Hat network, set KFP_GIT_URL to empty string"
               KFP_GIT_URL="https://gitlab.cee.redhat.com/osh/known-false-positives.git"
               ;;
             *)
-              echo "the task is running within Red Hat network, set KFP_GIT_URL to https://gitlab.cee.redhat.com/osh/known-false-positives.git"
+              echo "the task is not running within Red Hat network, set KFP_GIT_URL to empty string"
               KFP_GIT_URL=
               ;;
             esac
```

kdudka (changes requested):
@hongweiliu17 Thanks for the update!  A tiny fix is still needed though.

task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml
hongweiliu17: Updated. Thanks!

kdudka (approved):
@hongweiliu17 Thanks for the update!  The changes look good but I was unable to test this because I was not able to build any application with any of the internal Konflux instances.  I will give it another try next week.

kdudka:
Sorry for the delay!  Finally I was able to test this on https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com and it worked as expected.  I am putting this to the merge queue...

kdudka:
@jsztuka @dirgim I have submitted pull request #1883 to cover the other `sast-*-check` tasks.

hongweiliu17:
> Sorry for the delay! Finally I was able to test this on https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com and it worked as expected. I am putting this to the merge queue...

Thank you very much for your code and test!

PR #1859, opened 2025-01-27
Print build_push_task output

tkdchen:


hack/build-and-push.sh
```diff
@@ -482,6 +482,7 @@ build_push_tasks() {
             echo "info: push new bundle $task_bundle" 1>&2
 
             output=$(build_push_task "$task_dir" "$prepared_task_file" "$task_bundle" "$task_file_sha" "$has_migration")
+            echo "$output" >&2
```
chmeliik, line 485: Could we just echo an extra empty line here to make the output more readable? Currently, it all blends together and it's hard to tell where the output for one task ends and the next one begins

tkdchen: Updated.

chmeliik (approved):

PR #1860, opened 2025-01-27
Fix value of has_migration variable

tkdchen:
Using string "true" for consistence.

Avoid always annotating task bundles if there is no migration.


chmeliik (approved):

tkdchen:
:tada:

PR #1861, opened 2025-01-27
Link to migration bundles

tkdchen:
Any task bundles that have migrations are linked by the newer task bundles by a dedicated annotation. See the following example task bundles:

  tb0 -->  tb1 --> tb2 --> tb3 --> tb4 --> tb5
           |               |
           M               M

where, tb2 and tb3 point to tb1, and tb4 and tb5 point to tb3.

Linking task bundles helps to reduce the number of HTTP requests made to Quay.io that query which task bundles have migrations.

Terms:

- tb means task bundle.
- M means migration.


hack/build-and-push.sh
```diff
@@ -431,6 +434,34 @@ attach_migration_file() {
     return 0
 }
 
+# Find previous bundle that has a migration and output its digest.
+find_previous_migration_bundle_digest() {
+    local -r task_name=${1:?Missing task name}
+    local -r task_version=${2:?Missing task version}
+    local -r repo=${TEST_REPO_NAME:-task-${task_name}}
+
+    local -r url_list_recent_tags="https://quay.io/api/v1/repository/${QUAY_NAMESPACE}/${repo}/tag/?onlyActiveTags=true&limit=30"
```
chmeliik, line 443: Is there a specific reason for `limit=30`? Shall we just use Quay's maximum, i.e. `limit=100`?

chmeliik (commented):
Nice idea!

hack/build-and-push.sh
tkdchen: For the link, the first tag is enough to build the link. I selected 30 for my local testing by using a single test image repo. Updated it to 100 so that it can cover more cases for convenience.

tkdchen:
Rebased on main branch to fix code conflict.

hack/build-and-push.sh
```diff
@@ -431,6 +434,34 @@ attach_migration_file() {
...
+            prev_bundle_digest=$(jq -r '.annotations."dev.konflux-ci.task/previous-migration-bundle"' </tmp/manifest.json)
+            has_migration=$(jq -r '.annotations."dev.konflux-ci.task.migration"' </tmp/manifest.json)
+            if [ "$has_migration" == "true" ]; then
+                echo "$manifest_digest"
+                return
+            fi
+            if [ "$prev_bundle_digest" == "null" ] || [ -z "$prev_bundle_digest" ]; then
+                echo
+                return
+            fi
```
chmeliik, line 461: I'm not sure I fully understand the logic here. Why do we want to stop iterating through the manifest digests when we find one without the `dev.konflux-ci.task/previous-migration-bundle` annotation?

It would make more sense to me if the condition was backwards - if there's a non-empty `dev.konflux-ci.task/previous-migration-bundle` annotation, echo its value and return

hack/build-and-push.sh
```diff
@@ -431,6 +434,34 @@ attach_migration_file() {
...
+    local -r expr_filter_tags=".tags[] | select(.name | test(\"^(${task_name}-)?${task_version}-[0-9a-f]+\$\")) | .manifest_digest"
+
+    local prev_bundle_digest has_migration
+
+    curl --fail -sL "$url_list_recent_tags" | jq -r "$expr_filter_tags" | \
+        while read -r manifest_digest; do
+            curl --fail -sL -o /tmp/manifest.json \
+                "https://quay.io/v2/${QUAY_NAMESPACE}/${repo}/manifests/${manifest_digest}"
+            prev_bundle_digest=$(jq -r '.annotations."dev.konflux-ci.task/previous-migration-bundle"' </tmp/manifest.json)
+            has_migration=$(jq -r '.annotations."dev.konflux-ci.task.migration"' </tmp/manifest.json)
```
chmeliik, line 453: nitpick: maybe we should keep the annotation style consistent
```suggestion
            prev_bundle_digest=$(jq -r '.annotations."dev.konflux-ci.task.previous-migration-bundle"' </tmp/manifest.json)
            has_migration=$(jq -r '.annotations."dev.konflux-ci.task.migration"' </tmp/manifest.json)
```

hack/build-and-push.sh
mmorhun: We stop because missing annotation means older task version without any migration. So it doesn't make sense to look at older versions - no migrations anyway.

hack/build-and-push.sh
```diff
@@ -431,6 +434,34 @@ attach_migration_file() {
     return 0
 }
 
+# Find previous bundle that has a migration and output its digest.
+find_previous_migration_bundle_digest() {
+    local -r task_name=${1:?Missing task name}
+    local -r task_version=${2:?Missing task version}
+    local -r repo=${TEST_REPO_NAME:-task-${task_name}}
+
+    local -r url_list_recent_tags="https://quay.io/api/v1/repository/${QUAY_NAMESPACE}/${repo}/tag/?onlyActiveTags=true&limit=100"
```
mmorhun, line 443: Why do we need to get a 100 of bundles? If the latest one has the annotation, we know the bundle with the most recent migration from it's value.

hack/build-and-push.sh
chmeliik: Hmm, in that case we probably don't need to iterate through the manifests at all. Wouldn't it be enough to look at the latest one?

* if it has a migration, return the manifest digest
* otherwise, return the value of the `dev.konflux-ci.task/previous-migration-bundle` annotation

chmeliik: Ah, I see your other comment. +1

hack/build-and-push.sh
tkdchen: The reason to list multiple tags is for the testing where task and pipeline bundles are pushed into a single image repository. How the tasks and pipelines are mixed is unpredictable. 100 is too big actually. For convenience, how about reduce it into a smaller one, like 30 or 20? Hard to tell the new number can cover all possible cases, but it could for most of time I think.

tkdchen:
Added a new commit to address review comment and rebased to fix code conflict.

hack/build-and-push.sh
mmorhun: As discussed today with the team, let's assume that each task has own image repository.

tkdchen: Reduced it to 30.

Note: the limit may be needed to be increased when `build-and-push.sh` runs for a single task without setting `BUILD_TAG`.

tkdchen: @mmorhun PTAL

chmeliik (approved):

hack/build-and-push.sh
```diff
@@ -431,6 +436,34 @@ attach_migration_file() {
...
 }
 
+# Find previous bundle that has a migration and output its digest.
+find_previous_migration_bundle_digest() {
+    local -r task_name=${1:?Missing task name}
+    local -r task_version=${2:?Missing task version}
+    local -r repo=${TEST_REPO_NAME:-task-${task_name}}
+
+    local -r url_list_recent_tags="https://quay.io/api/v1/repository/${QUAY_NAMESPACE}/${repo}/tag/?onlyActiveTags=true&limit=30"
+    local -r expr_filter_tags=".tags[] | select(.name | test(\"^(${task_name}-)?${task_version}-[0-9a-f]+\$\")) | .manifest_digest"
```
mmorhun, line 446: I think we can check if we push into a test location (all bundles into the same image repo) and if not, just get only 1 tag without filtering by name.
Also comments would be appreciated. Thanks.

tkdchen:
Tested by tests https://github.com/tkdchen/build-definitions/blob/add-script-for-testing-migrations/hack/test-task-migrations.sh

hack/build-and-push.sh
```diff
@@ -38,9 +38,11 @@ VCS_URL=https://github.com/konflux-ci/build-definitions
 VCS_REF=$(git rev-parse HEAD)
 
 declare -r ARTIFACT_TYPE_TEXT_XSHELLSCRIPT="text/x-shellscript"
-declare -r ANNOTATION_TASK_MIGRATION="dev.konflux-ci.task.migration"
 declare -r TEST_TASKS=${TEST_TASKS:-""}
 
+declare -r ANNOTATION_TASK_MIGRATION="dev.konflux-ci.task.migration"
```
mmorhun, line 43: Does `dev.konflux-ci.task.migration-exists` makes it more clear?

hack/build-and-push.sh
```diff
@@ -38,9 +38,11 @@ VCS_URL=https://github.com/konflux-ci/build-definitions
...
+declare -r ANNOTATION_PREVIOUS_MIGRATION_BUNDLE="dev.konflux-ci.task.previous-migration-bundle"
```
mmorhun, line 44: I think a comment like "the annotation value points to the task bundle which has a migration that's most recent to the task with the annotation" would be helpful.

hack/build-and-push.sh
```diff
@@ -228,6 +230,7 @@ build_push_task() {
     local -r task_bundle=$3
     local -r task_file_sha=$4
     local -r has_migration=$5
+    local -r prev_bundle_digest=$6
```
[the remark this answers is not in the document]
mmorhun: yes, but you need to look at the assignment first to undestand it. Just using `previous_migration_bundle_digest` would remove any doubt I think.

hack/build-and-push.sh
```diff
@@ -431,6 +436,45 @@ attach_migration_file() {
...
 }
 
+# Find previous bundle that has a migration and output its digest.
+find_previous_migration_bundle_digest() {
+    local -r task_name=${1:?Missing task name}
+    local -r task_version=${2:?Missing task version}
+
+    local limit=1
+    if [ -n "$TEST_REPO_NAME" ]; then
+        limit=10
```
mmorhun, line 446: Is it enough? For testing purposes we can use a safe value.

hack/build-and-push.sh
```diff
@@ -431,6 +436,45 @@ attach_migration_file() {
...
+    fi
+
+    local bundle_ref
+    bundle_ref=$(generate_tagged_task_bundle "$task_name" "$task_version")
+    local ns_repo=${bundle_ref%:*}  # remove tag
+    ns_repo=${ns_repo#*/}  # remove registry
+    # Retain the tag used for matching tag names
+    local -r name_pattern="${bundle_ref##*:}-"
```
mmorhun, line 454: Do we need it for non testing use-case?

hack/build-and-push.sh
```diff
@@ -431,6 +436,45 @@ attach_migration_file() {
...
+    unset bundle_ref
+
+    local -r params="onlyActiveTags=true&filter_tag_name=like:${name_pattern}&limit=${limit}"
+    local -r url_list_tags="https://quay.io/api/v1/repository/${ns_repo}/tag/?${params}"
+    local prev_bundle_digest has_migration
+
+    curl --fail -sL "$url_list_tags" | jq -r '.tags[].manifest_digest' | \
+        while read -r manifest_digest; do
```
mmorhun, line 462: Do we need this loop for non testing use case?

hack/build-and-push.sh
tkdchen: The tag generated by `generate_tagged_task_bundle` is affected by the `TEST_REPO_NAME` variable. It covers both.

hack/build-and-push.sh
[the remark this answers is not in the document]
tkdchen: Updated

hack/build-and-push.sh
tkdchen: Updated

hack/build-and-push.sh
```diff
@@ -440,17 +440,28 @@ attach_migration_file() {
...
     local -r task_name=${1:?Missing task name}
     local -r task_version=${2:?Missing task version}
-    local -r repo=${TEST_REPO_NAME:-task-${task_name}}
 
-    local -r url_list_recent_tags="https://quay.io/api/v1/repository/${QUAY_NAMESPACE}/${repo}/tag/?onlyActiveTags=true&limit=100"
-    local -r expr_filter_tags=".tags[] | select(.name | test(\"^(${task_name}-)?${task_version}-[0-9a-f]+\$\")) | .manifest_digest"
+    local limit=1
+    if [ -n "$TEST_REPO_NAME" ]; then
+        limit=10
+    fi
```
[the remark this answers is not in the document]
tkdchen: Now, only list 1 tag.

hack/build-and-push.sh
mmorhun: If we don't do it now, later it will be too late, if some real bundles would have a migration

tkdchen: There is still space before the final delivery. Can we make this change in separate PRs for this script and the migration tool?

hack/build-and-push.sh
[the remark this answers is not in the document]
mmorhun: Not really

hack/build-and-push.sh
```diff
@@ -479,9 +530,15 @@ build_push_tasks() {
...
         else
+            echo "info: finding the previous task bundle that has a migration" 1>&2
+            previous_migration_bundle_digest=$(find_previous_migration_bundle_digest "$task_name" "$task_version")
+
             echo "info: push new bundle $task_bundle" 1>&2
 
-            output=$(build_push_task "$task_dir" "$prepared_task_file" "$task_bundle" "$task_file_sha" "$has_migration")
+            output=$(
+                build_push_task "$task_dir" "$prepared_task_file" "$task_bundle" "$task_file_sha" \
+                    "$has_migration" "$previous_migration_bundle_digest"
```
mmorhun, line 540: Is `previous_migration_bundle_digest` used anywhere?

hack/build-and-push.sh
mmorhun: yes, sure.

mmorhun (approved):
Approving with agreement to update the annotation name in a follow up PR.
Also, please take a look at the question above.

hack/build-and-push.sh
tkdchen: It is used in `build_push_task` https://github.com/konflux-ci/build-definitions/pull/1861/files#diff-78a47706e6b9aca3579f93bc34adac628f9c102de7257a2fe14939130786a807R234.

PR #1862, opened 2025-01-27
Add document for using task migrations

tkdchen:
[STONEBLD-2829](https://issues.redhat.com//browse/STONEBLD-2829)

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,3 +256,85 @@ Currently, there are three policy configurations.
 
 A build Task, e.g. one that produces a container image, must abide by both `all-tasks` and
 `build-tasks` policy configurations.
+
+## Task Migration
+
+Task migrations are Bash scripts defined in version-specific task
+directories. In general, a migration consists of a series of `yq` commands that
+modify pipeline in order to work with the new version of task. Developers can
+do more with task migrations on the pipelines, e.g. add/review a task,
+add/remove/update task parameters, change execution order of a task, etc.
+
+Historically, task maintainers write `MIGRATION.md` to notify users what changes
+have to be made to the pipeline. This mechanism is not deprecated. Besides
+writing the document, it is also recommended to write a migration script so that the
+updates can be applied to user pipelines automatically, that is done by the
+[pipeline-migration-tool](https://github.com/konflux-ci/pipeline-migration-tool).
+
+### Create a migration
+
+The following is the steps to write a migration:
+
+- Bump task version. Modify label `app.kubernetes.io/version` in the task YAML file.
+- Ensure `migrations/` directory exists in the version-specific task directory.
+- Create a migration file under the `migrations/` directory. Its name is in
+  form `<new task version>.sh`. Note that the version must match the bumped
+  version.
+
+The migration file is a normal Bash script file:
+
+- It accepts a single argument. The pipeline file path is provided via this
+  argument. The script must work with a Tekton Pipeline by modifying the
+  pipeline definition under the `.spec` field. In practice, regardless of whether
+  the pipeline definition is embedded within the PipelineRun by `pipelineSpec` or
+  extracted into a separate YAML file, the migration tool ensures that the
+  passed-in pipeline file contains the correct pipeline definition.
+- All modifications to the pipeline must be done in-place, i.e. using `yq
+  -i` to operate the pipeline YAML.
+- It should be simple and small as much as possible.
+- It should be idempotent as much as possible to ensure that the changes are
+  not duplicated to the pipeline when run the migration multiple times.
+- Pass the `shellcheck` without customizing the default rules.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -215,3 +215,55 @@ Currently, there are three policy configurations.
...
+## Task Migration
+
+Task migrations are Bash scripts defined in version-specific task
+directories. In general, a migration consists of a series of `yq` commands that
+modify pipeline in order to work with the new version of task.
+
+Historically, task maintainers write `MIGRATION.md` to notify users what changes
+have to be made to the pipeline. This mechanism is not deprecated. Besides
+writing the file, it is also recommended to write a migration so that the
+updates can be applied to user pipelines automatically.
```
chmeliik, line 228: It would also be good to mention how the migration files get applied - i.e. link to the migration tool and the configuration that enables it in our Renovate deployment  (Mintmaker)

README.md
```diff
@@ -215,3 +215,55 @@ Currently, there are three policy configurations.
...
+
+## Task Migration
+
+Task migrations are Bash scripts defined in version-specific task
+directories. In general, a migration consists of a series of `yq` commands that
+modify pipeline in order to work with the new version of task.
+
+Historically, task maintainers write `MIGRATION.md` to notify users what changes
+have to be made to the pipeline. This mechanism is not deprecated. Besides
+writing the file, it is also recommended to write a migration so that the
```
chmeliik, line 227:
```suggestion
writing the document, it is also recommended to write a script so that the
```

README.md
```diff
@@ -215,3 +215,55 @@ Currently, there are three policy configurations.
...
+
+- Bump task version. Modify label `app.kubernetes.io/version` in the task YAML file.
+- Ensure `migrations/` directory exists in the version-specific task directory.
+- Create a migration file under the `migrations/` directory. Its name is in
+  form `<new task version>.sh`. Note that the version must match the bumped
+  version.
+
+The migration file is a normal Bash script file. It should
+
+- Accept one argument. The pipeline file path is passed to via this argument.
```
chmeliik, line 240: We should also mention that this pipeline file is always a tekton Pipeline file. Even if the pipeline is defined inline in a PipelineRun file, the migration tool extracts the Pipeline from it and calls the migration script on the Pipeline.

This also means it's not possible to apply changes that would have to be done on the PipelineRun, e.g. changing the `volumeClaimTemplate` for a workspace or things like that

README.md
```diff
@@ -215,3 +215,55 @@ Currently, there are three policy configurations.
...
+- All modifications to the pipeline must be done in-place, i.e. using `yq
+  -i` to operate the pipeline YAML.
+- Be idempotent as much as possible.
+- Be simple and small as much as possible.
+- Pass the `shellcheck` without customizing the default rules.
+
+Here are example script to create a migration for a task `task-a`:
```
chmeliik, line 247:
```suggestion
Here are example steps to create a migration for a task `task-a`:
```

README.md
```diff
@@ -215,3 +215,55 @@ Currently, there are three policy configurations.
...
+cat >task/task-a/0.2/migrations/0.2.2.sh <<EOF
+#!/usr/bin/env bash
+set -e
+pipeline_file=\$1
+
+if ! yq -e '.spec.tasks[] | select(.name == "task-a") | .params[] | select(.name == "pipelinerun-name")' >/dev/null
+then
+  yq -i -e '
+    (.spec.tasks[] | select(.name == "task-a") | .params) +=
+    {"name": "pipelinerun-name", "value": "\$(context.pipelineRun.name)"}
```
chmeliik, line 261: Should we make this example idempotent? (I.e. only add the param if it's not already there)

chmeliik:
We should probably keep this open until the migration tool is ready for use?

README.md
tkdchen: It's already idempotent. Line 257 checks that.

README.md
chmeliik: I was thinking we can just add a link to the migration tool, e.g.

> ...so that the updates can be applied to user pipelines automatically (by the [pipeline-migration-tool](https://github.com/konflux-ci/pipeline-migration-tool))

README.md
chmeliik: Ah ok, I missed that. Maybe add a short comment about that above the `if`? It's not immediately obvious that that's why the check is there

README.md
tkdchen: Updated.

chmeliik (approved):

mmorhun (approved):

README.md
tkdchen: Added a comment for it.

mmorhun (approved):

PR #1865, opened 2025-01-28
Switch the default SBOM type to SPDX

chmeliik:
Bump the versions of all SBOM-related tasks
* For SBOM producers (prefetch-dependencies, build tasks), this is to mark a breaking change - the "sbom type" parameter now defaults to SPDX
* For SBOM consumers (source-build, deprecated-image-check), this simply marks them as SPDX ready (mainly to make the MIGRATION.md files saner - checking if your task is the right version is much easier than checking if the `sha256` digest of your task is new enough)

To minimize disruption, users have to upgrade all the SBOM-related tasks at the same time. This is why these changes need to be done in a single PR, so that all the updates are built and released together and then proposed all at once by Renovate/Mintmaker

### DIFF
--- a/task/deprecated-image-check/0.5/deprecated-image-check.yaml
+++ b/task/deprecated-image-check/0.5/deprecated-image-check.yaml
@@ -1,0 +1,215 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.3"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "konflux"
+  name: deprecated-image-check
+spec:
+  description: >-
+    Identifies the unmaintained and potentially insecure deprecated base images.
+    Pyxis API collects metadata from image repository, and Conftest applies supplied policy to identify the deprecated images using that metadata.
+  params:
+    - name: POLICY_DIR
+      description: Path to directory containing Conftest policies.
+      default: "/project/repository/"
+    - name: POLICY_NAMESPACE
+      description: Namespace for Conftest policy.
+      default: "required_checks"
+    - name: BASE_IMAGES_DIGESTS
+      description: Digests of base build images.
+      default: ""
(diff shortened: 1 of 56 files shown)


chmeliik:
/ok-to-test

I sure hope e2e-tests will work with SPDX :pray:

chmeliik:
As reflected in the prefetch task's MIGRATION.md doc, some of the more niche build tasks are still not SPDX ready, mainly:
* rpm-ostree: https://github.com/konflux-ci/build-definitions/pull/1826
* build-maven-zip: https://github.com/konflux-ci/build-definitions/pull/1866

But I don't think we should let that block us, because
1. Their user base is small
2. The impact shouldn't be much worse than a failing Mintmaker PR when it bumps the prefetch task to `0.2` without a corresponding update to the build task

chmeliik:
Rebased and "re-boostrapped" new task versions

chmeliik:
rebased, re-bootstrapped

task/deprecated-image-check/0.5/deprecated-image-check.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.3"
+    app.kubernetes.io/version: "0.5"
```
jsztuka, line 6: I'm little confused here, the Migration MR says migration from 0.4 to 0.5 but this differs. Looks like it was forgotten to increment a version to 0.4. Jut to be aware of this fact.

jsztuka (approved):
lgtm

ralphbean (approved):

mmorhun (approved):

chmeliik:
Waiting with merging until we're sure that all consumers are SPDX ready (discussing internally on Slack)

chmeliik:
Re-based, re-bootstrapped (to pick up latest changes in the prefetch task)

mmorhun (approved):

PR #1866, opened 2025-01-28
build-maven-zip: support SPDX

chmeliik:
Detect the format of the SBOM coming from the prefetch task. Use the right '--type' argument for 'cosign attach' based on the format.


chmeliik:
@ligangty @yma96 Hi, we'll be switching the default SBOM format to SPDX by the end of January (in https://github.com/konflux-ci/build-definitions/pull/1865)

Users of the build-maven-zip task will get a PR updating the prefetch task to version 0.2. If the build-maven-zip task isn't SPDX ready, that PR may break and/or merging that PR may cause issues later on.

No rush, the fix is here when you're ready to review

yma96 (approved):
LGTM

PR #1875, opened 2025-01-29
paketo: support generating SPDX SBOMs

chmeliik:
@cmoulliard context for this:

We'll be switching the default SBOM type to SPDX by the end of the month (i.e. this week).

I believe that a typical paketo-builder pipeline doesn't enable dependency prefetching via the `prefetch-dependencies` task, so this switch probably shouldn't break the users of this task. Otherwise, the prefetch-dependencies task and the build task (in this case paketo-builder) would have to switch to SPDX at the same time.

TLDR - it would be good to get this merged fairly soon, but we won't consider this a blocker to switch the other tasks to SPDX

### DIFF
--- a/task/build-paketo-builder-oci-ta/0.1/README.md
+++ b/task/build-paketo-builder-oci-ta/0.1/README.md
@@ -17,6 +17,7 @@ The task also produces the SBOM which is signed and added to the image.
 | SOURCE_CODE_DIR      | The subpath of the application source code.                                         | "."           | true     |
 | STORAGE_DRIVER       | Storage driver to configure for buildah                                             | vfs           | false    |
 | TLSVERIFY            | Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)       | true          | false    |
+| SBOM_TYPE            | Select the SBOM format to generate. Valid values: spdx, cyclonedx.                  | cyclonedx     | false    |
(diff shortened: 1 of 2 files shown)


chmeliik:
This mostly copies the changes done to the buildah task in https://github.com/konflux-ci/build-definitions/pull/1825

cmoulliard:
I will test this PR soon

cmoulliard:
I did a test using this PR = task within the pipeline that we use to verify it and it fails for different reasons
```
ERROR: failed to create builder: fetch lifecycle: downloading lifecycle: Get "https://github.com/buildpacks/lifecycle/releases/download/v0.20.4/lifecycle-v0.20.4+linux.x86-64.tgz": dial tcp: lookup github.com on 10.207.4.2:53: dial udp 10.207.4.2:53: connect: network is unreachable

AND

Pull the base image: docker.io/paketocommunity/build-ubi-base:0.0.90 using unshare
unshare: failed to execute buildah: No such file or directory
```

cmoulliard (dismissed):

chmeliik:
Rebased on main and dropped the incorrectly added commits. Also dropped the commit updating the sbom-utility-scripts image reference, because the image has been updated in main in the meantime.

cmoulliard (approved):

PR #1885, opened 2025-01-31
Remove unnecessary tasks from the RHTAP build

lcarva:
This changes removes various Tasks that have been added to template-build and thus automatically added to the docker-build-rhtap pipeline.

NOTE: At this point, it is probably not worth using kustomize to generate the docker-build-rhtap pipeline -- only three of the 20 tasks are reused. It probably also makes sense to remove this pipeline and all other RHTAP specific tasks out of this repository.


lcarva:
@mmorhun, @chmeliik, @tnevrlka ptal.

chmeliik (approved):

PR #1886, opened 2025-02-01
use git resolver for fips task stepaction

yashvardhannanavati:
Because the stepaction doesn't exist in the workspaces, the task image fails to run. Use the git resolver to fetch the stepaction in the task.

Refers to [KFLUXSPRT-1555](https://issues.redhat.com//browse/KFLUXSPRT-1555)

### DIFF
--- a/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
+++ b/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
@@ -190,7 +190,14 @@ spec:
           cpu: 100m
           memory: 256Mi
       ref:
-        name: fips-operator-check-step-action
+        params:
+          - name: url
+            value: https://github.com/konflux-ci/build-definitions
+          - name: revision
+            value: c366291f484f2510767a51d0f45d7bc37877e94e
+          - name: pathInRepo
+            value: stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml
+        resolver: git
     - name: parse-images-processed-result
       image: quay.io/redhat-appstudio/konflux-test:v1.4.12@sha256:b42202199805420527c2552dea22b02ab0f051b79a4b69fbec9a77f8832e5623
       env:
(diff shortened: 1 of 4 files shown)


yashvardhannanavati:
@arewm @dirgim Could you please help review this?

jsztuka:
I'm just curious about the specifying git resolver under **ref**. Has this been tested locally?

jsztuka (approved):
lgtm

task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
```diff
@@ -190,7 +190,14 @@ spec:
           cpu: 100m
           memory: 256Mi
       ref:
-        name: fips-operator-check-step-action
+        params:
+          - name: url
+            value: https://github.com/konflux-ci/build-definitions
+          - name: revision
+            value: c366291f484f2510767a51d0f45d7bc37877e94e
```
arewm, line 197: What is the plan for keeping this reference up to date? Will it be automated? The commit is already not the latest one (not that it makes sense for it to be the latest one since this stepaction hasn't changed).

Should we instead take a generic approach similar to how we inject the task bundles to pipelines? https://github.com/konflux-ci/build-definitions/blob/main/hack/build-and-push.sh#L193-L211

yashvardhannanavati: For now, the revision would have to be updated manually whenever we change the stepaction.

I think it does make sense to automate the process as we have an increasing number of stepactions. May I suggest tracking it in a separate Jira?

yashvardhannanavati: filed KONFLUX-6629

arewm (approved):

yashvardhannanavati:
I don't think the failing CI check is related to the PR.

dirgim (approved):
LGTM

PR #1898, opened 2025-02-05
fix: Increase computeResources for SL go scanning and also enable scanning debug mode

flacatus:
We still meet killed issue in the sealights container. I see sl go scan need more than 500M memory for scanning


flacatus:
The failure of the Validate PR - kustomize manifests / Check Kustomize Build of Task and Pipelines test does not seem to be related to the change

ascerra (approved):

PR #1900, opened 2025-02-05
Re-generate kustomized files

chmeliik:
We are now checking that kustomize-generated files are up to date in CI.

Some pull requests that would have failed this check were submitted before the check existed and merged without rebasing. Re-generate kustomized files to make the check pass again.


dirgim (approved):
+1

PR #1903, opened 2025-02-06
feat: execute e2e-tests only when needed

tisutisu:
PR adds a logic to determine when the e2e-tests needs to be executed. E2e tests will be executed only for the pipelines/tasks/files changes for which coverage is there and for unrelated changes, it will skip e2e-tests execution task.
This is part of the story: https://issues.redhat.com/browse/STONEBLD-3161

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -69,15 +69,12 @@ spec:
             value:
               - tasks := strings.any_prefix_match(input, ["task/", "hack/", ".tekton/"])
               - tasks_pipelines := strings.any_prefix_match(input, ["task/", "pipelines/", "hack/", ".tekton/"])
-              - |
-                e2e_tests if {
-                  some file in input
-                  strings.any_prefix_match(file, ["task/", "pipelines/", "hack/", ".tekton/"])
-                  not endswith(file, "/OWNERS")
-                }
               - check_partner_tasks := strings.any_prefix_match(input, ["partners/", "hack/", ".tekton/"])
         runAfter:
           - build-appstudio-utils
+        workspaces:
+          - name: source
+            workspace: workspace
       - name: task-lint-check
         when:
           - input: "tasks"
@@ -208,9 +205,9 @@ spec:
             - name: source
       - name: e2e-tests
         when:
-          - input: "e2e_tests"
+          - input: "$(tasks.task-switchboard.results.run-e2e)"
             operator: "in"
-            values: ["$(tasks.task-switchboard.results.bindings[*])"]
+            values: ["execute_e2e"]
         params:
           - name: e2e_test_namespace
             value: $(params.e2e_test_namespace)
@@ -286,9 +283,9 @@ spec:
     finally:
       - name: e2e-cleanup
         when:
-          - input: "e2e_tests"
+          - input: "$(tasks.task-switchboard.results.run-e2e)"
             operator: "in"
-            values: ["$(tasks.task-switchboard.results.bindings[*])"]
+            values: ["execute_e2e"]
         params:
           - name: e2e_test_namespace
             value: $(params.e2e_test_namespace)
@@ -306,9 +303,9 @@ spec:
                 oc delete --ignore-not-found eventlisteners --all -n $(params.e2e_test_namespace)
       - name: pull-request-status-message
         when:
-          - input: "tasks_pipelines"
+          - input: "$(tasks.task-switchboard.results.run-e2e)"
             operator: "in"
-            values: ["$(tasks.task-switchboard.results.bindings[*])"]
+            values: ["execute_e2e"]
         taskRef:
           resolver: git
           params:
--- a/.tekton/scripts/determine-if-e2e-execution-needed.py
+++ b/.tekton/scripts/determine-if-e2e-execution-needed.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import json
+import urllib.error
+import urllib.request
+import subprocess
+
+# Pipelines currently covered by e2e tests
+pipelines_covered_by_e2e = ["docker-build", "docker-build-oci-ta", "docker-build-multi-platform-oci-ta", "fbc-builder"]
+
+# Task list which are covered by e2e tests, generated dynamically 
+tasks_covered_by_e2e = []
+
+# Otherthan tasks and pipelines, related files for which e2e tests needs to be executed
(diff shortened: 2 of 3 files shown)


.tekton/tasks/task-switchboard.yaml
```diff
@@ -53,3 +56,6 @@ spec:
         'data[_]' \
         | jq '[.result.[].expressions.[].value | to_entries | .[] | select(.value == true) | .key]' \
         | tee "$(results.bindings.path)"
+
+        # Output is "execute_e2e" or "dont_execute_e2e" based on files changed in pr
```
chmeliik, line 60: Technically the output isn't `dont_execute_e2e` but `No need to run e2e tests\ndont_execute_e2e`, because the script prints everything to stdout. It might be good to clean that up - always print to stderr except for the actual output of the script

.tekton/pull-request.yaml
```diff
@@ -306,9 +303,9 @@ spec:
                 oc delete --ignore-not-found eventlisteners --all -n $(params.e2e_test_namespace)
       - name: pull-request-status-message
         when:
-          - input: "tasks_pipelines"
+          - input: "execute_e2e"
             operator: "in"
-            values: ["$(tasks.task-switchboard.results.bindings[*])"]
+            values: ["$(tasks.task-switchboard.results.run-e2e)"]
```
chmeliik, line 308: nitpick: it would make more sense to me to flip the order here

```yaml
input: "$(tasks.task-switchboard.results.run-e2e)"
operator: in
values: ["execute_e2e"]
```

"result in expected values" reads more naturally that "expected value in list of results"

.tekton/scripts/determine-if-e2e-execution-needed.py
```diff
@@ -0,0 +1,95 @@
...
+def add_only_unique_task_names(task_list):
+    for task_name in task_list:
+        if task_name not in tasks_covered_by_e2e:
+            tasks_covered_by_e2e.append(task_name)
+    
+def get_tasks_covered_by_e2e():
+    for pipeline_name in pipelines_covered_by_e2e:
+        pipeline_path = f"pipelines/{pipeline_name}/{pipeline_name}.yaml"
+        # Get the task names from pipeline spec
+        result = subprocess.run([f"yq -e '.spec.tasks[].taskRef.name' {pipeline_path}"], shell=True, capture_output=True, text=True)
```
chmeliik, line 28:
```suggestion
        result = subprocess.run(["yq", "-e", ".spec.tasks[].taskRef.name", pipeline_path], capture_output=True, text=True)
```
No need for `shell=True` here. That's generally unsafe (prone to shell injection if one doesn't properly escape variables like the `pipeline_path` here)

.tekton/scripts/determine-if-e2e-execution-needed.py
```diff
@@ -0,0 +1,95 @@
...
+            print(f"[ERROR] failed to get tasks inside .spec.finally: {result.stderr}")
+            sys.exit(1)
+        output = result.stdout
+        task_names = output.split()
+        add_only_unique_task_names(task_names)
+
+def get_changed_files_from_pr(pull_number):
+    updated_files = []
+    token = os.environ["GITHUB_TOKEN"]
+    headers = {"Authorization": "Bearer " + token}
```
chmeliik, line 47: To make this script runnable locally, could we make the GITHUB_TOKEN optional?

chmeliik (commented):
The approach looks reasonable, some nitpicks

.tekton/pull-request.yaml
tisutisu: Changed now.

.tekton/scripts/determine-if-e2e-execution-needed.py
chmeliik: To clarify, I think taking the list of changed files from the github API is OK. But it doesn't actually require authentication:

```
curl https://api.github.com/repos/konflux-ci/build-definitions/pulls/1903/files
```

^ works without the auth header. So my suggestion was just to not add the header if GITHUB_TOKEN is not defined.

But reworking it to use `git diff` is OK too

tisutisu: @chmeliik  removed dependency on the GITHUB_TOKEN now. Please take a look again.

chmeliik (dismissed):

tisutisu:
@chmeliik `task-switchboard` task fails for the CI run executed when added to merge queue with below error
```
no pull requests found for branch "https://github.com/konflux-ci/build-definitions/pull/"
[]
[ERROR] got error response: b'{"message":"Not Found","documentation_url":"https://docs.github.com/rest","status":"404"}' with status 404
```
I think it is due to `{{ revision }}` is resolving to "", do you have any idea why that happens?

tisutisu:
> @chmeliik `task-switchboard` task fails for the CI run executed when added to merge queue with below error
> 
> ```
> no pull requests found for branch "https://github.com/konflux-ci/build-definitions/pull/"
> []
> [ERROR] got error response: b'{"message":"Not Found","documentation_url":"https://docs.github.com/rest","status":"404"}' with status 404
> ```
> 
> I think it is due to `{{ revision }}` is resolving to "", do you have any idea why that happens?

Got the issue, it is due to empty `pr_number` , which is handled with [this commit](https://github.com/konflux-ci/build-definitions/pull/1903/commits/12c9a73d38ba5726e81ae142e598ba6cf216b272).
there is a potential bigger problem, for which I filled [a story](https://issues.redhat.com/browse/STONEBLD-3270) now.

chmeliik:
> > @chmeliik `task-switchboard` task fails for the CI run executed when added to merge queue with below error
> > ```
> > no pull requests found for branch "https://github.com/konflux-ci/build-definitions/pull/"
> > []
> > [ERROR] got error response: b'{"message":"Not Found","documentation_url":"https://docs.github.com/rest","status":"404"}' with status 404
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think it is due to `{{ revision }}` is resolving to "", do you have any idea why that happens?
> 
> Got the issue, it is due to empty `pr_number` , which is handled with [this commit](https://github.com/konflux-ci/build-definitions/pull/1903/commits/12c9a73d38ba5726e81ae142e598ba6cf216b272). there is a potential bigger problem, for which I filled [a story](https://issues.redhat.com/browse/STONEBLD-3270) now.

Oof. Good catch, thanks for filling the issue :+1:

chmeliik (approved):

PR #1904, opened 2025-02-06
prefetch-dependencies: Include SBOM summary for RPM

slimreaper35:

### DIFF
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -131,6 +131,31 @@ spec:
         subscription-manager unregister || true
       }
 
+      function is_json
+      {
+        jq . 2>/dev/null 1>&2 <<< "$1"
+      }
+
+      # The input JSON can be in one of these forms:
+      # 1) '[{"type": "gomod"}, {"type": "bundler"}]'
+      # 2) '{"packages": [{"type": "gomod"}, {"type": "bundler"}]}'
+      # 3) '{"type": "gomod"}'
+      function input_json_has_rpm
+      {
+        jq '
+          if (type == "array" or type == "object") | not then
+            false
+          elif type == "array" then
+            any(.[]; .type == "rpm")
+          elif has("packages") | not then
+            .type == "rpm"
+          elif (.packages | type == "array") then
+            any(.packages[]; .type == "rpm")
+          else
+            false
+          end' <<< "$1"
+      }
+
       function inject_ssl_opts
       {
         input="$1"
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -171,38 +171,19 @@ spec:
                       }
                     }'
                   )"
-        elif (jq . 2>/dev/null 1>&2 <<< "$input"); then
-          # The input JSON can be in one of these forms:
-          # 1) '[{"type": "gomod"}, {"type": "bundler"}]'
-          # 2) '{"packages": [{"type": "gomod"}, {"type": "bundler"}]}'
-          # 3) '{"type": "gomod"}'
-          input_json_has_rpm="$(jq '
-                                  if (type == "array" or type == "object") | not then
-                                    false
-                                  elif type == "array" then
-                                    any(.[]; .type == "rpm")
-                                  elif has("packages") | not then
-                                    .type == "rpm"
-                                  elif (.packages | type == "array") then
-                                    any(.packages[]; .type == "rpm")
-                                  else
-                                    false
-                                  end' <<< "$input")"
-
+        elif is_json "$input" && input_json_has_rpm "$input"; then
           # The output JSON may need the SSL options updated for the RPM backend
-          if [ "$input_json_has_rpm" == "true" ]; then
-              input="$(jq \
-                        --argjson ssl "$ssl_options" '
-                          if type == "array" then
-                            map(if .type == "rpm" then .options.ssl += $ssl else . end)
-                          elif has("packages") then
-                            .packages |= map(if .type == "rpm" then .options.ssl += $ssl else . end)
-                          else
-                            .options.ssl += $ssl
-                          end' \
-                          <<< "$input"
-                      )"
-          fi
+          input="$(jq \
+                    --argjson ssl "$ssl_options" '
+                      if type == "array" then
+                        map(if .type == "rpm" then .options.ssl += $ssl else . end)
+                      elif has("packages") then
+                        .packages |= map(if .type == "rpm" then .options.ssl += $ssl else . end)
+                      else
+                        .options.ssl += $ssl
+                      end' \
+                      <<< "$input"
+                  )"
         fi
         echo "$input"
       }
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -188,6 +188,41 @@ spec:
         echo "$input"
       }
 
+      function inject_rpm_summary_flag
+      {
+        input="$1"
+
+        if [ "$input" == "rpm" ]; then
+          input="$(jq -n '{type: "rpm", include_summary_in_sbom: true}')"
+
+        elif is_json "$input" && input_json_has_rpm "$input"; then
+            input="$(jq '
+                      if type == "array" then
+                        map(
+                          if .type == "rpm" and (has("include_summary_in_sbom") | not) then
+                            .include_summary_in_sbom = true
+                          else
+                            .
+                          end)
+
+                      elif has("packages") then
+                        .packages |= map(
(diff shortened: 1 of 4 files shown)


task/prefetch-dependencies/0.2/prefetch-dependencies.yaml
```diff
@@ -169,11 +170,11 @@ spec:
               input="$(jq \
                         --argjson ssl "$ssl_options" '
                           if type == "array" then
-                            map(if .type == "rpm" then .options.ssl += $ssl else . end)
+                            map(if .type == "rpm" then .options.ssl += $ssl | .include_summary_in_sbom //= "true" else . end)
```
chmeliik, line 173: > Add `include_summary_in_sbom` flag to ensure RPM package summaries appear in SBOM output **if it's not set already**.

If the prefetch-input already sets the option, did you intend to respect its value? Currently, this will flip `false` to `true`

```bash
$ jq -n '{include_summary_in_sbom: false} | .include_summary_in_sbom //= true'
{
  "include_summary_in_sbom": true
}
```

slimreaper35: Hmm, seems like I wrote bad tests.

slimreaper35:
Second attempt, PTAL. I have only updated `task/prefetch-dependencies/0.1/prefetch-dependencies.yaml` file for now.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -188,6 +188,48 @@ spec:
...
+
+        elif is_json "$input" && input_json_has_rpm "$input"; then
+            input="$(jq '
+                      if type == "array" then
+                        map(if .type == "rpm" then
+                        if has("include_summary_in_sbom") | not then
+                          .include_summary_in_sbom = "true"
+                        else
+                        .
+                        end
```
eskultety, line 206: You don't need `then if has("include...'`, you can chain the clauses with `and`. Moreover, the indentation on the double nested branching inside `map` seems off and my brain doesn't parse it well :) . I think we can improve upon that (applies for the whole hunk). Attaching a diff for convenience:
```diff
diff --git a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
index 23be27a0..01329026 100644
--- a/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
+++ b/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
@@ -198,32 +198,24 @@ spec:
         elif is_json "$input" && input_json_has_rpm "$input"; then
             input="$(jq '
                       if type == "array" then
-                        map(if .type == "rpm" then
-                        if has("include_summary_in_sbom") | not then
-                          .include_summary_in_sbom = "true"
-                        else
-                        .
-                        end
-                      else
-                        .
-                      end)
-
+                        map(
+                          if .type == "rpm" and (has("include_summary_in_sbom") | not) then
+                            .include_summary_in_sbom = "true"
+                          else
+                            .
+                          end
+                        )
                       elif has("packages") then
-                        .packages |= map(if .type == "rpm" then
-                        if has("include_summary_in_sbom") | not then
-                          .include_summary_in_sbom = "true"
-                        else
-                        .
-                        end
-                      else
-                        .
-                      end)
-
+                        .packages |= map(
+                          if .type == "rpm" and (has("include_summary_in_sbom") | not) then
+                            .include_summary_in_sbom = "true"
+                          else
+                            .
+                          end
+                        )
                       else
                         if has("include_summary_in_sbom") | not then
                           .include_summary_in_sbom = "true"
-                        else
-                          .
                         end
                       end' <<< "$input")"
         fi
```

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -188,6 +188,48 @@ spec:
...
+                      else
+                        .
+                      end)
+
+                      else
+                        if has("include_summary_in_sbom") | not then
+                          .include_summary_in_sbom = "true"
+                        else
+                          .
```
eskultety, line 226: You shouldn't need this. The 'else .' branch is only strictly needed for map which iterates over all elements in an array and if you don't define `else .` then it'll return `null` for objects that didn't match `map`'s own condition. Here we have a single object.

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -188,6 +188,48 @@ spec:
         echo "$input"
       }
 
+      function inject_rpm_summary_flag
+      {
+        input="$1"
+
+        if [ "$input" == "rpm" ]; then
+          input="$(jq -n '{type: "rpm", include_summary_in_sbom: "true"}')"
```
eskultety, line 196: JSON supports booleans, so we should probably format boolean fields as such instead of strings.

eskultety (commented):
I wish it didn't come to this and the task wouldn't become a massive `jq` black magic ++ , but well, what can you do. I guess we could think of a way of parametrizing the map conditions and corresponding actions and pass them as function arguments, but that's a topic for a different PR.
ACK with the minor comments addressed.

slimreaper35:
Ok, I still need to copy the changes to other prefetch tasks

eskultety:
> Ok, I still need to copy the changes to other prefetch tasks

Which ones? You only need this in the OCI counterpart, am I mistaken?

brunoapimentel:
Great job with the jq magic, but it makes me wonder if we should've added the summary toggle as a config option, so that it would be more easily configurable from server side. 

(This is a non-blocking comment, it is just something for us to keep in mind if something like this comes up again.)

eskultety:
> Great job with the jq magic, but it makes me wonder if we should've added the summary toggle as a config option, so that it would be more easily configurable from server side.

Definitely beyond the scope of this PR, however, we should track the idea as an issue in our repo. That said, I'd personally start with implementing this as an environment variable for 2 reasons:

- our config is embarrassingly simple and needs rework in order not to cause any more mess
- it's easier to consume env variables in container world, so the work should be related to: https://github.com/hermetoproject/cachi2/issues/573

task/prefetch-dependencies/0.1/prefetch-dependencies.yaml
```diff
@@ -188,6 +188,41 @@ spec:
...
+        input="$1"
+
+        if [ "$input" == "rpm" ]; then
+          input="$(jq -n '{type: "rpm", include_summary_in_sbom: true}')"
+
+        elif is_json "$input" && input_json_has_rpm "$input"; then
+            input="$(jq '
+                      if type == "array" then
+                        map(
+                          if .type == "rpm" and has("include_summary_in_sbom") | not then
```
eskultety, line 202: Have you tried this? ;)

I don't think this works as expected and I'm saying that because before submitting my original review comments I fell into the same trap of not using a set of parentheses around `has("include_summary_in_sbom") | not` . The pipe `|` operator has a smaller precedence than `and` which means that the following will happen:
```
# jq_test.sh contains the function code from above...
$ jq_test.sh '{"packages":[{"type":"rpm"}, {"type":"gomod", "foo":"bar"}]}'
{
  "packages": [
    {
      "type": "rpm",
      "include_summary_in_sbom": "true"
    },
    {
      "type": "gomod",
      "foo": "bar",
      "include_summary_in_sbom": "true"
    }
  ]
}
```

slimreaper35: Next time, I should copy the diff :) 
I only replaced "true" -> true based on the https://github.com/konflux-ci/build-definitions/pull/1904#discussion_r1948697934

slimreaper35:
> Which ones? You only need this in the OCI counterpart, am I mistaken?

IIRC, prefetch-task 0.1, 0.2 + prefetch-task-oci 0.1, 0.2

eskultety:
> > Which ones? You only need this in the OCI counterpart, am I mistaken?
> 
> IIRC, prefetch-task 0.1, 0.2 + prefetch-task-oci 0.1, 0.2

Oh, I forgot there's a new version due to SPDX!

eskultety (dismissed):
With the stylistic GH actions issue fixed -> ACK, though I don't remember having been bullied by this the last time I worked on the code, hmm, is it a new check or have I just not noticed it before?

slimreaper35:
> With the stylistic GH actions issue fixed, though I don't remember having been bullied by this the last time I worked on the code, hmm, is it a new check or have I just not noticed it before?

It seems like it only runs on `prefetch-dependencies-oci-ta`

eskultety (approved):

chmeliik (approved):

chmeliik:
@slimreaper35 you may need to rebase on main, the failing ec-checks were fixed in https://github.com/konflux-ci/build-definitions/commit/e25a938262983c07a456699016d020d33968730d

chmeliik:
Rebased again, the temporary fix for this e2e failure should be merged in main as well

```
Expected
    : 
to equal
    : enterprise-contract
```

MartinBasti:
https://github.com/konflux-ci/build-definitions/pull/1935

PR #1905, opened 2025-02-06
build-and-push: store all task bundle references

chmeliik:
The push pipeline runs the `build-and-push` script and the `build-acceptable-bundles` script (which takes `OUTPUT_TASK_BUNDLE_LIST` as input). It may happen that the `build-and-push` script succeeds while the `build-acceptable-bundles` script fails, resulting in a situation where new task bundles have been released but are not included in the `data-acceptable-bundles` image.

Previously, rerunning the push pipeline wouldn't help in those situtations. The `build-and-push` script doesn't re-push task bundles if the expected revision already exists in the registry and didn't use to record references of existing bundles in the `OUTPUT_TASK_BUNDLE_LIST`.

Always record the references of existing bundles to make it possible to fix the problem simply by rerunning the push pipeline.

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -478,6 +478,7 @@ build_push_tasks() {
         if [ -n "$digest" ]; then
             task_bundle_with_digest=${task_bundle}@${digest}
             echo "info: use existing $task_bundle_with_digest" 1>&2
+            echo "$task_bundle_with_digest" >> "$OUTPUT_TASK_BUNDLE_LIST"
         else
             echo "info: push new bundle $task_bundle" 1>&2
 


chmeliik:
@konflux-ci/ec the `ec track bundle` command will be able to handle this fine, right? https://github.com/konflux-ci/build-definitions/blob/f295a677727133a712fa4f3f6e95ca02b771345e/.tekton/scripts/build-acceptable-bundles.sh#L57-L63

("this" being `--bundle=$bundle_ref` params where the bundle refs are already present in the data-acceptable-bundles image)

lcarva:
@chmeliik, the `--freshen` parameter for the `ec track bundle` command will query the registry for anything that has been missed.

chmeliik:
> @konflux-ci/ec the `ec track bundle` command will be able to handle this fine, right?
> 
> https://github.com/konflux-ci/build-definitions/blob/f295a677727133a712fa4f3f6e95ca02b771345e/.tekton/scripts/build-acceptable-bundles.sh#L57-L63
> 
> ("this" being `--bundle=$bundle_ref` params where the bundle refs are already present in the data-acceptable-bundles image)

Per offline discussion, it may not be a good idea. When someone comments `/retest` on an old commit, this could result in EC treating an old $bundle_ref as the latest

PR #1906, opened 2025-02-06
deprecated-image-check/0.5: arbitrary change

chmeliik:
Needed to add the task to the data-acceptable-bundles image.

When merging https://github.com/konflux-ci/build-definitions/pull/1865, the build-acceptable-bundles failed. This meant the new versions of all tasks updated in that PR were missing from data-acceptable-bundles.

Due to the way the push pipeline is set up, making a change to the affected tasks is the only way to add them to data-acceptable-bundles.


dirgim (approved):
/lgtm

PR #1909, opened 2025-02-06
paketo/0.2: default to SPDX

chmeliik:
Switch the default SBOM format to SPDX.

Bump the version label, add the MIGRATION.md file.


chmeliik:
PTAL, to align with the rest of Konflux tasks we should switch the default SBOM format to SPDX

cmoulliard (approved):

PR #1912, opened 2025-02-06
Exclude step images

joejstuart:
Two images require auth and another two are parametized and EC is unable to handle these two cases.

https://issues.redhat.com/browse/EC-1038


lcarva (approved):

lcarva:
I think this needs an approval from @konflux-ci/buildah-maintainers

chmeliik (approved):

PR #1917, opened 2025-02-07
Deprecate old buildah tasks

chmeliik:
Deprecate buildah/0.2 on 1st March 2025

Deprecate buildah/0.3 on 31st March 2025

This mainly to "motivate" users to update to the version that switched the default SBOM type to SPDX

Also update the sast-coverity-check task to be based on buildah 0.4 rather than 0.3, to avoid inheriting the deprecation annotation. And fix the version label in the task.

### DIFF
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -5,6 +5,7 @@ metadata:
     app.kubernetes.io/version: "0.2.1"
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
+    build.appstudio.redhat.com/expires-on: "2025-03-01T00:00:00Z"
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "image-build, konflux"
   name: buildah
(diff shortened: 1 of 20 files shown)


task/buildah/0.2/buildah.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     app.kubernetes.io/version: "0.2.1"
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
+    build.appstudio.redhat.com/expires-on: "2025-03-01T00:00:00Z"
```
kdudka, line 8: Should we also define `expiry-message` as in the following annotation? https://github.com/konflux-ci/build-definitions/blob/54c1b06f77a1c8eba1baae789a810f60cd10d343/task/buildah-10gb/0.2/buildah-10gb.yaml#L5-L8

kdudka (approved):

task/buildah/0.2/buildah.yaml
chmeliik: The default message is ["Upgrade to a newer version of the Task."](https://github.com/enterprise-contract/ec-policies/blob/ce1475d3ab81a7ace697aee65e1e3bd0af895968/policy/release/tasks/tasks.rego#L265), which should work well enough here

tnevrlka (approved):

PR #1919, opened 2025-02-07
feat(): change go-version param in the sealights go instrumentation task to use by default latest version of ubi go-toolset

flacatus:


mmorhun (approved):
LGTM

ascerra (approved):

PR #1926, opened 2025-02-10
coverity-availability-check-oci-ta/0.2: fix name, improve migration doc

chmeliik:
Previously, coverity-availability-check-oci-ta.yaml was just a symlink to coverity-availability-check.yaml. Which means .metadata.name was 'coverity-availability-check'. This caused pipelines referencing coverity-availability-check-oci-ta:0.2 to fail with

    Couldn't retrieve Task ...
    could not find object in image with kind: task and name: coverity-availability-check-oci-ta

Generate coverity-availability-check-oci-ta.yaml from coverity-availability-check.yaml via kustomize, set the .metadata.name to the expected name.

Add kustomization.yaml to coverity-availability-check as well, otherwise coverity-availability-check-oci-ta wouldn't be able to use it as a base.

Explain how to switch from coverity-availability-check-oci-ta to
coverity-availability-check

### DIFF
--- a/policies/all-tasks.yaml
+++ b/policies/all-tasks.yaml
@@ -32,6 +32,7 @@ sources:
       exclude:
         # https://issues.redhat.com/browse/EC-1038
         - step_images.step_images_accessible:quay.io/redhat-services-prod/sast/coverity:202412.2
+        - step_images.step_images_accessible:registry.access.redhat.com/ubi8/nodejs-$(params.nodejs-version):latest
         - step_images.step_images_accessible:registry.access.redhat.com/ubi8/python-$(params.python-version):latest
         - step_images.step_images_accessible:registry.access.redhat.com/ubi8/go-toolset:$(params.go-version)
         - step_images.step_images_accessible:quay.io/redhat-services-prod/sast/coverity@sha256:0d1b96fb08a901b2d0e340599c7fee7e1de25e2d6ba58f3d95db4983f32b5a3c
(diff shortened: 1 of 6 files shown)


chmeliik:
Note: the `coverity-availability-check-oci-ta.yaml` diff in the second commit is just due to the way kustomize does formatting. The only real difference is the `.metadata.name`

dirgim (dismissed):
LGTM

task/coverity-availability-check/0.2/coverity-availability-check.yaml
```diff
@@ -1,88 +1,88 @@
...
+  annotations:
+    tekton.dev/pipelines.minVersion: 0.12.1
+    tekton.dev/tags: konflux
   labels:
     app.kubernetes.io/version: "0.1"
-  annotations:
-    tekton.dev/pipelines.minVersion: "0.12.1"
-    tekton.dev/tags: "konflux"
-  name: coverity-availability-check
+  name: coverity-availability-check-oci-ta
```
arewm, line 9: Hm. Shouldn't this be _without_ the `oci-ta`?

chmeliik: Oof, good catch! The output of kustomize went into the symlinked file. Fixed

dirgim (dismissed):
LGTM

kasemAlem (approved):

chmeliik:
The EC failure was not intermittent. Added a policy exception similar to those added in https://github.com/konflux-ci/build-definitions/pull/1912

dirgim (approved):
+1

PR #1918, opened 2025-02-07
Upgrade kubectl in appstudio-utils to v1.26

amisstea:
This should resolve a race condition seen in
`stepactions/eaas-create-ephemeral-cluster-hypershift-aws` when waiting for resource status which doesn't yet exist.

See https://github.com/kubernetes/kubectl/issues/1236

PR #1929, opened 2025-02-10
Upgrade appstudio-utils image in EaaS stepactions

amisstea:
This is a followup to #1918.

Commit tags are removed from the pull specs so renovate can identify newer versions.


yftacherzog (approved):

avi-biton (approved):
/lgtm

PR #1930, opened 2025-02-10
Fail acs-image-check task when image check fails

tnevrlka:
Failure during roxctl-image-check currently does not cause the
acs-image-check task to fail, resulting in a false negative.
Make rox-image-check step exit with non-zero code in case of a failure,
but continue with rest of the TaskRun in order to create a successful
report in the next report step.
Add a new step which fails if rox-image-check step failed so the whole
TaskRun ends with a failure

### DIFF
--- a/task/acs-image-check/0.1/acs-image-check.yaml
+++ b/task/acs-image-check/0.1/acs-image-check.yaml
@@ -66,11 +66,12 @@ spec:
           value: $(params.image)
         - name: PARAM_IMAGE_DIGEST
           value: $(params.image-digest)
+      onError: continue
       script: |
         #!/usr/bin/env bash
         set +x
 
-        # Check if rox API enpoint is configured
+        # Check if rox API endpoint is configured
         if test -f /rox-secret/rox-api-endpoint ; then
           export ROX_CENTRAL_ENDPOINT=$(</rox-secret/rox-api-endpoint)
         else
@@ -104,7 +105,7 @@ spec:
         fi
         received_filesize=$(stat -c%s ./roxctl)
         if (( $received_filesize < 10000 )); then
-          # Responce from ACS server is not a binary but error message
+          # Response from ACS server is not a binary but error message
           cat ./roxctl
           echo 'Failed to download roxctl'
           exit 2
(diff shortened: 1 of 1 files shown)


task/acs-image-check/0.1/acs-image-check.yaml
```diff
@@ -113,11 +113,15 @@ spec:
...
         echo "roxctl image check"
         IMAGE=${PARAM_IMAGE}@${PARAM_IMAGE_DIGEST}
-        ./roxctl image check \
+        if ! ./roxctl image check \
           $( [ "${PARAM_INSECURE_SKIP_TLS_VERIFY}" = "true" ] && \
           echo -n "--insecure-skip-tls-verify") \
           -e "${ROX_CENTRAL_ENDPOINT}" --image "$IMAGE" --output json --force \
           > roxctl_image_check_output.json
+        then
+          echo "roxctl image check failed, see the errors above"
```
chmeliik, line 122: Does roxctl print the errors to both stderr and stdout? Because we're redirecting stdout to a file

chmeliik: Also - what is "errors" in this context? Linter warnings etc.? Or some fatal failure in roxctl?

tnevrlka: Thanks for the review, I have updated the PR. 
- `roxctl` prints out errors to stderr, they are supposed to be displayed in the logs of the `rox-image-check` step, the file (stdout of roxctl) is supposed to be printed out in the `report` step
- I'm not exactly sure about what constitutes an error, that's handled by the `roxctl` authors and what they decide is serious enough to exit with a non-zero exit code

task/acs-image-check/0.1/acs-image-check.yaml
```diff
@@ -128,3 +135,17 @@ spec:
       script: |
         #!/usr/bin/env bash
         cat /steps-shared-folder/acs-image-check.json
+
+    - name: fail-if-rox-image-check-failed
+      image: registry.access.redhat.com/ubi8-minimal@sha256:d16d4445b1567f29449fba3b6d2bc37db467dc3067d33e940477e55aecdf6e8e
+      script: |
+        #!/usr/bin/env bash
+        ROX_IMAGE_CHECK_EXIT_CODE_PATH=$(steps.step-rox-image-check.exitCode.path)
```
chmeliik, line 143: Shouldn't there be something like `onError: continue` on the rox-image-check step?

chmeliik (approved):

PR #1935, opened 2025-02-11
Fix jq syntax error in prefetch-dependencies tasks

tnevrlka:
prefetch-dependencies has been changed recently and is currently failing on a jq syntax error

Allowing if without else-branch in jq seems to not be implemented in jq 1.6, which is used in the cachi2 image used in the prefetch-dependencies tasks

Add an else branch with '.' (identity) to fix the syntax error

This has been tested manually in the [base image used in the cachi2 image](https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/Dockerfile#L1)
image


arewm (approved):
LGTM, thanks!

MartinBasti (approved):

eskultety:
Ouch. I really didn't anticipate this kind of logical functionality to be a "recent" addition when I made the suggestion in the original PR, my bad, thanks for fixing this.

PR #1850, opened 2025-01-22
task: prefetch: Condense the RHSM entitlement RPM repositories logic to a single step

eskultety:
~**Depends on:**  https://github.com/containerbuildsystem/cachi2/pull/792~

Opening as an RFC draft since this solution currently relies on a custom-built cachi2 image which the task was tested with in a private Konflux deployment. Once the dependent PR is merged AND this solution is accepted I will lift the Draft status. Pasting a (publishable) snippet from the pipelines log:
```
prefetch
=======
Registering with Red Hat subscription manager.
The system has been registered with ID: cc44f731-1d42-44f2-9993-56454f63faa4
The registered system name is: konflux-test-repo-on-pull-rdc2a0834b9fb80d8cdb850c8f3586779-pod
2025-01-22 17:02:50,177 INFO Fetching the gomod dependencies at subpath hello
2025-01-22 17:02:50,186 INFO go.mod reported versions: '1.20'[go], '-'[toolchain]
2025-01-22 17:02:50,187 INFO Using Go release: go1.20
2025-01-22 17:02:50,195 INFO Downloading the gomod dependencies
2025-01-22 17:02:52,891 INFO Retrieving the list of packages
2025-01-22 17:02:53,195 INFO Reading RPM lockfile: /var/workdir/source/rpms.lock.yaml
2025-01-22 17:02:53,197 INFO Downloading files for 'x86_64' architecture.
2025-01-22 17:02:53,203 INFO Using client certificate auth.
2025-01-22 17:02:58,373 INFO All dependencies fetched successfully \o/
2025-01-22 17:02:59,454 INFO Creating repository metadata for repoid 'appstream': /var/workdir/cachi2/output/deps/rpm/x86_64/appstream
2025-01-22 17:02:59,520 INFO Creating /var/workdir/cachi2/output/deps/rpm/x86_64/repos.d/cachi2.repo
Unregistering from: subscription.rhsm.redhat.com:443/subscription
System has been unregistered.

build
====
...
[2/2] STEP 7/8: CMD ["/usr/local/bin/hello"]
[2/2] STEP 8/8: LABEL "build-date"="2025-01-22T17:03:57" "architecture"="x86_64" "vcs-type"="git" "vcs-ref"="a0c881286921d0ae047b932d3dc1a7645394c122" "quay.expires-after"="5d"
[2/2] COMMIT quay.io/<path>/konflux-test-repo:on-pr-a0c881286921d0ae047b932d3dc1a7645394c122
--> 60f0fc490100
[...]

PR #1938, opened 2025-02-12
prefetch-dependencies: Fix RPM check

slimreaper35:

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
@@ -240,7 +240,7 @@ spec:
                         }
                       }'
             )"
-          elif is_json "$input" && input_json_has_rpm "$input"; then
+          elif is_json "$input" && [[ $(input_json_has_rpm "$input") == true ]]; then
             # The output JSON may need the SSL options updated for the RPM backend
             input="$(
               jq \
@@ -264,7 +264,7 @@ spec:
           if [ "$input" == "rpm" ]; then
             input="$(jq -n '{type: "rpm", include_summary_in_sbom: true}')"
 
-          elif is_json "$input" && input_json_has_rpm "$input"; then
+          elif is_json "$input" && [[ $(input_json_has_rpm "$input") == true ]]; then
             input="$(jq '
                         if type == "array" then
                           map(
--- a/task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
@@ -240,7 +240,7 @@ spec:
                         }
                       }'
             )"
-          elif is_json "$input" && input_json_has_rpm "$input"; then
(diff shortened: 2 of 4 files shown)


chmeliik (approved):

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -211,7 +211,7 @@ spec:
         # 2) '{"packages": [{"type": "gomod"}, {"type": "bundler"}]}'
         # 3) '{"type": "gomod"}'
         function input_json_has_rpm {
-          jq '
+          jq -e '
```
eskultety, line 214: Not using it was a deliberate decision based on the following comment: https://github.com/konflux-ci/build-definitions/pull/1850#discussion_r1928624779 . That same PR introduced `set -e` to the bash exec options, so returning 1 means you'd fail the script.

chmeliik: They way `input_json_has_rpm` function is currently used (in an `if` condition), it does need to return 1 to work properly.

An alternative would be to switch from `if input_json_has_rpm "$INPUT"` to `if [[ $(input_json_has_rpm "$INPUT") == true ]]`

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
```diff
@@ -222,7 +222,7 @@ spec:
               any(.packages[]; .type == "rpm")
             else
               false
-            end' <<<"$1"
+            end' <<<"$1" 2>/dev/null 1>&2
```
eskultety, line 225: I don't think this is desirable for one reason - we're sending any valuable output to the sink, so if an error occurs (see https://github.com/konflux-ci/build-definitions/pull/1935), we wouldn't see anything in the logs. Errors in logs are paramount to investigation.
Note that the choice of redirecting the output in `function is_json` is also deliberate because we don't care about any errors, the sole purpose of that function is to figure out if input is JSON or not, no more semantics needed.

eskultety (commented):
I'm afraid it's a NACK for this PR unless I'm utterly wrong in any of my comments I made.

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
chmeliik: Good point, let's redirect only stdout not stderr (or not use `-e` and change the `if` condition)

task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml
eskultety: > They way the `input_json_has_rpm` function is currently used (in an `if` condition), it does need to return 1 to work properly.
> 
> An alternative would be to switch from `if input_json_has_rpm "$INPUT"` to `if [[ $(input_json_has_rpm "$INPUT") == true ]]`

Hmm, I stand corrected then.

eskultety (approved):

chmeliik (approved):

PR #1950, opened 2025-02-17
sast-shell-check: Mount certificates to avoid problems while cloning repo

jperezdealgaba:
Resolves: https://issues.redhat.com/browse/PSSECAUT-991
Related: https://issues.redhat.com/browse/OSH-738

Mount certificates to avoid problems with the certificates if the task is executed from external cluster. This prevented the sast-shell-check task to be executed successfully.


arewm (approved):
Simple enough fix!

jperezdealgaba:
As a summary, when this task was originally implemented, the certficates were never mounted. As we didn't clone the repo because the `KFP_REPO_URL` was empty, this was never tested.
After this commit: https://github.com/konflux-ci/build-definitions/commit/ea6b164caa438f28d5d42c237dbc9147f2b2ec4c , the problem surfaced as it tried to clone the repo

Josh-Everett (approved):

PR #1952, opened 2025-02-18
fix: CI execution in the merge queue

tisutisu:
Comparing with main branch of the origin to find changed files in the PR, when running in the merge queue it will help detecting the changes and execute CI accordingly.

This is part of the story: https://issues.redhat.com/browse/STONEBLD-3270

### DIFF
--- a/.tekton/tasks/task-switchboard.yaml
+++ b/.tekton/tasks/task-switchboard.yaml
@@ -54,7 +54,7 @@ spec:
         if [[ -z "${pr_number}" ]]; then
           # Fetch the main branch for comparison
           git fetch origin main
-          changed_files=$(git diff --name-only "origin/main..HEAD" | tr "\n" " ")
+          changed_files=$(git diff --name-only "origin/main...HEAD" | tr "\n" " ")
           echo "Changed files in PR: ${changed_files}"
           echo "${changed_files}" | jq -Rs 'rtrimstr("\n") | split(" ")' > input.json
           ec opa eval --v1-compatible --data "${rules}" --input input.json 'data[_]' \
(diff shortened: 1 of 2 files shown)


.tekton/tasks/task-switchboard.yaml
```diff
@@ -51,13 +51,28 @@ spec:
...
 
-        ec opa eval --v1-compatible --data "${rules}" --input \
-        <(gh pr view "https://github.com/konflux-ci/build-definitions/pull/${pr_number}" --json files --jq '[.files.[].path']) \
-        'data[_]' \
-        | jq '[.result.[].expressions.[].value | to_entries | .[] | select(.value == true) | .key]' \
-        | tee "$(results.bindings.path)"
+        if [[ -z "${pr_number}" ]]; then
+          # Fetch the main branch for comparison
+          git fetch origin main
+          changed_files=$(git diff --name-only "origin/main..HEAD" | tr "\n" " ")
```
chmeliik, line 57: Situation: my PR doesn't update any files that should trigger e2e-tests, but it is outdated compared to main. In main, some files that trigger e2e-tests have been updated. With this approach, we would trigger e2e-tests because the relevant files changed in main.

I think we need the three-dot diff
```suggestion
          changed_files=$(git diff --name-only "origin/main...HEAD" | tr "\n" " ")
```
```
       git diff [<options>] <commit>...<commit> [--] [<path>...]
           This form is to view the changes on the branch containing and up to the second <commit>, starting at a common ancestor of both <commit>.  git diff A...B is
           equivalent to git diff $(git merge-base A B) B. You can omit any one of <commit>, which has the same effect as using HEAD instead.
```

chmeliik (approved):

PR #1953, opened 2025-02-18
Improve the performance and validation of validate-fbc task

arewm:
## Reduce the resources required for the validate-fbc task

The computation of differential related images has been improved by removing jq operations which require large arrays to be loaded.

related: https://github.com/konflux-ci/konflux-test/pull/364

## Improve validation handling for FBC

* Check to make sure that a valid base image tag is found
* Properly catch failure to retrieve differential base iamges

resolves: [KONFLUX-6943](https://issues.redhat.com//browse/KONFLUX-6943)

### DIFF
--- a/task/validate-fbc/0.1/validate-fbc.yaml
+++ b/task/validate-fbc/0.1/validate-fbc.yaml
@@ -63,7 +63,7 @@ spec:
         name: workdir
   steps:
     - name: inspect-image
-      image: quay.io/konflux-ci/konflux-test:v1.4.15@sha256:23a7e8c745a037da283f18f5e5108598efeaeb5263d599411b6c4a25c42c6f94
+      image: quay.io/arewm/konflux-test:latest@sha256:ca4124c55acdff77804bcfd2e1ad1c726615eb087d30ec27ec7b1190e390c212
       # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
       # the cluster will set imagePullPolicy to IfNotPresent
       workingDir: /var/workdir/inspect-image
@@ -205,7 +205,7 @@ spec:
             exit 0
         fi
     - name: extract-and-validate
-      image: quay.io/konflux-ci/konflux-test:v1.4.15@sha256:23a7e8c745a037da283f18f5e5108598efeaeb5263d599411b6c4a25c42c6f94
+      image: quay.io/arewm/konflux-test:latest@sha256:ca4124c55acdff77804bcfd2e1ad1c726615eb087d30ec27ec7b1190e390c212
       # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
       # the cluster will set imagePullPolicy to IfNotPresent
       workingDir: /var/workdir/extract-and-validate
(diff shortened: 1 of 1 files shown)


.tekton/pull-request.yaml
```diff
@@ -237,18 +237,19 @@ spec:
...
-            operator: "in"
-            values: ["$(tasks.task-switchboard.results.bindings[*])"]
-        runAfter:
-          - task-switchboard
-        taskRef:
-          name: ec-checks
-        workspaces:
-          - name: source
-            workspace: workspace
+      # TODO: revert this
```
yashvardhannanavati, line 240: Probably have this in a separate commit so it's easier to revert?

yashvardhannanavati (commented):
One minor comment. lgtm code wise!

arewm:
@konflux-ci/integration-service-maintainers @konflux-ci/operator-foundry , CI has now passed on this task update. Would you be able to review? We can work to further reduce the memory requirements in the future.

Do you want me to try to reduce the memory requirements as they are? I suspect that some FBC components might need more memory than others due to the size of their generated fragments and the number of related images (this will therefore also increase as the age of a catalog increases as the number of package versions and therefore the number of related images will increase).

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -216,9 +216,9 @@ spec:
             - SETFCAP
       computeResources:
         limits:
-          memory: 4Gi
+          memory: 6Gi
```
yashvardhannanavati, line 219: Could we please set the ratio of memory request and limit to 1:1?

Setting it different can be detrimental to other tasks.

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -431,29 +440,33 @@ spec:
         # Get the new related images that are being added so that we can check them for validity
         echo "Rendering target index and finding unreleased related images."
         status=0
-        get_unreleased_fbc_related_images -i "$image_with_digest" > /shared/related-images.json || status=$?
+        related_images=$(get_unreleased_fbc_related_images -i "$image_with_digest") || status=$?
         if [ $status -ne 0 ]; then
+          echo $status
```
yashvardhannanavati, line 445: should this be removed? or accompanied with some more context?

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -216,9 +216,9 @@ spec:
             - SETFCAP
       computeResources:
         limits:
-          memory: 8Gi
+          memory: 4Gi
```
yashvardhannanavati: same as previous comment. can we set the ratio to 1:1?

yashvardhannanavati (changes requested):

task/validate-fbc/0.1/validate-fbc.yaml
arewm: Sure, I can update that. We didn't have that ratio before.

yashvardhannanavati: yeah, ideally we should have that ratio set to 1:1 in all tasks. well, one battle at a time

yashvardhannanavati (approved):

PR #1956, opened 2025-02-19
Add test-sast-snyk-check-no-token.yaml

sfowl:
Because we can't do a real run of snyk in tests due to the secret token requirement, this test just ensures the snyk check is skipped in the expected fashion. A test source code repository is not cloned for this test, as it would not change the behaviour of the skipped snyk check task.

Supports [PSSECAUT-954](https://issues.redhat.com//browse/PSSECAUT-954)

### DIFF
--- a/task/sast-snyk-check/0.3/tests/test-sast-snyk-check-no-token.yaml
+++ b/task/sast-snyk-check/0.3/tests/test-sast-snyk-check-no-token.yaml
@@ -1,0 +1,44 @@
+---
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
+  name: test-sast-snyk-check-no-token
+spec:
+  description: |
+    Test the sast-snyk-check task is skipped when no token is available
+  workspaces:
+    - name: tests-workspace
+  tasks:
+    - name: scan-with-snyk
+      workspaces:
+        - name: workspace
+          workspace: tests-workspace
+      taskRef:
+        name: sast-snyk-check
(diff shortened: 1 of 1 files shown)


task/sast-snyk-check/0.3/tests/test-sast-snyk-check-no-token.yaml
```diff
@@ -0,0 +1,48 @@
...
+              value: $(params.results)
+            script: |
+              #!/bin/bash
+
+              set -e
+
+              echo "Checking result:"
+              echo "$RESULTS"
+              # snyk check should skip because snyk_token is unavailable in ephemeral clusters
+              echo "$RESULTS" | jq -e '.result == "SKIPPED"'
```
jperezdealgaba, line 44: Do we want to show a list of `TRUE` values without the user knowing what they are for?

kdudka (approved):
Looks good!  Please squash the fix-up commit before merging this to avoid unnecessary changes back and forth in the upstream git history.

task/sast-snyk-check/0.3/tests/test-sast-snyk-check-no-token.yaml
sfowl: Fair enough. I consolidated these into a single `Expected result: true` message.

kdudka (approved):

dirgim (approved):
LGTM

PR #1957, opened 2025-02-19
Refine annotations

tkdchen:
The task bundles and migration artifacts are now annotated specifically.

As promised in https://github.com/konflux-ci/build-definitions/pull/1861#pullrequestreview-2621391050

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -40,7 +40,8 @@ VCS_REF=$(git rev-parse HEAD)
 declare -r ARTIFACT_TYPE_TEXT_XSHELLSCRIPT="text/x-shellscript"
 declare -r TEST_TASKS=${TEST_TASKS:-""}
 
-declare -r ANNOTATION_TASK_MIGRATION="dev.konflux-ci.task.migration"
+declare -r ANNOTATION_HAS_MIGRATION="dev.konflux-ci.task.has-migration"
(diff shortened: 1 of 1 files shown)


tkdchen:
Tested by https://github.com/tkdchen/build-definitions/blob/add-script-for-testing-migrations/hack/test-task-migrations.sh

hack/build-and-push.sh
```diff
@@ -40,7 +40,8 @@ VCS_REF=$(git rev-parse HEAD)
 declare -r ARTIFACT_TYPE_TEXT_XSHELLSCRIPT="text/x-shellscript"
 declare -r TEST_TASKS=${TEST_TASKS:-""}
 
-declare -r ANNOTATION_TASK_MIGRATION="dev.konflux-ci.task.migration"
+declare -r ANNOTATION_MIGRATION_EXISTS="dev.konflux-ci.task.migration-exists"
+declare -r ANNOTATION_IS_MIGRATION="dev.konflux-ci.task.is-migration"
```
chmeliik, line 44: optional: would `has-migration` be more obvious than `migration-exists`?

chmeliik (dismissed):

hack/build-and-push.sh
tkdchen: Yup, updated. This one should make Mykola happy as well. :smile:

chmeliik (approved):

PR #1960, opened 2025-02-19
feat(KONFLUX-6858): add support for generating docker mediaType

arewm:
While we still produce images with an OCI mediaType by default, some users may need to generate docker mediaTypes in order to remain compatible with older functionality. This change lets that setting be configured.

We are not exposing it as a pipeline parameter by default to encourage the continued use of OCI mediaTypes.

### DIFF
--- a/task/sast-coverity-check/0.2/sast-coverity-check.yaml
+++ b/task/sast-coverity-check/0.2/sast-coverity-check.yaml
@@ -131,6 +131,11 @@ spec:
       format.'
     name: SBOM_TYPE
     type: string
+  - default: oci
+    description: The format for the resulting image's mediaType. Valid values are
+      oci (default) or docker.
+    name: BUILDAH_FORMAT
+    type: string
   - name: image-url
     type: string
   - default: cov-license
@@ -171,8 +176,6 @@ spec:
         cpu: "1"
         memory: 1Gi
     env:
-    - name: BUILDAH_FORMAT
-      value: oci
     - name: STORAGE_DRIVER
       value: $(params.STORAGE_DRIVER)
     - name: HERMETIC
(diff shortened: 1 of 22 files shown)


kdudka (approved):
The changes in `sast-coverity-check` look fine.

chmeliik (approved):

arewm:
@konflux-ci/build-maintainers @konflux-ci/integration-service-maintainers @kdudka @jeremychoi @sfowl , I have fixed the issues with this PR. It is ready for review again.

sfowl (approved):
changes in sast-coverity-check LGTM

kdudka (approved):

task/sast-coverity-check/0.2/sast-coverity-check.yaml
```diff
@@ -171,8 +176,6 @@ spec:
         cpu: "1"
         memory: 1Gi
     env:
-    - name: BUILDAH_FORMAT
-      value: oci
```
dirgim, line 175: I may be wrong, but shouldn't this env variable be set to the value of `$(params.BUILDAH_FORMAT)` instead of being removed?

chmeliik: As far as I can tell, the format isn't relevant to the coverity task. The fact that it gets the new parameter makes it confusing, but it's probably not a big deal. It's not the first parameter that it ignores and likely won't be the last (considering it inherits all params from buildah)

dirgim: Okay, in that case I won't hold up the PR for this, thanks for the info.

dirgim (approved):
LGTM

task/sast-coverity-check/0.2/sast-coverity-check.yaml
arewm: This was removed everywhere because it doesn't matter what format is used when buildah runs to build since the format can just be specified when pushing.

I agree that it is confusing that coverity gets the new format though.

PR #1962, opened 2025-02-20
pnc-prebuild-git-clone: add build team to owners

chmeliik:
The task is generated from the git-clone task => changes to the git-clone task will always require corresponding changes to the PNC task.

Add build team to the CODEOWNERS of this task to avoid git-clone changes being blocked on reviews from the PNC team.

Also move this task to the same Renovate group as git-clone. Otherwise Renovate PRs would fail CI because they would only update the git-clone task and not the PNC task.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -25,6 +25,7 @@
 /task/git-clone                             @konflux-ci/build-maintainers
 /task/git-clone-oci-ta                      @konflux-ci/build-maintainers
 /task/init                                  @konflux-ci/build-maintainers
+/task/pnc-prebuild-git-clone-oci-ta         @konflux-ci/build-maintainers @rnc
 /task/push-dockerfile                       @konflux-ci/build-maintainers
 /task/push-dockerfile-oci-ta                @konflux-ci/build-maintainers
 /task/show-sbom                             @konflux-ci/build-maintainers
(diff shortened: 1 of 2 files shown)


chmeliik:
Would solve the failure in https://github.com/konflux-ci/build-definitions/pull/1932

But a better solution would be to rework the pnc prebuild task. @rnc why is it generated from git-clone? Would it work for the pnc prepare stuff to be a standalone task? `clone -> pnc-prebuild -> build`?

rnc:
> Would solve the failure in #1932
> 
> But a better solution would be to rework the pnc prebuild task. @rnc why is it generated from git-clone? Would it work for the pnc prepare stuff to be a standalone task? `clone -> pnc-prebuild -> build`?

At the time we needed it to be a single task to avoid using a PVC and as KONFLUX-4903 was not available.

chmeliik:
> > Would solve the failure in #1932
> > But a better solution would be to rework the pnc prebuild task. @rnc why is it generated from git-clone? Would it work for the pnc prepare stuff to be a standalone task? `clone -> pnc-prebuild -> build`?
> 
> At the time we needed it to be a single task to avoid using a PVC and as [KONFLUX-4903](https://issues.redhat.com//browse/KONFLUX-4903) was not available.

I don't think I understand. The `-oci-ta` variant of the pnc-prebuild task wouldn't need a PVC either way

chmeliik:
Discussed offline, decided to just go ahead with the current state of things for now. The need for the pnc-prebuild task may go away when https://issues.redhat.com//browse/KONFLUX-4903 is done

tkdchen (approved):

PR #1969, opened 2025-02-25
Add test-coverity-availability-check.yaml

sfowl:
Adds a test for the expected behaviour for `coverity-availability-check` when the expected secrets are not available.

Related, I'm not sure if it's intended that `coverity-availability-check` currently does not write anything to `TEST_OUTPUT`, if these secrets are not available. Because of this, this test currently does not verify anything in `TEST_OUTPUT`.

### DIFF
--- a/task/coverity-availability-check/0.2/tests/test-coverity-availability-check.yaml
+++ b/task/coverity-availability-check/0.2/tests/test-coverity-availability-check.yaml
@@ -1,0 +1,39 @@
+---
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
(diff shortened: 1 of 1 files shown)


kdudka (approved):
Looks good to me.  @jperezdealgaba Could you please have a quick look as the author of `coverity-availability-check` and confirm that the expected behavior verified by the test matches the intended one?

jperezdealgaba:
Just reviewed it. The testing pipeline looks good and the test makes sense. It is tested a negative use case of the image

dirgim (approved):
LGTM

PR #1970, opened 2025-02-25
chore(KFLUXVNGD-210): Add source dir to tkn bundle

avi-biton:
- Update recipe.yml to update workspace
- Add `SOURCE_CODE_DIR` to point to the source folder (defaults to `source`)
- Update `tkn-bundle-oci-ta` task

### DIFF
--- a/task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
+++ b/task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
@@ -28,6 +28,10 @@ spec:
       description: The Trusted Artifact URI pointing to the artifact with
         the application source code.
       type: string
+    - name: SOURCE_CODE_DIR
+      description: Source code directory in the working dir
+      type: string
+      default: source
     - name: STEPS_IMAGE
       description: An optional image to configure task steps with in the bundle
       type: string
@@ -47,6 +51,8 @@ spec:
     env:
       - name: HOME
         value: $(params.HOME)
+      - name: SOURCE_CODE_DIR
+        value: $(params.SOURCE_CODE_DIR)
       - name: TASK_FILE
         value: tekton_task_files
     volumeMounts:
@@ -60,7 +66,7 @@ spec:
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
     - name: modify-task-files
       image: quay.io/konflux-ci/konflux-test:latest@sha256:d596724343a31c201a5c2e79f233f9ef78ac65726ae4ed5ffa41d76b3dac630f
-      workingDir: /var/workdir/source
+      workingDir: /var/workdir
       env:
         - name: CONTEXT
           value: $(params.CONTEXT)
@@ -87,11 +93,11 @@ spec:
             neg=1
             path="${path#\!}"
           fi
-          if [[ -d "/var/workdir/source/${path}" ]]; then
+          if [[ -d "/var/workdir/${SOURCE_CODE_DIR}/${path}" ]]; then
             # for directories look for any .yaml or .yml files
             paths+=(
-              "/var/workdir/source/${path}"/**/*.yaml
-              "/var/workdir/source/${path}"/**/*.yml
+              "/var/workdir/${SOURCE_CODE_DIR}/${path}"/**/*.yaml
+              "/var/workdir/${SOURCE_CODE_DIR}/${path}"/**/*.yml
             )
           else
             # for files add the file to the collected paths
@@ -105,8 +111,8 @@ spec:
               # single negative path provided, first include everything then
               # subtract the negative elements
               FILES=(
-                "/var/workdir/source"/**/*.yaml
-                "/var/workdir/source"/**/*.yml
+                "/var/workdir/${SOURCE_CODE_DIR}"/**/*.yaml
+                "/var/workdir/${SOURCE_CODE_DIR}"/**/*.yml
               )
             fi
             for p in "${paths[@]}"; do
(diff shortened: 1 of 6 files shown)


arewm (commented):
is `source` not an implementation detail of the task itself? I would expect that users would want to provide the path to the source within the git clone. The location of that source within the context should be irrelevant. 

i.e. https://github.com/konflux-ci/build-definitions/blob/main/task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml#L65-L66

In the buildah tasks, this is achieved with the use of a separate environment variable, `SOURCE_CODE_DIR`: https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.4/buildah-oci-ta.yaml#L221-L222

avi-biton:
Thanks for the review @arewm, I will update the PR

avi-biton:
@arewm I added `SOURCE_CODE_DIR` parameter to the task with a default value of ""
In the pipeline definition I added this parameter with the value `source` 
PTAL and comment :)

task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
```diff
@@ -87,11 +93,11 @@ spec:
             neg=1
             path="${path#\!}"
           fi
-          if [[ -d "/var/workdir/source/${path}" ]]; then
+          if [[ -d "/var/workdir/source/${SOURCE_CODE_DIR}/${path}" ]]; then
```
chmeliik, line 96: I don't think this is correct for `oci-ta`. We download the source artifact like this

```
$(params.SOURCE_ARTIFACT)=/var/workdir/source
```

And the source artifact is just the repo, it doesn't have another `./source` subdirectory.

Most TA recipes do the replacement like this

https://github.com/konflux-ci/build-definitions/blob/4692a3811cf2cf319208729aa864b186f5f3743d/task/prefetch-dependencies-oci-ta/0.2/recipe.yaml#L51-L52

chmeliik: Or like this, this may be better

https://github.com/konflux-ci/build-definitions/blob/4692a3811cf2cf319208729aa864b186f5f3743d/task/buildah-oci-ta/0.4/recipe.yaml#L13-L14

avi-biton: Yes, I reverted the oci-ta to what it was and apply the changes only for `tkn-bundle`

avi-biton: I now see that the CI is failing because its running the `generate-ta-tasks.sh` and builds the `tkn-bundle-oci-ta` task from the `tkn-bundle` task

avi-biton: @chmeliik Can you PTAL.
I updated the recipe as in buildah task and updated the oci-ta task.

task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
```diff
@@ -28,6 +28,10 @@ spec:
       description: The Trusted Artifact URI pointing to the artifact with
         the application source code.
       type: string
+    - name: SOURCE_CODE_DIR
+      description: Source code directory in the working dir
+      type: string
+      default: /source
```
arewm, line 34: as a user, I would expect to input `source` over `/source`. Making this change would also require a change in the recipe to append the `/`.

avi-biton: I could use the recipe to append the `/` but it will only append it to the `oci-ta` task and not the non ta task.
However I updated the parameter to be `source` and added the `/` in the environment variable for both versions of the task.

task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
```diff
@@ -60,7 +66,7 @@ spec:
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
```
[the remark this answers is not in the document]
avi-biton: Thanks @dheerajodha 
Its not something I can change manually, when adding `use-source` in the recipe, it will automatically generate it as /var/workdir/source when generating a ta task as defined in the `ta.go` file:
https://github.com/konflux-ci/build-definitions/blob/main/task-generator/trusted-artifacts/ta.go#L313
Do you think we can do it in another way?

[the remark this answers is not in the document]
avi-biton: @arewm WDYT?

gbenhaim:
If there are no objections, I'll merge this PR tomorrow.

task/tkn-bundle-oci-ta/0.1/tkn-bundle-oci-ta.yaml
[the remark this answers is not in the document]
arewm: If you cannot change the directory in the TA task, then that implies to me that the source code directory should be some subdirectory within that. Is the `source` directory not just where the git source is cloned?

Should a user provide the directory _within_ that cloned source directory ... therefore never having to worry about `source`? `source` definitely feels like an implementation detail now.

gbenhaim (approved):

PR #1971, opened 2025-02-25
fix: retry git fetch --tags on fail

fmudry:
Git fetch tags currently fail, when a repo has submodules. 
Fetching for the second time works as a workaround.

### DIFF
--- a/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
+++ b/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
@@ -261,7 +261,10 @@ spec:
 
         if [ "${PARAM_FETCH_TAGS}" = "true" ]; then
           echo "Fetching tags"
-          git fetch --tags
+          if ! git fetch --tags; then
+            echo "Retrying fetch command..."
+            git fetch --tags
+          fi
         fi
       securityContext:
         runAsUser: 0
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -257,7 +257,10 @@ spec:
 
       if [ "${PARAM_FETCH_TAGS}" = "true" ] ; then
         echo "Fetching tags"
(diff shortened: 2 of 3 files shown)


chmeliik (commented):
> Git fetch tags currently fail, when a repo has submodules.

I think it only fails in very specific situations for repos with submodules, not always. I don't think anyone actually understands what causes those situtations. Let's just update the commit message to say "In some situtations, `git fetch tags` may fail for repos with submodules" :sweat_smile:

task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml
```diff
@@ -261,7 +261,17 @@ spec:
...
+          # Checking, if tags get fetched, when repo has submodules
+          if ! git fetch --tags; then
+            echo "Retrying fetch command..."
+            if git fetch --tags; then
+              echo "Fetching tags was successful"
+            else
+              EXIT_CODE="$?"
+              echo "fatal: Fetching tags was not successful, exit code: ${EXIT_CODE}"
+              exit "${EXIT_CODE}"
+            fi
```
chmeliik, line 273: IMO we can simplify this block to just this

```suggestion
            git fetch --tags
```

The success message doesn't seem necessary (and it's a bit weird that we would print it after a successful retry but not if the first attempt already succeeds). And for the error message, we can just let git print its own

fmudry: Thank you for a great review, I really appreciate it :smiley:

chmeliik (commented):
Otherwise LGTM :+1:

chmeliik (approved):

chmeliik:
@rnc PTAL, you're the sole owner of the pnc-prebuild task which blocks changes to the clone task

Please also take a look at https://github.com/konflux-ci/build-definitions/pull/1962

rnc (approved):

PR #1972, opened 2025-02-25
sast-coverity-check: do not mount the whole /opt

kdudka:
... as a volume during the build.  Mount only specific subdirectory (and a symbolic link) instead in order not to break container builds, such as `konflux-tempo`, that use /opt for other purposes.

Resolves: https://issues.redhat.com/browse/KFLUXSPRT-1945


kdudka:
@chmeliik Thanks for review!

PR #1883, opened 2025-01-31
sast-*-check: implement default KFP_GIT_URL

kdudka:
... for internal Konflux instances.  The default value of the `KFP_GIT_URL` parameter is now `SITE_DEFAULT`, which expands to https://gitlab.cee.redhat.com/osh/known-false-positives.git when the task is running on an internal Konflux instance.

This was initially implement for `sast-snyk-check` via #1848.  This commit duplicates the code/yaml changes for all other `sast-*check` tasks.

Related: https://github.com/konflux-ci/build-definitions/pull/1848
Resolves: https://issues.redhat.com/browse/OSH-763

PR #1974, opened 2025-02-26
sast-*-check: check connectivity to CEE GitLab

kdudka:
... when `KFP_GIT_URL` is set to `SITE_DEFAULT`, which is the default value.  This way we do not need to hardcode the list of internal Konflux clusters, as initially implemented in #1848 and #1883.

Related: https://issues.redhat.com/browse/KONFLUX-4530
Related: https://issues.redhat.com/browse/OSH-763
Resolves: https://issues.redhat.com/browse/OSH-814


sfowl (approved):
LGTM

kdudka:
@sfowl Thanks for review!

PR #1976, opened 2025-02-26
fix(KONFLUX-6999): improve logging output for FBC validation

arewm:
There is often a lot of text output when validating the related images. In order to enable better troubleshooting, we will augment the normal output to add additional details.


hongweiliu17 (approved):
lgtm

PR #1977, opened 2025-02-27
coverity-availability-check: always write TEST_OUTPUT

sfowl:
The intention of this change is write TEST_OUTPUT when coverity-availability-check results in failure, so that there is better visual indication in the konflux UI that user action is expected (to add coverity license and secret).


sfowl:
My understanding is that a version bump is not needed in this case, because there is no change to the parameters or migration steps needed by users.

dirgim (approved):
LGTM

kdudka (approved):

PR #1979, opened 2025-02-27
Allow specifying startup directory

tkdchen:
This change can benefit other scripts that want to reuse this script to build bundles at a differet location rather than from inside build-definitions repo. It may be for testing purpose without writing custom code to repeat the logic of build-and-push process. For example of a use case: https://github.com/konflux-ci/pipeline-migration-tool/pull/10

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -82,9 +82,16 @@ function should_skip_repo() {
     [ "$http_code" != "200" ]
 }
 
+: "${HERE:=}"
+
 # local dev build script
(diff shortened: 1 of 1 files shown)


chmeliik (approved):
Seems reasonable :+1:

hack/build-and-push.sh
```diff
@@ -82,9 +82,16 @@ function should_skip_repo() {
     [ "$http_code" != "200" ]
 }
 
+: "${HERE:=}"
```
mmorhun, line 85: I'd give it more meaningful name

mmorhun (approved):

PR #1981, opened 2025-02-27
Add wrapper script for generator scripts

chmeliik:
As of now, we generate files with 5 different scripts. What's worse,
they only work properly if run in a specific order.

Add a wrapper script that calls all the generator scripts in the right
order.

Rework the Local Development section in the README, mention this script
there.

Also mention the script in all the workflows that check if the generated
files are up to date.

---

Also improve the generate-buildah-remote script to set GOTOOLCHAIN=auto on its own, for consistency with the generate-ta-tasks script


hack/generate-everything.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+
+run_gen() {
+    echo -e "\n=== $* ===\n"
+    "$@"
+}
```
kdudka, line 7: It might be easier to set `-o xtrace`, perhaps in combination with setting `$PS4` if a custom marker is needed.

hack/generate-everything.sh
```diff
@@ -0,0 +1,18 @@
...
+# These 3 need to run in this order. Not for any logical reasons, but simply
+# because of the current state of dependence between the generated tasks
+# and the sources they are generated from.
+run_gen hack/build-manifests.sh
+run_gen hack/generate-ta-tasks.sh
+run_gen hack/generate-buildah-remote.sh
+
+run_gen hack/generate-pipelines-readme.py
+
+run_gen hack/update_renovate_json_based_on_codeowners.py -o renovate.json
```
kdudka, line 18: I am missing `hack/generate-sast-tasks.sh` in the list of scripts to run.

kdudka:
@chmeliik You may consider reusing the code that determines `$TASK_DIR` in the other `hack/*.sh` scripts so that the top-level script also works regardless of CWD.

hack/generate-everything.sh
chmeliik: Yup, changed to just use `xtrace`

hack/generate-everything.sh
chmeliik: It should be covered by `hack/build-manifests.sh` along with all other Kustomize-generated things

In fact, I think we can remove `hack/generate-sast-tasks.sh` now, WDYT?

kdudka: I have never used `build-manifests.sh` before but it seems to do what is needed.  So yes, I think we can safely remove `generate-sast-tasks.sh`.

chmeliik: Removed in latest commit

tnevrlka (approved):
Great quality of life feature! Thank you!

hack/generate-everything.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail -o xtrace
+
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+cd "$SCRIPTDIR/.."
```
kdudka, line 5: I am wondering whether this could be simplified like this?
```diff
--- a/hack/generate-everything.sh
+++ b/hack/generate-everything.sh
@@ -1,8 +1,7 @@
 #!/bin/bash
 set -o errexit -o nounset -o pipefail -o xtrace

-SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
-cd "$SCRIPTDIR/.."
+cd "$(dirname "${BASH_SOURCE[0]}")/.."

 # These 3 need to run in this order. Not for any logical reasons, but simply
 # because of the current state of dependence between the generated tasks
```

kdudka (approved):

PR #1983, opened 2025-02-27
Let acs-image-check task pass with policy violations

tnevrlka:
acs-image-check task currently fails when there are Critical or High policy violations detected
The task should pass if it succeeds in creating a report and let users handle the policy violations separately themselves

### DIFF
--- a/task/acs-image-check/0.1/acs-image-check.yaml
+++ b/task/acs-image-check/0.1/acs-image-check.yaml
@@ -124,6 +124,7 @@ spec:
 
         cp roxctl_image_check_output.json /steps-shared-folder/acs-image-check.json
 
+        # Exit code 3 means the error occurred during roxctl image check, but setup was successful
         if [ "$ROXCTL_CHECK_STATUS" -ne 0 ]; then
           exit 3
         fi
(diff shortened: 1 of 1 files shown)


task/acs-image-check/0.1/acs-image-check.yaml
```diff
@@ -135,18 +136,51 @@ spec:
...
+        then
+          echo "Error occurred during setup of roxctl"
+          exit "$ROX_IMAGE_CHECK_STATUS"
+        fi
+
+        # Number of policy violations with Critical and High severity parsed from the report
+        severe_violations=$(
+          jq '.summary |
+            with_entries(
+              select(.key | inside("CRITICAL", "HIGH"))
```
chmeliik, line 173: `inside` might not be exactly what we want: https://jqlang.org/manual/#inside

In this instance, it basically does substring search on each of the arguments
```bash
$ jq -n '"I" | inside("CRITICAL", "HIGH")'
true
true
```

Maybe `IN`, if our jq version has it?

```bash
$ jq -n '"I" | IN("CRITICAL", "HIGH")'
false

$ jq -n '"HIGH" | IN("CRITICAL", "HIGH")'
true
```

tnevrlka: Great catch, thank you!

PR #1988, opened 2025-03-03
use UBI9 buildah instead

Victoremepunto:

Use buildah from UBI instead of Fedora now that the `--artifact` param is present

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -214,13 +214,12 @@ spec:
           -v $BUILD_DIR/entitlement:/etc/pki/entitlement:Z \
           $BOOTC_BUILDER_IMAGE $IMAGE_TYPE_ARGUMENT --local $TAGGED_AS
 
-        # This script is currently using fedora image because buildah in UBI9 does not have the OCI support (no --artifact* switches)
         time sudo podman run --authfile=$BUILD_DIR/.docker/config.json --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t \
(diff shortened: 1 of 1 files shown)


task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -214,13 +214,12 @@ spec:
...
           $BOOTC_BUILDER_IMAGE $IMAGE_TYPE_ARGUMENT --local $TAGGED_AS
 
-        # This script is currently using fedora image because buildah in UBI9 does not have the OCI support (no --artifact* switches)
         time sudo podman run --authfile=$BUILD_DIR/.docker/config.json --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t \
           -v $BUILD_DIR/.docker:/.docker -v $(pwd)/output:/output \
           -v /var/lib/containers/storage:/var/lib/containers/storage \
           -v $BUILD_DIR/scripts/script-push.sh:/script-push.sh \
           -v $BUILD_DIR/tekton-results:/tekton-results \
-          registry.fedoraproject.org/fedora \
+          registry.access.redhat.com/ubi9/buildah:9.5 \
```
ralphbean, line 222: I think I'd like to move this to `latest`.

```suggestion
          registry.access.redhat.com/ubi9/buildah:latest \
```

I either want to pin to a specific digest or buidlah and let renovate update it (but I'm not ready to do that in this PR), or I want it tracking latest so there's zero toil to bump to 9.6 manually later.

Let's follow the `latest` tag for now.

Victoremepunto: sounds good !

PR #1990, opened 2025-03-03
use buildah image tag current latest

Victoremepunto:
Follow-up to #1988 , use specific image tag (current latest) for reproducible builds

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -33,7 +33,7 @@
         "task/acs-image-check/**",
         "task/acs-image-scan/**",
         "task/apply-tags/**",
-         "task/build-image-index/**",
+        "task/build-image-index/**",
         "task/build-image-manifest/**",
         "task/buildah-10gb/**",
         "task/buildah-20gb/**",
@@ -244,7 +244,7 @@
       "customType": "regex",
       "fileMatch": ["^task/build-vm-image/.*/build-vm-image.yaml"],
       "matchStrings": [
-        "default: (?<depName>.*):(?<currentValue>.*)"
+        "BUILDAH_IMAGE\\s*\\n\\s*value:\\s*['\"]?(?<depName>[^'\":]+):(?<currentValue>[^'\":\\s]+)['\"]?"
       ],
       "depNameTemplate": "registry.access.redhat.com/ubi9/buildah",
       "datasourceTemplate": "docker",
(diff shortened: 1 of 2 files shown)


ralphbean (approved):
Let's merge this but we need to watch to make sure renovate actually bumps it in the future.

I'll set a task to remind myself to do so.

ralphbean (dismissed):

Victoremepunto:
@ralphbean  I need some input here - I couldn't find a "better" way to integrate the automatic bump of buildah images, so I went ahead with this 2 step refactor:

- parametrize the buildah image on the build-vm-image task
- create a custom regex to match the `default:` value of the parameter

I'm not 100% sure we wanna go with this approach, since this rule is specific to this task, but like I said, I couldn't find anything "more suitable", hence my ask for input

secondly, we can avoid parametrizing the image, and simply use the regex to match the image in the YAML file instead

Please, LMK what you think. 

Also, I need input on what's the expected format for the image tag, I went ahead and used `redhat` format (as opposed to use a SHA digest for example).

thanks.

MartinBasti:
Parametrization of build image or task image is not allowed, users could inject whatever image there and konflux cannot guarantee safety of build. This was removed several times by Conforma team already.

you may save image into env var (either in task def or shell) and try to configure renovate to bump it there

renovate.json
```diff
@@ -33,7 +33,7 @@
         "task/acs-image-check/**",
         "task/acs-image-scan/**",
         "task/apply-tags/**",
-        "task/build-image-index/**",
+         "task/build-image-index/**",
```
MartinBasti, line 36: Unneeded whitespace change

Victoremepunto: amended, thanks!

Victoremepunto:
> Parametrization of build image or task image is not allowed, users could inject whatever image there and konflux cannot guarantee safety of build. This was removed several times by Conforma team already.
> 
> you may save image into env var (either in task def or shell) and try to configure renovate to bump it there

Hi @MartinBasti , based on the feedback I've added:

https://github.com/konflux-ci/build-definitions/pull/1990/commits/60ae1f9cf2b604d4dedd0497868cd84e36e6d30c

I went with the actual task environment variable suggestion.

I amended the corresponding regex to look for this value (whether if its quoted or not) since it may be really easy to break it inadvertently if someone updates the value and adds quotes for some reason.

MartinBasti (approved):
Thank you very much

ralphbean (approved):

PR #1997, opened 2025-03-05
chore(KFLUXVNGD-221): deprecate task verify-signed-rpms

yftacherzog:
The task was renamed and the task under the old name is not maintained anymore and should not be used.


avi-biton (dismissed):
/lgtm

avi-biton (approved):

PR #2007, opened 2025-03-06
fix(KONFLUX-6666) Select auth properly for preflight task

bcrochet:
Distill the auth down to the entry necessary for pulling an image before passing to preflight.

### DIFF
--- a/task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
+++ b/task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
@@ -22,6 +22,7 @@ spec:
       type: string
       description: The type of artifact. Select from application, operatorbundle, or introspect.
       default: "introspect"
+
   results:
     - name: TEST_OUTPUT
       description: Ecosystem checks pass or fail outcome.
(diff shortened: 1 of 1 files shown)


acornett21 (dismissed):

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -89,6 +90,25 @@ spec:
...
+      env:
+      - name: PARAM_IMAGE_URL
+        value: $(params.image-url)
+      results:
+        - name: auth-json-path
+          description: Path to auth.json
+      script: |
+        _AUTH_JSON_PATH="/auth/auth.json"
+        echo "Selecting auth for $PARAM_IMAGE_URL"
+        select-oci-auth "$PARAM_IMAGE_URL" > "${_AUTH_JSON_PATH}"
```
komish, line 104: In _this_ image, this script assumes the AUTHFILE to use is at ~/.docker/config.json - can we document that here, for the inevitability that this will change?

I don't think we have any expectations that this script's behavior won't change, given we don't create this image. I think it makes sense to document at least the expectation that the parse configuration file is the one I listed above.

komish: ref: https://github.com/konflux-ci/oras-container/blob/main/hack/select-oci-auth.sh

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -89,6 +90,25 @@ spec:
         printf "%s" "${_ARTIFACT_TYPE}" > "$(step.results.artifact-type.path)"
         echo "Introspection concludes that this artifact is of type \"${_ARTIFACT_TYPE}\"."
 
+    - name: generate-container-auth
```
komish, line 93: @bcrochet do you remember if Skopeo will properly select auth in the way `select-oci-auth` would?

If not, then I'd think the **generate-container-auth** step needs to happen first in this Task, given skopeo is used to read labels on the image in the **introspect** step

komish: Otherwise, the **generate-container-auth** can probably happen after **set-skip-for-bundles**

bcrochet: Since skopeo uses containers/image and containers/storage, it does in fact select the auth in the way that is expected. Preflight does not process the auth in the same way, hence the need for this in the first place.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
komish:
```suggestion
        # `select-oci-auth` here assumes the input credentials are at path ~/.docker/config.json
        select-oci-auth "$PARAM_IMAGE_URL" > "${_AUTH_JSON_PATH}"
```

acornett21 (approved):

komish (approved):
/lgtm

PR #2011, opened 2025-03-07
Fix mismatched task versions

simonbaird:
My original goal was to fix the missing label on the `rpms-signature-scan` task, since it causes problems for freshly onboarded components, see [here for example](https://github.com/enterprise-contract/ec-cli/pull/2368#pullrequestreview-2665779489). I've also seen multiple user support requests from users with the exact same problem.

I noticed there were a few other tasks with the version label missing, and some with copy/pasted and inconsistent values, so I decided to try fixing all of them. There might be some side effects of this change that I'm not aware of, so please give that it some consideration before merging.

The reason I think this label is important is because of [this code](https://github.com/konflux-ci/build-definitions/blob/eb11ac8800f6d42848e1635e9d0141fe2417b1a9/hack/build-and-push.sh#L244-L248), but I'm not sure exactly what the impact of incorrect or missing version labels.

See also:

* #2012 which contains a different fix for missing labels in pipeline definitions.
* The `hack/check-task-version-labels.sh` script added in #2048 which was used to help identify the version labels that are missing or incorrect.

### DIFF
--- a/task/buildah/0.2/buildah.yaml
+++ b/task/buildah/0.2/buildah.yaml
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.1"
+    app.kubernetes.io/version: 0.2.1
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
     build.appstudio.redhat.com/expires-on: "2025-03-01T00:00:00Z"
--- a/task/buildah/0.3/buildah.yaml
+++ b/task/buildah/0.3/buildah.yaml
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.1"
+    app.kubernetes.io/version: 0.3.1
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
     build.appstudio.redhat.com/expires-on: "2025-03-31T00:00:00Z"
--- a/task/acs-deploy-check/0.1/acs-deploy-check.yaml
+++ b/task/acs-deploy-check/0.1/acs-deploy-check.yaml
@@ -7,6 +7,8 @@ metadata:
     task.results.type: roxctl-deployment-check
     task.results.container: step-report
     task.output.location: logs
+  labels:
+    app.kubernetes.io/version: "0.1"
 spec:
   description: >-
     Policy check a deployment with StackRox/RHACS This tasks allows you to check
--- a/task/acs-image-check/0.1/acs-image-check.yaml
+++ b/task/acs-image-check/0.1/acs-image-check.yaml
@@ -7,6 +7,8 @@ metadata:
     task.results.type: roxctl-image-check
     task.results.container: step-report
     task.output.location: logs
+  labels:
+    app.kubernetes.io/version: "0.1"
 spec:
   description: Policy check an image with StackRox/RHACS This tasks allows you to
     check an image against build-time policies and apply enforcement to fail builds.
--- a/task/acs-image-scan/0.1/acs-image-scan.yaml
+++ b/task/acs-image-scan/0.1/acs-image-scan.yaml
@@ -8,6 +8,8 @@ metadata:
     task.results.key: SCAN_OUTPUT
     task.results.container: step-report
     task.output.location: logs
+  labels:
+    app.kubernetes.io/version: "0.1"
 spec:
   description: Policy check an image with StackRox/RHACS This tasks allows you to
     check an image against build-time policies and apply enforcement to fail builds.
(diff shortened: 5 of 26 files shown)


task/buildah/0.3/buildah.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.1"
+    app.kubernetes.io/version: "0.3.1"
```
simonbaird, line 5: Could consider "0.3" instead of "0.3.1", but idk.

simonbaird:
Getting some "Error: failed to parse string as a Tekton object:" when I try it locally. Will see if I can figure out why.

task/buildah/0.3/buildah.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.1"
+    app.kubernetes.io/version: 0.3.1
```
arewm, line 5: Should this just be X.Y?

task/generate-odcs-compose/0.1/generate-odcs-compose.yaml
```diff
@@ -3,9 +3,12 @@ apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
   name: generate-odcs-compose
-  build.appstudio.redhat.com/expires-on: "2025-03-15T00:00:00Z"
-  build.appstudio.redhat.com/expiry-message: 'This task is deprecated.
-    Please remove it from your pipeline.'
+  annotations:
+    build.appstudio.redhat.com/expires-on: "2025-03-15T00:00:00Z"
+    build.appstudio.redhat.com/expiry-message: 'This task is deprecated.
```
arewm, line 8: This is not consistently applied on all expired tasks... Not asking you to fix, necessarily, just noticing.

task/buildah/0.2/buildah.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.1"
+    app.kubernetes.io/version: 0.2.1
```
arewm, line 5: Do we want to be consistent with which is first, labels or annotations?

arewm (commented):
general comments: Many deprecated tasks are updated. Do we want to touch those for consistency? 

Should we update them for consistency but change the build+push logic to skip tasks if the expiration date has passed?

task/generate-odcs-compose/0.1/generate-odcs-compose.yaml
tkdchen: Just thinking about `build-and-push.sh` should skip building expired tasks. Does this make sense?

task/buildah/0.3/buildah.yaml
hongweiliu17: +1, some tasks are `X.Y`, some are some `X.Y.1`, but they folder are all `X.Y` format

yftacherzog:
> My original goal was to fix the missing label on the `rpms-signature-scan` task,

@simonbaird that task is maintained and tested on another repository and then copied to build-definitions.
I fixed those annotations now in the repo where the task is maintained.

simonbaird:
> > My original goal was to fix the missing label on the `rpms-signature-scan` task,
> 
> @simonbaird that task is maintained and tested on another repository and then copied to build-definitions. I fixed those annotations now in the repo where the task is maintained.

Okay.

task/buildah/0.2/buildah.yaml
simonbaird: :shrug:

simonbaird:
> > My original goal was to fix the missing label on the `rpms-signature-scan` task,
> 
> @simonbaird that task is maintained and tested on another repository and then copied to build-definitions. I fixed those annotations now in the repo where the task is maintained.

How about the missing `app.kubernetes.io/version` label though?

yftacherzog:
> > > My original goal was to fix the missing label on the `rpms-signature-scan` task,
> > 
> > 
> > @simonbaird that task is maintained and tested on another repository and then copied to build-definitions. I fixed those annotations now in the repo where the task is maintained.
> 
> How about the missing `app.kubernetes.io/version` label though?

We'll fix them as well.

simonbaird:
Looking at hack/build-and-push.sh, I'm not sure I understand how `concrete_task_version` and `task_version` are used when pushing the task bundles. So I might leave it up to the experts rather than risk the possibility of breaking something I'm not confident about troubleshooting.

I did split out the low risk annotation fix into a separate PR, see #2023.

task/buildah/0.2/buildah.yaml
simonbaird: Either way it's out of scope in this PR.

task/buildah/0.3/buildah.yaml
simonbaird: I'm not sure what the impact would be, but deciding if X.Y.Z should be disallowed is out of scope in this PR I guess.

simonbaird:
I moved the hack script into #2048 since I think it could be merged without any risk.

PR #1996, opened 2025-03-05
sast-coverity-check: another workaround for a Coverity bug

kdudka:
If `/usr/bin/file` does not exist and the `cmd-wrap.sh` script is not able to create the file, preload a shared object that will pretend that `/usr/bin/file` exists and make sure that a `file` executable is available in the search path.

Resolves: https://issues.redhat.com/browse/KFLUXSPRT-1983

PR #1766, opened 2024-12-19
remove references to jvm-build-service

tnevrlka:
- jvm-build-service is not used by anyone and will get reworked
- buildah task contains references to jvm-build-service
- Removal of the references is a breaking change
  * Remove `analyse-dependencies-java-sbom` step
  * Remove `SBOM_JAVA_COMPONENTS_COUNT` result
  * Remove `JAVA_COMMUNITY_DEPENDENCIES` result
- Release buildah version 0.3
- Upgrade buildah version in docker-build pipelines to 0.3

~~**Note for reviewers:**~~
~~I haven't found an easy way to compare different files in GitHub, I would probably checkout the branch locally and do something like:~~
```
diff task/buildah/0.2/buildah.yaml task/buildah/0.3/buildah.yaml
```
~~(I welcome different ideas)~~

PR #2018, opened 2025-03-10
fix(KONFLUX-6914): adding image digest to oras uploads

kasemAlem:
oras attach of the images in the security tasks was missing imgae-digest, as [discovered in thread](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1738860626486219) , for more details. 
we will add the image-digest parameter to tasks and add it to the command oras attach for tasks : 

sast-coverity-check
sast-shell-check
sast-snyk-check
sast-unicode-check

and  update accordingly pipelines and all required 
 

[KONFLUX-6914](https://issues.redhat.com/browse/KONFLUX-6914)

### DIFF
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -236,7 +236,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |caTrustConfigMapName| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
 |image-digest| Image digest to report findings for.| | '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
 |image-url| Image URL.| | '$(tasks.build-image-index.results.IMAGE_URL)'|
-### sast-snyk-check-oci-ta:0.3 task parameters
+### sast-snyk-check-oci-ta:0.4 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
 |ARGS| Append arguments.| | |
@@ -346,8 +346,8 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 ### prefetch-dependencies-oci-ta:0.2 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
-|CACHI2_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the prefetched dependencies.| build-images:0.4:CACHI2_ARTIFACT ; build-source-image:0.2:CACHI2_ARTIFACT ; sast-snyk-check:0.3:CACHI2_ARTIFACT ; sast-coverity-check:0.2:CACHI2_ARTIFACT ; sast-shell-check:0.1:CACHI2_ARTIFACT ; sast-unicode-check:0.1:CACHI2_ARTIFACT|
-|SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| build-images:0.4:SOURCE_ARTIFACT ; build-source-image:0.2:SOURCE_ARTIFACT ; sast-snyk-check:0.3:SOURCE_ARTIFACT ; sast-coverity-check:0.2:SOURCE_ARTIFACT ; sast-shell-check:0.1:SOURCE_ARTIFACT ; sast-unicode-check:0.1:SOURCE_ARTIFACT ; push-dockerfile:0.1:SOURCE_ARTIFACT|
+|CACHI2_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the prefetched dependencies.| build-images:0.4:CACHI2_ARTIFACT ; build-source-image:0.2:CACHI2_ARTIFACT ; sast-snyk-check:0.4:CACHI2_ARTIFACT ; sast-coverity-check:0.3:CACHI2_ARTIFACT ; sast-shell-check:0.1:CACHI2_ARTIFACT ; sast-unicode-check:0.2:CACHI2_ARTIFACT|
+|SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| build-images:0.4:SOURCE_ARTIFACT ; build-source-image:0.2:SOURCE_ARTIFACT ; sast-snyk-check:0.4:SOURCE_ARTIFACT ; sast-coverity-check:0.3:SOURCE_ARTIFACT ; sast-shell-check:0.1:SOURCE_ARTIFACT ; sast-unicode-check:0.2:SOURCE_ARTIFACT ; push-dockerfile:0.1:SOURCE_ARTIFACT|
 ### push-dockerfile-oci-ta:0.1 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
@@ -366,11 +366,11 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |TEST_OUTPUT| Tekton task test output.| |
-### sast-snyk-check-oci-ta:0.3 task results
+### sast-snyk-check-oci-ta:0.4 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |TEST_OUTPUT| Tekton task test output.| |
-### sast-unicode-check-oci-ta:0.1 task results
+### sast-unicode-check-oci-ta:0.2 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |TEST_OUTPUT| Tekton task test output.| |
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -7,12 +7,12 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 ## Parameters
 |name|description|default value|used in (taskname:taskrefversion:taskparam)|
 |---|---|---|---|
-|build-args| Array of --build-arg values ("arg=value" strings) for buildah| []| build-images:0.4:BUILD_ARGS ; sast-coverity-check:0.2:BUILD_ARGS|
-|build-args-file| Path to a file with build arguments for buildah, see https://www.mankier.com/1/buildah-build#--build-arg-file| | build-images:0.4:BUILD_ARGS_FILE ; sast-coverity-check:0.2:BUILD_ARGS_FILE|
+|build-args| Array of --build-arg values ("arg=value" strings) for buildah| []| build-images:0.4:BUILD_ARGS ; sast-coverity-check:0.3:BUILD_ARGS|
+|build-args-file| Path to a file with build arguments for buildah, see https://www.mankier.com/1/buildah-build#--build-arg-file| | build-images:0.4:BUILD_ARGS_FILE ; sast-coverity-check:0.3:BUILD_ARGS_FILE|
 |build-image-index| Add built image into an OCI image index| true| build-image-index:0.1:ALWAYS_BUILD_INDEX|
 |build-platforms| List of platforms to build the container images on. The available set of values is determined by the configuration of the multi-platform-controller.| ['linux/x86_64']| |
 |build-source-image| Build a source image.| false| |
-|dockerfile| Path to the Dockerfile inside the context specified by parameter path-context| Dockerfile| build-images:0.4:DOCKERFILE ; sast-coverity-check:0.2:DOCKERFILE ; push-dockerfile:0.1:DOCKERFILE|
+|dockerfile| Path to the Dockerfile inside the context specified by parameter path-context| Dockerfile| build-images:0.4:DOCKERFILE ; sast-coverity-check:0.3:DOCKERFILE ; push-dockerfile:0.1:DOCKERFILE|
 |git-url| Source Repository URL| None| clone-repository:0.1:url|
 |hermetic| Execute the build with network isolation| false| build-images:0.4:HERMETIC ; sast-coverity-check:0.2:HERMETIC|
 |image-expires-after| Image tag expiration time, time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | clone-repository:0.1:ociArtifactExpiresAfter ; prefetch-dependencies:0.2:ociArtifactExpiresAfter ; build-images:0.4:IMAGE_EXPIRES_AFTER ; build-image-index:0.1:IMAGE_EXPIRES_AFTER ; sast-coverity-check:0.2:IMAGE_EXPIRES_AFTER|
@@ -181,7 +181,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |image-digest| Image digest to scan| None| '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
 |image-url| Image URL| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
 |workdir| Directory that will be used for storing temporary files produced by this task. | /tmp| |
-### sast-coverity-check-oci-ta:0.2 task parameters
+### sast-coverity-check-oci-ta:0.3 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
 |ACTIVATION_KEY| Name of secret which contains subscription activation key| activation-key| |
@@ -291,9 +291,9 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |IMAGES| List of all referenced image manifests| |
-|IMAGE_DIGEST| Digest of the image just built| deprecated-base-image-check:0.5:IMAGE_DIGEST ; clair-scan:0.2:image-digest ; sast-snyk-check:0.3:image-digest ; clamav-scan:0.2:image-digest ; sast-shell-check:0.1:image-digest ; push-dockerfile:0.1:IMAGE_DIGEST ; rpms-signature-scan:0.2:image-digest|
+|IMAGE_DIGEST| Digest of the image just built| deprecated-base-image-check:0.5:IMAGE_DIGEST ; clair-scan:0.2:image-digest ; sast-snyk-check:0.4:image-digest ; clamav-scan:0.2:image-digest ; sast-coverity-check:0.3:image-digest ; sast-shell-check:0.1:image-digest ; sast-unicode-check:0.2:image-digest ; push-dockerfile:0.1:IMAGE_DIGEST ; rpms-signature-scan:0.2:image-digest|
 |IMAGE_REF| Image reference of the built image containing both the repository and the digest| |
-|IMAGE_URL| Image repository and tag where the built image was pushed| show-sbom:0.1:IMAGE_URL ; deprecated-base-image-check:0.5:IMAGE_URL ; clair-scan:0.2:image-url ; ecosystem-cert-preflight-checks:0.1:image-url ; sast-snyk-check:0.3:image-url ; clamav-scan:0.2:image-url ; sast-coverity-check:0.2:image-url ; sast-shell-check:0.1:image-url ; sast-unicode-check:0.1:image-url ; apply-tags:0.1:IMAGE ; push-dockerfile:0.1:IMAGE ; rpms-signature-scan:0.2:image-url|
+|IMAGE_URL| Image repository and tag where the built image was pushed| show-sbom:0.1:IMAGE_URL ; deprecated-base-image-check:0.5:IMAGE_URL ; clair-scan:0.2:image-url ; ecosystem-cert-preflight-checks:0.1:image-url ; sast-snyk-check:0.4:image-url ; clamav-scan:0.2:image-url ; sast-coverity-check:0.3:image-url ; sast-shell-check:0.1:image-url ; sast-unicode-check:0.2:image-url ; apply-tags:0.1:IMAGE ; push-dockerfile:0.1:IMAGE ; rpms-signature-scan:0.2:image-url|
 |SBOM_BLOB_URL| Reference of SBOM blob digest to enable digest-based verification from provenance| |
 ### buildah-remote-oci-ta:0.4 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
@@ -334,7 +334,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |CHAINS-GIT_COMMIT| The precise commit SHA that was fetched by this Task. This result uses Chains type hinting to include in the provenance.| |
 |CHAINS-GIT_URL| The precise URL that was fetched by this Task. This result uses Chains type hinting to include in the provenance.| |
 |SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| prefetch-dependencies:0.2:SOURCE_ARTIFACT|
-|commit| The precise commit SHA that was fetched by this Task.| build-images:0.4:COMMIT_SHA ; build-image-index:0.1:COMMIT_SHA ; sast-coverity-check:0.2:COMMIT_SHA|
+|commit| The precise commit SHA that was fetched by this Task.| build-images:0.4:COMMIT_SHA ; build-image-index:0.1:COMMIT_SHA ; sast-coverity-check:0.3:COMMIT_SHA|
 |commit-timestamp| The commit timestamp of the checkout| |
 |merged_sha| The SHA of the commit after merging the target branch (if the param mergeTargetBranch is true).| |
 |short-commit| The commit SHA that was fetched by this Task limited to params.shortCommitLength number of characters| |
@@ -358,7 +358,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGES_PROCESSED| Images processed in the task.| |
 |RPMS_DATA| Information about signed and unsigned RPMs| |
 |TEST_OUTPUT| Tekton task test output.| |
-### sast-coverity-check-oci-ta:0.2 task results
+### sast-coverity-check-oci-ta:0.3 task results
 |name|description|used in params (taskname:taskrefversion:taskparam)
 |---|---|---|
 |TEST_OUTPUT| Tekton task test output.| |
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -250,8 +250,9 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| None| '$(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)'|
 |caTrustConfigMapKey| The name of the key in the ConfigMap that contains the CA bundle data.| ca-bundle.crt| |
 |caTrustConfigMapName| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
-|image-url| Image URL.| | '$(tasks.build-image-index.results.IMAGE_URL)'|
-### sast-unicode-check-oci-ta:0.1 task parameters
+|image-digest| Digest of the image to scan.| None| '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
+|image-url| Image URL.| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
+### sast-unicode-check-oci-ta:0.2 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
 |CACHI2_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the prefetched dependencies.| | '$(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)'|
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -14,11 +14,11 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |build-source-image| Build a source image.| false| |
 |dockerfile| Path to the Dockerfile inside the context specified by parameter path-context| Dockerfile| build-images:0.4:DOCKERFILE ; sast-coverity-check:0.3:DOCKERFILE ; push-dockerfile:0.1:DOCKERFILE|
 |git-url| Source Repository URL| None| clone-repository:0.1:url|
-|hermetic| Execute the build with network isolation| false| build-images:0.4:HERMETIC ; sast-coverity-check:0.2:HERMETIC|
-|image-expires-after| Image tag expiration time, time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | clone-repository:0.1:ociArtifactExpiresAfter ; prefetch-dependencies:0.2:ociArtifactExpiresAfter ; build-images:0.4:IMAGE_EXPIRES_AFTER ; build-image-index:0.1:IMAGE_EXPIRES_AFTER ; sast-coverity-check:0.2:IMAGE_EXPIRES_AFTER|
-|output-image| Fully Qualified Output Image| None| init:0.2:image-url ; clone-repository:0.1:ociStorage ; prefetch-dependencies:0.2:ociStorage ; build-images:0.4:IMAGE ; build-image-index:0.1:IMAGE ; build-source-image:0.2:BINARY_IMAGE ; sast-coverity-check:0.2:IMAGE|
-|path-context| Path to the source code of an application's component from where to build image.| .| build-images:0.4:CONTEXT ; sast-coverity-check:0.2:CONTEXT ; push-dockerfile:0.1:CONTEXT|
-|prefetch-input| Build dependencies to be prefetched by Cachi2| | prefetch-dependencies:0.2:input ; build-images:0.4:PREFETCH_INPUT ; sast-coverity-check:0.2:PREFETCH_INPUT|
+|hermetic| Execute the build with network isolation| false| build-images:0.4:HERMETIC ; sast-coverity-check:0.3:HERMETIC|
+|image-expires-after| Image tag expiration time, time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | clone-repository:0.1:ociArtifactExpiresAfter ; prefetch-dependencies:0.2:ociArtifactExpiresAfter ; build-images:0.4:IMAGE_EXPIRES_AFTER ; build-image-index:0.1:IMAGE_EXPIRES_AFTER ; sast-coverity-check:0.3:IMAGE_EXPIRES_AFTER|
+|output-image| Fully Qualified Output Image| None| init:0.2:image-url ; clone-repository:0.1:ociStorage ; prefetch-dependencies:0.2:ociStorage ; build-images:0.4:IMAGE ; build-image-index:0.1:IMAGE ; build-source-image:0.2:BINARY_IMAGE ; sast-coverity-check:0.3:IMAGE|
+|path-context| Path to the source code of an application's component from where to build image.| .| build-images:0.4:CONTEXT ; sast-coverity-check:0.3:CONTEXT ; push-dockerfile:0.1:CONTEXT|
+|prefetch-input| Build dependencies to be prefetched by Cachi2| | prefetch-dependencies:0.2:input ; build-images:0.4:PREFETCH_INPUT ; sast-coverity-check:0.3:PREFETCH_INPUT|
 |privileged-nested| Whether to enable privileged mode, should be used only with remote VMs| false| build-images:0.4:PRIVILEGED_NESTED|
 |rebuild| Force rebuild image| false| init:0.2:rebuild|
 |revision| Revision of the Source Repository| | clone-repository:0.1:revision|
@@ -222,7 +222,8 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |YUM_REPOS_D_TARGET| Target path on the container in which yum repository files should be made available| /etc/yum.repos.d| |
 |caTrustConfigMapKey| The name of the key in the ConfigMap that contains the CA bundle data.| ca-bundle.crt| |
 |caTrustConfigMapName| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
-|image-url| | None| '$(tasks.build-image-index.results.IMAGE_URL)'|
+|image-digest| Digest of the image to which the scan results should be associated.| None| '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
+|image-url| URL of the image to which the scan results should be associated.| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
 ### sast-shell-check-oci-ta:0.1 task parameters
 |name|description|default value|already set by|
 |---|---|---|---|
@@ -264,6 +265,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |SOURCE_ARTIFACT| The Trusted Artifact URI pointing to the artifact with the application source code.| None| '$(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)'|
 |caTrustConfigMapKey| The name of the key in the ConfigMap that contains the CA bundle data.| ca-bundle.crt| |
 |caTrustConfigMapName| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
+|image-digest| Image digest| | '$(tasks.build-image-index.results.IMAGE_DIGEST)'|
 |image-url| Image URL.| | '$(tasks.build-image-index.results.IMAGE_URL)'|
 ### show-sbom:0.1 task parameters
 |name|description|default value|already set by|
--- a/pipelines/maven-zip-build/maven-zip-build.yaml
+++ b/pipelines/maven-zip-build/maven-zip-build.yaml
@@ -156,7 +156,7 @@ spec:
     - build-oci-artifact
     taskRef:
       name: sast-snyk-check
-      version: "0.3"
+      version: "0.4"
     when:
     - input: $(params.skip-checks)
       operator: in
--- a/pipelines/maven-zip-build/maven-zip-build.yaml
+++ b/pipelines/maven-zip-build/maven-zip-build.yaml
@@ -220,13 +220,15 @@ spec:
       workspace: workspace
   - name: sast-unicode-check
     params:
+    - name: image-digest
+      value: $(tasks.build-oci-artifact.results.IMAGE_DIGEST)
     - name: image-url
       value: $(tasks.build-oci-artifact.results.IMAGE_URL)
     runAfter:
     - build-oci-artifact
     taskRef:
       name: sast-unicode-check
-      version: "0.1"
+      version: "0.2"
     when:
     - input: $(params.skip-checks)
       operator: in
--- a/pipelines/maven-zip-build/maven-zip-build.yaml
+++ b/pipelines/maven-zip-build/maven-zip-build.yaml
@@ -167,13 +167,15 @@ spec:
       workspace: workspace
   - name: sast-coverity-check
     params:
+    - name: image-digest
+      value: $(tasks.build-oci-artifact.results.IMAGE_DIGEST)
     - name: image-url
       value: $(tasks.build-oci-artifact.results.IMAGE_URL)
     runAfter:
     - coverity-availability-check
     taskRef:
       name: sast-coverity-check
-      version: "0.2"
+      version: "0.3"
     when:
     - input: $(params.skip-checks)
       operator: in
--- a/task/clamav-scan/0.2/clamav-scan.yaml
+++ b/task/clamav-scan/0.2/clamav-scan.yaml
@@ -202,7 +202,7 @@ spec:
         echo "Selecting auth"
         select-oci-auth $IMAGE_URL > $HOME/auth.json
         echo "Attaching to ${IMAGE_URL}"
-         oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type application/vnd.clamav "${IMAGE_URL}" "${args[@]}"
+         oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type application/vnd.clamav "${IMAGE_URL}@${IMAGE_DIGEST}" "${args[@]}"
       volumeMounts:
         - mountPath: /work
           name: work
--- a/task/sast-coverity-check-oci-ta/0.3/MIGRATION.md
+++ b/task/sast-coverity-check-oci-ta/0.3/MIGRATION.md
@@ -1,0 +1,8 @@
+# Migration from 0.2 to 0.3
+
+- The `IMAGE_DIGEST` parameter has been added.
+
+## Action from users
+
+- All resources used task `sast-coverity-check-oci-ta` should be directed to use new `0.3` version.
+- The `IMAGE_DIGEST` parameter definition is required to be added for this task in the build pipeline.
--- a/task/sast-coverity-check/0.3/MIGRATION.md
+++ b/task/sast-coverity-check/0.3/MIGRATION.md
@@ -1,0 +1,9 @@
+# Migration from 0.2 to 0.3
+
+- The required `IMAGE_DIGEST` parameter has been added back.
+
+## Action from users
+
+- All resources used task `sast-coverity-check` should be directed to use new `0.3` version.
+- The `IMAGE_DIGEST` parameter definition is required to be added for this task in the build pipeline.
+
--- a/task/sast-coverity-check/0.3/README.md
+++ b/task/sast-coverity-check/0.3/README.md
@@ -2,18 +2,16 @@
 
 ## Description:
 
-The sast-coverity-check task uses Coverity tool to perform Static Application Security Testing (SAST). In this task, we use the buildless mode, where Coverity has the ability to capture source code without the need of building the product.
+The sast-coverity-check task uses Coverity tool to perform Static Application Security Testing (SAST).
 
 The documentation for this mode can be found here: https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/commands/topics/coverity_capture.html
 
 The characteristics of these tasks are:
 
-- Perform buildless scanning with Coverity
-- The whole source code is scanned (by scanning `$(workspaces.source.path)` )
+- Perform buildful scanning with Coverity
 - Only important findings are reported by default.  A parameter ( `IMP_FINDINGS_ONLY`) is provided to override this configuration.
 - The csdiff/v1 SARIF fingerprints are provided for all findings
 - A parameter ( `KFP_GIT_URL`) is provided to remove false positives providing a known false positives repository. By default, no repository is provided.
-- The stats of the scan are embedded into the result's SARIF file
 
 > NOTE: This task is executed only if there is a Coverity license set up in the environment. Please check coverity-availability-check task for more information.
 
@@ -25,10 +23,11 @@ The characteristics of these tasks are:
 | COV_LICENSE               | Name of secret which contains the Coverity license                                                                                    | cov-license               | no       |
 | AUTH_TOKEN_COVERITY_IMAGE | Name of secret which contains the authentication token for pulling the Coverity image                                                 | auth-token-coverity-image | no       |
 | IMP_FINDINGS_ONLY         | Report only important findings. Default is true. To report all findings, specify "false"                                              | true                      | no       |
-| KFP_GIT_URL               | Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled. | ""                        | no       |
+| KFP_GIT_URL               | Known False Positives (KFP) git URL (optionally taking a revision delimited by \#). Defaults to "SITE_DEFAULT", which means the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux instance and empty string for external Konflux instance. If set to an empty string, the KFP filtering is disabled.|SITE_DEFAULT|false|
 | PROJECT_NAME              | Name of the scanned project, used to find path exclusions. By default, the Konflux component name will be used.                       | ""                        | no       |
 | RECORD_EXCLUDED           | If set to `true`, excluded findings will be written to a file named `excluded-findings.json` for auditing purposes.                   | false                     | no       |
-
+|image-digest|Digest of the image to which the scan results should be associated.||true|
+|image-url|URL of the image to which the scan results should be associated.||true|
 ## Results:
 
 | name              | description              |
--- a/task/sast-coverity-check/0.3/README.md
+++ b/task/sast-coverity-check/0.3/README.md
@@ -1,0 +1,46 @@
+# sast-coverity-check task
+
+## Description:
+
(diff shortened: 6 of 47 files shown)


kdudka (changes requested):
The CI says that one more `-oci-ta` task needs to be updated and I agree:
```diff
--- a/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
+++ b/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
@@ -255,5 +255,5 @@ spec:
           echo "Selecting auth"
           select-oci-auth "$IMAGE_URL" >"$HOME/auth.json"
           echo "Attaching to ${IMAGE_URL}"
-          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
         done
```

kasemAlem:
> The CI says that one more `-oci-ta` task needs to be updated and I agree:
> 
> ```diff
> --- a/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
> +++ b/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
> @@ -255,5 +255,5 @@ spec:
>            echo "Selecting auth"
>            select-oci-auth "$IMAGE_URL" >"$HOME/auth.json"
>            echo "Attaching to ${IMAGE_URL}"
> -          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
> +          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
>          done
> ```

thanks @kdudka ,  
Done

task/sast-coverity-check/0.2/patch.yaml
```diff
@@ -460,7 +460,7 @@
         set -x
         UPLOAD_FILE="$1"
         MEDIA_TYPE="$2"
-        oras attach --no-tty --registry-config "${HOME}/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+        oras attach --no-tty --registry-config "${HOME}/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
kdudka, line 463: This will not work because the parameter was already removed in 8b62b416ce3332cf998d24c32d00061188d22f9c.  It needs to be reintroduced in the task, in the default build pipeline template, and a new version of the task (and migration to it) will be needed.

task/sast-unicode-check/0.1/sast-unicode-check.yaml
```diff
@@ -282,7 +282,7 @@ spec:
             echo "Selecting auth"
             select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
             echo "Attaching to ${IMAGE_URL}"
-            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+            oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
kdudka, line 285: Same problem here.  The parameter needs to be reintroduced in the task, in the default build pipeline template, and a new version of the task (and migration to it) will be needed.

kdudka (changes requested):

task/sast-coverity-check/0.2/patch.yaml
kdudka: Yes, I think we need to create `task/sast-coverity-check/0.3` to ensure smooth migration.

task/sast-coverity-check/0.3/patch.yaml
```diff
@@ -0,0 +1,467 @@
...
+
+      # upload scan results
+      echo "Selecting auth for upload of scan results"
+      select-oci-auth "${IMAGE_URL}" > "${HOME}/auth.json"
+
+      upload_file() (
+        set -x
+        UPLOAD_FILE="$1"
+        MEDIA_TYPE="$2"
+        oras attach --no-tty --registry-config "${HOME}/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
dirgim, line 458: I may be missing it, but I'm not seeing the new parameter in this file, I would have expected it to show up alongside the IMAGE_URL around [here](https://github.com/konflux-ci/build-definitions/pull/2018/files#diff-a400a090d26aaf7372823d01bf5fe4d22e8045b4a48822954610a9f41ba1df08R305-R306).

task/sast-unicode-check/0.2/MIGRATION.md
```diff
@@ -0,0 +1,6 @@
+# Migration from 0.1 to 0.2
+
+Version 0.2:
+
+- The `image-digest` parameter has been introduced back, to be used in ORAS uploading.
```
dirgim, line 5: Do the users need to add this parameter to their existing build pipelineRun definitions?

task/sast-coverity-check/0.3/patch.yaml
kasemAlem: nice catch, 
added

task/sast-unicode-check/0.2/MIGRATION.md
kasemAlem: yep, 
added the user's note

policies/all-tasks.yaml
```diff
@@ -31,6 +31,7 @@ sources:
         - trusted_artifacts
       exclude:
         # https://issues.redhat.com/browse/EC-1038
+        - step_images.step_images_accessible:quay.io/redhat-services-prod/sast/coverity:202412.3
```
kdudka, line 34: Why are you downgrading the `coverity` image to `202412.3` in the new version of `sast-coverity-check`?

This would make the workarounds for https://issues.redhat.com/browse/KFLUXSPRT-1983 etc. ineffective.

task/sast-snyk-check/0.4/sast-snyk-check.yaml
```diff
@@ -0,0 +1,306 @@
...
+      default: snyk-secret
+    - name: ARGS
+      type: string
+      description: Append arguments.
+      default: ""
+    - description: Image URL.
+      name: image-url
+      type: string
+      default: ""
+      # In a future 0.4 version of the task, drop the default to make this required
```
kdudka, line 34: The future 0.4 version is now.  So we should drop the default (and the comment).

task/sast-coverity-check/0.3/patch.yaml
```diff
@@ -0,0 +1,469 @@
...
+# Change build step image
+- op: replace
+  path: /spec/steps/0/image
+  # New image shoould be based on quay.io/konflux-ci/buildah-task:latest or have all the tooling that the original image has.
+  value: quay.io/redhat-services-prod/sast/coverity:202412.3
+
+# Change build step resources
+- op: replace
+  path: /spec/steps/0/computeResources/limits/cpu
+  value: 16
```
kdudka, line 79: The new version of `sast-coverity-check` effectively reverts #2034.  Please rebase this onto `main`.

task/sast-coverity-check/0.3/patch.yaml
```diff
@@ -0,0 +1,469 @@
...
+
+      # wrap the RUN command with "coverity capture" and record exit code of the wrapped command
+      /opt/coverity/bin/coverity --ticker-mode=no-spin capture --dir=/tmp/idir --project-dir="\$proj_dir" \
+        -- /bin/bash -c "PS4='@\\\${SECONDS}s: \\\${BASH_COMMAND} --> '
+        set -x
+        pwd >&2  # print CWD set by coverity
+        cd '\${PWD}'  # restore original CWD
+        \\"\\\$@\\"  # run the instrumented shell command
+        echo \\\$? >/tmp/idir/build-cmd-ec.txt" \
+        - "\$@"
```
kdudka, line 236: The new version of the task effectively reverts also #2022.  Please rebase this onto `main`.

task/sast-coverity-check/0.3/patch.yaml
```diff
@@ -0,0 +1,469 @@
...
+        cpu: 2
+    volumeMounts:
+      - name: trusted-ca
+        mountPath: "/mnt/trusted-ca"
+        readOnly: true
+    env:
+      - name: IMAGE_URL
+        value: $(params.image-url)
+      - name: IMAGE_DIGEST
+        value: $(params.image-digest)
```
kdudka, line 308: The referred parameter does not exist.  It needs to be added here: https://github.com/konflux-ci/build-definitions/blob/8a14fbc4ba92c0abd9eda2b41da00822a3481109/task/sast-coverity-check/0.2/patch.yaml#L87

task/sast-shell-check-oci-ta/0.2/sast-shell-check-oci-ta.yaml
```diff
@@ -0,0 +1,259 @@
+---
```
kdudka, line 1: Why are we introducing `sast-shell-check-oci-ta/0.2` when there is no `sast-shell-check/0.2`?

kdudka (changes requested):

policies/all-tasks.yaml
kasemAlem:
```
I've added it cause the EC check was failing due to the new version 0.3 of the task I created 
and I thought the fix would be by adding the image to the exclude : 
here is the error I had : 

- filepath: all_tasks-ec/0.3_sast-coverity-check.yaml
  success: false
  success-count: 9
  successes: null
  violations:
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 0 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 1 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 2 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  warnings: []
- filepath: all_tasks-ec/0.3_sast-coverity-check-oci-ta.yaml
  success: false
  success-count: 9
  successes: null
  violations:
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 1 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 2 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  - metadata:
      code: step_images.step_images_accessible
      effective_on: "2025-02-10T00:00:00Z"
    msg: Step 3 uses inaccessible image ref 'quay.io/redhat-services-prod/sast/coverity:202412.3'
  warnings: []
```

kasemAlem: ohh, 
my bad, I didn't paid enough attention to your merged [#2034](https://github.com/konflux-ci/build-definitions/pull/2034), I have to add no image in policies since it uses the latest one

task/sast-shell-check-oci-ta/0.2/sast-shell-check-oci-ta.yaml
kasemAlem: by mistake, 
agree with you, the parameter already exist so I only add it within the oras uploading command

policies/all-tasks.yaml
kdudka: The image reference was already updated in #1996.  The new version of the task needs to be based on what we have in the `main` branch now.

task/sast-coverity-check/0.3/patch.yaml
kasemAlem: rebased , 
now should be fixed

task/sast-snyk-check/0.4/sast-snyk-check.yaml
kasemAlem: fixed

policies/all-tasks.yaml
kasemAlem: fixed

task/sast-unicode-check/0.2/sast-unicode-check.yaml
```diff
@@ -0,0 +1,294 @@
...
+    - name: trusted-ca
+      configMap:
+        name: $(params.caTrustConfigMapName)
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        optional: true
+  steps:
+    - name: sast-unicode-check
+      image: quay.io/konflux-ci/konflux-test:v1.4.18@sha256:78c76869d0b171e9bfa5e931e5e6ecbe92b32f61c6ec627c57cbb9d1aad75aab
```
kdudka, line 69: This unnecessarily downgrades the image reference, compared to what we have in `sast-unicoce-check/0.1`.

task/sast-unicode-check/0.2/sast-unicode-check.yaml
```diff
@@ -0,0 +1,294 @@
...
+            echo "sast-unicode-check test failed because of the following issues:"
+            cat sast_unicode_check_out.json
+            TEST_OUTPUT=
+            parse_test_output "$(context.task.name)" sarif sast_unicode_check_out.sarif  || true
+            note="Task $(context.task.name) failed: For details, check Tekton task log."
+            ERROR_OUTPUT=$(make_result_json -r ERROR -t "$note")
+        fi
+        echo "${TEST_OUTPUT:-${ERROR_OUTPUT}}" | tee "$(results.TEST_OUTPUT.path)"
+    - name: upload
+      image: quay.io/konflux-ci/oras:latest@sha256:3536c6ca42a9168e0c7791b130fdee9793a317b306525bb609dca23d07186e9b
```
kdudka, line 260: Same here.  This reverts commit 115bccedd914ac9d803c9c073d36facba8173dc9.

task/sast-unicode-check/0.2/sast-unicode-check.yaml
kasemAlem: it was using the the old one , I've rebased to the main and now seems identical to main and my changes

task/sast-unicode-check/0.2/sast-unicode-check.yaml
kasemAlem: it was using the the old one , I've rebased to the main and now seems identical to main and my changes

task/sast-snyk-check/0.4/tests/test-sast-snyk-check-no-token.yaml
```diff
@@ -0,0 +1,47 @@
...
+  tasks:
+    - name: scan-with-snyk
+      workspaces:
+        - name: workspace
+          workspace: tests-workspace
+      taskRef:
+        name: sast-snyk-check
+      params:
+        - name: image-url
+          value: "https://github.com/konflux-ci/build-definitions.git"
```
kdudka, line 20: Although the test probably does not care, this does not look like a valid image URL.  Those usually look like this: `quay.io/redhat-user-workloads/kdudka-tenant/aap-konflux-pipelines@sha256:f672fb7e55e13b6874a86e42159d87b4d7cf3a2119a545b35d34ed66b07948f0`

kdudka (approved):
Looks good.  We should also update `app.kubernetes.io/version` in `sast-snyk-check` and `sast-coverity-check`.

kdudka (approved):

pipelines/docker-build-multi-platform-oci-ta/README.md
```diff
@@ -215,6 +215,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |YUM_REPOS_D_TARGET| Target path on the container in which yum repository files should be made available| /etc/yum.repos.d| |
 |caTrustConfigMapKey| The name of the key in the ConfigMap that contains the CA bundle data.| ca-bundle.crt| |
 |caTrustConfigMapName| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
+|image-digest| | | |
```
sonam1412, line 218: Shall we not write some words about image digest ?

14rcole (approved):
lgtm with the exception of Sonam's comment above

pipelines/docker-build-multi-platform-oci-ta/README.md
kdudka: There is no description for `image-url` either because we do not provide any description for these parameters in `task/sast-coverity-check/0.3/patch.yaml`.  The reason why the last two columns are empty is that we forgot to update `pipelines/docker-build/docker-build.yaml`:
```diff
--- a/pipelines/docker-build/docker-build.yaml
+++ b/pipelines/docker-build/docker-build.yaml
@@ -320,6 +320,8 @@ spec:
     params:
     - name: image-url
       value: $(tasks.build-image-index.results.IMAGE_URL)
+    - name: image-digest
+      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
     - name: IMAGE
       value: $(params.output-image)
     - name: DOCKERFILE
```

kasemAlem: thanks @kdudka 
you are right, that was to fix the patch param of sast-coverity-check

kdudka: @kasemAlem You did not include the patch from my previous comment, which I think is needed.  The changes you done in `pipelines/docker-build/patch.yaml` do not seem to be correct.  We do **not** need to define new parameters of the `sast-coverity-check` task for this.  The task already has `image-url` and `image-digest` parameters defined in `task/sast-coverity-check/0.3/patch.yaml`.  We only need to set the values in the pipeline template(s).

task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -193,12 +193,12 @@ spec:
...
           value: $(params.HOMEDIR)/snapshot.json
       image: registry.redhat.io/rhtas/ec-rhel9:0.6
-      onError: continue # progress even if the step fails so we can see the debug logs
+      onError: continue  # progress even if the step fails so we can see the debug logs
       command: [reduce-snapshot.sh]
 
     - name: validate
       image: registry.redhat.io/rhtas/ec-rhel9:0.6
-      onError: continue # progress even if the step fails so we can see the debug logs
+      onError: continue  # progress even if the step fails so we can see the debug logs
```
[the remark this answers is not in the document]
kasemAlem: yamlint was throwing a warning for due to this space , should I revert it 
and ignore any warnings ?

task/sast-snyk-check/0.4/README.md
```diff
@@ -0,0 +1,40 @@
...
+| IGNORE_FILE_PATHS  | Directories or files to be excluded from Snyk scan (Comma-separated). Useful to split the directories of a git repo across multiple components.  | ""            | false    |
+| IMP_FINDINGS_ONLY  | Report only important findings.  To report all findings, specify "false"                                                                         | true          | true     |
+| KFP_GIT_URL        | Known False Positives (KFP) git URL (optionally taking a revision delimited by \#). Defaults to "SITE_DEFAULT", which means the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux instance and empty string for external Konflux instance. If set to an empty string, the KFP filtering is disabled.|SITE_DEFAULT|false|
+| PROJECT_NAME       | Name of the scanned project, used to find path exclusions. By default, the Konflux component name will be used.                                  | ""            | false    |
+| RECORD_EXCLUDED    | Write excluded records in file. Useful for auditing.                                                                                             | false         | false    |
```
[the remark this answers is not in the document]
kasemAlem: fixed

task/sast-unicode-check/0.2/README.md
```diff
@@ -0,0 +1,30 @@
...
+| KFP_GIT_URL                  | Known False Positives (KFP) git URL (optionally taking a revision delimited by \#). Defaults to "SITE_DEFAULT", which means the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux instance and empty string for external Konflux instance. If set to an empty string, the KFP filtering is disabled.|SITE_DEFAULT|false|
+| PROJECT_NAME                 | Name of the scanned project, used to find path exclusions. If set to an empty string, the Konflux component name will be used.                | ""                                                                                              | No       |
+| RECORD_EXCLUDED              | Whether to record the excluded findings (defaults to false). If `true`, the the excluded findings will be stored in `excluded-findings.json`. | "false"
+| IMAGE_DIGEST | Image digest that will be uploaded with ORAS control.                                                                                                  | | No                                                                                          | true       |
```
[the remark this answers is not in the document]
kasemAlem: good point, 
fixed

kasemAlem:
> @kasemAlem We still need to pass the new `image-url` parameter to `sast-coverity-check` and `sast-unicode-check` in `template-build.yaml`:
> 
> ```diff
> --- a/pipelines/template-build/template-build.yaml
> +++ b/pipelines/template-build/template-build.yaml
> @@ -241,28 +241,30 @@ spec:
>      - name: sast-coverity-check
>        when:
>          - input: $(params.skip-checks)
>            operator: in
>            values: ["false"]
>          - input: $(tasks.coverity-availability-check.results.STATUS)
>            operator: in
>            values: ["success"]
>        runAfter:
>          - coverity-availability-check
>        taskRef:
>          name: sast-coverity-check
>          version: "0.3"
>        params:
> +        - name: image-digest
> +          value: $(tasks.build-image-index.results.IMAGE_DIGEST)
>          - name: image-url
>            value: $(tasks.build-image-index.results.IMAGE_URL)
>        workspaces:
>          - name: source
>            workspace: workspace
>      - name: coverity-availability-check
>        when:
>          - input: $(params.skip-checks)
>            operator: in
>            values: ["false"]
>        runAfter:
>          - build-image-index
>        taskRef:
>          name: coverity-availability-check
> @@ -288,28 +290,30 @@ spec:
>      - name: sast-unicode-check
>        when:
>          - input: $(params.skip-checks)
>            operator: in
>            values: ["false"]
>        runAfter:
>          - build-image-index
>        taskRef:
>          name: sast-unicode-check
>          version: "0.2"
>        workspaces:
>          - name: workspace
>            workspace: workspace
>        params:
> +        - name: image-digest
> +          value: $(tasks.build-image-index.results.IMAGE_DIGEST)
>          - name: image-url
>            value: $(tasks.build-image-index.results.IMAGE_URL)
>      - name: apply-tags
>        runAfter:
>          - build-image-index
>        taskRef:
>          name: apply-tags
>          version: "0.1"
>        params:
>        - name: IMAGE
>          value: $(tasks.build-image-index.results.IMAGE_URL)
>      - name: push-dockerfile
>        runAfter:
>        - build-image-index
> ```

I've added on both tasks in `template-build.yaml`,
thanks @kdudka

dirgim (dismissed):
LGTM

task/sast-unicode-check/0.2/sast-unicode-check.yaml
```diff
@@ -0,0 +1,294 @@
...
+    Scans source code for non-printable unicode characters in all text files.
+  results:
+    - description: Tekton task test output.
+      name: TEST_OUTPUT
+  params:
+    - name: image-url
+      type: string
+      description: Image URL.
+      default: ""
+    - name: IMAGE_DIGEST
```
kdudka, line 21: This change seems inconsistent.  It should have been `image-digest` I guess?

kdudka (changes requested):

task/sast-unicode-check/0.2/sast-unicode-check.yaml
kasemAlem: indeed , 
missed that one , now should be fixed

pipelines/maven-zip-build/maven-zip-build.yaml
```diff
@@ -167,13 +167,15 @@ spec:
       workspace: workspace
   - name: sast-coverity-check
     params:
-    - name: image-url
+    - name: image-digest
       value: $(tasks.build-oci-artifact.results.IMAGE_URL)
+    - name: image-url
+      value: $(tasks.build-image-index.results.IMAGE_URL)
```
kdudka, line 173: Also we are mixing results from `tasks.build-oci-artifat` and `tasks.build-image-index`.  Is this correct?

kdudka (changes requested):

kdudka (dismissed):
Looks good.  Thanks for the update!

hongweiliu17 (approved):
lgtm, thanks for the great work!

dirgim (approved):
LGTM

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
```diff
@@ -317,5 +317,5 @@ spec:
           echo "Selecting auth"
           select-oci-auth "${IMAGE_URL}" >"${HOME}/auth.json"
           echo "Attaching to ${IMAGE_URL}"
-          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
+          oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type "${MEDIA_TYPE}" "${IMAGE_URL}@${IMAGE_DIGEST}" "${UPLOAD_FILE}:${MEDIA_TYPE}"
```
mmorhun, line 320: Could it break user's pipeline if they were passing full image reference in `IMAGE_URL`?

task/sast-coverity-check-oci-ta/0.3/MIGRATION.md
```diff
@@ -0,0 +1,8 @@
+# Migration from 0.2 to 0.3
+
+- The `IMAGE_DIGEST` parameter has been added.
+
+## Action from users
+
+- NA
```
mmorhun, line 7: Could it break user's pipeline if they were passing full image reference in `IMAGE_URL`?

task/sast-coverity-check/0.3/MIGRATION.md
```diff
@@ -0,0 +1,9 @@
+# Migration from 0.2 to 0.3
+
+- The required `IMAGE_DIGEST` parameter has been added back.
+
+## Action from users
+
+- All resources used task `sast-coverity-check` should be directed to use new `0.3` version.
+- The `IMAGE_DIGEST` parameter definition is required to be added for this task in the build pipeline.
```
mmorhun, line 8: Does it make sense to [create a migration](https://github.com/konflux-ci/build-definitions?tab=readme-ov-file#task-migration) that would do it automatically?

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
kasemAlem: yes, 
user's pipeline will be broken in case of `IMAGE_URL` contains the full image

task/sast-coverity-check-oci-ta/0.3/MIGRATION.md
kasemAlem: same here

jsztuka (dismissed):
lgtm

simonbaird (approved):

kasemAlem:
> The migration overall looks good. Regarding the failure job "Check task migrations", the failure is
> 
> `Error from server (BadRequest): error when creating "/tmp/tmp.cwu3Hgbire-validate-migrations/pipelines/pushed/maven-zip-build-oci-ta.yaml.modified": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: couldn't add link between sast-coverity-check and build-image-index: task sast-coverity-check depends on build-image-index but build-image-index wasn't present in Pipeline: spec.tasks`
> 
> But, it succeeds in my local with this pr. Anyway, could you please check if the kustomization is correct? I'll make another check later.

I found the

tnevrlka:
@kasemAlem FYI I think you have an issue with your local Git config. [Your commit](https://github.com/konflux-ci/build-definitions/pull/2018/commits/338e7d8e10171574cdf32085068ed16059b950e2) is not getting associated with your GitHub account. From the signoff, it looks like you might not have your e-mail address setup correctly

task/sast-coverity-check/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,27 @@
...
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+
+# Determine the correct image-digest value based on the pipeline file
+if [[ "$pipeline_file" == "docker-build.yaml" ]]; then
+    image_digest_value="\$(tasks.build-image-index.results.IMAGE_DIGEST)"
+else
+    image_digest_value="\$(tasks.build-oci-artifact.results.IMAGE_DIGEST)"
+fi
```
tkdchen, line 16: A stable way for the purpose of this check is to check if task `build-oci-artifact` exists (my understanding is task `build-oci-artifact` is specific to the `maven-zip-build(-oci-ta)?` pipeline). The value passed to `$1` can be any string.

task/sast-coverity-check/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,27 @@
...
+
+
+# Check if image-digest parameter already exists using precise path
+if ! yq -e '.spec.tasks[] | select(.name == "sast-coverity-check").params[] | select(.name == "image-digest")' "$pipeline_file" >/dev/null; then
+    echo "Adding image-digest parameter to sast-coverity-check task"
+    yq -i "(.spec.tasks[] | select(.name == \"sast-coverity-check\")).params += [{\"name\": \"image-digest\", \"value\": \"\$image_digest_value\"}]" "$pipeline_file"
```
tkdchen, line 22: Replace `\$image_digest_value` with `$image_digest_value`

task/sast-coverity-check/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Created for task: sast-coverity-check@0.3
+# Creation time: $(date -u +"%Y-%m-%dT%H:%M:%SZ")
```
tkdchen, line 6: Please replace this with real date value

kasemAlem: fixed,

task/sast-coverity-check/0.3/migrations/0.3.sh
kasemAlem: replaced

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
mmorhun: Do you have any plan how to avoid breaking existing users?

tkdchen (commented):
Regarding the CI failure, the error is
`Error from server (BadRequest): error when creating "/tmp/tmp.tKEPH28Mzu-validate-migrations/pipelines/pushed/maven-zip-build-oci-ta.yaml.modified": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: couldn't add link between sast-coverity-check and build-image-index: task sast-coverity-check depends on build-image-index but build-image-index wasn't present in Pipeline: spec.tasks`

It is caused by the issue mentioned in thread https://github.com/konflux-ci/build-definitions/pull/2018/files#r2014124642.

In the mave-zip-build pipeline YAML, the `image-digest` is incorrect:

```yaml
    - name: sast-coverity-check
      params:
        - name: image-url
          value: $(tasks.build-oci-artifact.results.IMAGE_URL)
        - name: image-digest
          value: $(tasks.build-image-index.results.IMAGE_DIGEST)
```

tkdchen:
```yaml
    - name: sast-coverity-check
      params:
        - name: image-url
          value: $(tasks.build-image-index.results.IMAGE_URL)
        ...
        - name: image-digest
          value: ""
```

Checked result `docker-build-oci-ta` pipeline, parameter `image-digest` has this value. Is this expected?

dirgim (approved):
LGTM

tkdchen (commented):
Please take a look at https://github.com/konflux-ci/build-definitions/pull/2018#issuecomment-2757200583

tnevrlka (changes requested):
This PR is very hard to properly review.

What is the motivation behind this PR and the changes you are making? The issues in Jira you linked are not available outside of Red Hat, so there needs to be a description in the commits and/or the PR.

Please try to split the PR into more commits, so the PR can be reviewed by one logical chunk at a time. The [workaround with diff](https://github.com/konflux-ci/build-definitions/pull/2018#issuecomment-2751523206) could be avoided by committing a carbon copy of the latest existing task version and then making your changes in another commit. See https://github.com/konflux-ci/build-definitions/pull/1766/commits/dbcd3b79e6907deb3a4ad7aadd2b450ab6313212 as an example.

Please address the raised concerns by other reviewers.

Thank you for your contribution!

kasemAlem:
> ```yaml
>  sast-coverity-check
> ```

can you please add link to it,  I've reviewed the `pipelines/docker-build-oci-ta.yaml` , I couldn't find where it have `value: ""`
maybe I'm looking in the wrong place ?

task/sast-coverity-check/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,36 @@
...
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+# Determine the correct image-digest and image-url values based on task presence
+if yq -e '.spec.tasks[] | select(.name == "build-oci-artifact")' "$pipeline_file" >/dev/null; then
+    image_digest_value="$(yq '.tasks[] | select(.name == "build-oci-artifact").results[] | select(.name == "IMAGE_DIGEST").value' "$pipeline_file")"
+    image_url_value="$(yq '.tasks[] | select(.name == "build-oci-artifact").results[] | select(.name == "IMAGE_URL").value' "$pipeline_file")"
+elif yq -e '.spec.tasks[] | select(.name == "build-image-index")' "$pipeline_file" >/dev/null; then
+    image_digest_value="$(yq '.tasks[] | select(.name == "build-image-index").results[] | select(.name == "IMAGE_DIGEST").value' "$pipeline_file")"
+    image_url_value="$(yq '.tasks[] | select(.name == "build-image-index").results[] | select(.name == "IMAGE_URL").value' "$pipeline_file")"
```
tkdchen, line 16: Why get these values from task result? The passed-in pipelines are static YAML, there is no result.

tkdchen, line 16: In addition, there are missing `.spec` in above 4 YAML path.

kasemAlem: yes, 
missed by mistake, fixed in all places

task/sast-coverity-check/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,36 @@
...
+if yq -e '.spec.tasks[] | select(.name == "build-oci-artifact")' "$pipeline_file" >/dev/null; then
+    image_digest_value="$(yq '.spec.tasks[] | select(.name == "build-oci-artifact") | select(.name == "IMAGE_DIGEST").value' "$pipeline_file")"
+    image_url_value="$(yq '.spec.tasks[] | select(.name == "build-oci-artifact") | select(.name == "IMAGE_URL").value' "$pipeline_file")"
+elif yq -e '.spec.tasks[] | select(.name == "build-image-index")' "$pipeline_file" >/dev/null; then
+    image_digest_value="$(yq '.spec.tasks[] | select(.name == "build-image-index") | select(.name == "IMAGE_DIGEST").value' "$pipeline_file")"
+    image_url_value="$(yq '.spec.tasks[] | select(.name == "build-image-index") | select(.name == "IMAGE_URL").value' "$pipeline_file")"
+else
+    echo "Neither build-oci-artifact nor build-image-index tasks found."
+    exit 1
+fi
```
tkdchen, line 20:
```suggestion
if yq -e '.spec.tasks[] | select(.name == "build-oci-artifact")' "$pipeline_file" >/dev/null; then
    image_digest_value="\$(tasks.build-oci-artifact.results.IMAGE_DIGEST)"
    image_url_value="\$(tasks.build-oci-artifact.results.IMAGE_URL)"
elif yq -e '.spec.tasks[] | select(.name == "build-image-index")' "$pipeline_file" >/dev/null; then
    image_digest_value="\$(tasks.build-image-index.results.IMAGE_DIGEST)"
    image_url_value="\$(tasks.build-image-index.results.IMAGE_URL)"
else
    echo "Neither build-oci-artifact nor build-image-index tasks found."
    exit 0
fi
```

I think these values are what you need. Meanwhile, it's unnecessary to exit with 1 if neither is found. It's actually not an error. It's okay to just log a message properly then terminate the execution normally.

kasemAlem: I made the fix of the script you suggested , when I run `./hack/validate-migrations.sh ` it trow the below error : 

```
info: prepare Konflux standard pipelines
info: fetch pipeline fbc-builder from bundle quay.io/konflux-ci/tekton-catalog/pipeline-fbc-builder:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/fbc-builder.yaml
info: fetch pipeline docker-build from bundle quay.io/konflux-ci/tekton-catalog/pipeline-docker-build:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/docker-build.yaml
info: fetch pipeline docker-build-oci-ta from bundle quay.io/konflux-ci/tekton-catalog/pipeline-docker-build-oci-ta:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/docker-build-oci-ta.yaml
info: fetch pipeline docker-build-multi-platform-oci-ta from bundle quay.io/konflux-ci/tekton-catalog/pipeline-docker-build-multi-platform-oci-ta:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/docker-build-multi-platform-oci-ta.yaml
info: fetch pipeline maven-zip-build from bundle quay.io/konflux-ci/tekton-catalog/pipeline-maven-zip-build:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/maven-zip-build.yaml
info: fetch pipeline maven-zip-build-oci-ta from bundle quay.io/konflux-ci/tekton-catalog/pipeline-maven-zip-build-oci-ta:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/maven-zip-build-oci-ta.yaml
info: fetch pipeline tekton-bundle-builder from bundle quay.io/konflux-ci/tekton-catalog/pipeline-tekton-bundle-builder:922d8592c0ac5a5d43ff9b196f49228f87dfcd01 -> /tmp/tmp.Exg7jlCR17-validate-migrations/pipelines/pushed/tekton-bundle-builder.yaml
Traceback (most recent call last):
  File "<string>", line 3, in <module>
ModuleNotFoundError: No module named 'ruamel'
```

tkdchen: python3-ruamel-yaml package is required.

kasemAlem: so we need to add it to the workflow requirements as part of preparations ?

kasemAlem: I see that you added to the [script validation-migration.sh ](https://github.com/konflux-ci/build-definitions/blob/ae264d37c028b12e24d32f3326d808cab75d0f79/hack/validate-migration.sh#L53-L68) 
this is a better way printing out the  diffs 
thanks @tkdchen

tkdchen (dismissed):

tnevrlka:
Hi @kasemAlem. 
- What is the resolution of [Mykola's comments](https://github.com/konflux-ci/build-definitions/pull/2018#discussion_r2014345063)?
- The PR currently contains one feature commit and two merge commits updating the branch. Please rebase, get rid of the merge commits and try to split the feature commit.
- Thank you for updating the PR description. Please add a description to the commits too.

Thank you!

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
tkdchen: I missed this comment. Should making a new version 0.3 be enough? Users who are using older versions receive Renovate update PR and have a chance to upgrade to the latest version.

task/sast-coverity-check/0.3/MIGRATION.md
kasemAlem: Fixed and script is added and approved

kasemAlem:
> Hi @kasemAlem.
> 
> * What is the resolution of [Mykola's comments](https://github.com/konflux-ci/build-definitions/pull/2018#discussion_r2014345063)?
> * The PR currently contains one feature commit and two merge commits updating the branch. Please rebase, get rid of the merge commits and try to split the feature commit.
> * Thank you for updating the PR description. Please add a description to the commits too.
> 
> Thank you!

> * What is the resolution of [Mykola's comments](https://github.com/konflux-ci/build-definitions/pull/2018#discussion_r2014345063)?

I see that @tkdchen replied this and it sounds reasonable to me too 


> * Thank you for updating the PR description. Please add a description to the commits too.
Screenshot from 2025-04-06 19-49-58

task/sast-coverity-check-oci-ta/0.1/sast-coverity-check-oci-ta.yaml
tnevrlka: @kasemAlem You've created the version 0.3 of `sast-coverity-check-oci-ta` that includes the `IMAGE_DIGEST` change. Should the same change be included in this 0.1 version though?

task/clamav-scan/0.2/clamav-scan.yaml
```diff
@@ -202,7 +202,7 @@ spec:
         echo "Selecting auth"
         select-oci-auth $IMAGE_URL > $HOME/auth.json
         echo "Attaching to ${IMAGE_URL}"
-         oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type application/vnd.clamav "${IMAGE_URL}" "${args[@]}"
+         oras attach --no-tty --registry-config "$HOME/auth.json" --artifact-type application/vnd.clamav "${IMAGE_URL}@${IMAGE_DIGEST}" "${args[@]}"
```
tnevrlka, line 205: Is [this issue](https://github.com/konflux-ci/build-definitions/pull/2018/files#r2011681593) going to be a problem here too? (same for 0.1)

task/sast-coverity-check-oci-ta/0.3/MIGRATION.md
tnevrlka, line 7: Is there really no action required from users?

tnevrlka (changes requested):
Thank you for addressing many of the comments! 
I've left a few more comments and I think it would be great if you split the PR into multiple commits as described in [this review](https://github.com/konflux-ci/build-definitions/pull/2018#pullrequestreview-2720515156)

kdudka:
@kasemAlem Please make sure to include the fixes from #2142 in the 0.3 version of `sast-coverity-check*` so that users do not experience a regression after migrating to the next version.

task/clamav-scan/0.2/clamav-scan.yaml
kasemAlem: it can be a problem , as I see it's a users responsibility to enter correct parameters to task 
unless I'm missing something

tnevrlka:
> @kasemAlem Please make sure to include the fixes from #2142 in the 0.3 version of `sast-coverity-check*` so that users do not experience a regression after migrating to the next version.

Kudos to Kamil, great point. 

I would like to note that splitting the PR into several commits makes similar changes much easier. 
When using [this PR](https://github.com/konflux-ci/build-definitions/pull/1766/commits) as a reference (not my proudest work, but good enough for demonstrating my point), the possible workflow would be something like:

1. Rebase on main in order to get the latest changes
2. Do an interactive rebase and edit the first "scaffolding" commit. Make it so the old version and the new version are the same. So that the changes that were pulled in step 1 are included in the new version.
3. If there are no conflicts, you should be able to finish the interactive rebase

Please let me know if the explanation is not clear or you encounter any trouble

kasemAlem:
blocked by https://github.com/konflux-ci/build-definitions/pull/2181/files

task/sast-coverity-check/0.3/MIGRATION.md
hongweiliu17: Should we also add the script to the updated task folders as well, especially for the required tasks?

tnevrlka:
I'm taking a look at the CI failure

task/sast-coverity-check/0.3/patch.yaml
```diff
@@ -0,0 +1,491 @@
...
+  value: push
+- op: remove
+  path: /spec/steps/2
+
+# icm
+- op: test
+  path: /spec/steps/1/name
+  value: icm
+- op: remove
+  path: /spec/steps/1
```
tnevrlka, line 63: Why is this being added compared to `sast-coverity-check/0.2`?

kdudka: We discussed this on Slack with @kasemAlem.  I think this pull request needs to be rebased onto #2188 (which itself is broken because it caused [KFLUXSPRT-2702](https://issues.redhat.com/browse/KFLUXSPRT-2702)).

kasemAlem: thnaks @tnevrlka and @kdudka 
@kdudka already found it as he reported in ticket   [KONFLUX-6272](https://issues.redhat.com/browse/KONFLUX-6272) , I've rebased and made the changes in` 0.3/patch.yaml `to be aligned with `0.2/patch.yaml`  changes

kasemAlem: @kdudka @tnevrlka , cab I get approvals on this PR

task/sast-coverity-check/0.3/README.md
```diff
@@ -0,0 +1,47 @@
...
+| COV_ANALYZE_ARGS          | Append arguments to the cov-analyze CLI command                                                                                       | ""                        | no       |
+| COV_LICENSE               | Name of secret which contains the Coverity license                                                                                    | cov-license               | no       |
+| AUTH_TOKEN_COVERITY_IMAGE | Name of secret which contains the authentication token for pulling the Coverity image                                                 | auth-token-coverity-image | no       |
+| IMP_FINDINGS_ONLY         | Report only important findings. Default is true. To report all findings, specify "false"                                              | true                      | no       |
+| KFP_GIT_URL         | Report only important findings. Default is true. To report all findings, specify "false"                                              | SITE_DEFAULT                      | false       |
```
kdudka, line 28: This line is wrong and should be removed.  The description is for `IMP_FINDINGS_ONLY`, not for `KFP_GIT_URL`.  The next line needs to be updated to document `SITE_DEFAULT` as done in the 0.2 version of the task.

kdudka (changes requested):

task/sast-coverity-check/0.3/README.md
kdudka: Sorry for the confusion!  I am fine with keeping the `README.md` file in the 0.3 version of the task identical with the 0.2 version (because the `image-url` and `image-digest` params have been undocumented since the beginning).  But the change I pointed out is definitely not acceptable and needs to be removed from this merge request:
```diff
@@ -23,7 +25,8 @@
 | COV_LICENSE               | Name of secret which contains the Coverity license                                                                                    | cov-license               | no       |
 | AUTH_TOKEN_COVERITY_IMAGE | Name of secret which contains the authentication token for pulling the Coverity image                                                 | auth-token-coverity-image | no       | 
 | IMP_FINDINGS_ONLY         | Report only important findings. Default is true. To report all findings, specify "false"                                              | true                      | no       |
-| KFP_GIT_URL               | Known False Positives (KFP) git URL (optionally taking a revision delimited by \#). Defaults to "SITE_DEFAULT", which means the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux instance and empty string for external Konflux instance. If set to an empty string, the KFP filtering is disabled.|SITE_DEFAULT|false|
+| KFP_GIT_URL         | Report only important findings. Default is true. To report all findings, specify "false"                                              | SITE_DEFAULT                      | false       |
+| KFP_GIT_URL               | Known False Positives git URL, optionally taking a revision delimited by #; If empty, filtering of known false positives is disabled. | ""                        | no       |
 | PROJECT_NAME              | Name of the scanned project, used to find path exclusions. By default, the Konflux component name will be used.                       | ""                        | no       |
 | RECORD_EXCLUDED           | If set to `true`, excluded findings will be written to a file named `excluded-findings.json` for auditing purposes.                   | false                     | no       |
```

task/sast-snyk-check/0.4/sast-snyk-check.yaml
```diff
@@ -0,0 +1,307 @@
...
+    - name: trusted-ca
+      configMap:
+        name: $(params.caTrustConfigMapName)
+        items:
+          - key: $(params.caTrustConfigMapKey)
+            path: ca-bundle.crt
+        optional: true
+  steps:
+    - name: sast-snyk-check
+      image: quay.io/konflux-ci/konflux-test:v1.4.20@sha256:f9db697d8a45870b862252de61b3c29d9d6f79272ef8bf61ecb645f8bca27705
```
tnevrlka, line 83: This is an older version compared to the `v1.4.23` the current `0.3/sast-snyk-check` is using

task/sast-snyk-check/0.4/sast-snyk-check.yaml
```diff
@@ -0,0 +1,307 @@
...
+
+        SNYK_TOKEN_PATH="/etc/secrets/snyk_token"
+        if [ -f "${SNYK_TOKEN_PATH}" ] && [ -s "${SNYK_TOKEN_PATH}" ]; then
+          # SNYK token is provided
+          SNYK_TOKEN="$(cat ${SNYK_TOKEN_PATH})"
+          export SNYK_TOKEN
+        else
+          # According to shellcheck documentation, the following error can be ignored as it is ignored through indirection: https://www.shellcheck.net/wiki/SC2034
+          # shellcheck disable=SC2034
+          to_enable_snyk='[here](https://konflux-ci.dev/docs/how-tos/testing/build/snyk/)'
```
tnevrlka, line 147: Regression: the URL is invalid, it's correct in `0.3/sast-snyk-check`

task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
```diff
@@ -0,0 +1,329 @@
...
+        optional: true
+    - name: workdir
+      emptyDir: {}
+  stepTemplate:
+    volumeMounts:
+      - mountPath: /var/workdir
+        name: workdir
+  steps:
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:9b180776a41d9a22a1c51539f1647c60defbbd55b44bbebdd4130e33512d8b0d
```
tnevrlka, line 103: Wrong image URL

tnevrlka (changes requested):
I've found few small regressions (+probably some more), hopefully they should be resolved by updating the PR

task/sast-snyk-check/0.4/sast-snyk-check.yaml
tnevrlka: Instead of making changes one at a time (meaning all of the regressions would have to be found and changed manually), I would recommend reapplying your changes to the newest files from upstream/main.

task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
kasemAlem: fixed

task/sast-snyk-check/0.4/sast-snyk-check.yaml
kasemAlem: fixed

kasemAlem:
> @kasemAlem btw how are you creating the README.md files for tasks? They should be automatically generated with `hack/generate-readme.sh`. To be honest, I thought it was included in CI, but it seems like only PipelineRun READMEs are checked.

you are right,  I had to copy from former version then modify manually

tkdchen (approved):
The migration LGTM

dirgim:
Hi @kdudka can you re-review your requested changes and see if the PR looks good to you now?

task/sast-snyk-check/0.4/sast-snyk-check.yaml
```diff
@@ -0,0 +1,307 @@
...
+      default: snyk-secret
+    - name: ARGS
+      type: string
+      description: Append arguments.
+      default: ""
+    - description: Image URL.
+      name: image-url
+      type: string
+    - name: image-digest
+      description: Digest of the image to scan.
```
kdudka, line 34: The description is misleading.  `sast-snyk-check` does not scan the given image.  These params are only used to specify to which image the scan results should be attached to.

task/sast-snyk-check/0.3/sast-snyk-check.yaml
```diff
@@ -46,11 +46,7 @@ spec:
       default: "true"
     - name: KFP_GIT_URL
       type: string
-      description: Known False Positives (KFP) git URL (optionally taking
-        a revision delimited by \#). Defaults to "SITE_DEFAULT", which means
-        the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux
-        instance and empty string for external Konflux instance.
-        If set to an empty string, the KFP filtering is disabled.
+      description: Known False Positives (KFP) git URL (optionally taking a revision delimited by \#). Defaults to "SITE_DEFAULT", which means the default value "https://gitlab.cee.redhat.com/osh/known-false-positives.git" for internal Konflux instance and empty string for external Konflux instance. If set to an empty string, the KFP filtering is disabled.
```
kdudka, line 49: This change was not needed and introduces an unnecessary difference between the 0.3 and 0.4 versions of the task.

kdudka (approved):
Thanks for the update!  I spotted only some minor issues and a few image references being downgraded in a newer version of Tekton tasks, which will hopefully be fixed automatically by the Renovate bot.  I propose to merge this as it is to avoid additional divergence.

cuipinghuo:
not really change behaviors of conforma in [task/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml](https://github.com/konflux-ci/build-definitions/pull/2018/files#diff-6e14df40f8ecb3b4b1b8de218f16acbfdb62a97f29321b8cd600a3be38097f3e).

PR #2021, opened 2025-03-10
feat: stop checking related image accessibility in validate-fbc

arewm:
With EC-965, we created a policy which can verify a set of related images that have been idenified by this task and attached to the FBC image. Since we can check validity with Conforma, we no longer need to check validity here.

This enables users to produce a FBC catalog while ignoring unreleased related images until it actually matters (i.e. until the fragment is about to be pushed to production).

### DIFF
--- a/task/validate-fbc/0.1/validate-fbc.yaml
+++ b/task/validate-fbc/0.1/validate-fbc.yaml
@@ -450,27 +450,10 @@ spec:
         else
           echo -n "$related_images" > /shared/related-images.json
 
-          # A while loop executed in a subshell so we will check for failure based
-          # on whether /tmp/failedimages exists.
+          # We have moved verification of the related images to a Conformal policy. This enables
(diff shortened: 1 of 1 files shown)


dirgim (approved):
LGTM

task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -450,27 +450,10 @@ spec:
         else
           echo -n "$related_images" > /shared/related-images.json
 
-          # A while loop executed in a subshell so we will check for failure based
-          # on whether /tmp/failedimages exists.
+          # We have moved verification of the related images to a Conformal policy. This enables
```
grokspawn, line 453: s/Conformal/Conforma/ ?

arewm: Yes, that was a mistake but I don't think it is enough to warrant a new task version.

PR #2025, opened 2025-03-10
fix(KONFLUX-7202): access IDMS only if original pullspec is unavailable

yashvardhannanavati:
currently, the stepaction directly tries to access the mirror from IDMS, if present. This commit fixes that behavior by accessing the original pullspec first and resorting to the mirror only if required. It also allows users to provide more than one mirrors for their images.


arewm (approved):
Thanks

nmars (approved):
LGTM

PR #2026, opened 2025-03-10
access IDMS in FIPS tasks only if required

yashvardhannanavati:
like the stepaction, the tasks also used IDMS directly instead of
attempting to access the original pullspec. This commit fixes that
behavior and uses the updated stepaction. It also allows users to
provide more than one mirrors

Refers to [KONFLUX-7202](https://issues.redhat.com//browse/KONFLUX-7202) and [KONFLUX-7203](https://issues.redhat.com//browse/KONFLUX-7203)


nmars (approved):
LGTM

PR #2027, opened 2025-03-11
coverity-availability-check: result status ERROR -> FAILURE

sfowl:
The intention of this change is the coverity-availability-check task should report failure status to users, to indicate that action is needed (add coverity secrets), but it should NOT cause the pipeline to fail.


kdudka (approved):
Looks good.  My only worry is that we should not cause any disruption for users of public Konflux clusters, who will never be able to add the secrets.

hongweiliu17 (approved):
lgtm

PR #2029, opened 2025-03-11
fix: avoid attaching migration to irrelevant task bundle

tkdchen:
[STONEBLD-3353](https://issues.redhat.com//browse/STONEBLD-3353)

The problem is:

* commit 1 includes a migration, which is attached to the task bundle built from commit 1.
* then, commit 2 includes updates to the same task without bumping the version in label '.metadata.labels."app.kubernetes.io/version'.
* as a result, the migration is attached to the last bundle.

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -530,7 +530,9 @@ build_push_tasks() {
         migration_file="${task_dir}/migrations/${concrete_task_version}.sh"
 
         has_migration=false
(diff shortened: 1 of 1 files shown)


mmorhun (commented):
Please add a comment in the code that explains the condition

tkdchen:
> Please add a comment in the code that explains the condition

Updated. PTAL.

mmorhun (approved):

PR #2032, opened 2025-03-11
fix: generated files

yftacherzog:
Running generate-everything.sh with no changes results in changes. This fails CI for other PRs.


tisutisu (approved):
LGTM

PR #2034, opened 2025-03-11
sast-coverity-check: request 4 GiB of RAM

kdudka:
... in the post-process step, too, and do not set any limit on CPUs.

The increased resources are needed in the `post-process` step, too, because Coverity runs there in case the buildful capture fails.  We have reports about Coverity being OOM killed in the step.  I wish we could ask for more memory and CPUs but then the `sast-coverit-check` task is not started at all.

Related: https://github.com/konflux-ci/build-definitions/pull/1821
Related: https://issues.redhat.com/browse/KFLUXSPRT-2167
Related: https://issues.redhat.com/browse/KFLUXSPRT-2181

### DIFF
--- a/task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
+++ b/task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
@@ -733,7 +733,6 @@ spec:
         done
       computeResources:
         limits:
-          cpu: "16"
           memory: 16Gi
         requests:
           cpu: "4"
@@ -913,7 +912,7 @@ spec:
         fi
       computeResources:
         limits:
-          memory: 4Gi
+          memory: 16Gi
         requests:
-          cpu: "2"
-          memory: 2Gi
+          cpu: "4"
+          memory: 4Gi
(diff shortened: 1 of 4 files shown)


dirgim (approved):
LGTM

task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
```diff
@@ -733,7 +733,6 @@ spec:
         done
       computeResources:
         limits:
-          cpu: "16"
```
sfowl, line 736: Are you sure removing this won't fallback to using the default value? Maybe set by a LimitRange?

kdudka: @sfowl This is exactly what I asked on #1821 when this change was applied to all the other tasks and these were the answers I was given: https://github.com/konflux-ci/build-definitions/pull/1821#discussion_r1916242807

sfowl: nice diagram!

sfowl (approved):

tnevrlka (approved):

PR #2043, opened 2025-03-12
KFLUXDP-159: Add stepaction to gather cluster resources

JoaoFula:
Add the stepaction to gather the cluster resources by running the gather-extra.sh script and give the user the possibility to run a script of their own that gathers resources that matter to their team.


JoaoFula:
Moved to https://github.com/konflux-ci/tekton-integration-catalog/pull/133

PR #2050, opened 2025-03-13
feat(KONFLUX-7127): Setup limits and resource for ecosystem-cert-preflight-checks task

rh-rahulshetty:
### Description

This PR updates resource limits and requests for the task **ecosystem-cert-preflight-checks** as part of this effort - [KONFLUX-7127](https://issues.redhat.com/browse/KONFLUX-7127). 

The new request/limit values have been defined taking into account usages observed historically (tracked in this [spreadsheet](https://docs.google.com/spreadsheets/d/1d4Tn7PYkY2EVgpNyT29xL7y62akZNPIjy18VTzw54iQ/edit?gid=1735880927#gid=1735880927)) in the **stone-prd-rh01** cluster.

### DIFF
--- a/task/ecosystem-cert-preflight-checks/0.1/ecosystem-cert-preflight-checks.yaml
+++ b/task/ecosystem-cert-preflight-checks/0.1/ecosystem-cert-preflight-checks.yaml
@@ -23,7 +23,7 @@ spec:
     - name: check-container
       image: quay.io/opdev/preflight:stable@sha256:3f5b988200a3fd490e0a102a4851e0a81f2f41e724e7482a726b79b9d5ccd3b6
       args: ["check", "container", "$(params.image-url)"]
-      computeResources:
+      resources:
         limits:
           memory: 4Gi
         requests:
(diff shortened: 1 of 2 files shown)


arewm:
Do we know what reasonable requests/limits should be for this task? @jhutar proposed some different numbers in #2055.

rh-rahulshetty:
> Do we know what reasonable requests/limits should be for this task? @jhutar proposed some different numbers in #2055.

In my PR, I've set higher resource limits to prevent potential OOM issues or throttling, ensuring the task runs smoothly under varying conditions. However, if the task doesn't require such high limits, I'm open to adjusting them based on @jhutar's suggestions. Let me know your thoughts!

rh-rahulshetty:
Hello @arewm

Please help with review this change and to re-run the PR tests.

arewm:
I suspect that this PR is hitting issues similar to some others. It isn't specific to changes being introduced here.

Since these changes are just trying to increase the limits by default to improve user experience, I am going to just push it through without CI as there is no functionality change.

PR #2053, opened 2025-03-13
Add bootc-rechunk task to build-definitions

gursewak1997:
Add the `bootc-rechunk` task to `build-definitions` to enable the build pipeline to run the [rechunker](https://coreos.github.io/rpm-ostree/experimental-build-chunked-oci/) as a separate Konflux task.

### DIFF
--- a/task/bootc-rechunk/0.1/README.md
+++ b/task/bootc-rechunk/0.1/README.md
@@ -1,0 +1,22 @@
+# bootc-rechunk Task
+
+## Overview
+The `bootc-rechunk` task is a Tekton task for running the `rpm-ostree experimental build-chunked-oci` command as a separate step in the pipeline. It generates a chunked OCI image from the root filesystem.
+
+## Parameters
+- **`IMAGE_TAG`** (string) - The tag for the OCI image.
+- **`OUTPUT_PATH`** (string) - The directory to store the output OCI image (default: `/buildcontext`).
+
+## Usage
+Add this task to your Tekton pipeline to run the `bootc-rechunk` step before the final image build.
+
+Example:
+```yaml
+- name: bootc-rechunk
+  taskRef:
+    name: bootc-rechunk
+  params:
+    - name: IMAGE_TAG
+      value: "$(params.IMAGE_TAG)"
+    - name: OUTPUT_PATH
+      value: "/buildcontext"
--- a/task/bootc-rechunk/0.1/bootc-rechunk.yaml
+++ b/task/bootc-rechunk/0.1/bootc-rechunk.yaml
@@ -1,0 +1,26 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: bootc-rechunk
+  namespace: konflux-ci
+spec:
+  params:
+    - name: IMAGE_TAG
+      type: string
+      description: "Tag for the OCI image"
+    - name: OUTPUT_PATH
+      type: string
+      description: "Path to store the chunked OCI image"
+      default: "/buildcontext"
+  steps:
+    - name: rechunk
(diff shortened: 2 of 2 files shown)


task/bootc-rechunk/0.1/README.md
```diff
@@ -0,0 +1,45 @@
+# Bootc-Rechunk task
+
+## Overview
+
+The `bootc-rechunk` task is a Tekton task used to chunk an OCI image into smaller chunks for more efficient management, with the goal of reducing image size. This task is designed to work in conjunction with other container build tasks (e.g., `buildah-oci-ta`). It takes an existing image, processes it, and outputs a chunked OCI archive.
+
+## Inputs
+
+- **IMAGE**: The reference of the image to be chunked (required). This image will be used as the input for the rechunking process.
```
cgwalters, line 9: This means we pull it from a registry always?

task/bootc-rechunk/0.1/README.md
```diff
@@ -0,0 +1,45 @@
...
+  
+## Outputs
+
+- **OCI Archive**: A chunked OCI image that is output after the process completes. The output is saved to the specified path in the task parameters.
```
cgwalters, line 13: There's several different output formats here; an OCI archive (aka `oci-archive:` in [containers-transport terms](https://github.com/containers/image/blob/main/docs/containers-transports.5.md)) is just one.

I think it's generally better to output `oci:` personally.

Although this relates to the above - if we pull from a registry it'd probably make sense to push to a registry?

task/bootc-rechunk/0.1/bootc-rechunk.yaml
```diff
@@ -0,0 +1,28 @@
...
+        #!/bin/sh
+        set -eux
+        mkdir -p $(params.OUTPUT_PATH)
+        
+        # Running the rechunking process
+        rpm-ostree experimental compose build-chunked-oci \
+          --bootc \
+          --format-version=1 \
+          --rootfs /target-rootfs \
+          --output oci-dir:$(params.OUTPUT_PATH)/out.ociarchive
```
cgwalters, line 28: There's no `oci-dir:` (AFAIK) only `oci:`.

Also but most importantly...we need to step through and consider how this should work in the context of a pipeline. There's two big options:

- Fetch the input image from a registry, and output to a registry (note this will require indirecting through a local path)
- Fetch the input image from a local path (maybe a `containers-storage:` or maybe `oci:`), and output to a local path, and then push from there to a registry

I think we can safely assume in general right now we're operating after e.g. the `buildah-remote-oci-ta` task, which always outputs to a registry.

So it will be a bit silly of course to build an image, push it to a registry, then pull it, and re-push it again...but probably not fatal.

If we care to fix that in the medium term, the simplest thing may be to add an `rpmostree-rechunk: bool` flag to `buildah-remote-oci-ta`.

arewm: How is the output of this task used? Are we meaning to push it to a registry or store it in a PVC?

cgwalters: The intent is to use this right after `buildah-oci-remote-ta` to push a per-arch chunked image, which we then can just pass to `task-build-image-manifest` to stitch together, so today that has to be done via push to registry.

arewm: Will this use trusted artifacts to share data between tasks or a PVC? Or should this be functionality that is injected as a new step in the buildah tasks to optionally chunk it?

I am asking because I assume it would also have some impact on our ability to generate provenance and verify the build chain for these artifacts.

MartinBasti:
Please add CODEOWNERS

PR #2055, opened 2025-03-13
feat: Set default compute resources for ecosystem-cert-preflight-checks task steps

jhutar:


jhutar:
Oops, Rahul already have similar PR: https://github.com/konflux-ci/build-definitions/pull/2050

PR #2065, opened 2025-03-14
Don't skip prefetch by default

ralphbean:
I think it may be confusing users that if they pass `""` then the task is skipped (which may cause conforma to fail with "no skipped tasks"), even when they have no _real_ content that they want to prefetch, say for a data container like a bundle or something that just contains basic shell scripts.

With this, the prefetch task will run, but with the input of `""`, it will exit early with exit code 0.

Related: https://github.com/release-engineering/rhtap-ec-policy/pull/128

### DIFF
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -219,11 +219,6 @@ spec:
     taskRef:
       name: prefetch-dependencies
       version: "0.2"
-    when:
-    - input: $(params.prefetch-input)
-      operator: notin
-      values:
-      - ""
     workspaces:
     - name: source
       workspace: workspace
--- a/pipelines/docker-build-oci-ta/patch.yaml
+++ b/pipelines/docker-build-oci-ta/patch.yaml
@@ -75,8 +75,6 @@
     value: $(params.image-expires-after)
 - op: remove
   path: /spec/tasks/2/workspaces/0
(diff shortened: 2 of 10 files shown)


tkdchen (approved):

pipelines/core-services/core-services.yaml
```diff
@@ -219,11 +219,6 @@ spec:
     taskRef:
       name: prefetch-dependencies
       version: "0.2"
-    when:
-    - input: $(params.prefetch-input)
-      operator: notin
-      values:
-      - ""
```
arewm, line 226: I'll make the comment here then since the other PR was closed as a duplicate. Should we make these conditional operations consistent with the build tasks? i.e. 
```suggestion
    when:
    - input: $(tasks.init.results.build)
      operator: in
      values:
      - "true"
```

simonbaird (dismissed):

pipelines/core-services/core-services.yaml
simonbaird: IIUC the goal is to always run prefetch, so I'm in favor of removing the when condition entirely, (i.e. this PR as it is now), rather than reworking the logic.

arewm (approved):

avi-biton (approved):
lgtm

chmeliik (approved):

PR #2074, opened 2025-03-17
feat(KONFLUX-7125): Change buildah task to use overlay2 storage driver

zxiong:
It aims to make it able to build image with overlay2 storage driver.

As of right now, buildah-remote-oci-it can only build images using the VFS storage driver. We'd like it's able to support overlay2 as well, because overlay2 performs better than VFS in performance. We did a performance test to compare both (see https://issues.redhat.com/browse/KONFLUX-6577).

Arches | Duration - vfs | Duration overlay2 | % Change
-- | -- | -- | --
linux/amd64 | 0:00:34 | 0:00:20 | -41.52%
linux/arm64 | 0:01:57 | 0:01:12 | -38.30%
linux/ppc64le | 0:09:05 | 0:05:59 | -34.18%

Above result shows, compared to vfs storage drivers, duration of building image with overlay2 is reduced by around 30% to 40%.

**Note that the PR is not completed yet**.  It requires "buildah-task" modification [1], which makes buildah run with overlay2 storage by default, and "rh-syft" change[2], which can execute "buildah mount $container" to feed data for syft. 

After the two dependencies complete, the buildah-task image used by "build" step will be updated with the new image generated by "buildah-task" modification [1], and the "syft-rhel9" image used by "sbom-syft-generate" step will be replaced by the new image generated by "rh-syft" change[2] as well.

[1] https://github.com/konflux-ci/buildah-container/pull/116.
[2] https://github.com/redhat-appstudio/rh-syft/pull/124.

### DIFF
--- a/task/buildah-min/0.4/buildah-min.yaml
+++ b/task/buildah-min/0.4/buildah-min.yaml
@@ -105,7 +105,7 @@ spec:
       as per --squash
     name: SQUASH
     type: string
-  - default: vfs
+  - default: overlay
     description: Storage driver to configure for buildah
     name: STORAGE_DRIVER
     type: string
@@ -368,7 +368,7 @@ spec:
         UNSHARE_ARGS+=("--net")
 
         for image in $BASE_IMAGES; do
-          unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -- buildah pull "$image"
+          unshare -Ufp --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 --mount -- buildah pull "$image"
         done
         echo "Build will be executed with network isolation"
       fi
@@ -538,22 +538,15 @@ spec:
       echo "[$(date --utc -Ins)] Run buildah build"
       echo "[$(date --utc -Ins)] ${command}"
 
-      unshare -Uf "${UNSHARE_ARGS[@]}" --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w "${SOURCE_CODE_DIR}/$CONTEXT" -- sh -c "$command"
+      unshare -Uf "${UNSHARE_ARGS[@]}" --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w "${SOURCE_CODE_DIR}/$CONTEXT" --mount -- sh -c "$command"
 
       echo "[$(date --utc -Ins)] Add metadata"
-
-      container=$(buildah from --pull-never "$IMAGE")
 
       # Save the SBOM produced by Cachi2 so it can be merged into the final SBOM later
       if [ -f "/tmp/cachi2/output/bom.json" ]; then
         echo "Making copy of sbom-cachi2.json"
         cp /tmp/cachi2/output/bom.json ./sbom-cachi2.json
       fi
-
-      buildah mount "$container" | tee /shared/container_path
-      # delete symlinks - they may point outside the container rootfs, messing with SBOM scanners
-      find $(cat /shared/container_path) -xtype l -delete
-      echo $container > /shared/container_name
 
       touch /shared/base_images_digests
       echo "Recording base image digests used"
@@ -565,6 +558,10 @@ spec:
           echo "$image $base_image_digest" | tee -a /shared/base_images_digests
         fi
       done
+
+      image_name=$(echo "${IMAGE##*/}" | tr ':' '-')
+      buildah push "$IMAGE" oci-archive:"/shared/$image_name.tar"
+      echo "/shared/$image_name.tar" > /shared/container_path
 
       echo "[$(date --utc -Ins)] End build"
     securityContext:
@@ -687,8 +684,8 @@ spec:
 
       echo "Running syft on the source directory"
       syft dir:"$(workspaces.source.path)/$SOURCE_CODE_DIR/$CONTEXT" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-source.json"
-      echo "Running syft on the image filesystem"
-      syft dir:"$(cat /shared/container_path)" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-image.json"
+      echo "Running syft on the image"
+      syft oci-archive:"$(cat /shared/container_path)" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-image.json"
 
       echo "[$(date --utc -Ins)] End sbom-syft-generate"
     volumeMounts:
@@ -696,6 +693,10 @@ spec:
       name: varlibcontainers
     - mountPath: /shared
       name: shared
+    - mountPath: /etc/pki/tls/certs/ca-bundle.crt
+      name: trusted-ca
+      readOnly: true
+      subPath: ca-bundle.crt
     workingDir: $(workspaces.source.path)/source
   - args:
     - --additional-base-images
--- a/task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
+++ b/task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
@@ -129,7 +129,7 @@ spec:
       as per --squash
     name: SQUASH
     type: string
-  - default: vfs
+  - default: overlay
     description: Storage driver to configure for buildah
     name: STORAGE_DRIVER
     type: string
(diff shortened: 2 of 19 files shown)


zxiong:
@mmorhun @MartinBasti @mshaposhnik please review it.

mshaposhnik (approved):

MartinBasti:
`./hack/generate-buildah-remote.sh` this script is used to generate tasks (that's why CI is failing), that script must be updated (and the underlying go code) to generate tasks

task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
```diff
@@ -829,8 +825,24 @@ spec:
...
+      # delete symlinks - they may point outside the container rootfs, messing with SBOM scanners
+      find $(cat /shared/container_path) -xtype l -delete
       echo "Running syft on the image filesystem"
       syft dir:"$(cat /shared/container_path)" --output "$syft_sbom_type"="/var/workdir/sbom-image.json"
+      SBOMIMAGEEOF
+      chmod a+x ./script-sbom-image.sh
+
+      unshare -Uf --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 --mount ./script-sbom-image.sh
+    securityContext:
+      runAsUser: 0
```
MartinBasti, line 842: why do we need root?

zxiong:
Hi @chmeliik, added change (export the image to an oci-archive and scan the oci-archive) for [STONEBLD-3020](https://issues.redhat.com/browse/STONEBLD-3020), please review.

task/buildah-min/0.4/buildah-min.yaml
```diff
@@ -638,7 +636,7 @@ spec:
       echo "Running syft on the source directory"
       syft dir:"$(workspaces.source.path)/$SOURCE_CODE_DIR/$CONTEXT" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-source.json"
       echo "Running syft on the image filesystem"
-      syft dir:"$(cat /shared/container_path)" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-image.json"
+      syft oci-archive:"$(cat /shared/container_path)" --output "$syft_sbom_type"="$(workspaces.source.path)/sbom-image.json"
```
chmeliik, line 639: I just realized that `syft` can probably read the image directly from the buildah storage if we just do `syft 
image:$IMAGE"`

Images can be gigantic, so if we could avoid pushing the image into the `/shared` volume that would be great

task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
chmeliik: +1, @zxiong could you add more info about the changes in the commit messages?

chmeliik:
The buildah task explicitly sets the `STORAGE_DRIVER=vfs` variable, wouldn't that also need an update?

task/buildah-min/0.4/buildah-min.yaml
arewm: For my proposed content-sets change, will we still hit this same performance hit?

chmeliik: It will re-introduce the performance hit of starting a container and mounting its filesystem that this PR is removing. But it shouldn't be worse than the current state of the buildah task (prior to this PR), and supposedly we will one day be able to get rid of the ICM generation altogether

chmeliik: Ideally, I think all we have should be the $IMAGE

> Would it be any better if we add the ICM generation script into the buildah-task image to run it in the same step where we build the container?

Maybe... :bulb: 
We could then, instead of using `buildah commit`, modify the Containerfile before building and add something like:

```Dockerfile
COPY content-sets.json /usr/share/buildinfo/content-sets.json
```

And because we're building with `--layers=false` (the default), that means the layers should get squashed the same way that `podman build --squash` would do it

zxiong:
> The buildah task explicitly sets the `STORAGE_DRIVER=vfs` variable, wouldn't that also need an update?

Yes, but before that we required `quay.io/konflux-ci/buildah-task` runs with overlay2 storage driver by default. Currently, it run with fuse-overlayfs storage driver by default. I have a PR for it https://github.com/konflux-ci/buildah-container/pull/116.

zxiong:
> ./hack/generate-buildah-remote.sh

Updated. I missed the comment before.

chmeliik:
The PR has merge conflicts, otherwise the changes LGTM

chmeliik:
Re: the failing buildah-remote check - it may be necessary to update the generator to still including that code hunk for versions < 0.4

zxiong:
> Re: the failing buildah-remote check - it may be necessary to update the generator to still including that code hunk for versions < 0.4

I think it may not be able to include that code hunk for version 0.1 and 0.2 since the step `analyse-dependencies-java-sbom` exists in the two versions and depends on the image container mount path ("/shared/container_path"). So, I updated the buildah-remote generator to exclude the code hunk for versions < 0.4.

chmeliik:
e2e-tests are failing with:

```
  pod: test-comp-lpcd-on-pull-request-j4qt8-build-container-pod | container step-sbom-syft-generate: 
  [2025-04-10T04:45:46,426338959+00:00] Generate SBOM
  Running syft on the source directory
  [0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
  Running syft on the image
  [0000] ERROR could not determine source: errors occurred attempting to resolve 'quay.io/redhat-user-workloads/build-templates-e2e/test-comp-lpcd:on-pr-f23e58d861e8b7bd5895b81ca82446d1b5d40c71':
    - docker: docker not available: failed to connect to Docker daemon. Ensure Docker is running and accessible
    - podman: podman not available: no host address
    - containerd: containerd not available: no grpc connection or services is available: unavailable
    - oci-registry: failed to get image descriptor from registry: Get "https://quay.io/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority
    - additionally, the following providers failed with file does not exist: docker-archive, oci-archive, oci-dir, singularity, oci-dir
```

zxiong:
> `  e2e-tests are failing with:
>   - oci-registry: failed to get image descriptor from registry: Get "https://quay.io/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority`

Added CA to fix this certificate verification failure.

But I have no idea about what's the new failure:
`nexpected error:
    <*errors.errorString | 0xc00096cb10>: 
    timed out when waiting for image-controller annotations to be updated on component test-comp-bble in namespace build-templates-e2e. component: {
      "metadata": {`

chmeliik:
> Added CA to fix this certificate verification failure.

Hmm, I don't think that was it. If the image is already present in the local container storage, then syft shouldn't have to connect to quay.io at all. Maybe we need to run the syft command with `unshare` as well?

Btw, it would probably be more effective to test the changes directly in some repo that you have onboarded to Konflux. Waiting for e2e-tests results takes a long time and it's hard to analyze the failures. https://github.com/konflux-ci/build-definitions?tab=readme-ov-file#pushing-task-bundles-for-inner-loop-development

chmeliik:
> Can you elaborate it a bit more about run the syft command with unshare?

I'm wondering if syft is having trouble accessing the buildah storage and needs `unshare` for similar reasons as buildah itself. Just a guess, I didn't think very deeply about it :smile: 

If we can't make `syft image:"$IMAGE"` work, we can go back to exporting an oci-archive I suppose

zxiong:
> If we can't make `syft image:"$IMAGE"` work, we can go back to exporting an oci-archive I suppose
Yeah, I agree completely :)

task/buildah/0.4/buildah.yaml
```diff
@@ -535,6 +528,10 @@ spec:
         fi
       done
 
+      image_name=$(echo "${IMAGE##*/}" | tr ':' '-')
+      buildah push "$IMAGE" oci-archive:"./$image_name.tar"
+      echo "$(pwd)/$image_name.tar" > /shared/container_path
```
chmeliik, line 533: What is the `$PWD` at this point? If I'm not mistaken, it's the root of the git repo, so we're writing the oci-archive into the root of the git repo?

Could we avoid that and instead write it to e.g. `/shared/image-archive.tar`?

chmeliik:
Mostly LGTM, but there is still an unresolved comment: 
* https://github.com/konflux-ci/build-definitions/pull/2074#discussion_r2055699074

chmeliik (approved):
LGTM, please just squash the commits before merging

zxiong:
> LGTM, please just squash the commits before merging

Done. Thanks!

PR #2082, opened 2025-03-18
fix: update konflux-tooling image in patch.yaml

tkdchen:
[STONEBLD-3351](https://issues.redhat.com//browse/STONEBLD-3351)

The customeManager for recipe.yaml is extended to work for patch.yaml files under task/ directory.

The regex is rewritten since the original for reciple.yaml does not work for images which do not include tag. For example, for given image app@sha256:1234, the matched depName is app@sha256.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -239,12 +239,12 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "^task/.*/recipe\\.yaml$"
+        "^task/.*/recipe\\.yaml$",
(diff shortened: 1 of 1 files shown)


mmorhun (approved):
LGTM

MartinBasti:
https://github.com/konflux-ci/build-definitions/pull/2036 it didn't generate update for patch.yaml there

MartinBasti:
it works actually, it just doesn't re-generate task yaml

PR #2092, opened 2025-03-20
Add support for self-signed OCI registries to clamav-scan task

amisstea:
The upload step required a fix to test this. It was silently failing to find any logs because the `find` command isn't present in the oras image.


amisstea:
@konflux-ci/integration-service-maintainers please review when you get a chance. Thanks!

jencull (approved):
lgtm

PR #2093, opened 2025-03-20
Add support for self-signed OCI registries to sast-coverity-check-oci-ta task

amisstea:


amisstea:
I'm not able to test this without requesting a coverity license which feels like too much effort for such a small addition. I've made similar changes (#2063, #2087, #2088)  to the `use-trusted-artifact` step in other tasks which have all been tested.

kdudka (approved):

dirgim (approved):
LGTM

PR #2095, opened 2025-03-21
feat(KONFLUX-7127): Update limits/resources for task build-image-index

zxiong:
This PR updates resource limits and requests for the task build-image-index as part of this effort - [KONFLUX-7127](https://issues.redhat.com/browse/KONFLUX-7127).

The new request/limit values have been defined taking into account usages observed historically (tracked in this [spreadsheet](https://docs.google.com/spreadsheets/d/1d4Tn7PYkY2EVgpNyT29xL7y62akZNPIjy18VTzw54iQ/edit?gid=1735880927#gid=1735880927)) in the stone-prd-rh01 cluster.

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -82,9 +82,9 @@ spec:
     name: build
     computeResources:
       limits:
-        memory: 4Gi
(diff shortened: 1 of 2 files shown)


zxiong:
As it has default value, let's observe more and update it later if need.

PR #2096, opened 2025-03-21
fix(KONFLUX-7346): buildah: check if cachi2.repo file exists

sfowl:
This fixes a bug in the coverity-check task. When builds are run with prefetched rpms, the cachi2.repo file is first copied by the original buildah task. Later when coverity-check repeats the build, the cachi2.repo file is already there, so `cp --no-clobber` exits with an error.

Fix tested here: https://github.com/openshift-kni/dpdk-base/pull/17

### DIFF
--- a/task/buildah/0.4/buildah.yaml
+++ b/task/buildah/0.4/buildah.yaml
@@ -367,7 +367,9 @@ spec:
         if [ -f "$prefetched_repo_for_my_arch" ]; then
           echo "Adding $prefetched_repo_for_my_arch to $YUM_REPOS_D_FETCHED"
           mkdir -p "$YUM_REPOS_D_FETCHED"
-          cp --no-clobber "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          if [ ! -f "${YUM_REPOS_D_FETCHED}/cachi2.repo" ]; then
+            cp "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          fi
         fi
       fi
 
--- a/task/buildah-min/0.4/buildah-min.yaml
+++ b/task/buildah-min/0.4/buildah-min.yaml
@@ -379,7 +379,9 @@ spec:
         if [ -f "$prefetched_repo_for_my_arch" ]; then
           echo "Adding $prefetched_repo_for_my_arch to $YUM_REPOS_D_FETCHED"
           mkdir -p "$YUM_REPOS_D_FETCHED"
-          cp --no-clobber "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          if [ ! -f "${YUM_REPOS_D_FETCHED}/cachi2.repo" ]; then
+            cp "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          fi
         fi
       fi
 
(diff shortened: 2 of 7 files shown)


kdudka (approved):
This was actually a bug in `coreutils-9.4`, which has already been [fixed](https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commitdiff;h=9907b6f5dc3faf370c916c2f81bfcc3d06028e83).  `cp --no-clobber` did not work this way when I [implemented](https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commitdiff;h=d01338eb3d30e5634f1b4d4179c229f54eea0b44) it 16 years ago, and it does not work this way in the latest upstream version either.

sfowl:
> This was actually a bug in coreutils-9.4, which has already been [fixed](https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commitdiff;h=9907b6f5dc3faf370c916c2f81bfcc3d06028e83). cp --no-clobber did not work this way when I [implemented](https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commitdiff;h=d01338eb3d30e5634f1b4d4179c229f54eea0b44) it 16 years ago, and it does not work this way in the latest upstream version either.

Wow, impressive!

Perhaps it would be better to just wait for the coreutils fix to reach our container images, but it looks like the latest RHEL9 is release still on v8.32, so I reckon we still need this workaround for the short term.

kdudka:
Luckily, this bug does not exist on RHEL-9 (nor any other RHEL).  It was a bug in Fedora 40, which the `buildah` image is based on.  I was only curious where this error came from but your fix is perfectly fine.

dirgim (approved):
LGTM for sast check changes

kdudka:
@tnevrlka Could you please review/approve this?  The fix has been [confirmed](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1743066723695309?thread_ts=1741795054.227499&cid=C04PZ7H0VA8) by users who are blocked on this.

task/buildah/0.4/buildah.yaml
```diff
@@ -367,7 +367,9 @@ spec:
         if [ -f "$prefetched_repo_for_my_arch" ]; then
           echo "Adding $prefetched_repo_for_my_arch to $YUM_REPOS_D_FETCHED"
           mkdir -p "$YUM_REPOS_D_FETCHED"
-          cp --no-clobber "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          if [ ! -f "${YUM_REPOS_D_FETCHED}/cachi2.repo" ]; then
+            cp "$prefetched_repo_for_my_arch" "$YUM_REPOS_D_FETCHED"
+          fi
```
tnevrlka, line 372: I think you should be able to avoid the if block by using `cp --update=none`. (if it's available in `quay.io/redhat-services-prod/sast/coverity` too)

tnevrlka (approved):

task/buildah/0.4/buildah.yaml
sfowl: Nice idea, I was tempted to use it, but I figured the if/else approach is slightly more resilient to change (in the unlikely event the base image changes). To get the fix out to users, I will merge as-is.

PR #2102, opened 2025-03-21
sast-shell-check: request more memory

kdudka:
... to avoid OOM failures on projects with complex shell scripts.

Related: https://issues.redhat.com/browse/KFLUXSPRT-2338

This pull request needs to be rebased once #2018 is merged.

### DIFF
--- a/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
+++ b/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
@@ -225,6 +225,11 @@ spec:
         note="Task $(context.task.name) completed successfully."
         TEST_OUTPUT=$(make_result_json -r SUCCESS -t "$note")
         echo "${TEST_OUTPUT}" | tee "$(results.TEST_OUTPUT.path)"
+      computeResources:
+        limits:
+          memory: 16Gi
+        requests:
+          memory: 8Gi
     - name: upload
       image: quay.io/konflux-ci/oras:latest@sha256:d9fea2ee280880feef6909bef3e18318444231c83736bcc41d54b4e5064f23c9
       workingDir: /var/workdir/source
--- a/task/sast-shell-check/0.1/sast-shell-check.yaml
+++ b/task/sast-shell-check/0.1/sast-shell-check.yaml
@@ -68,6 +68,11 @@ spec:
       # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
       # the cluster will set imagePullPolicy to IfNotPresent
       workingDir: $(workspaces.workspace.path)/hacbs/$(context.task.name)
+      computeResources:
+        requests:
+          memory: "8Gi"
+        limits:
+          memory: "16Gi"
       volumeMounts:
         - name: trusted-ca
           mountPath: /mnt/trusted-ca


sfowl:
Hmm, got the same error on two different runs:

```
 task scan-with-shellcheck has failed: TaskRun "test-sast-shell-check-run-zpnzh-scan-with-shellcheck" was cancelled. TaskRun cancelled as the PipelineRun it belongs to has timed out.
failed to get logs for task scan-with-shellcheck : task scan-with-shellcheck failed: pods "test-sast-shell-check-run-zpnzh-scan-with-shellcheck-pod" not found. Run tkn tr desc test-sast-shell-check-run-zpnzh-scan-with-shellcheck for more details
PipelineRun "test-sast-shell-check-run-zpnzh" failed to finish within "1h0m0s"
ERROR: Pipelinerun test-sast-shell-check failed
Error: Process completed with exit code 1.
```

Maybe the [shellcheck test pipeline](https://github.com/konflux-ci/build-definitions/blob/main/task/sast-shell-check/0.1/tests/test-sast-shell-check.yaml) is not running because of insufficient resources in the test cluster?

kdudka:
@sfowl I think you are correct.  I will try to lower the requirement for now but the proper solution would be to allocate enough resources for testing.

MartinBasti:
> I tried to decrease the memory requirements down to 8/16 GiB and the task is still not starting on the test cluster. I do not think we should go even lower because that would defeat the only purpose of this pull request.
> 
> @konflux-ci/build-maintainers How can one requests more memory on the test cluster?

That's probably question for QE or infra, we don't maintain testing infrastructure.

I'm also shocked, that there is a task that needs 16GB, it's more than majority of builds, this doesn't look very economic

kdudka:
> That's probably question for QE or infra, we don't maintain testing infrastructure.

What does it mean in this specific case?  What is the preferred communication channel to request the resources?

> I'm also shocked, that there is a task that needs 16GB, it's more than majority of builds, this doesn't look very economic

The test that stopped working does not need 16 GiB of memory.  We need 16+ GiB of memory on the production instance.  Then the question is whether we want to have separate task definitions for testing and production.  I would say no.

MartinBasti:
16GB for static linter sounds too much even for prod, isn't there something wrong with shellcheck to consume such amount of memory?

kdudka:
I do not think so.  This is how it works.  I tried to make it clear in the [design document](https://docs.google.com/document/d/1Hg1beFd_2ecxVljqNJsmnhyVc1Mn_kDAttWE_EPeiFQ/edit?disco=AAABR9U7JRA) a year ago that SAST scanning requires more resources than the native build.  Perhaps you are comparing it with Snyk Code?  The key difference is that `sast-snyk-check` uses a remote service to analyze the code (and even the `snyk` client is [known to crash](https://issues.redhat.com/browse/OSH-333) with OOM).

We could make the `sast-shell-check` task require less memory by reducing the number of available CPUs at the cost of increasing the waiting time, which does not help with cost efficiency much (because manpower is more expensive than machine power).

sfowl:
@kdudka I'm wondering maybe if might be better to require users to set there own resource limits for this task, and include that in the docs (maybe event a hint in the script itself, if possible)? 

The required resources seems like it would be somewhat proportional to the number and size/complexity of shell scripts in a given component, and I would imagine that would be relatively low in most cases. If it is uncommon to have large, complex shell scripts, then maybe it is reasonable to expect the owners of those components to bump the resource limits as needed.

kdudka:
That is fine by me, too.  You can re-evaluate this based on the feedback later if needed.  Another potential solution to explore is https://issues.redhat.com/browse/PSSECAUT-1062.

Anyway, I am closing this pull request out as it turned out to be way more complicated that I had expected.

PR #2108, opened 2025-03-24
Force fetching git tags on some projects with go submodule

fmudry:
Some repos with go submodules would fail during prefetch dependencies task. 
Cachi2 would fail and this serves as a workaround, since we didn't find the root cause.

### DIFF
--- a/task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
+++ b/task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
@@ -372,6 +372,13 @@ spec:
 
         INPUT=$(inject_rpm_summary_flag "$INPUT")
 
+        # Some repos with go submodules would fail during prefetch dependencies task
+        # Forcing fetching tags serves as a workaround
+        if [[ "$(cd "/var/workdir/source" && ! git fetch --tags)" ]]; then
+          echo "Retrying fetch command..."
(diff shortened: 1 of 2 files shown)


rcerven (approved):

task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
```diff
@@ -372,6 +372,11 @@ spec:
 
         INPUT=$(inject_rpm_summary_flag "$INPUT")
 
+        if [[ "${INPUT}}" == *"gomod"* ]] && [[ "$(cd "/var/workdir/source" && ! git fetch --tags)" ]]; then
```
rcerven, line 375: why only for 'gomod' ?

rcerven (changes requested):

task/prefetch-dependencies-oci-ta/0.2/prefetch-dependencies-oci-ta.yaml
brunoapimentel: Only gomod is affected by the bug we're trying to solve here (no other package manager fetches tags, not at the moment at least).

rcerven: right, but it wouldn't hurt to do it always

rcerven (approved):

MartinBasti (approved):

PR #2114, opened 2025-03-24
fix(KFLUXSPRT-2387) Fix mirrors array population in FIPS tasks

nmars:
In FIPS tasks, fixed the way lists of IDMS mirrors are populated into bash arrays.


arewm (approved):

hongweiliu17 (approved):
Need to fix warnings as below. Other lgtm
```
warning: Expanding an array without an index only gives the first element.
   ┌─ stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml:66:47
   │
66 │           echo "Mirrors for $reg_and_repo are $mirrors"
```

dirgim (approved):
LGTM

hongweiliu17 (approved):
lgtm

PR #2119, opened 2025-03-26
Add more content for task migrations

tkdchen:
This update provides more information about task migrations, and a new section describing fictional task update for adding tasks to Konflux pipelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,17 +280,22 @@ A build Task, e.g. one that produces a container image, must abide by both `all-
 
 ## Task Migration
 
-Task migrations are Bash scripts defined in version-specific task
-directories. In general, a migration consists of a series of `yq` commands that
-modify pipeline in order to work with the new version of task. Developers can
-do more with task migrations on the pipelines, e.g. add/review a task,
-add/remove/update task parameters, change execution order of a task, etc.
+Task migrations allow task maintainers to introduce changes to Konflux standard
+pipelines according to the task updates. By creating migrations, task
+maintainers are able to add/remove/update task parameters, change task
+execution order, add/remove mandatory task to/from pipelines, etc.
 
 Historically, task maintainers write `MIGRATION.md` to notify users what changes
 have to be made to the pipeline. This mechanism is not deprecated. Besides
 writing the document, it is also recommended to write a migration script so that the
 updates can be applied to user pipelines automatically, that is done by the
 [pipeline-migration-tool](https://github.com/konflux-ci/pipeline-migration-tool).
+
+Task migrations are Bash scripts defined in version-specific task
+directories. In general, a migration consists of a series of `yq` commands that
+modify pipeline in order to work with the new version of task. Developers can
+do more with task migrations on the pipelines, e.g. add/review a task,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -280,18 +280,23 @@ A build Task, e.g. one that produces a container image, must abide by both `all-
...
 ## Task Migration
 
-Task migrations are Bash scripts defined in version-specific task
-directories. In general, a migration consists of a series of `yq` commands that
-modify pipeline in order to work with the new version of task. Developers can
-do more with task migrations on the pipelines, e.g. add/review a task,
-add/remove/update task parameters, change execution order of a task, etc.
+Task migrations allow task maintainers to introduce changes to Konflux standard
+pipelines according to the task updates. By creating migrations, task
+maintainers are able to add/remove mandatory task to/from pipelines,
```
mmorhun, line 285: I'd put task adding/removal at the end since it's less often case.

README.md
```diff
@@ -316,6 +321,13 @@ The migration file is a normal Bash script file:
 - It should be idempotent as much as possible to ensure that the changes are
   not duplicated to the pipeline when run the migration multiple times.
 - Pass the `shellcheck` without customizing the default rules.
+- Check whether the migration is for all kinds of Konflux pipelines or not. If
+  no, skip the pipeline properly in the script, e.g. skip FBC pipeline due
```
mmorhun, line 325: Would be helpful to describe with an example how to skip FBC (but probably not here, maybe just leave a note like "see below").

README.md
```diff
@@ -316,6 +321,13 @@ The migration file is a normal Bash script file:
...
+  to [many tasks are removed](https://github.com/konflux-ci/build-definitions/blob/main/pipelines/fbc-builder/patch.yaml)
+  from template-build.yaml.
+- The pipeline file path and name can be arbitrary. Please do not use the input
```
mmorhun, line 328: What is input value? Why a user shouldn't do it?

tkdchen: In build-definitions pr 2018, the author matches pipeline file to determine pipeline types for conditional operation.

It can ends with a user-defined extracted pipeline file name or be a temporary file. It's the implementation details. So, as mentioned in the doc, from user's point of view, the pipeline file value can be arbitrary and should not be relied on to match anything.

README.md
tkdchen: Updated.

README.md
tkdchen: Removed this item. This one should be over-thought. An example is the pr 2018, fbc-bundle does not have task sast-coverity-check, then there will not sast-coverity-check upgrade.

mmorhun: I think this would be quite a common case, we need to have it in the docs somewhere.

README.md
mmorhun: I think the explanation from your second paragraph should be included into the doc.

tkdchen: I don't think so. It's implementation details which can be changed by developers at any time when need. It should not be exposed to users, if yes, some migrations may try to match file path according to this document.

README.md
tkdchen: Added it back.

PR #2124, opened 2025-03-26
sast-coverity-check: create two more tasks

kdudka:
- sast-coverity-check-build - separate task for the instrumented build
- sast-coverity-check-upload - separate task for upload of scan results

The first task is needed if we want to ask Tekton for more resources while the second task is needed if we want to run Coverity in parallel with the native build task.  See the following Jira issue for details.

Related: https://issues.redhat.com/browse/PSSECAUT-1032


kdudka:
The CI is [failing](https://github.com/konflux-ci/build-definitions/actions/runs/14132835998/job/39597265127?pr=2124) with:
```
Error: Task is using a workspace(s): sast-results, to share data and needs a corresponding Trusted Artifacts Task variant in task/sast-coverity-check-build-oci-ta-oci-ta/0.4/
Error: Task is using a workspace(s): sast-results, to share data and needs a corresponding Trusted Artifacts Task variant in task/sast-coverity-check-oci-ta-oci-ta/0.4/
Error: Task is using a workspace(s): sast-results, to share data and needs a corresponding Trusted Artifacts Task variant in task/sast-coverity-check-upload-oci-ta/0.4/
```
... which is not really expected.  I have introduced the `sast-results` workspace on purpose so that the tasks can pass the SAST scanning results to each other.  This is completely separated from the `source` workspace and the tasks that use the `source` workspace have their own `-oci-ta` copies to handle this properly.

PR #2125, opened 2025-03-26
remove when condition for prefetch-dependencies task

rcerven:

### DIFF
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -219,11 +219,6 @@ spec:
     taskRef:
       name: prefetch-dependencies
       version: "0.2"
-    when:
-    - input: $(params.prefetch-input)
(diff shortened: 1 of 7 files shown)


pipelines/core-services/core-services.yaml
```diff
@@ -219,11 +219,6 @@ spec:
     taskRef:
       name: prefetch-dependencies
       version: "0.2"
-    when:
```
arewm, line 222: Should we not just make this consistent with the build tasks?
```
      operator: in
      values:
      - "true"
```

rcerven:
based on https://github.com/release-engineering/rhtap-ec-policy/pull/128#issuecomment-2724501208

ralphbean:
This might be a duplicate of https://github.com/konflux-ci/build-definitions/pull/2065

PR #2130, opened 2025-03-27
Better output from migrations validation

tkdchen:
Without -x, script behaves more informative to tell what is being done and necessary output from the command executed.

### DIFF
--- a/hack/validate-migration.sh
+++ b/hack/validate-migration.sh
@@ -38,6 +38,45 @@ declare -r DEFAULT_BRANCH
 : "${IN_CLUSTER:=""}"
 declare -r IN_CLUSTER
 
+info() {
+    echo "info: $*" >&2
+}
+
+error() {
+    echo "error: $*" >&2
+}
+
+warning() {
+    echo "warning: $*" >&2
+}
+
+format_yaml_in_python() {
+    local -r filepath=$1
+    python3 -c "
+import sys
+from ruamel.yaml import YAML
+yaml = YAML()
+# These settings are same as pipeline-migration-tool
+yaml.preserve_quotes = True
+yaml.width = 8192
+yaml_file = sys.argv[1]
+with open(yaml_file, 'r', encoding='utf-8') as f:
+    data = yaml.load(f)
+with open(yaml_file, 'w', encoding='utf-8') as f:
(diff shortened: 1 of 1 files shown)


hack/validate-migration.sh
```diff
@@ -323,32 +371,43 @@ check_apply_in_real_cluster() {
...
 main() {
     if git status --porcelain | grep -qv "^??"; then
-        echo "There are uncommitted changes. Please commit them and run again." >&2
+        info "There are uncommitted changes. Please commit them and run again."
         exit 1
     fi
 
     if ! is_on_topic_branch; then
-        echo "Script must run on a topic branch rather than the main branch." >&2
+        info "Script must run on a topic branch rather than the main branch."
```
mmorhun, line 379: As a user, I might not get what is topic branch.

hack/validate-migration.sh
```diff
@@ -111,12 +152,15 @@ check_apply_on_pipelines() {
     local updated=
     while read -r file_path; do
         cp "$file_path" "${file_path}.copy"
+        info "apply migration $migration_file to pipeline ${file_path}.copy"
```
mmorhun, line 155: Does it make sense to have a separate variable for `${file_path}.copy"`?

tkdchen: Updated. PTAL.

hack/validate-migration.sh
```diff
@@ -109,16 +150,21 @@ check_apply_on_pipelines() {
     local failed=
     local file_path=
     local updated=
+    local tmp_pl_file=
     while read -r file_path; do
-        cp "$file_path" "${file_path}.copy"
-        if ! bash -x "$migration_file" "${file_path}.copy" 2>"$run_log_file" >&2; then
-            echo "error: failed to run migration file $migration_file on pipeline $file_path:" >&2
+        tmp_pl_file="${file_path}.copy"
+        cp "$file_path" "$tmp_pl_file"
```
mmorhun, line 156: From the name of variables I cannot conclude the second is a copy of the first. Please make them match, like `pl_file_path` and `pl_file_copy_path` or similar.

mmorhun (approved):

PR #2133, opened 2025-03-27
fix(KONFLUX-5907): make clamav work on kind cluster

jsztuka:

This change makes it possible to extract image even on kind cluster, the behaviour of scan itself remains the same.


jsztuka:
Set it for older version, let me correct that.

hongweiliu17 (approved):
LGTM

PR #2134, opened 2025-03-27
fix(KFLUXVNGD-210): update SOURCE_CODE_DIR env var

avi-biton:
- update `SOURCE_CODE_DIR` environment variable to get `source` as a fixed value 
- remove `SOURCE_CODE_DIR` parameter


arewm (approved):
This looks good as it is consistent with buildah

PR #2142, opened 2025-03-28
sast-coverity-check: fix scan failures reported by users

kdudka:
See the individual commit messages for details.

Resolves: https://issues.redhat.com/browse/PSSECAUT-1051
Resolves: https://issues.redhat.com/browse/KFLUXSPRT-2270

### DIFF
--- a/task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
+++ b/task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
@@ -360,6 +360,9 @@ spec:
         # avoid tracing "dnf --installroot" because it would break the build due to unreachable /tmp/idir in the chroot
         export COVERITY_DISENGAGE_EXES="dnf;microdnf;yum"
 
+        # always use a fresh temp directory for Coverity to avoid "Permission denied" errors on reuse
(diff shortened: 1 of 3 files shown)


sfowl:
Could you add some brief comment about where these vars come from or how they are used? Just for reference to future maintainers, as it seems they are not present in coverity docs.

task/sast-coverity-check-oci-ta/0.2/sast-coverity-check-oci-ta.yaml
```diff
@@ -349,6 +349,12 @@ spec:
           fi
         fi
 
+        # avoid tracing "dnf --installroot" because it would break the build due to unreachable /tmp/idir in the chroot
+        export COVERITY_DISENGAGE_EXES="dnf"
```
sfowl, line 353: Wondering if this should also include `microdnf` and `yum`?

 Also, do you know if coverity will combine this with the default value? I.e. if coverity would have set it to "qemu;ldd", and we set it to "dnf", will the final value used at runtime be "qemu;ldd;dnf"?

kdudka:
Both the issues have been confirmed as fixed by this pull request.

dirgim (approved):
LGTM

PR #2159, opened 2025-04-02
[prefetch] Switch the cachi2 image to the new name hermeto

shi2wei3:
Move to quay.io/konflux-ci/hermeto since I see 0.23.0 is only pushed to the new repo instead of quay.io/konflux-ci/cachi2


eskultety (approved):
Yes, this is the right patch, we'll be sending out and announcement momentarily, thank you.

slimreaper35 (approved):

cgwalters (commented):
Thanks for this!

PR #2167, opened 2025-04-04
fix(CLOUDDST-26534): update fips stepaction

nmars:
Updated the fips stepaction to use konflux-test:v1.4.22, which has check-payload 0.3.5.


dirgim (approved):
LGTM

asergienk (approved):
lgtm

PR #2169, opened 2025-04-07
run-script-oci-ta: fix .metadata.name

chmeliik:
The .metadata.name in run-script-oci-ta.yaml doesn't match the filename. Because of that, the CI gets confused:

- The create-task-pipeline-bundle-repos.sh script reads the name from the yaml, creates quay.io/konflux-ci/tekton-catalog/task-run-script
- The build-and-push.sh script takes the name from the file name, pushes to quay.io/konflux-ci/tekton-catalog/task-run-script-oci-ta. Pushing to a non-existent repo creates it as private. The script then fails to read from the private repo.

Fix the run-script-oci-ta task name in the YAML metadata, to make sure both the scripts will use the same name.

### DIFF
--- a/task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
+++ b/task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
@@ -2,7 +2,7 @@
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
-  name: run-script
+  name: run-script-oci-ta
   annotations:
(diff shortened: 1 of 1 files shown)


chmeliik:
I believe this is not a breaking change, because currently https://quay.io/konflux-ci/tekton-catalog/task-run-script-oci-ta is private so nobody can use the task yet

I am not sure if this change alone will be enough to un-private the repository, we may need an admin to do that. We'll see after merging

PR #2177, opened 2025-04-09
AIPCC-779(feat): build multi-arch disk images

gnaponie:
Some products need to build multi-arch disk images.
In order to do that, a matrix is used in the Pipeline definition for some parameters.
This commit introduces:
- support for those parameters as `matrix`.
- sanitizing the PLATFORM string, as tags don't not accept `_` chars.
- new result IMAGE_REFERENCE: see comment in the code for more details on why this is needed.

### DIFF
--- a/task/build-vm-image/0.1/build-vm-image.yaml
+++ b/task/build-vm-image/0.1/build-vm-image.yaml
@@ -14,6 +14,10 @@ spec:
     - description: The platform to build on
       name: PLATFORM
       type: string
+    - default: "false"
+      description: Whether to append a sanitized platform architecture on the IMAGE tag
+      name: IMAGE_APPEND_PLATFORM
+      type: string
     - name: OUTPUT_IMAGE
       type: string
       description: The output manifest list that points to the OCI artifact of the zipped image
@@ -43,6 +47,8 @@ spec:
       name: IMAGE_DIGEST
     - description: Image repository where the built manifest list was pushed
       name: IMAGE_URL
+    - description: Image reference (IMAGE_URL + IMAGE_DIGEST)
+      name: IMAGE_REFERENCE
   stepTemplate:
     env:
       - name: OUTPUT_IMAGE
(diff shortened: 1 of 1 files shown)


gnaponie:
@arewm @ralphbean @fabiendupont PTAL

task/build-vm-image/0.1/build-vm-image.yaml
```diff
@@ -249,23 +263,23 @@ spec:
         if [ -f "/output/qcow2/disk.qcow2" ]; then
           echo -e "Found qcow2 image."
           # don't zip qcow2 images, it is already compressed.
-          buildah manifest add --arch $(arch) --os linux --artifact --artifact-type application/vnd.diskimage.qcow2.gzip $OUTPUT_IMAGE /output/qcow2/disk.qcow2
+          buildah manifest add --arch $(arch) --os linux --artifact --artifact-type application/vnd.diskimage.qcow2.gzip $OUTPUT_IMAGE /output/qcow2/disk.qcow2 --all
```
arewm, line 266: These commands just add a single artifact. I don't think that we need `--all` here. That parameter is necessary when we are generating an image index of image references which are themselves image indexes.

gnaponie:
Let me add that I still haven't test this.... :'(

task/build-vm-image/0.1/build-vm-image.yaml
gnaponie: Makes sense. Removed.

ralphbean:
From slack, giulia is working out a flow to test this with some of her builds. @gnaponie, when you're ready for a round of review, let us know.

---

I notice, in the description of the PR it mentions something about add `--all` to buildah invocations, but I don't see that in the diff. Did that change?

gnaponie:
Yeah, let me write an update on the testing: tekton is complaining on a weird issue.
I'm using the matrix, seems like tekton runs a task multiple times in parallel with different parameters, to use matrix. The [*] I'm using is not a valid tekton syntax for result aggregation. I'm trying to adjust the task/pipeline accordingly.

About the `--all`: yes, I'll update the commit message so it's not confusing!

gnaponie:
@arewm @ralphbean @fabiendupont I believe this PR is finally ready for the last round of reviews. I was able to test it successfully. I'll share the details of my tests privately.

fabiendupont:
Looks good to me.

PR #2181, opened 2025-04-10
Remove workspace from show-summary task

tkdchen:
workspace was removed from the pipeline when creating tekton-bundle-builder-oci-ta pipeline.

workspace is optional for task summary, so this removal just works.

This issue is found during review on #2018. That makes migration validation fails.


kasemAlem (approved):

kasemAlem:
LGTM

PR #2186, opened 2025-04-10
buildah-remote: fix ssh shell quoting

chmeliik:
When passing a variable to SSH, it gets evaluated by two shells: the shell that's currently running and the shell on the remote server. What does this mean?

If you run this example locally:

    message='hello there'
    printf '%s.' $message

    # result: hello.there.

The message breaks into two arguments, because it's not properly quoted. Fix by adding quotes:

    message='hello there'
    printf '%s.' "$message"

    # result: hello there.

But then, when you add ssh into the mix:

    message='hello there'
    ssh "$some_host" printf '%s.' "$message"

    # result: hello.there.

The message broke again. That's because ssh received a list of arguments already expanded by your shell and concatenated them together, resulting in a command that may look something like this:

    printf %s. hello there

When passing arbitrary strings to ssh, they need two layers of quoting. The result of variable expansion on the client side must be a string that, when evaluated by a shell again, expands to the original value of the variable. In bash, this is achieved with the ${var@Q} operator; see the Bash manual [1] (refer to the "parameter@operator" section).

Apply the `@Q` operator when passing the $@ argument array to ssh, as well as other variables.

[1]: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

### DIFF
--- a/task-generator/remote/main.go
+++ b/task-generator/remote/main.go
@@ -268,7 +268,7 @@ if ! [[ $IS_LOCALHOST ]]; then
 		ret += "\n  echo \"[$(date --utc -Ins)] Build via ssh\""
 		podmanArgs += "    -v \"$BUILD_DIR/scripts:/scripts:Z\" \\\n"
 		podmanArgs += "    \"${PRIVILEGED_NESTED_FLAGS[@]}\" \\\n"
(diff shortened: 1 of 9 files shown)


chmeliik:
This was preventing users from passing `BUILD_ARGS`, `LABELS` or `ANNOTATIONS` containing spaces or other special characters, because we were passing them over `ssh` as `$@`.

My tests that this resolves the issue are in an internal gitlab repo (because I have no public repo capable of using the multi-arch feature :disappointed: )

chmeliik:
Fixed the rest of the ssh shell quoting and re-tested that this doesn't break anything.

mmorhun (approved):

PR #2189, opened 2025-04-11
Add SBOM_DIGEST result to build-index-image task

Allda:
The SBOM digest is exposed in the task result to be later used by the Konflux UI to redirect user to SBOM uploaded to Atlas UI.

This is done because a new version of Atlas no longer supports URL templates based on custom fields. The SBOM digest is now used as primary identifier.

JIRA: [ISV-5734](https://issues.redhat.com//browse/ISV-5734)

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -54,6 +54,9 @@ spec:
   - name: SBOM_BLOB_URL
     description: Reference of SBOM blob digest to enable digest-based verification from provenance
     type: string
+  - name: SBOM_DIGEST
+    description: A reference to the digest of the SBOM blob
+    type: string
   volumes:
     - name: shared-dir
       emptyDir: {}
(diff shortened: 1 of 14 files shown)


task/build-image-index/0.1/build-image-index.yaml
```diff
@@ -54,6 +54,9 @@ spec:
   - name: SBOM_BLOB_URL
     description: Reference of SBOM blob digest to enable digest-based verification from provenance
     type: string
+  - name: SBOM_DIGEST
+    description: A reference to the digest of the SBOM blob
+    type: string
```
chmeliik, line 59: AFAIK the `SBOM_BLOB_URL` is already in `quay.io/foo/bar@sha256:digest` format. Would the UI be able to use the existing digest instead?

Allda: The aim of this result was to make it easier for a frontend to not have to parse the existing URL and have the logic spread between backend and frontend.

Allda:
@chmeliik is the 502 error common for the tests? Since Friday, I wasn't able to pass the tests, and the only result I see is the Quay 502.

task/build-image-index/0.1/build-image-index.yaml
chmeliik: Hmm, I don't see a need to duplicate this info in the task results just for the UI. The logic should be pretty much "take the part after `@`"

Allda: ok agree, I'll drop this and sync with the frontend team to implement this logic.

PR #2191, opened 2025-04-11
Improve log-based buildah debugging

arewm:
As a user, I have often wanted to know what architecture a build is running on at the start as well as what the actual buildah command used was. We can easily output both of these to the log.


arewm:
Here is a sample build log that I got from one of the e2e tests: [text.log](https://github.com/user-attachments/files/19712496/text.log)

rcerven (approved):

PR #2192, opened 2025-04-11
Set build-source-image param default to true

simonbaird:
When a new component is onboarded, it's likely it will immediately fail the Conforma/EC check because the source image is required.

Let's save some time and effort and make the source image built by default.

Notes:
* There maybe be a good reason that I'm not aware of why this hasn't been done already, but I figured I'd make a PR anyhow to kick off the discussion.
* I'm not sure what tests will need to be modified due to this change, but let's see what happens in the PR.
* The pipelines/template-build/template-build.yaml file was the one edited manually, all the other changes are in the generated files, updated by running hack/build-mannifests.sh.

Ref: https://issues.redhat.com/browse/KFLUXSPRT-2694


chmeliik:
I recently found out that not all Conforma policies require source containers. Even some of those used at Red Hat for releasing (but only to place other than registry.redhat.io I think). Setting the default to true may not be what we want, since for those users it would just unnecessarily make the build take longer

PR #2197, opened 2025-04-14
fix(KONFLUX-7549): update refs for fips tasks

nmars:
Updated fips tasks refs to use the latest fips stepaction with logging and resiliency improvements.


asergienk (approved):
lgtm

PR #2200, opened 2025-04-15
fix(KFLUXVNGD-212): update reference to FILES array

avi-biton:
Updating FILES[@] to FILES[0] to only pick the first element. 
All elements are referencing to the same task dir.

### DIFF
--- a/task/tkn-bundle-oci-ta/0.2/tkn-bundle-oci-ta.yaml
+++ b/task/tkn-bundle-oci-ta/0.2/tkn-bundle-oci-ta.yaml
@@ -199,7 +199,9 @@ spec:
         task_version=$(yq '.metadata.labels."app.kubernetes.io/version"' "${FILES[@]}" | sed '/null/d' | tr -d '[:space:]')
         ANNOTATIONS+=("org.opencontainers.image.version=${task_version}")
 
-        task_dir=$(echo "${FILES[@]}" | awk -F'/' '{NF--; print $0}' OFS='/')
+        # task_dir is where all the tasks definitions resides
+        # all tasks definitions are in the same task_dir, so extracting it from the first element is enough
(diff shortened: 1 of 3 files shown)


task/tkn-bundle-oci-ta/0.2/tkn-bundle-oci-ta.yaml
```diff
@@ -199,7 +199,7 @@ spec:
         task_version=$(yq '.metadata.labels."app.kubernetes.io/version"' "${FILES[@]}" | sed '/null/d' | tr -d '[:space:]')
         ANNOTATIONS+=("org.opencontainers.image.version=${task_version}")
 
-        task_dir=$(echo "${FILES[@]}" | awk -F'/' '{NF--; print $0}' OFS='/')
+        task_dir=$(echo "${FILES[0]}" | awk -F'/' '{NF--; print $0}' OFS='/')
```
lcarva, line 202: IIRC, this is being done because `task_dir` is used as the value for some of the annotations?
I understand the reasoning but it's a bit odd. Can we document it somewhere? Maybe in the README of this Task?

At some point, we may want to re-consider dropping the multiple tasks per bundle feature.

lcarva (dismissed):
Make

task/tkn-bundle-oci-ta/0.2/tkn-bundle-oci-ta.yaml
avi-biton: @lcarva I added a comment in the script and updated the readme file.
Do you think something else is needed?

lcarva: Maybe in the README, it should list the annotations that are added and explain that the path of the first task definition is used to derive its value.

Not a blocker for this PR. Just a nice to have.

PR #2202, opened 2025-04-15
Fix additonal_secret incorrect temporal path

jordigilh:
This PR fixes the incorrect path where the addtional secret is mounted, based on the build task description:
```
    - name: ADDITIONAL_SECRET
      description: Name of a secret which will be made available to the build
        with 'buildah build --secret' at /run/secrets/$ADDITIONAL_SECRET
```

But it is actually mounting it in `/tmp/additional-secret`
https://github.com/konflux-ci/build-definitions/blob/941a9636be801fe11c197ec879b4e55294d09f09/task/buildah-oci-ta/0.4/buildah-oci-ta.yaml#L554

@arewm PTAL

### DIFF
--- a/task/buildah/0.4/buildah.yaml
+++ b/task/buildah/0.4/buildah.yaml
@@ -470,11 +470,11 @@ spec:
       echo "[$(date --utc -Ins)] Add secrets"
 
       ADDITIONAL_SECRET_PATH="/additional-secret"
-      ADDITIONAL_SECRET_TMP="/tmp/additional-secret"
+      ADDITIONAL_SECRET_TMP="/run/secrets/${ADDITIONAL_SECRET}"
       if [ -d "$ADDITIONAL_SECRET_PATH" ]; then
         cp -r --preserve=mode -L "$ADDITIONAL_SECRET_PATH" $ADDITIONAL_SECRET_TMP
         while read -r filename; do
-          echo "Adding the secret ${ADDITIONAL_SECRET}/${filename} to the build, available at /run/secrets/${ADDITIONAL_SECRET}/${filename}"
+          echo "Adding the secret ${ADDITIONAL_SECRET}/${filename} to the build, available at ${ADDITIONAL_SECRET_TMP}/${filename}"
           BUILDAH_ARGS+=("--secret=id=${ADDITIONAL_SECRET}/${filename},src=$ADDITIONAL_SECRET_TMP/${filename}")
         done < <(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;)
       fi


task/buildah/0.4/buildah.yaml
```diff
@@ -470,11 +470,11 @@ spec:
       echo "[$(date --utc -Ins)] Add secrets"
 
       ADDITIONAL_SECRET_PATH="/additional-secret"
-      ADDITIONAL_SECRET_TMP="/tmp/additional-secret"
+      ADDITIONAL_SECRET_TMP="/run/secrets/${ADDITIONAL_SECRET}"
```
chmeliik, line 473: This is the path on the OpenShift Pod, but is unrelated to the path that you can access in your Containerfile. That one is defined by `"--secret=id=${ADDITIONAL_SECRET}/${filename},src=$ADDITIONAL_SECRET_TMP/${filename}"`

The confusion may come from the fact that `/run/secrets/$ADDITIONAL_SECRET` is a directory, with files corresponding to the keys in the Secret

jordigilh: I noticed that the secret was not mounted in `/run/secrets` but the parameter was captured in the task log. I thought that the error was that it was being mounted in the wrong path. I will investigate further to see where the secret is mounted, but insofar it's not in `/run/secrets`:
```
Build will be executed with network isolation
[2025-04-16T13:20:41,617701109+00:00] Setup prefetched
[2025-04-16T13:20:41,620706490+00:00] Register sub-man
[2025-04-16T13:20:41,621806789+00:00] Add secrets
Adding the secret external-registry-pull-secret/pull to the build, available at /run/secrets/external-registry-pull-secret/pull
[2025-04-16T13:20:41,631166920+00:00] Run buildah build
[2025-04-16T13:20:41,632197508+00:00] ip link set lo up && buildah build --pull=never --secret=id=external-registry-pull-secret/pull\,src=/tmp/additional-secret/pull --label build-date=2025-04-16T13:20:41 --label architecture=x86_64 --label vcs-type=git --label vcs-ref=57a12f1d404a28aee4cffd0e5e19c4b976e6af2b --label quay.expires-after=5d --tls-verify=true --no-cache --ulimit nofile=4096:4096 -f /tmp/operator.Dockerfile.Evkq7K -t quay.io/redhat-user-workloads/storage-scale-releng-tenant/controller-rhel9-operator:on-pr-57a12f1d404a28aee4cffd0e5e19c4b976e6af2b . 

```
and further down I have a RUN statement in the Dockerfile to list the contents of `/run/secrets`:

```
+ ls -laR /run/secrets/
/run/secrets/:
total 4
drwxr-xr-x.  2 root root    6 Apr 16 13:20 .
drwxr-xr-x. 14 root root 4096 Apr 16 13:20 ..
```

chmeliik: Hmm, another idea: how exactly does your RUN instruction look? I think it may need to look something like this:

```Dockerfile
RUN --mount=type=secret,id=external-registry-pull-secret/pull ls /run/secrets
```

With the `id` having to match exactly

jordigilh: It doesn't have the `--mount` parameter. I will add it and retest. I think that's probably the root cause of the issue. I will drop the PR since it's clearly not a problem in the task scrpt.

arewm: @jordigilh , does our documentation need to be improved? Ah, @Jared-Sprague has this PR to document it (but it needs a rebase after we changed the structure): https://github.com/konflux-ci/docs/pull/232

PR #2204, opened 2025-04-16
Add new pipeline docker-build-operator-bundle

spotlesstofu:
Create a new pipeline to build the operator bundle.

The new pipeline is a clone of the docker-build pipeline, with the unneeded tasks removed:
  - ecosystem-cert-preflight-checks
  - sast-coverity-check
  - coverity-availability-check

Enables building operator bundle images using a pipeline reference (see KONFLUX-2500 for context).


pipelines/docker-build-operator-bundle/patch.yaml
```diff
@@ -0,0 +1,44 @@
...
+- op: remove
+  path: /spec/tasks/8
+
+# sast-coverity-check Task
+- op: remove
+  path: /spec/tasks/10
+
+# coverity-availability-check Task
+- op: remove
+  path: /spec/tasks/10
```
arewm, line 44: These are effectively the only changes in this pipeline. Would it not be better (i.e. less maintenance) for us to add additional conditional execution to these which are exposed via pipeline parameters instead of maintaining a whole new pipeline customization?

We can forward specific pipeline parameters to be made available on the pipeline definition that is pushed out using `additional-params`: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml#L12-L13

@konflux-ci/build-maintainers , how hard would it be to allow these default parameters to be customized with the pipeline config? Ideally, we would also have name metadata in this config but I think that we have some technical debt around that as well.

arewm (changes requested):
While I understand the motivation for this change (you want to simplify the process for onboarding), we also want to be careful with the creation of additional pipelines. We already have a lot and it can be challenging for users to be able to properly decide which pipeline is an appropriate one. 

We made a decision when we started supporting bundles to _NOT_ create a separate flow, instead just treating them like container images.

This pipeline that you are proposing isn't special for a bundle image. It just doesn't run "problematic" tasks which also happen to not be required by the Conforma policy.

pipelines/docker-build-operator-bundle/patch.yaml
chmeliik: > how hard would it be to allow these default parameters to be customized with the pipeline config? Ideally, we would also have name metadata in this config but I think that we have some technical debt around that as well.

What kind of customization are we talking about? The user would need the ability to customize some aspects of the Pipeline while creating their Component?

spotlesstofu: > Would it not be better (i.e. less maintenance) for us to add additional conditional execution to these

You mean something like the `skip-checks` param, for example a new `build-operator-bundle` param?

The new param would be propagated to all the other pipelines that inherit from `docker-build`, would that be OK?

Initially i did that, then i realized this could've been a cleaner solution.

arewm: @spotlesstofu , I was thinking something like `skip-optional-checks` but the challenge with that is that the optional checks could change in time. If we push out pipelines to newly onboarded teams that have specific values for the conditionals (i.e. indicating that specific tasks are part of the optional checks) then we can have users in a future situations where there are migration challenges to enable those "optional" tasks.

@chmeliik , We have the ability to "forward" params from the pipeline definition to the pipelinerun definition. This just sets the param value to its default. This would be a further refinement to that process enabling the pipeline config to set specific non-default values for those parameters which are added (in a way which is unique to specific pipelines configured).

chmeliik: > This would be a further refinement to that process enabling the pipeline config to set specific non-default values for those parameters which are added (in a way which is unique to specific pipelines configured).

If I understand correctly, that would avoid the need to have another pipeline yaml here in build-definitions, but wouldn't address this concern you mentioned? Because it would still be a separate entry in the pipelines list?

> We also want to be careful with the creation of additional pipelines. We already have a lot and it can be challenging for users to be able to properly decide which pipeline is an appropriate one.

chmeliik: FWIW I'd say having a new entry in the pipelines list is not so bad. The UI doesn't currently provide any information about the available pipelines, but IIRC there were plans to have it show the `.spec.description`, which could help

PR #2205, opened 2025-04-16
Add hermetic support to run-script

Zokormazo:
Add support for hermeto prefetched artifacts to run-script .

This is required for rhdh dynamic plugins, the script itself is the one triggering the builds

### DIFF
--- a/task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
+++ b/task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
@@ -173,7 +173,7 @@ spec:
           echo "The script will be executed with network isolation"
           UNSHARE_ARGS+=("--net")
           HERMETO_MOUNT="--volume /var/workdir/cachi2:/cachi2"
-          HERMETO_SOURCE="source /cachi2/cachi2.env &&"
+          HERMETO_SOURCE=". /cachi2/cachi2.env &&"
         fi
 
         # Create the script file


arewm (dismissed):

task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
```diff
@@ -26,6 +26,11 @@ spec:
     - name: ociStorage
       description: The OCI repository where the Trusted Artifacts are stored.
       type: string
+    - name: CACHI2_ARTIFACT
```
arewm, line 29: Should we rename these HERMETO_ARTIFACT so that we don't have to rename them later? Or maybe just prefetch?

@chmeliik @mmorhun , what are your thoughts about a potential future for this variable name?

chmeliik: PREFETCH_ARTIFACT sounds good to me, the task is called `prefetch-dependencies` after all

Zokormazo: Will not a partial renaming be worse? I can rename here now, but the artifact coming from prefetch-dependencies will continue being `CACHI2_ARTIFACT` and integrating the task on a build pipeline will look like this:

```
        - name: PREFETCH_ARTIFACT
          value: $(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)
```

But then we go back to `CACHI2_ARTIFACT` on buildah task. The cognitive complexity to recall where is renamed and where is not is going to be higher this way.

Moreover, prefetch-dependencies will still assume `/cachi2` as the mountpoint and on the edge cases where the script writer needs to know the implementation details, she will need to refer to artifact on `/cachi2` .

For example, I have a case where I need to install yarn globally before going further, on that case I need to know the location of the tgz and install it directly from there:

```
npm install --global /cachi2/output/deps/npm/yarnpkg-cli-dist-3.8.3.tgz
```

For the internal variables the user does not need to know about, I already used hermeto based names, but here, I'm hesitating to make the change tbh, is going to be super confusing. The task requires dealing with the pipeline yaml to use it and the result with partial renaming is going to be a mess.

I'm all about moving to hermeto or the more generic prefetch_artifact naming everywhere, but that should be done across the board and probably with a version bump of the tasks and a migration document. Thoughts?

eskultety: > Moreover, prefetch-dependencies will still assume /cachi2 as the mountpoint and on the edge cases where the script writer needs to know the implementation details, she will need to refer to artifacts on /cachi2

NB this will change soon because Konflux will be properly flipped to the hermeto usage from 'cachi2'. That's just the legacy entrypoint in the `hermeto` image Konflux is using just to give Conforma time to catch up on the labels it's currently matching in the SBOM. I agree that `PREFETCH_ARTIFACT` is a sane and vendor agnostic choice and honestly, with the pending cachi2->hermeto renaming flip we may as well use `prefetch` for the mounted dir instead.

> I'm all about moving to hermeto or the more generic prefetch_artifact naming everywhere, but that should be done across the board

Yes, this rename will be done across the whole board, soon...we don't need to wait with using  `PREFETCH_ARTIFACT` until then IMO, it's a matter of a couple of weeks at most.

Zokormazo: I forgot to update the readme last week. Updated now.

Let me know if you prefer to keep it as `CACHI2_ARTIFACT` for now or move to `HERMETO_ARTIFACT` :)

Zokormazo: @eskultety thanks for the info! I moved towards `PREFETCH_ARTIFACT` :)

The script itself will later need adjustments to move toward `/prefetch` as the mount point and so on

task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
```diff
@@ -107,6 +112,7 @@ spec:
       args:
         - use
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
+        - $(params.PREFETCH_ARTIFACT)=/var/workdir/cachi2
```
Zokormazo, line 115: Should I use `/var/workdir/prefetch` here now? Or wait for the changes on prefetch dependencies?

Zokormazo: sounds good :+1:

task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
```diff
@@ -161,9 +167,13 @@ spec:
         echo 'root:1:4294967294' | tee -a /etc/subuid >>/etc/subgid
 
         UNSHARE_ARGS=()
+        HERMETO_MOUNT=""
+        HERMETO_SOURCE=""
         if [ "${HERMETIC}" == "true" ]; then
           echo "The script will be executed with network isolation"
           UNSHARE_ARGS+=("--net")
+          HERMETO_MOUNT="--volume /var/workdir/cachi2:/cachi2"
+          HERMETO_SOURCE="source /cachi2/cachi2.env &&"
```
chmeliik, line 176:
```suggestion
          HERMETO_SOURCE=". /cachi2/cachi2.env &&"
```

In the past, we ran into an issue where some container images had shells without `source`, only `.`

chmeliik (commented):
LGTM, minor suggestion for improved compatibility

task/run-script-oci-ta/0.1/run-script-oci-ta.yaml
chmeliik: I'll note that the `/cachi2` part of `--volume /var/workdir/cachi2:/cachi2` is the more impactful decision here since, as @Zokormazo pointed out in https://github.com/konflux-ci/build-definitions/pull/2205#discussion_r2053395888, users may sometimes depend on the directory name explicitly.

But it makes sense to me to keep this consistent with the buildah task for now and then make the changes all at once

Zokormazo:
https://github.com/Zokormazo/dynamic-plugins-custom-authentication-module/pull/3 to test it. It works :)

chmeliik (approved):

PR #2207, opened 2025-04-16
feat: Add support for ANNOTATIONS_FILE

CryptoRodeo:
# Feature

This change will allow users of this task to specify a path to a file
containing annotations for their build.

This is especially useful in scenarios where the annotation value is dynamically generated by another task and the resulting value exceeds the limit set by Tekton task results.

## Usage:
Users can have a basic `annotations.txt` file in their repo with contents such as the following:
```
test.annotation.a=abc123
test.annotation.b=def456
```

And specify the param in the following format:
```
- name: ANNOTATIONS_FILE
  value: "annotations.txt"
```

Or dynamically generate and export the annotation file in a script using the `run-script-oci-ta` task.

### DIFF
--- a/task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
@@ -39,6 +39,11 @@ spec:
         to the image
       type: array
       default: []
+    - name: ANNOTATIONS_FILE
+      description: Path to a file with additional key=value annotations that
+        should be applied to the image
+      type: string
+      default: ""
     - name: BUILDAH_FORMAT
       description: The format for the resulting image's mediaType. Valid values
         are oci (default) or docker.
@@ -212,6 +217,8 @@ spec:
         value: $(params.ADDITIONAL_SECRET)
       - name: ADD_CAPABILITIES
         value: $(params.ADD_CAPABILITIES)
+      - name: ANNOTATIONS_FILE
+        value: $(params.ANNOTATIONS_FILE)
       - name: BUILD_ARGS_FILE
         value: $(params.BUILD_ARGS_FILE)
       - name: CONTEXT
@@ -364,6 +371,17 @@ spec:
 
         LABELS=()
         ANNOTATIONS=()
+        # Append any annotations from the specified file
+        if [ -n "${ANNOTATIONS_FILE}" ] && [ -f "${SOURCE_CODE_DIR}/${ANNOTATIONS_FILE}" ]; then
+          echo "Reading annotations from file: ${SOURCE_CODE_DIR}/${ANNOTATIONS_FILE}"
+          while IFS= read -r line || [[ -n "$line" ]]; do
+            # Skip empty lines and comments
+            if [[ -n "$line" && ! "$line" =~ ^[[:space:]]*# ]]; then
+              ANNOTATIONS+=("--annotation" "$line")
+            fi
+          done <"${SOURCE_CODE_DIR}/${ANNOTATIONS_FILE}"
+        fi
+
         # Split `args` into two sets of arguments.
         while [[ $# -gt 0 ]]; do
           case $1 in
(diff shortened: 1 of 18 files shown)


CryptoRodeo:
@lcarva

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
```diff
@@ -398,6 +405,17 @@ spec:
           esac
         done
 
+        # Append any additional annotations from the specified file
```
chmeliik, line 408: IMO `ANNOTATIONS` should take precedence over `ANNOTATIONS_FILE` the same way that `BUILD_ARGS` take precedence over `BUILD_ARGS_FILE`. We would probably achieve that by parsing the annotations file first and then appending the regular annotations?

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
```diff
@@ -398,6 +405,17 @@ spec:
...
+        if [ -n "${ANNOTATIONS_FROM_PATH}" ] && [ -f "${SOURCE_CODE_DIR}/${ANNOTATIONS_FROM_PATH}" ]; then
+          echo "Reading annotations from file: ${SOURCE_CODE_DIR}/${ANNOTATIONS_FROM_PATH}"
+          while IFS= read -r line || [[ -n "$line" ]]; do
```
chmeliik, line 411: What does `|| [[ -n "$line" ]]` do here?

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
```diff
@@ -39,6 +39,11 @@ spec:
         to the image
       type: array
       default: []
+    - name: ANNOTATIONS_FROM_PATH
+      description: Path to a file with additional key=value annotations that
+        should be applied to the image
+      type: string
+      default: ""
```
chmeliik, line 46: Consider renaming to `ANNOTATIONS_FILE` for symmetry with `BUILD_ARGS` -> `BUILD_ARGS_FILE`

chmeliik (commented):
Seems like a reasonable feature, thanks for contributing :taco: 

Just a question on the motivation for adding:

> This is especially useful in scenarios where the annotation value exceeds the limit set by Tekton task results.

I'm not sure I understand this. The result size limit is relevant if you're generating your annotations dynamically in another Task. But if you're just passing a static list of annotations, AFAIK the parameter array can be as large as you want. And supporting a file won't help you with dynamically generated annotations, will it?

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
CryptoRodeo: It should handle the case where the last line doesn't end with a newline character

chmeliik:
> If more context helps, this other PR shows what I am trying to achieve: https://github.com/redhat-appstudio/rhtap-backstage-plugins/pull/7/files

Ah, you're going to use the run-script task to generate the annotations file. Makes sense :+1:

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
chmeliik: Oh wow, TIL. I wonder how many subtle bugs caused by this we have :smile:

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
CryptoRodeo: Like this? -> https://github.com/konflux-ci/build-definitions/pull/2207/files#diff-8ececcbe0dae0061a00d374353e73b4c75b606c645b90e1546ea1dcff69bb748R314-R349

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
CryptoRodeo: Yeah it's very subtle, here is a quick demo if you want to see this bug in action:
```bash
printf "first line\nsecond line" > test.txt

# test.sh
while IFS= read -r line; do
  echo "Line: '$line'"
done <test.txt

# Run
bash ./test.sh

# Output
Line: 'first line'

# test.sh (updated)
while IFS= read -r line || [[ -n "$line" ]]; do
  echo "Line: '$line'"
done <test.txt

# Run
bash ./test.sh

# Output
Line: 'first line'
Line: 'second line'
```

lcarva (approved):

chmeliik (approved):
LGTM, thanks!

Could you please just squash the commits into logical units (here, that would probably mean just one commit)?

task/buildah-oci-ta/0.4/buildah-oci-ta.yaml
chmeliik: Yup, that should work :+1:

CryptoRodeo:
@chmeliik commits have been squashed 🙂

chmeliik (approved):

PR #2210, opened 2025-04-16
fix: fix jq queries for FIPS annotations in the task

yashvardhannanavati:
The query to get the subscription label from the bundle CSV failed if the value key was a string and not an object. This commit fixes that behavior by filtering out only objects.


hongweiliu17 (approved):
lgtm

dirgim (approved):
LGTM

PR #2212, opened 2025-04-17
task/modelcar-oci-ta: Parametrize olot --remove-originals param

jrusz:
Adding a parameter for olot to be able to pass `--remove-originals`.
Also adjusting permissions on the model files so that they can actually
be remove. Also updating olot version to support this new param.


Victoremepunto (approved):
nicely done!!

Victoremepunto (approved):

PR #2214, opened 2025-04-17
deprecate old tasks owned by the build team

fmudry:
Expiry date is set to the 31st of May.
After the deprecation period, they should be removed.

### DIFF
--- a/task/init/0.1/init.yaml
+++ b/task/init/0.1/init.yaml
@@ -4,6 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/version: "0.1"
   annotations:
+    build.appstudio.redhat.com/expires-on: "2025-05-31T00:00:00Z"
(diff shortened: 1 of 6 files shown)


chmeliik (approved):
LGTM :+1:

Note for others: buildah tasks < 0.4 are already deprecated

task/init/0.1/init.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/version: "0.1"
   annotations:
+    build.appstudio.redhat.com/expires-on: "2025-06-30T00:00:00Z"
```
mmorhun, line 7: I think we can make them expire in shorter term, like 31-st of May.

fmudry: Updated.

PR #2215, opened 2025-04-17
fix(KONFLUX-7841): Increase memory for fbc pruning check

nmars:
Increased memory for fbc-target-index-pruning-check.


dirgim (approved):
LGTM

PR #2216, opened 2025-04-17
fix(KONFLUX-7846): fix parsing of related imgs in fips stepaction

nmars:
Fixed the way the fips stepaction parses `unique_related_images.txt`.

### DIFF
--- a/stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml
+++ b/stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml
@@ -38,7 +38,7 @@ spec:
       exit 0
     fi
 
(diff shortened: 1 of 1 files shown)


stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml
```diff
@@ -60,6 +60,7 @@ spec:
 
     declare -i count=0
     for related_image in "${related_images[@]}"; do
+      related_image="${related_image//$'\n'/}"
```
nmars, line 63: In testing this locally, the `-t` in the `mapfile -d ' ' -t` on line 41 above did not seem to remove the newline from the final element in the `related_images` array. So I introduced this variable expansion here to strip the newline if it exists.

yashvardhannanavati (approved):

PR #2222, opened 2025-04-20
chore(KFLUXVNGD-242): add kustomization support to tkn-bundle

avi-biton:
Add support to kustomization file.
In case a `kustomiztion.yaml` file is located in `CONTEXT` it will be used to generate the task definition and all other files in `CONTEXT` will be ignored

### DIFF
--- a/task/tkn-bundle/0.2/tkn-bundle.yaml
+++ b/task/tkn-bundle/0.2/tkn-bundle.yaml
@@ -172,6 +172,37 @@ spec:
           printf '"%s"' "$escaped_arg"
         }
 
+        # Check if task is kustomized
+        # A task considered to be kustomized when a kustomization file is found in the task dir
+        is_kustomized_task() {
+            local -r task_dir=$1
+            if [[ -f "$task_dir/kustomization.yaml" || -f "$task_dir/kustomization.yml" ]] ; then
+                return 0
+            fi
+            return 1
+        }
+
+        # Generate a task from the kustomization file
+        generate_kustomized_task() {
+            local -r task_dir=$1
+            local -r prepared_task_file="$task_dir/generated-kustomized-task.yaml"
+            kubectl kustomize "$task_dir" >"$prepared_task_file"
+            echo "$prepared_task_file"
+        }
+
+        # task_dir is where all the tasks definitions reside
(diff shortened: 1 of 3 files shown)


task/tkn-bundle/0.2/tkn-bundle.yaml
```diff
@@ -172,6 +172,38 @@ spec:
           printf '"%s"' "$escaped_arg"
         }
 
+        # Check if task is kustomized
+        # A task considered to be kustomized when a kustomization file is found in the task dir
+        is_kustomized_task() {
+            local -r task_dir=$1
+            local -r kustomiztion_file="$task_dir/kustomization.yaml"
```
yftacherzog, line 179: For the task itself we mention that it finds both yaml and yml files, but for the kusotmization files here we only support one extension. I think it's worth adding a check for both `$task_dir/kustomization.yaml` and `$task_dir/kustomization.yml` to avoid potential misunderstandings.

(and also changing the echo command later -- maybe it doesn't have to include the extension at all there)

task/tkn-bundle/0.2/tkn-bundle.yaml
```diff
@@ -172,6 +172,38 @@ spec:
...
+
+        # Generate a task from the kustomization file
+        generate_kustomized_task() {
+            local -r task_dir=$1
+            local -r prepared_task_file="$task_dir/generated-kustomized-task.yaml"
+            kubectl kustomize "$task_dir" >"$prepared_task_file"
+            echo "$prepared_task_file"
+        }
+
+        # task_dir is where all the tasks definitions resides
```
yftacherzog, line 194:
```suggestion
        # task_dir is where all the tasks definitions reside
```

task/tkn-bundle/0.2/tkn-bundle.yaml
avi-biton: I added a check for `kustomization.yaml` and `kustomization.yml` and updated the echo command

task/tkn-bundle/0.2/tkn-bundle.yaml
avi-biton: Done

PR #2228, opened 2025-04-22
Work around broken cosign auth in build tasks

chmeliik:
Cosign uses the go-containerregistry module for authentication. This module doesn't implement containers-auth.json handling properly [1]. In particular, the keys in the auth file must either be full image names or just hostnames:

    quay.io/my-org/my-image
    quay.io

Partial paths are not supported:

    quay.io/my-org

Use the select-oci-auth script as a workaround to make cosign work for partial paths in the auth file.

[1]: https://github.com/google/go-containerregistry/pull/1966

### DIFF
--- a/task/show-sbom/0.1/show-sbom.yaml
+++ b/task/show-sbom/0.1/show-sbom.yaml
@@ -45,6 +45,7 @@ spec:
         status=-1
         max_try=5
         wait_sec=2
+        local extra_args=("$@")
 
         # Pre-select the correct credentials to work around cosign not supporting the containers-auth.json spec
         mkdir -p /tmp/auth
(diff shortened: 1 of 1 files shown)


mmorhun (approved):
Would be nice to have a comment in the code that explain why we do this workaround

chmeliik:
Before setting this ready for review, we'll need to merge #2242 to add the required script to the utils image.

Tested in https://github.com/chmeliik/multi-arch-konflux-sample/pull/20 that after updating the appstudio-utils image, this will do what we need

chmeliik:
The stderr from the `select-oci-auth` script was breaking the e2e-tests when trying to parse the output of the show-sbom task. Fixed.

chmeliik:
`timed out when waiting for image-controller annotations to be updated on component ...`

/retest

mmorhun (approved):

PR #2233, opened 2025-04-23
chore(KFLUXVNGD-283): use scoped auth for helm push

yftacherzog:
Helm is unable to select the correct auth when pushing to OCI registries, so need to help it select the right one.


avi-biton (approved):
/lgtm

PR #1931, opened 2025-02-10
Set preflight to SUCCESS for operator bundles

dhaiducek:
Having a SKIPPED result for `ecosystem-cert-preflight-checks` for operator bundles doesn't align with what Conforma expects.

ref: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1739212062777609

PR #2237, opened 2025-04-23
Set preflight to SUCCESS for operator bundles

aleskandro:
`ecosystem-cert-preflight-checks` is skipped for operator bundles. However, conforma checks fail if the result of it is not `SUCCESS`. This commit changes the task code to return `SUCCESS` when skipping due to the image under test being an operator bundle.

Originally-implemented-by: Dale Haiducek <19750917+dhaiducek@users.noreply.github.com>

Fixes #1931


komish:
@aleskandro Why a second PR for this when https://github.com/konflux-ci/build-definitions/pull/1931 exists? I assume it's because you're adding the requested comment?

komish (approved):
/lgtm

PR #2239, opened 2025-04-23
[ISV-5704] Used parent image and runtime arch is identified.

BorekZnovustvoritel:
# Identify parent image

This PR adds variables to the `buildah` task identifying the parent image and arch used at runtime.

To do this, PR simply stores the last value of the `BASE_IMAGES` array if no `TARGET_STAGE` is specified.

If this argument is specified, then it looks for an image alias (attribute `As` in the `dockerfile-json` output) with value of `$TARGET_STAGE`. If not found, the target must be a pullspec.

If the image alias is found, the script tries to look for an image which has the alias set to the `BaseName` of the previous image. This way even transitive references are resolved (example [here](https://github.com/keilerkonzept/dockerfile-json?tab=readme-ov-file#json-output)).

The `buildah-oci-ta` and `buildah-remote` tasks have been updated by the scripts in the `hack` directory, I have skipped Buildah 0.1 as there is no parsed Dockerfile there.

This has been tested on [this repository](https://github.com/BorekZnovustvoritel/Konflux-testrepo/pull/2) and a private Konflux app.

Ticket for tracking: [ISV-5704](https://issues.redhat.com//browse/ISV-5704)

### DIFF
--- a/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
+++ b/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
@@ -394,7 +394,43 @@ spec:
 
         if [ -n "${TARGET_STAGE}" ]; then
           BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
+
+          # Search for parent image
+
+          # step 1: try to find the build step with the target as an alias
+          LAST_IMAGE_REF=$(
+            jq --arg alias "${TARGET_STAGE}" '.Stages[] |
+              select(.As==$alias) | .BaseName |
+              select(test("^oci-archive:") | not)' \
+              /shared/parsed_dockerfile.json |
+              tr -d '"' |
+              tr -d "'"
+          )
+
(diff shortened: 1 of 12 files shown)


task/buildah-oci-ta/0.2/buildah-oci-ta.yaml
```diff
@@ -394,8 +394,44 @@ spec:
 
         if [ -n "${TARGET_STAGE}" ]; then
           BUILDAH_ARGS+=("--target=${TARGET_STAGE}")
+
```
Allda, line 397: I assume this is a start and prepareation fo contextual SBOMs. I am not really sure if adding more bash code for simple logic like this is way to go forward. With this code it can't be tested properly and we can't verify if the logic is correct. How about wrapping up the logic into some Python scirpts that handles the overall contextual SBOM logic.

BorekZnovustvoritel: I will reimplement this. Closing this PR.

PR #2242, opened 2025-04-24
appstudio-utils: add select-oci-auth script

chmeliik:
This script is already installed in the 'oras' and 'build-trusted-artifacts' container images to work around oras not supporting containers-auth.json properly [1].

Add the script to the appstudio-utils image as well, which will enable us to also use this workaround for cosign. Cosign uses the go-containerregistry module for authentication, which works a bit better than oras, but still not well enough [2].

The script was copied from the build-trusted-artifacts repo [3].

[1]: https://github.com/oras-project/oras-go/issues/840
[2]: https://github.com/google/go-containerregistry/pull/1966
[3]: https://github.com/konflux-ci/build-trusted-artifacts/blob/ddb050d092df562860fe2522d4de234ed49dabd3/select-oci-auth.sh

### DIFF
--- a/appstudio-utils/Dockerfile
+++ b/appstudio-utils/Dockerfile
@@ -27,3 +27,4 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     && dnf clean all
 
 COPY util-scripts /appstudio-utils/util-scripts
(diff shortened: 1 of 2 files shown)


chmeliik:
This is a prerequisite for #2228

mmorhun (approved):

PR #2261, opened 2025-04-29
RHTAP-4382 - Use ca bundle from tekton-custom-certs if available.

rhopp:


tnevrlka (commented):
The changes look good to me.
Would you mind adding some reasoning for this change into the commit description? Thank you!

tnevrlka (approved):

PR #2269, opened 2025-05-02
buildah: add WORKINGDIR_MOUNT parameter

chmeliik:
Allows to mount the current working directory into the build using
--volume $PWD:/$WORKINGDIR_MOUNT. Note that the $PWD will be the context
directory for the build, because we set the workdir to the context dir
before calling 'buildah build'.

The primary use case for this parameter is for builds that need to write
some outputs into a shared directory and reference the output in a later
FROM instruction, e.g.

    FROM oci-archive:./out.ociarchive

See https://github.com/konflux-ci/buildah-container/issues/134 for more details.

### DIFF
--- a/task/buildah-min/0.4/buildah-min.yaml
+++ b/task/buildah-min/0.4/buildah-min.yaml
@@ -153,6 +153,12 @@ spec:
       image_reference_with_digest strings
     name: ADDITIONAL_BASE_IMAGES
     type: array
+  - default: ""
+    description: Mount the current working directory into the build using --volume
+      $PWD:/$WORKINGDIR_MOUNT. Note that the $PWD will be the context directory for
+      the build (see the CONTEXT param).
+    name: WORKINGDIR_MOUNT
+    type: string
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
@@ -216,6 +222,8 @@ spec:
       value: $(params.SBOM_TYPE)
     - name: ANNOTATIONS_FILE
       value: $(params.ANNOTATIONS_FILE)
+    - name: WORKINGDIR_MOUNT
+      value: $(params.WORKINGDIR_MOUNT)
     volumeMounts:
     - mountPath: /shared
       name: shared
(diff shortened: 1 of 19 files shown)


chmeliik:
Testing in https://gitlab.com/acmiel/centos-bootc/-/merge_requests/11

task/buildah-min/0.4/buildah-min.yaml
```diff
@@ -490,6 +498,13 @@ spec:
         echo "Adding the entitlement to the build"
       fi
 
+      if [ -n "$WORKINGDIR_MOUNT" ]; then
+        # ${SOURCE_CODE_DIR}/${CONTEXT} will be the $PWD when we call 'buildah build'
+        # (we set the workdir using 'unshare -w')
+        context_dir=$(realpath "${SOURCE_CODE_DIR}/${CONTEXT}")
+        VOLUME_MOUNTS+=(--volume "$context_dir:${WORKINGDIR_MOUNT}")
```
cgwalters, line 505: This is totally not a new issue however note that `${WORKINGDIR_MOUNT}` could in theory contain `:` which could allow injecting options like `:Z` or whatever, i.e. `WORKINGDIR_MOUNT=/buildcontext:Z`. Maybe simplest to reject mount destinations containing `:`.

cgwalters (commented):
Looks sane to me, just one comment

cgwalters:
Also just for extra visibility you need https://gitlab.com/acmiel/centos-bootc/-/merge_requests/11#note_2482149950 when testing this.

task/buildah-min/0.4/buildah-min.yaml
chmeliik: Good suggestion, added a check :+1:

chmeliik:
Tested successfully in https://gitlab.com/acmiel/centos-bootc/-/merge_requests/11

tnevrlka (approved):

kdudka (approved):
The changes in `sast-coverity-check*` are implied by changes in the `buildah` task.  Hence approving this part.

PR #2278, opened 2025-05-05
delete deprecated tasks owned by build team

fmudry:
Removing all deprecated tasks owned by build team to stop Renovate updates. 
Only expired tasks will be deleted; the rest will expire at the end of May.


tnevrlka:
@fmudry I think you'll have to change [this for loop](https://github.com/konflux-ci/build-definitions/blob/71bfb9c62936d3db3fcdb723d0be951a16b9aaeb/hack/generate-buildah-remote.sh#L9)

PR #2283, opened 2025-05-06
fix: check cachi2 dir existence/PREFETCH_ARTIFACT being non-empty

fmudry:
some build would fail, because cachi2 directory doesn't exist in the image


chmeliik (commented):
I think the two conditions are functionally equivalent (`/var/workdir/cachi2` is a directory <=> `PREFETCH_ARTIFACT` is not empty) so we only need one of them

> some build would fail, because cachi2 directory doesn't exist in the image

More specifically, this task would fail any time someone uses it with HERMETIC=true but without prefetching.

PR #2314, opened 2025-05-13
Add note to sast tasks about KFP directory names

sfowl:
Supports https://issues.redhat.com/browse/PSSECAUT-1147 but I also intend to update other docs.

Note sure if I missed something, but I assumed that if I just updated the sast non-trusted-artifact tasks the `hack/generate-everything.sh` script would update the -oci-ta tasks for me. That didn't seem to occur, so I updated the -oci-ta tasks manually.

I also did some light markdown linting, like wrapping bare URLs in `<>` and whitespace tidy.

### DIFF
--- a/task/sast-coverity-check/0.3/README.md
+++ b/task/sast-coverity-check/0.3/README.md
@@ -4,7 +4,7 @@
 
 The sast-coverity-check task uses Coverity tool to perform Static Application Security Testing (SAST).
 
-The documentation for this mode can be found here: https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/commands/topics/coverity_capture.html
+The documentation for this mode can be found here: <https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/commands/topics/coverity_capture.html>
 
 The characteristics of these tasks are:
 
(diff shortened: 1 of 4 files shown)


hongweiliu17 (dismissed):
lgtm

task/sast-coverity-check/0.3/README.md
```diff
@@ -38,8 +41,7 @@ The characteristics of these tasks are:
 
 // TODO: Add reference to private repo for the container image once the task is migrated to repo
 
-
 ## Additional links:
 
-* https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/commands/topics/coverity_capture.html
-* https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/cli/topics/options_reference.html
+* <https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/commands/topics/coverity_capture.html>
```
hongweiliu17, line 46: Why `<>` is added here? I can't see the difference when I read https://github.com/konflux-ci/build-definitions/tree/sast-kfp-note/task/sast-coverity-check/0.3

kdudka (dismissed):

task/sast-coverity-check/0.3/README.md
sfowl: It was just a linting suggestion from my editor:

https://github.com/DavidAnson/markdownlint/blob/main/doc/md034.md

I can see why it's distracting though. I'll avoid such changes in future PRs.

sfowl:
@kdudka @hongweiliu17 I pushed a small update that tweaks the note from "ignore rules" -> "path exclusions" because I learned the top level directory names in KFP only affect `exclude-paths.txt` not `ignore.err`

I also undid the oci-ta task readme changes, as it seems these readmes are generated from the task yaml in the non -ta tasks.

Both of your approvals seem to have been lost because of these changes, PTAL :pray:

kdudka (approved):
Looks good.  Thanks for the update!

hongweiliu17 (approved):
lgtm

PR #2327, opened 2025-05-16
fix(KONFLUX-8337): Local storage directory names should be unique to …

yashvardhannanavati:
…the image

If an operator uses multiple versions of the same image in the relatedImages section, the directory names used by the stepaction are not unique enough and result in false positives. This commit fixes that issue


asergienk (approved):
LGTM

nmars (approved):
LGTM

PR #2332, opened 2025-05-19
fix(EET-4670): override arch/os in ecosystem-cert-preflight-checks

caxu-rh:
During introspection, check if the provided image is a multi-arch image. If so, explicitly override the arch/os during the subsequent inspect step to the first image/manifest in the list.

skopeo will fail to inspect a multi-arch image if the image does not cover the host system where skopeo is running, unless an explicit arch/os override is specified.

---

Some of this logic could probably be improved/simplified if there is progress on https://github.com/containers/skopeo/issues/1283.

### DIFF
--- a/task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
+++ b/task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
@@ -76,9 +76,55 @@ spec:
           exit 0
         fi
 
+        # If the image URL points to a manifest list (a multi-arch image), check the labels on any of the child
+        # images (don't fail in the case where the list does not include an image for the arch of the system
+        # where this pipeline is running).
+
+        declare -a _SKOPEO_INSPECT_ARGS
+
+        echo "Checking the media type of the OCI artifact..."
+        _RAW_IMAGE_MANIFEST=$(skopeo inspect --raw "docker://${PARAM_IMAGE_URL}")
+        _IMAGE_MEDIA_TYPE=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.mediaType')
+        echo "The media type of the OCI artifact is ${_IMAGE_MEDIA_TYPE}."
+
+        if [[ "${_IMAGE_MEDIA_TYPE}" == "application/vnd.docker.distribution.manifest.list.v2+json" || "${_IMAGE_MEDIA_TYPE}" == "application/vnd.oci.image.index.v1+json" ]]; then
+          _CURRENT_ARCH=$(uname -m)
+          _CURRENT_OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+          # The archs returned by uname are not always the same as the archs used by OCI manifests, so we need
+          # to map them.
+          case ${_CURRENT_ARCH} in
+            "aarch64")
+              _CURRENT_ARCH="arm64"
+              ;;
+            "x86_64")
+              _CURRENT_ARCH="amd64"
+              ;;
+            *)
+              ;;
+          esac
+
+          # If the manifest list contains an image for the current OS and architecture, prefer to test that.
+          _MATCHING_IMAGE_COUNT=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r "[.manifests[] | select(.platform.os == \"${_CURRENT_OS}\" and .platform.architecture == \"${_CURRENT_ARCH}\")] | length")
+          if [[ "${_MATCHING_IMAGE_COUNT}" -gt 0 ]]; then
+            echo "Found an image in the manifests for the current OS and architecture (${_CURRENT_OS}/${_CURRENT_ARCH})."
+          else
+            # If there is no image for the current OS and architecture, just use the first one in the list.
+            _INSPECT_OVERRIDE_IMAGE_OS=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.os')
+            _INSPECT_OVERRIDE_IMAGE_ARCH=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.architecture')
+            _SKOPEO_INSPECT_ARGS+=("--override-os=${_INSPECT_OVERRIDE_IMAGE_OS}")
+            _SKOPEO_INSPECT_ARGS+=("--override-arch=${_INSPECT_OVERRIDE_IMAGE_ARCH}")
+
+            echo "Could not find an image in the manifests for the current OS and architecture (${_CURRENT_OS}/${_CURRENT_ARCH}), inspecting the image for ${_INSPECT_OVERRIDE_IMAGE_OS}/${_INSPECT_OVERRIDE_IMAGE_ARCH} instead."
+          fi
+        fi
+
         # Introspect based on minimum count of operator-framework related bundle labels.
         echo "Looking for image labels that indicate this might be an operator bundle..."
-        skopeo inspect "docker://${PARAM_IMAGE_URL}" \
+
+        # We purposely do not quote the array elements here, so that they are expanded by the shell as separate args.
+        # shellcheck disable=SC2068
+        skopeo inspect ${_SKOPEO_INSPECT_ARGS[@]} "docker://${PARAM_IMAGE_URL}" \
           | jq '.Labels | keys | .[]' -r \
           | { grep operators.operatorframework.io.bundle || true ;} \
           | tee /tmp/ecosystem-image-labels


task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,21 @@ spec:
...
+        _IMAGE_MEDIA_TYPE=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.mediaType')
+        if [[ "${_IMAGE_MEDIA_TYPE}" == "application/vnd.docker.distribution.manifest.list.v2+json" ]]; then
+          _INSPECT_OVERRIDE_IMAGE_ARCH=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.architecture')
+          _INSPECT_OVERRIDE_IMAGE_OS=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.os')
+        fi
+
         # Introspect based on minimum count of operator-framework related bundle labels.
         echo "Looking for image labels that indicate this might be an operator bundle..."
-        skopeo inspect "docker://${PARAM_IMAGE_URL}" \
+        skopeo inspect --override-arch "${_INSPECT_OVERRIDE_IMAGE_ARCH}" --override-os "${_INSPECT_OVERRIDE_IMAGE_OS}" "docker://${PARAM_IMAGE_URL}" \
```
acornett21, line 93: If the `if` block above doesn't execute the cmd ends up looking like the below correct? and there is no issue with empty flag values?

```
skopeo inspect --override-arch "" --override-os ""
```

caxu-rh: In my local testing, this seems to work fine (`--override-arch "" --override-os ""` or `--override-arch= --override-os=` both seem to just have the effect of not setting those respective overrides).

This could be adjusted to avoid emitting the flags at all if those variables are empty, if that would make more sense.

komish: +1, I would prefer this did not emit a flag if it would emit it with an empty value.

Mostly a nit, but a reader would see this as written and probably assume the value of `"${_INSPECT_OVERRIDE_IMAGE_ARCH}"` always has _some_ value.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,21 @@ spec:
...
         fi
 
+        # If the image URL points to a manifest list (a multi-arch image), check the labels on any of the child
+        # images (don't fail in the case where the list does not include an image for the arch of the system
+        # where this pipeline is running).
+        _INSPECT_OVERRIDE_IMAGE_ARCH=
+        _INSPECT_OVERRIDE_IMAGE_OS=
+        _RAW_IMAGE_MANIFEST=$(skopeo inspect --raw "docker://${PARAM_IMAGE_URL}")
+        _IMAGE_MEDIA_TYPE=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.mediaType')
+        if [[ "${_IMAGE_MEDIA_TYPE}" == "application/vnd.docker.distribution.manifest.list.v2+json" ]]; then
```
komish, line 86: You probably need to also match the OCI Index mediatype, as the pipelines (and tooling) seem to default to it. I see it here: https://specs.opencontainers.org/image-spec/image-index/

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,21 @@ spec:
           exit 0
         fi
 
+        # If the image URL points to a manifest list (a multi-arch image), check the labels on any of the child
+        # images (don't fail in the case where the list does not include an image for the arch of the system
+        # where this pipeline is running).
+        _INSPECT_OVERRIDE_IMAGE_ARCH=
+        _INSPECT_OVERRIDE_IMAGE_OS=
```
komish, line 83: I think it might be worthwhile to optionally accept a parameter here for either of these values, but it might complicate the logic a bit more than necessary for the nearly-zero count of bugs we've seen for issues related to this.

No immediate action related to this comment. Just something to think about.

komish (changes requested):
Some changes requested in-line.

Overall, this should get the ball rolling. The only concern I have is that we default to resolving the os/arch values, even for cases where the manifest list _has_ an entry matching the runtime platform of this task. In other words, it could have us pulling a non-matching image when one would otherwise exist, which isn't my favorite.

Curious to hear your thoughts on it.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,21 @@ spec:
...
+        _RAW_IMAGE_MANIFEST=$(skopeo inspect --raw "docker://${PARAM_IMAGE_URL}")
```
komish, line 84: Forgot one thing: would like to see an `echo` log line emitted just before this call so that, should we see a skopeo error in the log, we can differentiate between _this_ call and the call that comes later.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
komish: Might also want to log the media type if it's not already being emitted somewhere, just for debugging.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
acornett21: I was thinking this might need to be a param too, but then that yields potential weird results. If we run into this often, we can just have the user add the params in their execution environment, no sense in over optimizing now. That's where I landed when I thought about it.

komish: Users can't add params if the Task doesn't support those params. That said, +1 to waiting until we have more issues related to this to where it makes sense to work around this problem. The workaround of "skip introspection" seemed reasonable enough for the case we have.

acornett21: I forgot this comes from an image and not raw tasks in a cluster, so you're correct a user couldn't edit the task itself, only the pipeline calling the task.

caxu-rh: I've held off from adding this as a param in the Tekton task for now; I feel like there's enough moving parts in this PR to leave that further change for another PR.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,57 @@ spec:
...
+        echo "The media type of the OCI artifact is ${_IMAGE_MEDIA_TYPE}."
+
+        if [[ "${_IMAGE_MEDIA_TYPE}" == "application/vnd.docker.distribution.manifest.list.v2+json" || "${_IMAGE_MEDIA_TYPE}" == "application/vnd.oci.image.index.v1+json" ]]; then
+          _CURRENT_ARCH=$(uname -m)
+          _CURRENT_OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+          # The archs returned by uname are not always the same as the archs used by OCI manifests, so we need
+          # to map them.
+          case ${_CURRENT_ARCH} in
+          "aarch64")
```
komish, line 97: Indentation issue here?

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,57 @@ spec:
...
+          if [[ "${_MATCHING_IMAGE_COUNT}" -gt 0 ]]; then
+            echo "Found an image in the manifests for the current OS and architecture (${_CURRENT_OS}/${_CURRENT_ARCH})."
+          else
+            # If there is no image for the current OS and architecture, just use the first one in the list.
+            _INSPECT_OVERRIDE_IMAGE_OS=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.os')
+            _INSPECT_OVERRIDE_IMAGE_ARCH=$(printf "%s" "${_RAW_IMAGE_MANIFEST}" | jq -r '.manifests[0].platform.architecture')
+            _SKOPEO_INSPECT_ARGS+=("--override-os=${_INSPECT_OVERRIDE_IMAGE_OS}")
+            _SKOPEO_INSPECT_ARGS+=("--override-arch=${_INSPECT_OVERRIDE_IMAGE_ARCH}")
+
+            echo "Could not find an image in the manifests for the current OS and architecture (${_CURRENT_OS}/${_CURRENT_ARCH}), inspecting the image for ${_CURRENT_OS}/${_CURRENT_ARCH} instead."
```
komish, line 118:
```suggestion
            echo "Could not find an image in the manifests for the current OS and architecture (${_CURRENT_OS}/${_CURRENT_ARCH}), inspecting the image for ${_INSPECT_OVERRIDE_IMAGE_OS}/${_INSPECT_OVERRIDE_IMAGE_ARCH} instead."
```

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -76,9 +76,57 @@ spec:
...
+          fi
+        fi
+
+        _SKOPEO_INSPECT_ARGS+=("docker://${PARAM_IMAGE_URL}")
```
komish, line 122: Nit: remove this, and instead just add it to the skopeo invocation below on L129-130 (or so).

komish (changes requested):
Overall, LGTM. Just a couple of small issues remain.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
komish: Works for me. Resolving this.

komish (approved):
/lgtm

acornett21 (approved):

PR #2333, opened 2025-05-19
fix: remove new lines from local storage dir names

asergienk:
1. Fix retrieval of the version label to extract only the label that starts with the word version. Currently, it matches multiple labels that contain the word version:
* io.buildah.version
* upstream-version
* version

2. Sanitize the component, version, and release labels by removing newlines and whitespace.


yashvardhannanavati (approved):

nmars (approved):
LGTM

PR #2342, opened 2025-05-21
pipelines: pass correct image url to source task

chmeliik:
Previously, it was possible to set parameters in such a way that the image url passed to the build-source-image task wouldn't exist.

In particular:

- IMAGE_APPEND_PLATFORM=true for the buildah-remote task; this causes the output image to become {output-image}-{platform}
- ALWAYS_BUILD_INDEX=false for the build-index-image task; this causes the build-index-image task to skip index image generation if there is only a single image.
- build-plaforms=[only one platform]

This combination of parameters causes the pipeline to not push the {output-image} at all, there's only {output-image}-{platform}. That's likely not what the user intended, but nevertheless, there's no reason the build-source-image task should fail because of it.

Pass the IMAGE_URL result from the build-index-image task instead of passing the output-image param. When building the index is disabled, the build-index-image task just forwards the result of the per-platform build task. That ensures the image passed to the source-build task exists, regardless of the parameter combination.

### DIFF
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -286,7 +286,7 @@ spec:
   - name: build-source-image
     params:
     - name: BINARY_IMAGE
-      value: $(params.output-image)
+      value: $(tasks.build-image-index.results.IMAGE_URL)
     runAfter:
     - build-image-index
     taskRef:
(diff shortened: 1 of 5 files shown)


chmeliik:
> timed out when waiting for image-controller annotations to be updated on component

PR #2344, opened 2025-05-22
Introduce 'archived-tasks' directory and create symlinks to old tasks

eskultety:
Several tasks have several revisions present in their respective task directory. The rule of thumb is that only a single task revision is maintained at all times, although there isn't a formal statement on this anywhere. Some tasks deleted their old revisions, but not all.

Having unmaintained task revisions present is just painful to work with especially using tools like 'git grep' when the number of revisions isn't known upfront and so filtering is impossible. Without having a global formal statement on the task deprecation policy it's unsafe to just delete a bunch of tasks right away.
Instead, this patch introduces a top-level 'archived-tasks' directory where all old tasks are to be hosted until it is decided when they should be deleted. A symlink is provided in the original location to retain a working reference and hence making this change transparent and non-intrusive.

Having old task revisions moved out of the 'task' directory suddenly makes working with 'git grep <pattern> task/' a breeze because symlinks are skipped unless explicitly told.


eskultety:
Since v1:
- added a change to `hack/missing-ta-tasks.sh` check to skip symlinked paths which globstar follows for some reason

cgwalters (approved):

eskultety:
> I think the next step btw is automation that checks for tasks that are unused and fully removes them.

Absolutely, but that is a story for another day and like I mentioned in the description requires a uniform guideline on task deprecation policy which the check can then follow and enforce.

chmeliik (commented):
Currently, the mechanism to indicate task deprecation is the `build.appstudio.redhat.com/expires-on` label. Going by the presence of this label, not all the archived tasks are deprecated

```bash
for task_dir in archived_tasks/*; do
  taskname=$(basename "$task_dir")
  for taskfile in "$task_dir"/*/"$taskname.yaml"; do
    echo "$taskfile"
  done
done | xargs yq '{"file": filename, "is_deprecated": .metadata.annotations["build.appstudio.redhat.com/expires-on"] != null}'
```

And the `expires-on` value of some tasks is in the future, but IMO it's fair to consider those deprecated already (though maybe not "archived" yet?)

chmeliik:
> @konflux-ci/build-maintainers why does the
> 
> https://github.com/konflux-ci/build-definitions/blob/9f8fa5cfd36a41c1d5627d2ef28738919747a1b9/.tekton/pull-request.yaml#L8
> pipeline run definition rely on a sast-snyk-check task version 0.2 when 0.4 is the latest maintained one?

Probably an oversight

kdudka (changes requested):
All directories maintained in the `build-definitions` repository use `-` as the word separator.  So I think it would be more systematic to name the directory `archived-tasks` (with dash instead of underscore).

eskultety:
> Currently, the mechanism to indicate task deprecation is the build.appstudio.redhat.com/expires-on label. Going by the presence of this label, not all the archived tasks are deprecated

So, I'm currently rebranding cachi2 -> hermeto/prefetch in all places and IIUC I should only assume the latest revision task files. That is non-deterministic however when using sed and so with this change I just tried to make my life easier to make sense out of the diff the automation produced for me.

> And the expires-on value of some tasks is in the future, but IMO it's fair to consider those deprecated already (though maybe not "archived" yet?)

Note that some tasks, e.g. `sast-snyk-check` has 4 revisions and none of them defines any expiration metadata so even if I wanted to update the archiving piece of code I used, sast tasks would have to be addressed manually.

Let me ask differently:  If the overall sentiment is to only update the latest task revisions why is it a problem to have tasks that DO HAVE a newer revision available but have not reached the expiration date in an archive directory if work is not supposed to happen on them? IIUC only the latest stuff should be deployed to production with the old maintained stuff pretty much just serving as a reachable reference.
If we can't reach an agreement here then I guess the only other way (the proper way) would be to remove all tasks which are past their expiration already first, lookup all tasks which don't have an expiration date and set one and then skip archiving completely, which is definitely the correct solution, but also one involving much more manual and tedious work. Archiving of old tasks is just an intermediary step towards that future goal.

kdudka (dismissed):

chmeliik:
The overall sentiment *within the Build team* is to only update the latest task version. But if you get the approvals from all the code owners here, then we can take that as a sign that everyone agrees with it :)

> IIUC only the latest stuff should be deployed to production with the old maintained stuff pretty much just serving as a reachable reference.

We build and push all task files when they get updated. For the old versions, the only updates they get are typically image updates from Renovate. So even deprecated tasks continue getting CVE fixes until they're deleted from the repo (as the expired Build tasks have been).

But if moving the tasks to the `archived-tasks` directory doesn't break either Renovate or the build-and-push.sh script, then I have no issue with this PR

chmeliik (dismissed):

eskultety:
So what does this tell me?

> [User error] Validation failed for pipelinerun build-definitions-pull-request-z2t4r with error invalid input params for task : missing values for these params which have no default values: [image-url image-digest]

chmeliik:
> So what does this tell me?
> 
> > [User error] Validation failed for pipelinerun build-definitions-pull-request-z2t4r with error invalid input params for task : missing values for these params which have no default values: [image-url image-digest]

Required parameters are missing, likely due to bumping the sast-* task versions in the `.tekton/` pipeline

eskultety:
> > Any way the `build-and-push` script can be tested to be sure?
> 
> Here's a command from my shell history:
> 
> ```
> QUAY_NAMESPACE=acmiel/tekton-catalog BUILD_TAG=$(date --iso-8601=date) SKIP_BUILD=1 SKIP_INSTALL=1 hack/build-and-push.sh
> ```
> 
> The part of QUAY_NAMESPACE before `/` needs to be something you can access on quay.io (e.g. your username), otherwise it should work.
> 
> The first time your run it, it should push one task bundle for each version of each task. The second time, it should push only tasks that have changed (the script figures this out by looking at the latest commit sha that updated the task file)

@chmeliik so here's a snippet of the stdout log with this PR in effect and local change to an archived task version:
```
<----------------- 0.1 is archived------------------>
info: build and push task task/clair-scan/0.1
info: finding the previous task bundle that has a migration
info: push new bundle quay.io/eskultety/tekton-catalog/test:clair-scan-0.1
Creating Tekton Bundle:
	- Added Task: clair-scan to image

Pushed Tekton Bundle to quay.io/eskultety/tekton-catalog/test@sha256:996587df819c92664b655b106fa4d791242ec28c355eceaeee2fbb8acfc1b6f2
Created:
quay.io/eskultety/tekton-catalog/test:clair-scan-0.1@sha256:996587df819c92664b655b106fa4d791242ec28c355eceaeee2fbb8acfc1b6f2
Copying blob sha256:2c1e9efa1ac2e65f086aad861bb97a62de05697c63f704eb3bf39d9d46be142b
Copying config sha256:8c4127897b7d1cf71609a86e5eb4884eebdf484e719ff886c1980cf8291f2231

info: inject task bundle to pipelines quay.io/eskultety/tekton-catalog/test:clair-scan-0.1@sha256:996587df819c92664b655b106fa4d791242ec28c355eceaeee2fbb8acfc1b6f2

<----------------- 0.2 is current ------------------>
info: build and push task task/clair-scan/0.2
info: use existing quay.io/eskultety/tekton-catalog/test:clair-scan-0.2@sha256:dc5f4414e61b1602dc448020c97a4e0c47a9c25a17129f353ce1e60ed176e9f6
info: inject task bundle to pipelines quay.io/eskultety/tekton-catalog/test:clair-scan-0.2@sha256:dc5f4414e61b1602dc448020c97a4e0c47a9c25a17129f353ce1e60ed176e9f6
```

However, I had to introduce the following to the build script:
```diff
diff --git a/hack/build-and-push.sh b/hack/build-and-push.sh
index a49a7dc7..97c67806 100755
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -522,7 +522,7 @@ build_push_tasks() {
     local build_new_bundle
     local regex
 
-    find task/*/* -maxdepth 0 -type d | awk -F '/' '{ print $0, $2, $3 }' | \
+    find -L task/*/* -maxdepth 0 -type d | awk -F '/' '{ print $0, $2, $3 }' | \
     while read -r task_dir task_name task_version
     do
         if [ -n "$TEST_TASKS" ] && echo "$TEST_TASKS" | grep -qv "$task_name" 2>/dev/null; then
```

attaching my quay.io test repo for completeness (will remove it once this PR is either accepted or closed as rejected):
https://quay.io/repository/eskultety/tekton-catalog/test

eskultety:
Since v2:
- squashed the `.tekton` changes to the 1st commit where the change belongs (since CI passed for the previous revision)
- updated `hack/build-and-push.sh` such that `find` follows symlinks when looking for tasks.

chmeliik (commented):
LGTM, there just seems to be an extraneous testing commit included

chmeliik (approved):

eskultety:
Can someone translate the e2e test error for me please?

chmeliik (dismissed):

chmeliik (dismissed):

cgwalters (dismissed):
Note the rpm-ostree task is gone so I am no longer a code owner here.

(Also really I think an admin should just force this one in instead of having people continually reapprove)

eskultety:
> (Also really I think an admin should just force this one in instead of having people continually reapprove)

Yeah, policies need some love I guess...

eskultety:
Ping. So, how many approvals does this PR need, realistically?

chmeliik:
> Ping. So, how many approvals does this PR need, realistically?

Let's just have at least someone from @konflux-ci/integration-service-maintainers take a look. They own most of the to-be-archived tasks

jsztuka:
I'm for this idea of archiving the older versions of the tasks and having just one maintained task present there.

dirgim (dismissed):
Agree with the approach

jsztuka (dismissed):
lgtm

PR #2349, opened 2025-05-26
Revert "feat(ISV-5868): Use Mobster to generate index image sboms"

chmeliik:
This reverts commit f2284b41234eafd9f1152a9030e51c816bf002ac.

Mobster seems to generate SBOMs based the SPDX spec version 2.3.1. In particular, Mobster uses referenceCategory=PACKAGE_MANAGER, which has become valid in 2.3.1 [1] but was not yet valid in 2.3 [2].

Conforma checks SBOMs against version 2.3 of the spec.

Revert back to the custom script solution until we can fix this either on the Mobster side or the Conforma side.

[1]: https://github.com/spdx/spdx-spec/blob/support/2.3.1/schemas/spdx-schema.json#L332
[2]: https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json#L325

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -192,7 +192,7 @@ spec:
         add:
           - SETFCAP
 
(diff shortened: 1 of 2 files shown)


mmorhun:
cc @Allda

PR #2355, opened 2025-05-27
feat(STONEBLD-3484): Ensure {task}-oci-ta migrations match {task}

kyngs:
If the regular task has a migration script and the oci-ta version doesn't, that's a mistake. This PR adds a check in hack/validate-migration.sh to check whether this is the case.

### DIFF
--- a/hack/validate-migration.sh
+++ b/hack/validate-migration.sh
@@ -371,6 +371,36 @@ check_apply_in_real_cluster() {
     rm "$apply_logfile"
 }
 
+# Check that the OCI-TA variant of this task has the same set of migration scripts
+# Arguments: the normal-task migration file path
+check_oci_ta_migration() {
+    local migration_file=$1
+    # e.g. migration_file=task/foo/0.2/migrations/0.2.1.sh
+    local parent_dir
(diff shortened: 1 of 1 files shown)


hack/validate-migration.sh
```diff
@@ -371,6 +371,31 @@ check_apply_in_real_cluster() {
...
+    local task_dir="${parent_dir%/*}"                               # => task/foo
+    local task_name
+    task_name=$(basename "$task_dir")                               # => foo
+    local version
+    version=$(basename "$parent_dir")                               # => 0.2
+
+    # path to the OCI-TA variant migrations:
+    local oci_mig_dir="${task_dir}-oci-ta/${version}/migrations"
+
+    if [ ! -d "$oci_mig_dir" ] || [ -z "$(ls -A "$oci_mig_dir")" ]; then
```
chmeliik, line 392: This checks that `{task}-oci-ta` has at least 1 migration script, which is a good start. But IMO we should ensure that both tasks have the same set of migration scripts.

Using the example from the comments: `migration_file=task/foo/0.2/migrations/0.2.1.sh`
IMO the check should validate that this file also exists: `task/foo-oci-ta/0.2/migrations/0.2.1.sh`

kyngs: You're right; resolved.

PR #2356, opened 2025-05-27
partners/task/cosa-build: initialize

jcapiitao:
Initializing the Tekton task that will build CoreOS image with coreos-assembler [1].

[1] https://github.com/coreos/coreos-assembler/


jcapiitao:
Maybe this can be moved to `task/`, let me know where is the best place.

PR #2357, opened 2025-05-27
do not merge

lmilbaum:


cgwalters:
The change makes sense to me; that said there's something important to bear in mind here is that this type of thing everyone trying to replicate with `podman build` locally unless they apply the same change.

This isn't the *only* instance of skew between konflux and local, and I'm definitely in favor of this one.

task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
```diff
@@ -400,6 +400,16 @@ spec:
       chown root:root /var/lib/containers
 
       sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
+      cat << 'EOF' >> /etc/containers/registries.conf
+      [[registry]]
+      prefix   = "registry.redhat.io"
+      location = "registry.redhat.io"
+
+        [[registry.mirror]]
+        location = "registry.stage.redhat.io"
```
MartinBasti, line 409: this is not a mirror of registry.redhat.com

task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
```diff
@@ -400,6 +400,16 @@ spec:
       chown root:root /var/lib/containers
 
       sed -i 's/^\s*short-name-mode\s*=\s*.*/short-name-mode = "disabled"/' /etc/containers/registries.conf
+      cat << 'EOF' >> /etc/containers/registries.conf
+      [[registry]]
+      prefix   = "registry.redhat.io"
```
MartinBasti, line 405: Why do we need this? It's recommended to use FQDN in Containerfiles.

Also the idea of hardcoding RH specific paywall registry in upstream project, isn't something, I'm in favor of.

PR #2392, opened 2025-06-06
Pass image digest to apply-tags task

mkosiarc:
See commit messages for details. If you want to test the migration script locally you would need to change/update ".spec.tasks[]" for ".spec.pipelineSpec.tasks[]" in the 0.2.sh.

Then, if you run it on a example pipelineRun that contains the current version of apply-tags, looking like:
```
  - name: apply-tags
    params:
      - name: ADDITIONAL_TAGS
        value: ["tag1", "tag2"]
      - name: IMAGE
        value: $(tasks.build-image-index.results.IMAGE_URL)
```

It will update it to:
```
  - name: apply-tags
    params:
      - name: ADDITIONAL_TAGS
        value: ["tag1", "tag2"]
      - name: IMAGE_URL
        value: $(tasks.build-image-index.results.IMAGE_URL)
      - name: IMAGE_DIGEST
        value: $(tasks.build-image-index.results.IMAGE_DIGEST)
```


task/apply-tags/0.2/MIGRATION.md
```diff
@@ -0,0 +1,3 @@
+# Migration from 0.1 to 0.2
+
+- The parameter `IMAGE_DIGEST` is added and the parameter `IMAGE` is renamed to `IMAGE_URL`.These changes will be added to build pipeline definition file automatically by script migrations/0.2.sh when MintMaker runs [pipeline-migration-tool](https://github.com/konflux-ci/pipeline-migration-tool).
```
chmeliik, line 3: Can we also describe what changes the user would need to do in case the migration fails to apply for any reason?

task/apply-tags/0.2/migrations/0.2.sh
```diff
@@ -0,0 +1,47 @@
...
+# Determine the correct image-digest and image-url values based on task presence
+if yq -e '.spec.tasks[] | select(.name == "build-oci-artifact")' "$pipeline_file" >/dev/null; then
+    image_digest_value="\$(tasks.build-oci-artifact.results.IMAGE_DIGEST)"
+    image_url_value="\$(tasks.build-oci-artifact.results.IMAGE_URL)"
+elif yq -e '.spec.tasks[] | select(.name == "build-image-index")' "$pipeline_file" >/dev/null; then
+    image_digest_value="\$(tasks.build-image-index.results.IMAGE_DIGEST)"
+    image_url_value="\$(tasks.build-image-index.results.IMAGE_URL)"
+else
+    echo "Neither build-oci-artifact nor build-image-index tasks found."
+    exit 0
```
chmeliik, line 19: One reason why the migration could fail to apply would be if users created their Components a long time ago, before even the single-arch pipeline started using the build-image-index task.

We can add handling for that (handle the `build-container` as fallback) or just document the necessary changes in the MIGRATION.md

task/apply-tags/0.2/migrations/0.2.sh
```diff
@@ -0,0 +1,47 @@
...
+# Check if IMAGE_DIGEST parameter already exists using precise path
+if ! yq -e '.spec.tasks[] | select(.name == "apply-tags").params[] | select(.name == "IMAGE_DIGEST")' "$pipeline_file" >/dev/null; then
+    echo "Adding image-digest parameter to apply-tags task"
+    yq -i "(.spec.tasks[] | select(.name == \"apply-tags\")).params += [{\"name\": \"IMAGE_DIGEST\", \"value\": \"$image_digest_value\"}]" "$pipeline_file"
+else
+    echo "IMAGE_DIGEST parameter already exists in apply-tags task. No changes needed."
+fi
+
+
+
```
chmeliik, line 47: nitpick: empty lines

chmeliik:
> This is the initial commit. Its only purpose is to
make the job easier for reviewers to verify only the actual changes
in subsequent commits.

:heart:

task/apply-tags/0.2/migrations/0.2.sh
mkosiarc: Added build-container task as well to the script + updated the MIGRATION.md as requested in different thread

task/apply-tags/0.2/migrations/0.2.sh
mkosiarc: fixed

simonbaird (approved):
Lgtm.

PR #2310, opened 2025-05-12
sast-coverity-check: fix IMAGE param description

chmeliik:
Users were confused about the IMAGE parameter of the sast-coverity-check task. Explain that the parameter is mostly irrelevant.

PR #2397, opened 2025-06-06
Fix for pipeline maven-zip-build: Add missing param of IMAGE for sast-coverity-check

ligangty:
According to the [PR-2310](https://github.com/konflux-ci/build-definitions/pull/2310)


chmeliik:
> @konflux-ci/build-maintainers hi. one question for the failed check. Seems it requires me to run the ./hack/build-manifests.sh, but when I ran it, it will remove the fix I gave in this PR. How should I resolve this issue?

Pipelines are generated using kustomize, so this will need a fix either in the template file that's used as the base for the maven pipeline, or in the patch for the maven pipeline

ligangty:
> > @konflux-ci/build-maintainers hi. one question for the failed check. Seems it requires me to run the ./hack/build-manifests.sh, but when I ran it, it will remove the fix I gave in this PR. How should I resolve this issue?
> 
> Pipelines are generated using kustomize, so this will need a fix either in the template file that's used as the base for the maven pipeline, or in the patch for the maven pipeline

@chmeliik Oh forgot this. Just added the patch change. Thanks for the help!

PR #2421, opened 2025-06-12
docker-build: add dev-package-manager globally

jcapiitao:
We add the `dev-package-manager` parameter as 'global' parameter in order to be able to customize the `docker-build` (and `-multi-platform`) pipeline when pulling it with `pipelineRef` and bundle.

This change should not break backward compatibility as it's still setting an env var with same default value.

### DIFF
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -12,6 +12,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |build-image-index| Add built image into an OCI image index| true| build-image-index:0.1:ALWAYS_BUILD_INDEX|
 |build-platforms| List of platforms to build the container images on. The available set of values is determined by the configuration of the multi-platform-controller.| ['linux/x86_64']| |
 |build-source-image| Build a source image.| false| |
+|dev-package-managers| Enable in-development package managers| false| prefetch-dependencies:0.2:dev-package-managers|
 |dockerfile| Path to the Dockerfile inside the context specified by parameter path-context| Dockerfile| build-images:0.4:DOCKERFILE ; sast-coverity-check:0.3:DOCKERFILE ; push-dockerfile:0.1:DOCKERFILE|
 |git-url| Source Repository URL| None| clone-repository:0.1:url|
 |hermetic| Execute the build with network isolation| false| build-images:0.4:HERMETIC ; sast-coverity-check:0.3:HERMETIC|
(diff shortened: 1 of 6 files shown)


jcapiitao:
> @jcapiitao the pipeline files are autogenerated, If I am not mistaken you need to add the changes in the `patch.yaml` file

Thank you @mmartinv for the heads-up here, I amended my commit accordingly :+1:

ralphbean:
This makes sense to me. It's quite the hoop you have to jump through to get rpm lockfiles enables in prefetch and it's been a year now that we've considered that tech preview without a clear sign that the dnf manifest plugin will be done soon (the expected successor to the rpm-lockfile-prototype). Perhaps we should re-characterize the rpm-lockfile-prototype feature in hermeto as _not_ a dev package manager anymore?

ralphbean:
.. then again, presumably there will be other dev package managers in the future and having a pipeline-level param to switch those on and off sounds generally useful.

jcapiitao:
I guess the E2E test failure run in Konflux is not related to this change:

`timed out when waiting for image-controller annotations to be updated on component test-comp-ooqc in namespace build-templates-e2e. component`

jcapiitao:
> > .. then again, presumably there will be other dev package managers in the future and having a pipeline-level param to switch those on and off sounds generally useful.
> 
> In Hermeto terms, dev package managers are those that are still in development and not fully supported yet.
> 
> The other day, we had a user reporting that Hermeto's handling of Cargo is completely broken. After too much time spent debugging, we realized the user had simply been using an outdated version of Hermeto that didn't support Cargo yet. The user had the dev-package-managers flag enabled which made it look like Cargo was already supported.
> 
> TL;DR, I'm not in favor of blessing this flag by making it a pipeline-level param.

I see the point being made here, but leveling up this flag as pipeline param, and keeping it disable by default, would just make it more visible but won't change the current behavior. I also do understand that this might increase the probability of issues, but if the end-user want to enable hermertic builds with e.g rpms, cargo, it will enable the `dev-package-managers` no matter the level (i.e `task` or `pipeline` level), leading the to same potential issues.

For CoreOS projects we want to build hermetically and are pulling the pipeline definition from bundle with `pipelineRef` i.e https://github.com/coreos/coreos-assembler/pull/4140 to simplify config.
Not merging this PR will force us to fork and own our pipeline, which will kind of impede the 'upstream first' model workflow.

jcapiitao:
> > > it's an equivalent of using Beta functionality in production - big NO-NO! The flag should have never been seen in production in any way and the sole reason it is being done so is RPM support (that's a topic for another day).
> > 
> > 
> > What I get from that is that konflux is not production ready for any build that needs to pull RPMs ?
> 
> Not quite. That's the
> 
> > that's a topic for another day
> 
> part of my response which I don't think is appropriate to explore and discuss here. Long story short it is about the current implementation of the solution that is used, its overall maintainability, extensibility and where the solution was introduced. Based on this we may need to start thinking of how `dev-package-managers` would not be used for the RPM functionality anymore and have it just as a "best effort" type of solution that we can get rid of at any time once the proper one done by the DNF team matures.

I see, then I guess we'll have to wait for https://github.com/rpm-software-management/libpkgmanifest + the DNF plugin to be ready in order to enable hermetic build.
Also, maybe we should update the https://github.com/konflux-ci/rpm-lockfile-prototype/ README with all your explanations to warn the end-users, wdyt @eskultety @chmeliik ?

jcapiitao:
> Unfortunately, the reality here is that the adoption of the DNF plugin will take time (upstream takes time), in fact it's taking much longer than anyone has anticipated and the RPM build functionality in Konflux is being utilized heavily today, hence my remark on actually dropping the flag from the current implementation and instead just figure out how hermeto + konflux could transition to a better solution in the future. @jcapiitao I went ahead and filed [hermetoproject/hermeto#986](https://github.com/hermetoproject/hermeto/issues/986) on our side; IIUC its completion should NOT be a hard blocker for anything in this PR and in build-definitions in general, correct?

Thank you @eskultety for initiating the discussion there about this :100: 
Well, it would not be a hard blocker if we fork the pipeline (and add `dev-package-managers` at pipeline level) in the meantime until https://github.com/hermetoproject/hermeto/issues/986 is completed. But maintaining our fork is something we can do without too much friction TBH as we're still in development mode.

PR #2427, opened 2025-06-13
Update konflux-test version for FIPS tasks

yashvardhannanavati:


task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
```diff
@@ -217,7 +217,7 @@ spec:
             value: stepactions/fips-operator-check-step-action/0.1/fips-operator-check-step-action.yaml
         resolver: git
     - name: parse-images-processed-result
-      image: quay.io/redhat-appstudio/konflux-test:v1.4.27@sha256:5f3200bb9f52c888be99d3eeacde6445b21b731b9216e20259522eba99f9766e
+      image: quay.io/redhat-appstudio/konflux-test:v1.4.29@sha256:3619ffde751d337d02f1f61c83c5c39eb87d8091dbe5a9af58ea98577fa09461
```
nmars, line 220: In addition to this, do we need to also update the `revision` field for the git resolver to the stepAction on line 215?

yashvardhannanavati: ooh! nice catch. Fixed it

PR #2438, opened 2025-06-16
fix: attach apply-tags 0.2 migration

tkdchen:
Something is wrong with the build-and-push.sh that makes the migration not attached previously. This PR cherry-picks the migration to a new task bundle so that users are able to get it with a newer bundle update.


mkosiarc:
So the only change here is that 0.2 was changed to 0.2.1 without any code changes? Is that actually something we want?

mkosiarc:
Please add a commit message as well, so in the future it is clear that no changes were really made and the purpose of why we needed to do this.

tkdchen:
@mkosiarc commit is updated based on my previous comment.

PR #2448, opened 2025-06-20
feat(CLOUDDST-27673): improve preflight check performance using matrix

zxiong:
Introduce a matrix configuration to run preflight checks in parallel for each architecture, reducing overall execution time.

Performance tests comparing the original task and the new one with the PR. 
 |   | original (baseline) | Matrix | Improved
-- | -- | -- | --
Task duration | 0:10:43 | 0:03:07| ~71%

Note: The image used in the test is a multi-arch image with 4 architectures.


pipelines/docker-build-multi-platform-oci-ta/patch.yaml
```diff
@@ -54,6 +54,14 @@
   value:
   - build-images
 
+# buildah-remote-oci-ta
```
arewm, line 57: Is this accurate? Should this actually be 
```suggestion
# ecosystem-cert-preflight-checks
```
?

pipelines/docker-build-oci-ta/README.md
```diff
@@ -119,6 +119,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |ca-trust-config-map-key| The name of the key in the ConfigMap that contains the CA bundle data.| ca-bundle.crt| |
 |ca-trust-config-map-name| The name of the ConfigMap to read CA bundle data from.| trusted-ca| |
 |image-url| Image url to scan.| None| '$(tasks.build-image-index.results.IMAGE_URL)'|
+|platform| The platform the image is built on.| | |
```
arewm, line 122: You are adding this parameter here in the README but you are not adding this to the pipeline, correct?

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -22,6 +22,10 @@ spec:
       type: string
       description: The type of artifact. Select from application, operatorbundle, or introspect.
       default: "introspect"
+    - name: platform
```
arewm, line 25: @mshaposhnik , can you help review this from the multi-platform controller perspective? I don't think that we would be able to use the `build-platforms` parameter directly as this isn't guaranteed to be _just_ an architecture specification.

Does this parameter also work with multi-platform controller (I don't think that we want VMs provisioned here... the analysis can all run within the cluster pods).

arewm (commented):
@mshaposhnik , can you confirm whether this would conflict with the multi-platform logic? I think we want to use the platforms

pipelines/docker-build-oci-ta/README.md
zxiong: Yes, that’s right. I believe there’s no need to set it in the pipeline, since the image being checked is expected to be a single-architecture image—we don’t need to run it with a matrix.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -213,6 +216,29 @@ spec:
...
+
+        if [ -n "$platform" ]; then
+          arch="${platform#*/}"
+          if [ "$arch" = "x86_64" ]; then
+            arch="amd64"
+          fi
+          /usr/local/bin/preflight check container "$image_url" --platform "$arch"
+        else
+          /usr/local/bin/preflight check container "$image_url"
+        fi
```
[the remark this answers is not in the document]
acornett21: preflights `--platform` flag maps to the manifest platform of an OCI image.

```bash
--platform string                     Architecture of image to pull. Defaults to runtime platform. (default "amd64")
```

```json
❯ crane manifest quay.io/opdev/preflight:stable | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 595,
      "digest": "sha256:99311338e7cae27a822afb23f50d61fc2c93691255b1b9aada93d691258027c6",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 595,
      "digest": "sha256:afd3dfc12f0d7d17f021157967eebb509eb859e414610b77aeafc65e1529c989",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 595,
      "digest": "sha256:59905cc5cd76ee1cb6ed50c8ae06d9e7e208ce355e7da97ea5bd84f887db5d7a",
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 595,
      "digest": "sha256:c24f5edc451f2c7e09b7609d566a2d6c6bde471ead90f23c345a116f8671c576",
      "platform": {
        "architecture": "s390x",
        "os": "linux"
      }
    }
  ]
}
```

ie: amd64, ppc64le, arm64, s390x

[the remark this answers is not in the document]
arewm: Right, that's what I would have expected. The task would need to do a little more pre-computing to figure out what the proper architecture is to pass. I think it could be as simple as searching for those strings in the platform... defaulting back to x86 if nothing matches. This would also be beneficial for returning the IMAGES_PROCESSED result as you have to return the digest of the architectures that you processed (one of the tasks would also need to return the digest for the image index).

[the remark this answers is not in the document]
zxiong: @acornett21, I extracted the digest for each processed architecture from /artifacts/$arch/cert-image.json (e.g., jq '.docker_image_digest' /artifacts/amd64/cert-image.json) to populate `IMAGES_PROCESSED`. Does it seem okay to you?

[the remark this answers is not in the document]
acornett21: `docker_image_digest` in the `cert-image.json` is the digest of the tag for the image, so this seems fine to take from there. If this is being stored how konflux expects, I can't speak to that.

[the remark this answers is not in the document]
zxiong: Thanks for confirming. `IMAGES_PROCESSED` stores the digests of the specific architecture images that were processed. Based on what I’ve observed, this aligns with Konflux’s expected format.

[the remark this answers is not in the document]
zxiong: @arewm could you review it again?

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -213,6 +219,40 @@ spec:
...
+          fi
+
+          # Validate against supported arch list. If it's not a known arch, fallback to amd64
+          case "$arch" in
+            amd64|ppc64le|arm64|s390x)
+              ;;
+            *)
+              arch="amd64"
+              ;;
+          esac
```
arewm, line 250: @mshaposhnik , do you expect that this pattern will be followed? Should we document the assumption of this pattern somewhere in the future to ensure that we keep track of the dependency?

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -213,6 +219,40 @@ spec:
...
+
+          /usr/local/bin/preflight check container "$image_url" --platform "$arch"
```
arewm, line 252: Do we need to handle the case where `$arch` isn't set in case there is a malformed platform? Maybe we should return an error result if we cannot parse the platform properly?

zxiong: Currently, we silently fall back to amd64 for unknown or malformed platforms.
Do you suggest that we should return an error result instead of falling back in those cases?

arewm: If you want to fall back, then I wouldn't do it silently. This would prevent issues where users are missing a digest from being processed unknowingly

zxiong: Updated. Instead of  falling back to `amd64` silently, an error result is returned for malformed or unsupported architecture.

acornett21 (approved):

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -24,6 +24,10 @@ spec:
       type: string
       description: The type of artifact. Select from application, operatorbundle, or introspect.
       default: "introspect"
+    - name: platform
+      type: string
```
enkeefe00, line 28: Since this parameters is directly taking in the pipelines parameter `build_platforms`, shouldn't this be an array since that's what `build_platforms` is defined as?

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Check if the task already has a matrix
+if yq -e '.spec.tasks[] | select(.name == "'"ecosystem-cert-preflight-checks"'") | has("matrix")' "$pipeline_file" >/dev/null 2>&1; then
+  echo "Matrix already exists for task 'ecosystem-cert-preflight-checks'. No changes made."
+else
+  echo "Adding matrix for task 'ecosystem-cert-preflight-checks'..."
+  yq -i "
+  (.spec.tasks[] 
+    | select(.name == \"ecosystem-cert-preflight-checks\" and .matrix == null)
+  ).matrix = {
```
enkeefe00, line 30: See above comment about treating this as an array.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
```diff
@@ -225,6 +231,41 @@ spec:
...
+        #!/usr/bin/env bash
+
+        set -o errexit
+        set -o nounset
+        set -o pipefail
+
+        image_url="${PARAM_IMAGE_URL}"
+        platform="${PARAM_PLATFORM}"
+
+        if [ -n "$platform" ]; then
```
enkeefe00, line 248: See above comment about treating this as an array.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
acornett21: No, this should be a string, since the `matrix` value in the pipeline will fan this out automatically and the task will only see one value from the array since the pipeline will create N of this tasks.

Matrix docs [here](https://tekton.dev/docs/pipelines/matrix/)

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
zxiong: As @acornett21 noted in the comment above, it's a string rather than an array.

task/ecosystem-cert-preflight-checks/0.2/ecosystem-cert-preflight-checks.yaml
zxiong: As @acornett21 noted in the comment above, it's a string rather than an array.

zxiong:
> I have no issues with these changes, but the failing checks need to be resolved

Sure. I’ve just rebased the PR — all checks have passed except for the CI (tried for 3 times), which is still waiting on a fix from konfluxbld-green.

arewm (approved):
Thanks!

arewm:
@konflux-ci/build-maintainers , I think we need an approval from your team for this PR?

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Check if the pipeline has 'build-platforms' parameter
+if ! yq -e '.spec.params[] | select(.name == "build-platforms")' "$pipeline_file" >/dev/null 2>&1; then
+  echo "Matrix will not be added because the dependent parameter 'build-platforms' is not defined in the pipeline."
+  exit 0
+fi
+
+# Check if the task exists
+if ! yq -e '.spec.tasks[] | select(.name == "'"ecosystem-cert-preflight-checks"'")' "$pipeline_file" >/dev/null 2>&1; then
+  echo "Task '$TASK_NAME' does not exist in the pipeline."
```
tkdchen, line 18: `TASK_NAME` is not defined.

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,41 @@
...
+declare -r pipeline_file=${1:?missing pipeline file}
+
+# Check if the pipeline has 'build-platforms' parameter
+if ! yq -e '.spec.params[] | select(.name == "build-platforms")' "$pipeline_file" >/dev/null 2>&1; then
+  echo "Matrix will not be added because the dependent parameter 'build-platforms' is not defined in the pipeline."
+  exit 0
+fi
+
+# Check if the task exists
+if ! yq -e '.spec.tasks[] | select(.name == "'"ecosystem-cert-preflight-checks"'")' "$pipeline_file" >/dev/null 2>&1; then
```
tkdchen, line 17:
```suggestion
if ! yq -e '.spec.tasks[] | select(.name == "ecosystem-cert-preflight-checks")' "$pipeline_file" >/dev/null 2>&1; then
```

Suggest to make this change to simplify it.

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "Task '$TASK_NAME' does not exist in the pipeline."
+  exit 0
+fi
+
+# Check if the task already has a matrix
+if yq -e '.spec.tasks[] | select(.name == "'"ecosystem-cert-preflight-checks"'") | has("matrix")' "$pipeline_file" >/dev/null 2>&1; then
```
tkdchen, line 23:
```suggestion
if yq -e '.spec.tasks[] | select(.name == "ecosystem-cert-preflight-checks") | has("matrix")' "$pipeline_file" >/dev/null 2>&1; then
```

Likewise.

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
zxiong: Fixed

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
zxiong: Accepted and updated.

task/ecosystem-cert-preflight-checks/0.2/migrations/0.2.1.sh
zxiong: Accepted and updated.

tkdchen (approved):
The migration LGTM.

PR #2449, opened 2025-06-20
buildah-remote-oci-ta: pass a HERMETIC build-arg

jbtrystram:
Being aware in the Container build that the hermetic mode is active is useful in some scenarios when we try to use already existing resources that point to remotes.

For ex in rpm-ostree the repo files are defined alongside the manifests and rpm-ostree use those to compose the rootfs. Having the HERMETIC value set here allow to fetch the repo files made available by hermeto under /etc/yum.repos.d instead.

### DIFF
--- a/task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
+++ b/task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
@@ -468,6 +468,10 @@ spec:
         BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
       done
 
+      if [ "${HERMETIC}" == "true" ]; then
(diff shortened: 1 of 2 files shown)


task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml
```diff
@@ -468,6 +468,10 @@ spec:
         BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
       done
 
+      if [ "${HERMETIC}" == "true" ]; then
+        BUILD_ARG_FLAGS+=("--build-arg=HERMETIC=true")
```
jcapiitao, line 472: I'd put it inside the exiting `if [ "${HERMETIC}" == "true" ]` block a few lines after L.481 for consistency.

PR #2453, opened 2025-06-23
sast-shellcheck: add 0.2 task with TARGET_DIR param

sfowl:
[PSSECAUT-1148](https://issues.redhat.com/browse/PSSECAUT-1148)

Though I believe this should be a backwards compatible change, the README guidance seems to say that any param change requires a new task.

The diff between task versions is rather small:

```
$ diff task/sast-shell-check/0.{1,2}/sast-shell-check.yaml
5c5
<     app.kubernetes.io/version: "0.1"
---
>     app.kubernetes.io/version: "0.2"
48a49,52
>     - name: TARGET_DIR
>       type: string
>       description: Target directory in component's source code
>       default: "."
121a126
>         TARGET_DIR=${SOURCE_CODE_DIR}/${TARGET_DIR}
124c129
<         /usr/share/csmock/scripts/run-shellcheck.sh "$SOURCE_CODE_DIR"
---
>         /usr/share/csmock/scripts/run-shellcheck.sh "${TARGET_DIR}"
```

I intend to make a similar change to all sast tasks, but I have limited this PR to just the shellcheck task to keep the review size manageable.

PS: There are also some minor 0.2 README grammar/formatting tweaks included


task/sast-shell-check-oci-ta/0.2/MIGRATION.md
```diff
@@ -0,0 +1,4 @@
+# Migration from 0.1 to 0.2
+
+* A new, optional parameter `TARGET_DIR` has been added, which defaults to `"."`. This can be set to the name of a valid
```
kdudka, line 3: If the default behavior is compatible with the previous version of the task, do we need to create a new version of the task and bother end users with the migration?

task/sast-shell-check-oci-ta/0.2/sast-shell-check-oci-ta.yaml
```diff
@@ -0,0 +1,286 @@
...
+        If `true`, the excluded findings will be stored in `excluded-findings.json`.
+      type: string
+      default: "false"
+    - name: SOURCE_ARTIFACT
+      description: The Trusted Artifact URI pointing to the artifact with
+        the application source code.
+      type: string
+    - name: TARGET_DIR
+      description: Target directory in component's source code
+      type: string
```
kdudka, line 50: Will a single directory be sufficient for all projects?  I think it would be more flexible to accept a list of directories.  `run-shellcheck.sh` accepts multiple arguments already.

task/sast-shell-check-oci-ta/0.2/MIGRATION.md
sfowl: D'oh! I should have read the readme closer, at the ends of the Versioning sections it says:

> Adding a new parameter with a default value does not require a task version increase.

Fixed now.

PR #2462, opened 2025-06-24
fix: update deprecated image to use new konflux-test version

sonam1412:
This is to fix the issue with deprecated-image-check failing for the below error:
`"Unauthorized: access to the requested resource is not authorized" in latest deprecated-base-image-check`


dirgim (approved):
LGTM

PR #2463, opened 2025-06-25
sast-unicode: add 0.3 with required image-digest

sfowl:
[PSSECAUT-1203](https://issues.redhat.com/browse/PSSECAUT-1203)

It seems as though in https://github.com/konflux-ci/build-definitions/pull/2018 the `image-digest` param was not made required for sast-unicode-check and had a default value of `""`. This caused upload to fail for pipelines that did not have image-digest set for this task.

There is also a small bugfix due to incorrect var name `UPLOAD_FILE` vs `UPLOAD_FILES`.

To fix I've creaated a new version, 0.3, and migration scripts. Also related to https://issues.redhat.com/browse/STONEINTG-1208

### DIFF
--- a/task/sast-unicode-check-oci-ta/0.3/migrations/0.3.sh
+++ b/task/sast-unicode-check-oci-ta/0.3/migrations/0.3.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+tasks_root=".spec.tasks[]"
+if yq -e ".spec.pipelineSpec.tasks" "$pipeline_file" &>/dev/null; then
+  tasks_root=".spec.pipelineSpec.tasks[]"
+fi
+
+# check if image-url and image-digest is missing for sast-unicode-check-oci-ta task and early exit
+if yq -e "$tasks_root"' | select(.name == "sast-unicode-check-oci-ta").params[] | has(.name == "image-url") && has(.name == "image-digest"' "$pipeline_file" &>/dev/null; then
+  echo "image-url and image-digest already set for sast-unicode-check-oci-ta"
+  exit 0
+fi
+
+# get image-url and image-digest value from other tasks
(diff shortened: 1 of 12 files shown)


task/sast-unicode-check-oci-ta/0.3/migrations/0.3.sh
```diff
@@ -0,0 +1,38 @@
...
+elif yq -e "$tasks_root"' | select(.name == "build-image-index")' "$pipeline_file" &>/dev/null; then
+    image_url_value="\$(tasks.build-oci-artifact.results.IMAGE_URL)"
+    image_digest_value="\$(tasks.build-image-index.results.IMAGE_DIGEST)"
+else
+    echo "Neither build-oci-artifact nor build-image-index tasks found. Can't get image-url and image-digest."
+    exit 0
+fi
+
+# add params to sast-unicode-check-oci-ta
+if ! yq -e "$tasks_root"' | select(.name == "sast-unicode-check-oci-ta").params[] | select(.name == "image-url")' "$pipeline_file" &>/dev/null; then
```
hongweiliu17, line 31: [Here](https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check-oci-ta/0.3/migrations/0.3.sh) is the migration script, we may also check task `-oci-ta` and non `oci-ta` task

kdudka (approved):

task/sast-unicode-check-oci-ta/0.3/migrations/0.3.sh
sfowl: Discussed offline. To better matching the approach used for similar migrations, I've combined the separate migration scripts for -oci-ta and non-oci-ta tasks into single script, and removed the oci-ta specific script.

sfowl: Hmm, pipeline wants a script for both variants, so I've made a copy.

hongweiliu17 (approved):
lgtm

PR #2477, opened 2025-06-26
feat: Add slack-webhook-notification-oci-ta task

pastequo:
Adding the equivalent of https://github.com/konflux-ci/build-definitions/tree/main/task/slack-webhook-notification/0.1 for trusted artifact pipelines

This allows users to send a slack notification when a merge build pipeline fails, or when a new artifact is available for example.

T̶h̶i̶s̶ ̶t̶a̶s̶k̶ ̶w̶a̶s̶ ̶c̶r̶e̶a̶t̶e̶d̶ ̶m̶a̶n̶u̶a̶l̶l̶y̶ ̶b̶y̶ ̶a̶d̶a̶p̶t̶i̶n̶g̶ ̶t̶h̶e̶ ̶n̶o̶n̶-̶o̶c̶i̶-̶t̶a̶ ̶e̶q̶u̶i̶v̶a̶l̶e̶n̶t̶.̶

This task was genered with a recipe from the non-oci-ta equivalent.

I had no idea who should be the CODEOWNER so I mimic-ed slack-webhook-notification

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -26,6 +26,7 @@
 /task/show-sbom                             @konflux-ci/build-maintainers
 /task/show-sbom-rhdh                        @konflux-ci/build-maintainers
 /task/slack-webhook-notification            @konflux-ci/build-maintainers
+/task/slack-webhook-notification-oci-ta     @konflux-ci/build-maintainers
 /task/source-build                          @konflux-ci/build-maintainers
 /task/source-build-oci-ta                   @konflux-ci/build-maintainers
 /task/summary                               @konflux-ci/build-maintainers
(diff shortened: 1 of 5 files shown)


tnevrlka:
Thank you for the contribution!

Could you please describe in the commit message and the PR description why is this change being made - what problem is this solving?

Furthermore, how was this created? Manually or via a script?

pastequo:
Done.

tnevrlka:
@konflux-ci/build-maintainers
This is a new version of a task that the Build team will have to maintain. It's been created manually from the non-oci-ta version. Does anybody have something against merging this? As far as I know, it shouldn't have to change often, so the maintenance burden should be pretty low

pastequo:
> oci-ta tasks should be created automatically using a recipe.yaml - see other oci-ta tasks in this repo

Done

PR #2481, opened 2025-06-27
define the TASKRUN_NAME variable in tasks that reference it

mikebonnet:
Several tasks contained scripts that reference `$TASKRUN_NAME` but don't define it. This resulted in confusing log messages about pushing images with missing tags.

Define `TASKRUN_NAME` from `$(context.taskRun.name)` in all the tasks that reference it. Standardize on using `$TASKRUN_NAME` in those scripts rather than referencing value via context directly, in accordance with Tekton scripting best-practices.


chmeliik (commented):
Needs a rebase, otherwise LGTM :+1:

mikebonnet:
Rebased.

mmorhun (approved):

cmoulliard (approved):

PR #2491, opened 2025-07-01
Update fips stepaction to use check-payload 0.3.8

nmars:
Updated fips stepaction to latest konflux-test image containing check-payload 0.3.8.


dirgim (approved):
LGTM

PR #2347, opened 2025-05-22
[WIP] Mitigate quay.io instabilities by adding retries to commands using registry

rcerven:
STONEBLD-3488

# Before you complete this pull request ...

Look for any open pull requests in the repository with the title "e2e-tests update" and 
see if there are recent e2e-tests updates that will be applicable to your change.

PR #2497, opened 2025-07-02
Deprecate some tasks

eskultety:
~~This PR deprecates some tasks that are no longer in use.~~

_Edit: Deprecates tasks that have been replaced by a single `validate-fbc` task rendering these redundant._


eskultety:
@arewm since you have the most context and I'm opening this based on our private discussion I'd really like to add some WHY to the commits wrt/ this deprecation, can you please share the backstory here why these are no longer needed and can be deprecated as a whole other than trying to unblock https://github.com/konflux-ci/build-definitions/pull/2347?

rcerven:
sast-coverity-check needs also fix, because 0.3 version (in task) is using in kustomization 0.2 version (in archived)

eskultety:
> sast-coverity-check needs also fix, because 0.3 version (in task) is using in kustomization 0.2 version (in archived)

@rcerven that doesn't sound right: https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check/0.3/kustomization.yaml

When I opened this PR I ran `git grep -C30 resources task/**/kustomization.yaml` to double check any other suspicious relative path references.

rcerven:
> > sast-coverity-check needs also fix, because 0.3 version (in task) is using in kustomization 0.2 version (in archived)
> 
> @rcerven that doesn't sound right: https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check/0.3/kustomization.yaml
> 
> When I opened this PR I ran `git grep -C30 resources task/**/kustomization.yaml` to double check any other suspicious relative path references.

actually problem is with version 0.2 (in archived) which references 0.4 buildah (in task)

eskultety:
> > > sast-coverity-check needs also fix, because 0.3 version (in task) is using in kustomization 0.2 version (in archived)
> > 
> > 
> > @rcerven that doesn't sound right: https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check/0.3/kustomization.yaml
> > When I opened this PR I ran `git grep -C30 resources task/**/kustomization.yaml` to double check any other suspicious relative path references.
> 
> actually problem is with version 0.2 (in archived) which references 0.4 buildah (in task)

That's not a problem, nothing in archived should be used anywhere so broken references should ideally not matter at all and should not be a blocker for any work, this or yours. That said, I guess I know where this is coming from - manifest validator which follows the symlinks and causes the CI validator checks to fail. I know about it and that's follow up work, actually, let me open the PR immediately.

arewm:
@eskultety , I missed that comment in the PR, apologies.

> I'd really like to add some WHY to the commits wrt/ this deprecation, can you please share the backstory here why these are no longer needed and can be deprecated

None of these tasks are needed anymore as I wrapped all of the functionality into the `validate-fbc` task in #1534. I wanted to post a migration message to users that had the other tasks since they were becoming deprecated, so I artificially bumped those versions.

eskultety:
Commit messages adjusted. @konflux-ci/build-maintainers, @konflux-ci/integration-service-maintainers  anything else missing or can we proceed with merging this?

PR #2499, opened 2025-07-03
README: Add a brief task deprecation process to follow

eskultety:
Until we adopt a formal task deprecation (and most importantly task version support) notice, let's document the best-effort process of properly deprecating a given task version if a new one becomes available.

**Related to: https://github.com/konflux-ci/build-definitions/pull/2497**


README.md
```diff
@@ -420,3 +420,33 @@ is the workflow:
   review process.
 
 The migration will be applied during next Renovate run scheduled by MintMaker.
+
+## Task Deprecation
+
+Often times when a new version of task is introduced, the old version of the task doesn't need to
```
fmudry, line 426: I believe there should be an article. Otherwise great job.
```suggestion
Often times when a new version of a task is introduced, the old version of the task doesn't need to
```

README.md
```diff
@@ -444,3 +444,33 @@ is the workflow:
...
+  that you have a brand new task definition YAML in place, this is so that the symlink we are about
+  to introduce doesn't confuse various validators and CI checks that are in place
+- move the old task to the `archived-tasks` top-level directory
+
+    ```bash
+    $ mkdir archived-tasks/<task_name> 2>/dev/null
+    $ git mv task/<task_name>/<old_version> archived-tasks/<task_name>
+    ```
+
+- symlink the old task from the main task location for easy historical version tracking
```
chmeliik, line 470: Hmm, now that we've merged https://github.com/konflux-ci/build-definitions/pull/2498, the only purpose of the symlink is historical tracking? The tasks don't get updates anymore?

I thought that the proposal in https://github.com/konflux-ci/build-definitions/pull/2344 was that it was only a file structure change to more easily see which tasks are not latest. My understanding was that the archived tasks would still get built and released on push.

chmeliik (commented):
The current deprecation process relies on Conforma to notify the user that their task is deprecated.

To deprecate a task, one has to:
- set the expire label, e.g. `build.appstudio.redhat.com/expires-on: "2025-03-15T00:00:00Z`
- optionally, set the expiry message
  ```
      build.appstudio.redhat.com/expiry-message: 'Instead of using this task, switch
        to task rpms-signature-scan
        (https://quay.io/repository/konflux-ci/tekton-catalog/task-rpms-signature-scan)'
  ```
- build and relese new version of the task that has these labels

Without this, I don't think Conforma will ever flag the older version of the task as deprecated

eskultety:
@chmeliik I reworded both the README section as well as the commit message. Let me know if I captured your remarks correctly.
Additionally, now that I tried to capture the reality including Conforma, do you think this now satisfies https://github.com/konflux-ci/build-definitions/issues/2509? Now, the latest changes do not mention a global policy on number of supported task versions, instead it leaves this to the task maintainers which is in conflict with the issue (though when I filed it I didn't know the exact specifics).

chmeliik:
> Additionally, now that I tried to capture the reality including Conforma, do you think this now satisfies #2509? Now, the latest changes do not mention a global policy on number of supported task versions, instead it leaves this to the task maintainers which is in conflict with the issue (though when I filed it I didn't know the exact specifics).

Mostly yes, thanks for adding this :heart: 

But a statement on how many task versions we want to support would be nice, so we can keep #2509 just for that

eskultety:
> Or undo the changes that skip CI for archived tasks and just apply the labels.

Why do we need to skip CI if we just wanted to apply the labels, I'm lost. Does this have anything to do with the fact that we want to re-release those tasks with the labels set or is there a different reason behind this particular comment?

chmeliik:
> > Or undo the changes that skip CI for archived tasks and just apply the labels.
> 
> Why do we need to skip CI if we just wanted to apply the labels, I'm lost. Does this have anything to do with the fact that we want to re-release those tasks with the labels set or is there a different reason behind this particular comment?

We want to release the tasks with the labels. If I understand correctly, merging a change will not achieve that because the CI has been changed to ignore archived tasks. So we'll need to undo those changes.

eskultety:
> > > Or undo the changes that skip CI for archived tasks and just apply the labels.
> > 
> > 
> > Why do we need to skip CI if we just wanted to apply the labels, I'm lost. Does this have anything to do with the fact that we want to re-release those tasks with the labels set or is there a different reason behind this particular comment?
> 
> We want to release the tasks with the labels. If I understand correctly, merging a change will not achieve that because the CI has been changed to ignore archived tasks. So we'll need to undo those CI changes.

Created https://github.com/konflux-ci/build-definitions/issues/2527 to track the effort. If that's all I think this PR is ready and set :) .

PR #2504, opened 2025-07-07
prevent build context path to escape source code directory

fmudry:
The build pipeline now validates that the CONTEXT parameter points to a location within the SOURCE_CODE directory.

It checks that the CONTEXT doesn't escape from SOURCE_CODE directory, otherwise the build fails at the early stage.

Previously, the script did not check the provided path. This could lead the build process to use a context in a way it shouldn't, causing (possibly security) issues. This change ensures the build context is always within its intended scope, making our builds more reliable and predictable.

---

I've updated the base buildah task and `./hack/generate-everything.sh` generated other buildah versions (except `buildah-rhtap`, which I edited manually).

I've also edited other tasks, that contain duplicate code from buildah:
- `build-paketo-builder-oci-ta` (not a duplicate, but builds an image and uses context)
- `push-dockerfiles` and generated `oci-ta` version
-  `patch.yaml` in sast-coverity-check

---

### Tests

Tested the changes with `buildah-oci-ta` in a separate repository:
- https://github.com/fmudry-konflux/devfile-sample-hello-world/pull/10
- https://github.com/fmudry-konflux/devfile-sample-hello-world/pull/11


dirgim (dismissed):
LGTM

tkdchen (approved):

task/build-paketo-builder-oci-ta/0.2/build-paketo-builder-oci-ta.yaml
```diff
@@ -144,6 +144,23 @@ spec:
         set -eu
         set -o pipefail
 
+        echo "[$(date --utc -Ins)] Validate context path"
+
+        source_dir_path=$(realpath "$SOURCE_CODE_DIR")
+        context_dir_path=$(realpath "$SOURCE_CODE_DIR/$CONTEXT")
+
+        case "$context_dir_path" in
+          "$source_dir_path"*)
```
chmeliik, line 153:
```suggestion
          "$source_dir_path/"*)
```

Like this I think

`/tmp/source-foo` is not a subpath of `/tmp/source` but the current check would think it is

fmudry: Thank you for catching this edge case. Updated. Also, if the context is empty, it defaults to '.'

chmeliik (approved):

kasemAlem (approved):

PR #2506, opened 2025-07-08
sast-shell-check: set ShellCheck jobs based on cgroup v2 CPU limits

sfowl:
- [PSSECAUT-1196](https://issues.redhat.com//browse/PSSECAUT-1196)
- set sast-shell-check default CPU limit to 8
- set SC_JOBS equal to CPU limit (rounded up) based on cgroups v2 max
- fix issue with IFS=, affecting word splitting (which impacted `read -r quota period < /sys/fs/cgroup/cpu.max`)

~Marked as draft for now~, as it requires an updated version of the quay.io/konflux-ci/konflux-test image with a new epel9 release of csmock-plugins-shellcheck-core with this [change](https://github.com/csutils/csmock/pull/205).

EDIT: konflux-test image has been updated, ready for merge

### DIFF
--- a/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
+++ b/task/sast-shell-check-oci-ta/0.1/sast-shell-check-oci-ta.yaml
@@ -149,8 +149,8 @@ spec:
 
         # generate full path for each dirname separated by comma
         declare -a ALL_TARGETS
-        IFS=","
-        for d in $TARGET_DIRS; do
+        IFS="," read -ra TARGET_ARRAY <<<"$TARGET_DIRS"
+        for d in "${TARGET_ARRAY[@]}"; do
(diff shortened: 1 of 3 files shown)


kdudka (dismissed):
Looks good.  The updated `csmock` packages are now in `epel9-testing`: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2025-f4112db701

sfowl:
Waiting on update to konflux-test image:

https://github.com/konflux-ci/konflux-test/pull/512 

side note: it seems that in epel only the latest version of each package is kept, meaning older package versions are removed, which cause builds depending on older versions to fail (e.g. hermetic builds). Example:

https://github.com/konflux-ci/konflux-test/runs/45881914589

kdudka (approved):

PR #2518, opened 2025-07-09
hack: build-manifests: Replace pattern glob for native options in find

eskultety:
`$ find task/*/*/*.yaml -maxdept 0`

is only stable until symlinks are added to the equation. Everything is fine as long as one wants the symlinks to be included, however, since we gave find a fully resolved list of files based on the glob (which isn't capable of skipping symlinks) and limited its max descend level to 0 (which is questionable since we gave it a list of resolved paths to individual YAML files) it won't apply any further filtering options specified on the command line.

In our case, we don't intend to provide updates to tasks that have been moved to the 'archived-tasks' directory which happen to be symlinked to the 'task' directory.
Therefore, manifests would continue to be built AND verified for archived tasks as well. To prevent that, let's use filtering native to find to give us only paths to task YAML files (also limiting the descend level to 3 which seems to be the norm) and ignoring symlinks along the way:

`$ find task -maxdept 3 -type f -name "*.yaml"`

### DIFF
--- a/hack/build-manifests.sh
+++ b/hack/build-manifests.sh
@@ -33,7 +33,7 @@ warning_message="# WARNING: This is an auto generated file, do not modify this f
 main() {
     cd "$SCRIPT_DIR/.."
     local ret=0
-    find task/*/*/*.yaml -maxdepth 0 | awk -F '/' '{ print $0, $2, $3, $4 }' | \
+    find task -maxdepth 3 -mindepth 3 -type f -name "*.yaml" | awk -F '/' '{ print $0, $2, $3, $4 }' | \
     while read -r task_path task_name task_version file_name
     do
         if [[ "$file_name" == "kustomization.yaml" ]]; then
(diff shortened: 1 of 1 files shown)


hack/build-manifests.sh
```diff
@@ -33,7 +33,7 @@ warning_message="# WARNING: This is an auto generated file, do not modify this f
 main() {
     cd "$SCRIPT_DIR/.."
     local ret=0
-    find task/*/*/*.yaml -maxdepth 0 | awk -F '/' '{ print $0, $2, $3, $4 }' | \
+    find task -maxdepth 3 -type f -name "*.yaml" | awk -F '/' '{ print $0, $2, $3, $4 }' | \
```
chmeliik, line 36: To make sure it's equivalent to the previous behavior, maybe `-mindepth` should also be 3? I don't _think_ we have any yaml file at levels 1 or 2, but I'm not sure

chmeliik: > isn't it actually better to either fail

Good point. I'm not sure if the current behavior would result in a sensible failure though, given that the `awk` on the same line expects paths with exactly 4 components. I suspect we'd just end up skipping the files anyway, but in a more obfuscated way

PR #2560, opened 2025-07-17
Describe why setup-tektoncd fails in GH Actions

tnevrlka:
GitHub Actions using the setup-tektoncd step have recently started intermittently failing with a somewhat cryptic error message `jq: error (at <stdin>:1): Cannot iterate over null (null)`

Upstream issue for setup-tektoncd has been filed at https://github.com/tektoncd/actions/issues/9

Until the issue is fixed in upstream, inform users that the issue is known, link it and tell them that they should re-run the workflow.


fmudry (approved):
Great job!

PR #2562, opened 2025-07-17
fix(KONFLUX-8786): make jq error user friendly

nmars:
Fixed to print user-friendly message when no mirror entries found in image mirror map. Also modified to print the entries in images-mirror-set only once at the top.


nmars:
Looks like `e2e-tests` is broken at the moment.

I tested the fix in the Konflux stage environment. 

Before the fix a `jq` error would be displayed:

```
jq: error (at <stdin>:1): Cannot iterate over null (null)
```

Now, after the fix, a user-friendly message is displayed. For example, from my stage-environment test:

```
Could not inspect original pullspec registry.redhat.io/container-native-virtualization/hco-bundle-registry-rhel9@sha256:6b690ca4d3e99732e334d1b88ef6e6635b116f44374592755817e02c2cb26c4f. Checking if there's a mirror present
No mirrors found in image mirror map for registry.redhat.io/container-native-virtualization/hco-bundle-registry-rhel9
...
```

PR #2570, opened 2025-07-21
sast-unicode: also get digest from build-container task

sfowl:
I noticed in one of other pipelines that this migration script did not handle the [case](https://gitlab.cee.redhat.com/osh/cov-sa-container/-/merge_requests/142#note_16828986) where there wasn't a build-oci-artifact or build-image-index task, but there was a build-container task.

We can also grab the image_digest from the build-container task, so added that to the migration script.


sfowl:
Oh dang:

```
error: It is not allowed to delete or modify existing migration file: task/sast-unicode-check-oci-ta/0.3/migrations/0.3.sh
```

I don't think this change is worth a new migration, so I guess this is just a small amount of pain we will have to inflict on some user users :(

Lesson learned, I hope.

PR #2582, opened 2025-07-28
fix: capture wrong digest for helm chart artifact

yftacherzog:
The build-helm-chart task was capturing a digest from the output of the skopeo copy command which was not the actual digest of the artifact. This change addresses that.

### DIFF
--- a/task/build-helm-chart-oci-ta/0.2/build-helm-chart-oci-ta.yaml
+++ b/task/build-helm-chart-oci-ta/0.2/build-helm-chart-oci-ta.yaml
@@ -291,13 +291,15 @@ spec:
           exit 1
         fi
 
-        # Extract digest from the skopeo copy output
-        digest=$(echo "$skopeo_output" | grep "Copying config" | awk '{print $3}' 2>/dev/null || echo "")
-        if [ -n "$digest" ]; then
-          echo "Successfully retrieved digest from skopeo copy: $digest"
-        else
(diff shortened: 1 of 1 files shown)


task/build-helm-chart-oci-ta/0.2/build-helm-chart-oci-ta.yaml
```diff
@@ -291,13 +291,15 @@ spec:
...
 
-        # Extract digest from the skopeo copy output
-        digest=$(echo "$skopeo_output" | grep "Copying config" | awk '{print $3}' 2>/dev/null || echo "")
-        if [ -n "$digest" ]; then
-          echo "Successfully retrieved digest from skopeo copy: $digest"
-        else
-          echo "Could not retrieve digest from skopeo copy output"
+        # Get the manifest digest of the pushed chart
+        # The manifest digest is the SHA256 of the entire manifest JSON
+        if ! manifest_digest=$(skopeo inspect --raw "docker://$pushed" | sha256sum | awk '{print "sha256:" $1}' 2>/dev/null); then
```
ifireball, line 296: it seems we mgiht be able to do this directly with curl without locally calcluating the digest. See:

```
curl -s -H "Accept: application/vnd.oci.image.manifest.v1+json" \
  https://quay.io/v2/redhat-user-workloads/konflux-vanguard-tenant/caching/squid-helm/manifests/0.1.140_2679b7c-on-pr | jq .
```

PR #2589, opened 2025-07-28
Add proxy support to buildah tasks

amisstea:
If specified, the proxy is only configured for buildah pull or build operations to prevent unintentional use by other processes.

Assisted-by: cursor (claude-3.5-sonnet)

### DIFF
--- a/task/buildah-min/0.4/buildah-min.yaml
+++ b/task/buildah-min/0.4/buildah-min.yaml
@@ -163,6 +163,25 @@ spec:
     description: Determines if the image inherits the base image labels.
     name: INHERIT_BASE_IMAGE_LABELS
     type: string
+  - default: ""
+    description: HTTP/HTTPS proxy to use for the buildah pull and build operations.
+      Will not be passed through to the container during the build process.
+    name: HTTP_PROXY
+    type: string
+  - default: ""
+    description: Comma separated list of hosts or domains which should bypass the
+      HTTP/HTTPS proxy.
+    name: NO_PROXY
+    type: string
+  - default: proxy-ca-bundle
+    description: The name of the ConfigMap to read proxy CA bundle data from.
+    name: PROXY_CA_TRUST_CONFIG_MAP_NAME
+    type: string
+  - default: ca-bundle.crt
+    description: The name of the key in the ConfigMap that contains the proxy CA bundle
+      data.
+    name: PROXY_CA_TRUST_CONFIG_MAP_KEY
+    type: string
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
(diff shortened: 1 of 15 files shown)


yftacherzog (dismissed):

amisstea:
@chmeliik the pending ADR (https://github.com/konflux-ci/architecture/pull/236) is now mentioned in the commit msg.

task/buildah-min/0.4/buildah-min.yaml
```diff
@@ -252,18 +270,47 @@ spec:
...
     image: quay.io/konflux-ci/buildah-task:latest@sha256:121ccc64ade7c25fa85e9476d6a318d0020afb159cfc0217c082c04261b3bfdf
     name: build
     script: |
       #!/bin/bash
       set -euo pipefail
 
+      function set_proxy {
+        if [ -n "${BUILDAH_HTTP_PROXY}" ]; then
+          echo "[$(date --utc -Ins)] Setting proxy to ${BUILDAH_HTTP_PROXY}"
+          export HTTP_PROXY="${BUILDAH_HTTP_PROXY}"
```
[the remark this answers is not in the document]
amisstea: Buildah supports both upper and lowercase varieties (mentioned [here](https://github.com/containers/buildah/blob/main/docs/buildah-build.1.md)).

task/buildah-min/0.4/buildah-min.yaml
```diff
@@ -252,18 +270,47 @@ spec:
...
-        cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors
+      proxy_ca_bundle=/mnt/proxy-ca-bundle/ca-bundle.crt
+      if [ -f "$ca_bundle" ] || [ -f "$proxy_ca_bundle" ]; then
+        if [ -f "$ca_bundle" ]; then
+          echo "[$(date --utc -Ins)] Using mounted CA bundle: $ca_bundle"
+          cp -vf $ca_bundle /etc/pki/ca-trust/source/anchors/ca-bundle.crt
+        fi
+        if [ -f "$proxy_ca_bundle" ]; then
+          echo "[$(date --utc -Ins)] Using mounted proxy CA bundle: $proxy_ca_bundle"
+          cp -vf $proxy_ca_bundle /etc/pki/ca-trust/source/anchors/proxy-ca-bundle.crt
```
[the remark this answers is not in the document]
amisstea: I've added an additional condition and refactored this section in the process.

task/buildah-remote/0.4/buildah-remote.yaml
```diff
@@ -656,11 +715,15 @@ spec:
       # disable host subcription manager integration
       find /usr/share/rhel/secrets -type l -exec unlink {} \;
 
+      set_proxy
+
```
MartinBasti, line 719: By default buildah propagates proxy variables into build, see `buildah build --http-proxy` option, which is true by default

```
       --http-proxy

       By default proxy environment variables are passed into the container if  set  for  the  Buildah  process.
       This  can  be  disabled by setting the --http-proxy option to false.  The environment variables passed in
       include http_proxy, https_proxy, ftp_proxy, no_proxy, and also the upper case versions of those.

       Defaults to true
```

However ADR says, that only layers will be cached, thus proxy shoudln't be propagated to build container

amisstea: I see your point now. I'll try setting `--http-proxy` to `false` to stop the env variable propagation.

amisstea: It worked! All the `buildah build` commands now have `--http-proxy=false` set. I verified that when defined, the proxy is used to pull image layers but not used for a `RUN` directive (see [PR #11](https://github.com/amisstea/testrepo/pull/11/files) in my test repo). The build also still works when no proxy is defined. I used the squid proxy logs to validate which traffic was passing through it.

MartinBasti (approved):

PR #2628, opened 2025-08-05
fix: rollback clamav-scan from 0.3 to 0.2

14rcole:
clamav-scan 0.3 introduced parallel scanning with clamdscan. Unfortunately, Conforma does not support matrixed pipeline results. This PR rolls back the upgrade to 0.3.  It can be reverted once the issue is fixed in Conforma


zxiong:
@simonbaird has raised a [PR](https://github.com/conforma/policy/pull/1461) to fix the issue that Conforma can not handle matrix results and it is expected to be merged soon, also there is a workaround that just removing existing matrix section from clamav-scan task in the PipelineRun, so I think this revert may no longer be necessary.

PR #2630, opened 2025-08-06
Remove migration script from clamav-scan v0.3

zxiong:
Currently, Conforma doesn't process `IMAGES_PROCESSED` results correctly when matrix is used in a PipelineRun (see [1]). To prevent users from adopting matrix, we are removing the migration script that otherwise enable it.

[1] https://issues.redhat.com/browse/KONFLUX-9576


zxiong:
It's not allow to remove and delete the migration script, this might not be good way. Close the PR.

PR #2631, opened 2025-08-06
Remove matrix for clamav-scan task if present

zxiong:
Currently, Conforma doesn't process `IMAGES_PROCESSED` results correctly when matrix is used in a PipelineRun (see [1]). Therefore, the matrix section for clamav-scan is removed from the multi-arch PipelineRun, and the migration script is updated to remove the matrix section from user existing PipelineRuns if presents.

[1] https://issues.redhat.com/browse/KONFLUX-9576


zxiong:
@simonbaird has raised a [PR](https://github.com/conforma/policy/pull/1461) to fix the issue that Conforma can not handle matrix results and it is expected to be merged soon, also there is a workaround that just removing existing matrix section from clamav-scan task in the PipelineRun, so I think this PR may no longer be necessary?

PR #2647, opened 2025-08-11
fix(KONFLUX-9618): Manually pass Multiarch arguments to dockerfile-json

BorekZnovustvoritel:
Dockerfile-json doesn't recognize multiarch arg values by itself. This PR parses the values of these arguments from the output of `buildah info` and makes them available to the Dockerfile for a more durable Dockerfile parsing.

### DIFF
--- a/task/buildah-min/0.4/buildah-min.yaml
+++ b/task/buildah-min/0.4/buildah-min.yaml
@@ -454,7 +454,28 @@ spec:
         BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
       done
 
-      dockerfile-json "${BUILD_ARG_FLAGS[@]}" "$dockerfile_copy" > /shared/parsed_dockerfile.json
+      # Dockerfile-json cannot parse Buildah's host variables, we have to pass them manually
+      BUILDAH_INFO=$(buildah info)
+      BUILDAH_OS=$(jq -r '.host.os' <<< "$BUILDAH_INFO")
+      BUILDAH_ARCH=$(jq -r '.host.arch' <<< "$BUILDAH_INFO")
+      BUILDAH_VARIANT=$(jq -r '.host.variant' <<< "$BUILDAH_INFO")
+      BUILDAH_PLATFORM="${BUILDAH_OS}/${BUILDAH_ARCH}"
+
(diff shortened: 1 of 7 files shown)


BorekZnovustvoritel:
As stated in [the ticket](https://issues.redhat.com/browse/KONFLUX-9618), we had some issues with these values not being present in Dockerfiles that use them. In the ticket there is a linked Dockerfile which caused issues.

Mobster receives an invalid image reference because the reference `cuda-base-$TARGETARCH` doesn't match any previous stages, while there are defined stages for `cuda-base-amd64` as well as `cuda-base-arm64`. So the expanded value `cuda-base-` matches nothing and is interpreted as a pullspec.

jedinym (approved):

task/buildah-min/0.4/buildah-min.yaml
```diff
@@ -367,7 +367,28 @@ spec:
         BUILD_ARG_FLAGS+=("--build-arg=$build_arg")
       done
 
-      dockerfile-json "${BUILD_ARG_FLAGS[@]}" "$dockerfile_copy" > /shared/parsed_dockerfile.json
+      # Dockerfile-json cannot parse Buildah's host variables, we have to pass them manually
+      BUILDAH_INFO=$(buildah info)
+      BUILDAH_OS=$(jq '.host.os' <<< "$BUILDAH_INFO" | tr -d '"' | tr -d "'")
+      BUILDAH_ARCH=$(jq '.host.arch' <<< "$BUILDAH_INFO" | tr -d '"' | tr -d "'")
+      BUILDAH_VARIANT=$(jq '.host.variant' <<< "$BUILDAH_INFO"  | tr -d '"' | tr -d "'")
```
chmeliik, line 374: Nitpick: since the intent is to get raw strings, a more obvious way to do that is `jq`'s `-r/--raw-output` flag rather than the `tr -d '"'` calls

chmeliik (approved):
Small nitpick, otherwise seems reasonable

task/buildah-min/0.4/buildah-min.yaml
BorekZnovustvoritel: Good suggestion, applied.

PR #2678, opened 2025-08-19
sast-coverity: pin base buildah task dep to specific commit

sfowl:
This is to reduce churn so that changes to buildah task don't require updating the coverity task at the same time. This should prioritize stability for the coverity task, and prevent cases when buildah task updates introduces changes that rely on recent buildah image versions, that may not yet also be in the coverity image.

### DIFF
--- a/task/sast-coverity-check/0.3/kustomization.yaml
+++ b/task/sast-coverity-check/0.3/kustomization.yaml
@@ -2,7 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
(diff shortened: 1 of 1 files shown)


hongweiliu17 (approved):

sfowl:
> Seems like a good idea. What is the plan for keeping the task up to date? I don't think we want to let them drift too much either, do we?

My intention is to periodically update the commit, and refresh the task. That said, discussion elsewhere is leading towards the coverity task possibly being deprecated entirely, so this maintenance might end up being short lived.

chmeliik (approved):

PR #2687, opened 2025-08-20
sast-snyk: don't fail task when KFP unreachable

sfowl:
The main change here is changing the csfilter-kfp --kfp-git-url param to --kfp-dir and doing a git clone of KFP earlier in the task. If the git clone succeeds a new var KFP_CLONED=1 will be set, and if this is unset, csfilter-kfp will not run, and the scan results will be unfiltered.

In case there is some issue with this logic, the KFP_DIR will still exist (although empty) and csfilter-kfp should run without errors.

Also improves some of the log messages with log level prefixes.

[PSSECAUT-1297](https://issues.redhat.com//browse/PSSECAUT-1297)

### DIFF
--- a/task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
+++ b/task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
@@ -150,7 +150,7 @@ spec:
           PROJECT_NAME=${COMPONENT_LABEL}
         fi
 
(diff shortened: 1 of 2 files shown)


sfowl:
I want to test a few different manual scenarios with this before merging, so please don't merge on my behalf yet (though reviews and approvals very much welcome now)

kasemAlem (approved):

jsztuka (approved):
lgtm

PR #2700, opened 2025-08-24
feat: build appstudio-utils for arm

arewm:
konflux-ci/konflux-ci can install on arm, but we cannot run our pipelines on arm unless all step images are built for both x86 and arm.

This change starts to build the appstudio-utils image for both architectures.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -5,7 +5,7 @@ metadata:
   name: build-definitions-pull-request
   annotations:
     pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main" && ( !has(body.pull_request) || !body.pull_request.draft) ) || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/") && body.head_commit != "null" )
-    pipelinesascode.tekton.dev/task: "[task/git-clone/0.1/git-clone.yaml, task/buildah-remote/0.4/buildah-remote.yaml, task/build-image-index/0.1/build-image-index.yaml, .tekton/tasks/task-lint.yaml, .tekton/tasks/e2e-test.yaml, task/sast-snyk-check/0.4/sast-snyk-check.yaml, task/sast-unicode-check/0.2/sast-unicode-check.yaml, .tekton/tasks/task-switchboard.yaml]"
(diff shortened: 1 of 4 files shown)


tkdchen (commented):
Is this ready for review?

.tekton/push.yaml
```diff
@@ -66,30 +66,52 @@ spec:
...
+          - name: IMAGE
+            value: quay.io/konflux-ci/appstudio-utils:{{ revision }}
+          - name: IMAGES
+            value:
+            - $(tasks.build-appstudio-utils.results.IMAGE_REF[*])
+        runAfter:
+          - build-appstudio-utils
+        taskRef:
+          name: build-image-index
+          version: "0.1"
```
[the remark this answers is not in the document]
arewm: I am not sure I understand what you mean by this comment. Do we just need to remove the version from these?

arewm:
The EC image is not multi-arch. We can either wait for the PR to enable it as multi-arch (conforma/cli/pull/2704) or we can `curl` the binaries from the releases.

.tekton/pull-request.yaml
```diff
@@ -190,11 +216,12 @@ spec:
             description: "custom bundle for fbc-builder pipeline"
           steps:
             - name: build-bundles
-              image: quay.io/konflux-ci/pull-request-builds:appstudio-utils-{{revision}}
+              image: $(params.image-target)
               workingDir: $(workspaces.source.path)/source
               script: |
                 #!/usr/bin/env bash
                 set -e
+                git config --global --add safe.directory /workspace/source/source
```
arewm, line 224: I needed to add this because we were hitting an issue when trying to run the included script. This started happening after I changed the buildah task from the custom one for appstudio-utils to our buildah-remote task. I suspect that the issue is that the buildah-remote task has the user defined: https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-remote/0.4/buildah-remote.yaml#L1120-L1122

tisutisu (approved):

chmeliik (approved):

.tekton/pull-request.yaml
```diff
@@ -37,26 +37,48 @@ spec:
...
+        matrix:
+          params:
+          - name: PLATFORM
+            value:
+            - linux/x86_64
+            - linux/arm64
         params:
           - name: IMAGE
-            value: quay.io/konflux-ci/pull-request-builds:appstudio-utils-{{revision}}
+            value: quay.io/konflux-ci/appstudio-utils:{{ revision }}
```
chmeliik: I just now noticed - what is the purpose of this change? I don't think we want to push to the "production" repo from the PR pipeline

chmeliik: Ah, I think it's fixed in a later commit

PR #2701, opened 2025-08-25
Add needed volume mounts for certs.

ggallen:
Add needed volume mounts for certs.


bcrochet:
@arewm Can you offer some advice here?

ggallen:
> You can test this change by either producing your own Tekton bundle `tkn bundle push` and then modifying the bundle resolver target, or by changing the bundle resolver to a git resolver.
> 
> Even if this wouldn't create a "passing" conforma policy, it should be sufficient for testing on a cluster.

I will try this when I get a chance, but I currently do not have a component to build.  Thanks, @arewm and @bcrochet.

PR #2702, opened 2025-08-25
feat: change the core services pipeline to multi-arch

arewm:
We have made a concerted effort to ensure that Konflux controllers can run on both amd64 and arm64 hosts. While this is beneficial, none of the pipelines that we produce can currently run on both hosts since the step images are not all being built for both platforms. In order to assist with making all tasks support both architectures, we will make the pipeline which is focused on building Konflux components default to building both architectures.

Use of this pipeline now requires the multi-platform controller to be deployed on the cluster.

### DIFF
--- a/pipelines/core-services/README.md
+++ b/pipelines/core-services/README.md
@@ -1,1 +1,3 @@
-An extension of build pipelines for Stonesoup core services to allow updating infra-deployments after successful build in CI.
+An extension of build pipelines for Konflux core services to allow updating infra-deployments after successful build in CI.
+
+Since [konflux-ci/konflux-ci](github.com/konflux-ci/konflux-ci) supports running on arm64 and x86, this pipeline will produce a multi-arch image by default as well.
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -11,10 +11,10 @@ metadata:
   name: docker-build
 spec:
   description: |
-    This pipeline is ideal for building container images from a Containerfile while reducing network traffic.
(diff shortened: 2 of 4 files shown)


arewm:
I don't see too many uses of this pipeline directly within the konflux-ci github org. While it is possible that some components may have been onboarded and use this pipeline without the pipeline bundle references (therefore preventing me from easily detecting it), this is not an available pipeline in the build configmap so it is likely that [these](https://github.com/search?q=org%3Akonflux-ci+pipeline-core-services-docker-build&type=code) are all the users.

That being said, the primary use case for me to want to change this default was to enable arm for hermeto without pushing to that repo: https://github.com/arewm/hermeto/blob/main/.tekton/push.yaml

I am happy to discuss this change as we could just change the hermeto pipeline to multi-arch (but it would be more complex as that repo appears to depend on the added [slack notification](https://github.com/arewm/hermeto/blob/main/.tekton/push.yaml#L21-L22) task).

chmeliik (commented):
Enabling arm behind the scenes has the potential to break builds that used to work (even if multi-platform-controller _is_ available). More conservatively, we could default to just x86 and leave it to the users of this pipeline to enable arm explicitly

And if this was primarily for Hermeto, it may not be needed. @brunoapimentel was working on onboarding a "midstream" of Hermeto to Konflux, and looking at the attestations of the images in https://quay.io/repository/konflux-ci/hermeto?tab=tags&tag=latest, that may be done now

chmeliik:
Should I enable the index image by default or make that optional, defaulting to off?

I would say enable by default, so that builds can be made multi-platform just by adding another arch to the `build-platforms` param

PR #2706, opened 2025-08-26
feat(CLOUDDST-29100): Optimize clair-scan performance with matrix & report convert

zxiong:
This PR improves performance of the clair-scan task by:
- Introduce Tekton’s matrix to run clair-scan TaskRuns in parallel across image architectures.
- Reuse the first scan result by converting its output format for the second report. This avoids redundant scans, reduces resource usage, and speeds up execution.

Performance tests comparing between original task and the task after optimization.
Tests | Original | After optimization
-- | -- | --
Average Duration | 0:13:08 | 0:03:30


task/clair-scan/0.3/MIGRATION.md
```diff
@@ -0,0 +1,18 @@
+# Migration from 0.2 to 0.3
+
+Version 0.3:
+
+This version improves performance:
+- Parallel execution: TaskRuns are executed in parallel across each image architecture using Tekton’s matrix.
+- Single scan execution: The scan is performed once, and the second report is generated by converting the first report’s output format, reducing redundant scanning.
+
+Changes:
+- Added a new parameter `image-platform` (default: "").
```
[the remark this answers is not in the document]
zxiong: The `image-platform` parameter is primarily used when a matrix is applied to the taskrun within a pipelinerun.
An example value is "linux/x86_64". Other all possible values can be found here: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml#L10-L46

hongweiliu17 (approved):
LGTM

jsztuka (dismissed):
lgtm

pipelines/docker-build-multi-platform-oci-ta/docker-build-multi-platform-oci-ta.yaml
```diff
@@ -256,7 +256,12 @@ spec:
       operator: in
       values:
       - "false"
-  - name: clair-scan
+  - matrix:
+      params:
+      - name: image-platform
```
chmeliik, line 261: Why does every task use a different name for the platform param?

preflight uses `platform`, clamav uses `image-arch` and now clair uses `image-platform`

task/clair-scan/0.3/MIGRATION.md
```diff
@@ -0,0 +1,18 @@
...
+- Optimized scanning: run the scan once and reuse results for the second report.
+- Introduced matrix in the build pipeline definition to support parallel TaskRuns for multi-arch builds.
+
+## Action from users
+
+Renovate bot PR will be created with warning icon for a clair-scan which is expected, no actions from users are required.
+
+For multi-arch build, `matrix` will be added to build pipeline definition file automatically by script migrations/0.3.sh when MintMaker runs [pipeline-migration-tool](https://github.com/konflux-ci/pipeline-migration-tool).
```
chmeliik, line 18: Please always add manual migration instructions. The migration may fail to apply. In some Konflux deployments, Mintmaker may not even be configured to use the migration tool.

pipelines/docker-build-multi-platform-oci-ta/docker-build-multi-platform-oci-ta.yaml
zxiong: Good question.

Originally, I used the name `image-arch` for clamav-scan, but in Clair I realized it might not be the best choice since the value is actually a platform (e.g., linux/amd64). In any case, the parameter name should be consistent across tasks. I can create a PR later to update them all to use the same name `image-platform`.

chmeliik: Yes, that would be nice :pray:

PR #2714, opened 2025-08-27
Add volume mount to sast-snyk-check-oci-ta.

ggallen:
Add volume mount to sast-snyk-check.

I am running Konflux locally on a kind cluster on my laptop. When running my pipeline, the ecosystem-cert-preflight-checks task fails with TLS errors. This is rectified by adding `useTAVolumeMount: true` to the recipe.yaml files which will add the volume mounts for the certificates in the task when it is generated.


sfowl:
Could you share a bit more context for this change? This task has been part of user pipelines for a fair while without this volume mount, not sure why it's needed now.. (and only the oci-ta variant?)

ggallen:
@sfowl?

sfowl (approved):
Thanks, LGTM!

Issue #2719, opened 2025-08-29
apply-tags: jq missing from image


mh21:
running apply-tags with something like

    - name: apply-tags
      params:
        - name: IMAGE_URL
          value: $(tasks.build-image-index.results.IMAGE_URL)
        - name: IMAGE_DIGEST
          value: $(tasks.build-image-index.results.IMAGE_DIGEST)
      runAfter:
        - build-image-index
      taskRef:
        params:
          - name: name
            value: apply-tags
          - name: bundle
            value: quay.io/konflux-ci/tekton-catalog/task-apply-tags:0.2@sha256:70881c97a4c51ee1f4d023fa1110e0bdfcfd2f51d9a261fa543c3862b9a4eee9
          - name: kind
            value: task
        resolver: bundles

has a failure in the logs like

    /tekton/scripts/script-1-m7jpf: line 11: jq: command not found

source code:

        if manifest_digest=$(jq -e -r '.manifests[0].digest' <<< "$raw_inspect"); then

confirming it is missing from the skopeo image:

```console
$ podman run -it --rm registry.access.redhat.com/ubi9/skopeo:9.6-1754871306@sha256:e59e2cb3fd8d7613798738fb06aad5aab61f32c18aed595df16a46a8e078dfa6 jq
Error: crun: executable file `jq` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

this failure
- should fail the task, but does not because the jq execution is hidden in the if condition which swallows the failure
- should not happen in the first place

PR #2748, opened 2025-09-04
fix: change image used in  apply-tags task

tisutisu:
Use `appstudio-utils` image which contains both `skopeo` and `jq`
Fixes: https://github.com/konflux-ci/build-definitions/issues/2719

### DIFF
--- a/task/apply-tags/0.2/apply-tags.yaml
+++ b/task/apply-tags/0.2/apply-tags.yaml
@@ -43,7 +43,7 @@ spec:
         requests:
           cpu: 100m
           memory: 256Mi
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Have you tested this? As we've learned, changing the image can always break something unexpected

tisutisu:
> Have you tested this? As we've learned, changing the image can always break something unexpected

@chmeliik we already have test for the apply-tags task [here](https://github.com/konflux-ci/build-definitions/pull/2748#pullrequestreview-3183706123), it should be running as part of the CI, by testing you mean the `apply-tags` task functionality or any other testing?

PR #2774, opened 2025-09-10
fix(KONFLUX-9947): add idms file path param

asergienk:
The new parameter allows you to pass a file path (for example, /workspace/shared-data/mirrorSet.yaml) instead of embedding the entire YAML as a string.

Reason for this change:
Previously, the mirror set YAML file had to be saved as a Tekton task result and then passed to the `eaas-create-ephemeral-cluster-hypershift-aws` StepAction as a string. This worked fine for small strings (<4096 bytes). However, Tekton enforces a 4096-byte limit on task results, so larger YAML files could not be saved or passed this way.

The solution was to store the YAML as a file in a shared workspace. For this to work, the eaas-create-ephemeral-cluster-hypershift-aws StepAction needed the ability to consume a file path directly, which was not previously supported. The new parameter adds this capability.


stepactions/eaas-create-ephemeral-cluster-hypershift-aws/0.1/eaas-create-ephemeral-cluster-hypershift-aws.yaml
```diff
@@ -71,6 +79,18 @@ spec:
     #!/bin/bash
     set -eo pipefail
 
+    # Resolve imageContentSources from file (if provided) or fallback to inline string
+    ICS_VALUE=""
+    if [[ -n "$IMAGE_CONTENT_SOURCES_FILE" && -f "$IMAGE_CONTENT_SOURCES_FILE" ]]; then
```
amisstea, line 84: I think we should also check for the case where the ICS file param is provided but the file doesn't exist. Right now this will silently continue without configuring the ICS. Users _may_ only notice after waiting for the cluster to provision. I think we should fail instead.

PR #2775, opened 2025-09-10
Remove show-summary from default pipelines

lcarva:
As discussed in a previous community meeting:
* This task is not marked as a required task since May 17th, 2024.
* The -oci-ta pipelines already do not include it.

This commit removes the Task from the default pipelines but leaves the Task definition around for those that do want to continue using it.


pipelines/docker-build-rhtap/docker-build-rhtap.yaml
```diff
@@ -8,30 +8,7 @@ metadata:
...
   name: docker-build-rhtap
 spec:
-  finally:
-  - name: show-sbom
-    params:
-    - name: IMAGE_URL
-      value: $(tasks.build-container.results.IMAGE_URL)
-    taskRef:
-      name: show-sbom-rhdh
-      version: "0.1"
```
chmeliik, line 18: It looks like the RHTAP pipelines also lost the show-sbom task, was that intended?

avi-biton (approved):
/lgtm

PR #2795, opened 2025-09-16
prefetch-dependencies (Hermeto) files update

slimreaper35:


task/sast-coverity-check/0.3/sast-coverity-check.yaml
```diff
@@ -728,23 +728,23 @@ spec:
 
       echo "[$(date --utc -Ins)] Setup prefetched"
 
-      if [ -f "$(workspaces.source.path)/cachi2/cachi2.env" ]; then
+      if [ -f "$(workspaces.source.path)/cachi2/prefetch.env" ]; then
```
sfowl, line 731: I'm a bit surprised changes to the sast-coverity task in this PR. The coverity task is based on the buildah task, but was pinned to a [specific commit](https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check/0.3/kustomization.yaml#L6) some time ago. Perhaps this PR was based on an old version of the main branch?

slimreaper35: > Perhaps this PR was based on an old version of the main branch?
No.

Looks like the only way to pass the CI is to update the git ref.

task-generator/trusted-artifacts/golden/prefetch-dependencies/base.yaml
```diff
@@ -98,17 +98,20 @@ spec:
...
       --for-output-dir=/cachi2/output
 
       # NOTE: Compatibility hack, hermeto will create a hermeto.repo when processing RPMs (1 for
-      # each architecture) which may break users expecting cachi2.repo
+      # each architecture) which may break users expecting prefetch.repo or legacy cachi2.repo
       find /var/workdir/cachi2/output \
         -type f \
         -name hermeto.repo \
-        -execdir mv {} cachi2.repo \;
+        -execdir sh -c 'mv "$1" prefetch.repo && ln -s prefetch.repo cachi2.repo' shell {} \;
```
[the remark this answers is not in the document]
slimreaper35: yes :disappointed:

task/prefetch-dependencies/0.2/prefetch-dependencies.yaml
```diff
@@ -356,6 +356,9 @@ spec:
       --for-output-dir=/cachi2/output \
       --output "$(workspaces.source.path)/cachi2/prefetch.env"
 
+      # backwards compatibility with legacy cachi2.env
```
eskultety, line 359: @chmeliik @mmorhun what would be a reasonable time frame to remove compatibility hacks like ^this one?

archived-tasks/buildah-min/0.4/buildah-min.yaml
```diff
@@ -567,11 +567,11 @@ spec:
             "$dockerfile_copy"
         echo "Prefetched content will be made available"
 
-        prefetched_repo_for_my_arch="/tmp/cachi2/output/deps/rpm/$(uname -m)/repos.d/cachi2.repo"
```
eskultety, line 570: **Commit subject:**
>  Rename cachi2.env file to more generic prefetch.env 

I believe this patch is about `.repo` not `.env` :)

eskultety (commented):
LGTM, but I'd like to wait for other authoritative build-team ACK for more comments on other aspects I may not be seeing.

eskultety (approved):

archived-tasks/buildah-min/0.4/buildah-min.yaml
slimreaper35: fixed

task/sast-coverity-check/0.3/sast-coverity-check.yaml
slimreaper35: Everything has been replaced automatically. I replaced all occurrences of the file, and then I ran the `generate-everything.sh` script, which reverted my change.

chmeliik: Yes, should be safe

chmeliik:
> We're doing backwards compatibility only one way. The prefetch task keeps creating the older filenames, so it will be compatible with older buildah tasks.
> 
> But the buildah task expects the new names only, so it will not be compatible with older prefetch tasks. Either we should make the compatibility work both ways, or bump the versions of both tasks and do incompatible changes

I don't think this was addressed

chmeliik:
> > But the buildah task expects the new names only, so it will not be compatible with older prefetch tasks.
> 
> What do _older tasks_ mean in this context?

Any tasks older than the ones that will be released when this PR is merged. If we merge it in the current state, than `buildah:0.5 [released >= when this PR got merged]` will be incompatible with `prefetch-dependencies:0.2 [released < when this PR got merged]`

slimreaper35:
> Any tasks older than the ones that will be released when this PR is merged. If we merge it in the current state, then buildah:0.5 [released >= when this PR got merged] will be incompatible with prefetch-dependencies:0.2 [released < when this PR got merged]

Okay, in that case, it might be easier to create a new version for both tasks and replace all `cachi2` occurrences.

chmeliik:
> So, I also wanted to update the Hermeto output path: `$(workspaces.source.path)/cachi2/output/bom.json` -> `$(workspaces.source.path)/prefetch/output/bom.json`. However, I realized that I would have to bump the version of many tasks. Am I right?

Yeah, it is one of the outputs of the task, so changing the path would be a backwards incompatible change. Providing a symlink for backwards compatibility might work though

PR #2799, opened 2025-09-16
Add  rsync command arguments to improve transfer speed back and forth on remote build machines

mshaposhnik:
In this PR:

Adding `-W` key to all transfers, as we're always sending forward to a clean machine, and no files are modified when sending back, only new ones added. There is no need to calculate the deltas.

Adding `-z`. Even though it doesn't add much profit, it still cuts 2-3-5%, which is notable on slow IBM conn-s.

Adding `--stats` output, to analyze the network perf. and collect some more stats


mshaposhnik:
@chmeliik @MartinBasti hit that merge button if you're ok, I don't have it..

PR #2806, opened 2025-09-18
fix(buildah): correct the buildah build ignorefile precedence

tommartensen:
The new buildah task in v0.5 picks `.dockerignore`  over `.containerignore` if both files are present in the build context.
We have both files in our repository (https://github.com/stackrox/stackrox). We relied on a `.containerignore` to pass the build context as we need it for Konflux, and still need `.dockerignore` for upstream builds (so deleting it is not an option).

This works, because the default behaviour for ignore files in buildah build  is

> If  the  .containerignore/.dockerignore  file  exists  in  the context directory, buildah build reads its
       contents. If both exist, then .containerignore is used.

(Source: https://manpages.ubuntu.com/manpages/jammy/man1/buildah-build.1.html)

Hence, `buildah build` will pick `.containerignore` over `.dockerignore` if both files are present. --> this was the default behaviour up to v0.4 of the task, ie before https://github.com/konflux-ci/build-definitions/commit/6522f6024f3cf784caeda639eac763731771415f#diff-d1179f56596679d45c7cdd228dd957593a14133175499194ccb3b0f7ddf9e781R665-R681 was added.

We believe that this is a bug/regression and the buildah  tasks should have the same precedence configuration as the `buildah build` to prevent incorrect ignore settings to be applied during the build.

### DIFF
--- a/task/buildah-min/0.5/buildah-min.yaml
+++ b/task/buildah-min/0.5/buildah-min.yaml
@@ -677,10 +677,10 @@ spec:
 
       # Make sure our labels.json file isn't filtered out
       containerignore=""
-      if [ -f "$SOURCE_CODE_DIR/$CONTEXT/.dockerignore" ]; then
-        containerignore="$SOURCE_CODE_DIR/$CONTEXT/.dockerignore"
-      elif [ -f "$SOURCE_CODE_DIR/$CONTEXT/.containerignore" ]; then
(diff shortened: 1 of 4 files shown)


chmeliik (commented):
Thanks for the fix! Do you have a reference for this being the correct precedence order? It would be worth including a link in the commit message

PR #2810, opened 2025-09-19
fix(OCPBUGS-60987): change default mediaType from oci to docker

MartinBasti:
OpenShift's documentation indicates that disconnected clusters can use any container registry that supports Docker v2-2. This means that disconnected mirroring will fail if the registry doesn't support OCI.

https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/disconnected_environments/connected-to-disconnected

In order to have the easiest default support for all images produced, we should just change the default mediaType. If any users want OCI specific mediaTypes, they can still enable that.

FBC build must continue to use OCI format.

Migration script is created to add BUILDAH_FORMAT param if doesn't exist with value "docker" (or "oci" for FBC build).

### DIFF
--- a/task/init/0.2/migrations/0.2.1.sh
+++ b/task/init/0.2/migrations/0.2.1.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Created for task: init@0.2.1
+# Creation time: 2025-09-19T10:00:53Z
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+
+buildah_format="docker"
+
+# Check if the run-opm-command task exists, FBC pipeline must be OCI by default
+if yq -e '.spec.tasks[] | select(.name == "run-opm-command")' "$pipeline_file" >/dev/null; then
+    echo "FBC pipeline detected, skipping"
+    exit 0
+fi
+
+# Tekton bundle build pipeline must be OCI by default
+if yq -e '.spec.tasks[] | select(.taskRef.params[] | (.name == "name" and (.value == "tkn-bundle" or .value == "tkn-bundle-oci-ta")))' "$pipeline_file" >/dev/null; then
+    echo "Tekton bundle build pipeline detected, skipping"
+    exit 0
+fi
+
+
+buildah_format_tasks=( \
+  "buildah-min" "buildah" "buildah-oci-ta" \
+  "buildah-remote" "buildah-remote-oci-ta" \
+  "build-image-index" "build-image-manifest" \
+  # sast tasks have BUILDAH_FORMAT param, but we can ignore them, as image is ephemeral for testing only
+  # "sast-coverity-check" "sast-coverity-check-oci-ta" \
+)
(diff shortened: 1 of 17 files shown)


task/init/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,76 @@
...
+# Determine all tasks which should be updated
+tasks_to_be_updated=()
+
+if yq -e '.spec.tasks[] | select(.name == "build-images")' "$pipeline_file" >/dev/null; then
+    tasks_to_be_updated+=( "build-images" )
+fi
+
+if yq -e '.spec.tasks[] | select(.name == "build-container")' "$pipeline_file" >/dev/null; then
+    tasks_to_be_updated+=( "build-container" )
+fi
```
chmeliik, line 35: I think we should include these task names only if they use the buildah task ref. Buildah and build-image-index are the only ones that have the BUILDAH_FORMAT param, so it wouldn't make sense to add the param if it's a different task.

That way we would exclude the tekton-bundle-builder pipeline, for example (which we do want to exclude IMO)

task/init/0.2/migrations/0.2.1.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Update all relevant tasks
+for build_taskname_value in "${tasks_to_be_updated[@]}"; do
+    echo "processing task ${build_taskname_value}"
+    # Check if the task already has the BUILDAH_FORMAT parameter
+    if ! yq -e ".spec.tasks[] | select(.name == \"$build_taskname_value\").params[] | select(.name == \"BUILDAH_FORMAT\")" "$pipeline_file" >/dev/null; then
+        echo "Adding BUILDAH_FORMAT parameter to $build_taskname_value task"
+        yq -i "(.spec.tasks[] | select(.name == \"$build_taskname_value\")).params += [{\"name\": \"BUILDAH_FORMAT\", \"value\": \"\$(params.buildah-format)\"}]" "$pipeline_file"
+    else
+        echo "BUILDAH_FORMAT parameter already exists in $build_taskname_value task. No changes needed."
```
chmeliik, line 74: There's also a chance that someone is already setting this parameter, but with a different value than `$(params.buildah-format)`. We should probably overwrite the value in that case

pipelines/fbc-builder/fbc-builder.yaml
```diff
@@ -64,6 +64,11 @@ spec:
     description: Add built image into an OCI image index
     name: build-image-index
     type: string
+  - default: oci
+    description: The format for the resulting image's mediaType. Valid values are
+      oci or docker.
+    name: buildah-format
+    type: string
```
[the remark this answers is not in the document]
MartinBasti: fixed

task/init/0.2/migrations/0.2.1.sh
MartinBasti: reworked

chmeliik (dismissed):

PR #2813, opened 2025-09-23
sast-coverity: pin base buildah task in 0.2 to specific commit

sfowl:
Previously done for sast-coverity 0.3 version in 6d7eee3b4b0d65755d26097aa550930d1be23262.


sfowl:
you're right, it's not necessary, closing.

PR #2828, opened 2025-09-25
task: add wait-for-pipelinerun-image task

JasonPowr:
This pr adds a task called wait-for-pipelinerun-image the task waits for the most recent Tekton PipelineRun matching a given label selector to finish, then emits the image URL and digest that the PipelineRun published in its results.

Use case:
RHTAS has a system for building the operator manager and its bundle within the same PR. The process works by running the manager build pipeline and the bundle build pipeline simultaneously. The bundle build pipeline includes a task that waits for the manager build pipeline to finish, then retrieves the manager artifact and generates an up-to-date bundle.

We recently have ran into conforma failing due to the task not being trusted, according to these docs, this is the solution https://conforma.dev/docs/policy/trusting_tasks.html

Original task: https://github.com/securesign/pipelines/blob/main/tasks/wait-for-pipelinerun.yaml
MR Ref: https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/10589


chmeliik:
> Hi @chmeliik, yes in our pipelines we use Trusted Artifacts tasks, from my understanding the violation is being triggered because the results of the task is being consumed by the build task [here](https://github.com/securesign/pipelines/blob/main/pipelines/bundle-build-oci-ta.yaml#L234)

That shouldn't cause a violation, Trusted Artifacts were designed to allow things like this :thinking: 

Only the SOURCE_ARTIFACT and PREFETCH_ARTIFACT parameters should be "protected", and you're passing those from a trusted task: https://github.com/securesign/pipelines/blob/d138d9d23a3f86dafdbbb572fa656e7cc7ed299f/pipelines/bundle-build-oci-ta.yaml#L237-L240

What exact violation are you getting?

JasonPowr:
We are seeing three in total

```
Image "quay.io/securesign/model-validation-operator@sha256:3e607ed16b14d6f449c09b3b17e8ae1debe981582c86bc47018dfd9fd0eb47a4" not built by a trusted task: Build Task(s) "wait-for-pipelinerun" are not trusted
```

The two of these one for each component
```
Code tampering detected, untrusted PipelineTask "get-manager-image" (Task "wait-for-pipelinerun") was included in build chain comprised of: get-manager-image
```

The naming being a bit different as I updated the task for a broader use

JasonPowr:
Hmm OK Ill give this a try and get back to you, thanks @chmeliik :)

chmeliik:
> @chmeliik That does seem to have solved the trusted tasks issues :) the conforma failures have went away, I am seeing a different error now:
> 
> ```
> The "BUILD_ARGS" parameter of the "build-container" PipelineTask includes an untrusted digest: sha256:caf425898bfbc5d081da5694de3a45a84c3a1bd2930b12bc8f8177b1aeb94021`
> ```
> 
> I understand that this is probaly not the place to do this but any ideas? or is this just something we will need a exception for?

Huh, I didn't know this rule existed :sweat_smile: 

Off the top of my head, I can't come up with a reason to disallow this. I would suggest contacting the Conforma team in #forum-konflux-conforma or #forum-konflux-developer in our internal Slack

JasonPowr:
No problem @chmeliik, thanks again for your help

PR #2832, opened 2025-09-28
fix(KFLUXSPRT-5245): fix task sast-snyk-check finds no sarif files

kasemAlem:
**Root Cause Analysis**
The failure occurs in the file merging step of the sast-snyk-check task. Here's what's happening:
**The Problem**
1. Individual Snyk scans succeed: The task scans each target directory and generates individual SARIF files with names like `sast_snyk_check_out_${d//\//_}.json` (line 204).
2. File merging fails: The task then tries to merge all these individual files into a single sast_snyk_check_out.json file using this command (line 208)

[KFLUXSPRT-5245](https://issues.redhat.com/browse/KFLUXSPRT-5245)


task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
```diff
@@ -228,8 +228,14 @@ spec:
           snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$resolved_path" --max-depth=1 --sarif-file-output="${resolved_path}/sast_snyk_check_out_${d//\//_}.json" 1>&2 >>stdout.txt || true
         done
 
-        # Merge all SARIF outputs
-        find "$SOURCE_CODE_DIR" -name "sast_snyk_check_out_*.json" -exec cat {} + >"${SOURCE_CODE_DIR}/sast_snyk_check_out.json"
```
kdudka, line 232: I do not understand how you can merge JSON files with `cat`.  That will not produce a valid JSON file.  This problem was already introduced in commit 244f032aa70f2d1d5fd1b3f2fe033b5c22a69d48.

kdudka (changes requested):

task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
sfowl: That is a great point, not sure how that was missed earlier.. we will need to fix.

kdudka: > I'll fix using jq

I do not think this would be easy:

- If you get valid JSON, it does not mean that you get valid SARIF.
- If you get valid SARIF, it does not mean that `csgrep` will be able to process it correctly (because `csgrep` implements only a small subset of the SARIF specification).

I would suggest to rather use `csgrep` directly to read all the SARIF files you need to process.

task/sast-snyk-check/0.4/sast-snyk-check.yaml
```diff
@@ -204,8 +204,9 @@ spec:
           snyk code test $ARGS --severity-threshold="$SEVERITY_THRESHOLD" "$resolved_path" --max-depth=1 --sarif-file-output="${resolved_path}/sast_snyk_check_out_${d//\//_}.json" 1>&2>> stdout.txt || true
         done
 
+
         # Merge all SARIF outputs
-        find "$SOURCE_CODE_DIR" -name "sast_snyk_check_out_*.json" -exec cat {} + > "${SOURCE_CODE_DIR}/sast_snyk_check_out.json"
+        find "$SOURCE_CODE_DIR" -name "sast_snyk_check_out_*.json" -exec jq -s '{"version": "2.1.0", "runs": map(.runs[])}' {} + >"${SOURCE_CODE_DIR}/sast_snyk_check_out.json" 2>/dev/null || echo '{"version": "2.1.0", "runs": []}' >"${SOURCE_CODE_DIR}/sast_snyk_check_out.json"
```
kdudka: As I said in my previous comment, this is not going to work because `csgrep` does not support multiple runs in a SARIF file: https://github.com/csutils/csdiff/blob/2791d449cede54a0e687d8c820da04bb4166d97b/src/lib/parser-json-sarif.cc#L149

task/sast-snyk-check-oci-ta/0.4/sast-snyk-check-oci-ta.yaml
```diff
@@ -228,8 +228,19 @@ spec:
...
+        # Collect all SARIF files
+        mapfile -t sarif_array < <(find "$SOURCE_CODE_DIR" -name "sast_snyk_check_out_*.json" -type f 2>/dev/null | sort)
```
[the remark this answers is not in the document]
kdudka: The script is interpreted by `bash`, so you can expand this as `**/sast_snyk_check_out_*.json`.  This way you avoid the troubles with forking a separate process, handling the array in shell, quoting of special chars while expanding shell variables etc.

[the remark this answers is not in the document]
kasemAlem: hmmm, 
never used this way, 
wdyt about the way I refactored with `**/sast_snyk_check_out_*.json`

ccronca:
Hi @kasemAlem, are you still working on this? If not, I’d like to continue this work and create a PR, if that’s okay with you. Could you also share the current status?

kasemAlem:
@ccronca , I've made the fixes was requested at last comment by Kamil, 
can you please review

sfowl:
@kasemAlem Thanks for looking at this again. Since this PR was made, the sast task have been moved to a new repo: https://github.com/konflux-ci/konflux-sast-tasks 

Could you please open this PR there instead? Additionally, the new repo supports tests for the snyk task, you could include a new test that validates this PR works as expected.

kasemAlem:
> @kasemAlem Thanks for looking at this again. Since this PR was made, the sast task have been moved to a new repo: https://github.com/konflux-ci/konflux-sast-tasks
> 
> Could you please open this PR there instead? Additionally, the new repo supports tests for the snyk task, you could include a new test that validates this PR works as expected.

Thanks @sfowl for the important correction, on it a PR will be created soon , for the test of the task 
I can create a separate Ticket for it so it be handled within our team dashboard, 
what do you think ?

MartinBasti:
closing, this task doesn't exist in this repo anymore

PR #2837, opened 2025-09-29
fix: correct paths for init migration

dhaiducek:
The 0.2.2 script pointed to `spec.params[]` where it needed to point to `spec.pipelineSpec.params[]`.

Followup to:
- #2836

Slack thread:
- https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1759111303625859

Issue: https://issues.redhat.com/browse/KFLUXSPRT-5270

### DIFF
--- a/task/init/0.2/init.yaml
+++ b/task/init/0.2/init.yaml
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.2.2"
+    app.kubernetes.io/version: "0.2.3"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "konflux"
--- a/task/init/0.2/migrations/0.2.3.sh
+++ b/task/init/0.2/migrations/0.2.3.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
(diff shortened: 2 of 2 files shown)


dhaiducek:
> @dhaiducek Ok, I think I'm starting to understand this. #2836 applied the fix only for `Pipeline`s while you're using a `PipelineRun`?

They're targeting the same pipelines in the repo #2386 was targeting `.spec.params[]`, but those are user parameters. Instead, it should have used `.spec.pipelineSpec.params[]` where the pipeline parameters and defaults are defined.

I tried it out on our repos and it seems to work, though verification would also be good since I may not be aware of all the parts to this issue--I just noticed the paths were wrong.

dhaiducek:
> Bear with me because I don't have original context for the whole issue.
> 
> There's no mention of `pipelineSpec` in the original [0.2.1.sh](https://github.com/konflux-ci/build-definitions/blob/83f029a0d5861a12a4f15f90cd473520f632703b/task/init/0.2/migrations/0.2.1.sh)?
> 
> My understanding is that `0.2.1.sh` added the `buildah-format` parameter. How did `buildah-format` get to `pipelineSpec` parameters then? Do you have an example I could take a look at?

Oh, that's a good point. On my end in our repos we have `pipelineSpec`. But if the flow somehow happens before the spec lands in `pipelineSpec`, then the original hotfix is indeed what will work and this one is irrelevant.

PR #2840, opened 2025-09-30
fix(init): serious flaw in migration

MartinBasti:
0.2.2.sh migration was adding element into doc root and causes failure.

New migration 0.2.3.sh has been added.

0.2.2 was removed from existence, to avoid issues if somebody runs it accidentally and also
image was removed.


MartinBasti:
New changes just fixes the comment with time and version

PR #2845, opened 2025-10-01
Temporarily disable FBC e2e test

chmeliik:
The repository used for testing FBC builds [1] uses a base image from registry.redhat.io.

To make it possible to pull this base image, the OpenShift namespace where the tests run has a registry.redhat.io pull secret. This pull secret is linked to the 'appstudio-pipeline' Service Account.

The 'appstudio-pipeline' Service Account is no longer used for running the test build pipelines. Until recently, this wasn't a problem, because the build-service operator would link the same secrets to the Service Accounts used today. But, after a recent change [2], build-service doesn't do that anymore so the test is broken.

Disable the test while we work on a fix.

[1]: https://github.com/konflux-qe-bd/fbc-sample-repo
[2]: https://github.com/konflux-ci/build-service/commit/e7349da3fd900eed14130aa0953626faaca150a5

### DIFF
--- a/.tekton/tasks/e2e-test.yaml
+++ b/.tekton/tasks/e2e-test.yaml
@@ -60,7 +60,7 @@ spec:
       - name: APP_SUFFIX
         value: "$(params.app_suffix)"
       - name: COMPONENT_REPO_URLS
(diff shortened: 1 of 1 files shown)


chmeliik:
@konflux-ci/operator-foundry PTAL, this temporarily disables the only FBC test

Alternatively, if you have ideas for a non- registry.redhat.io base image for https://github.com/konflux-qe-bd/fbc-sample-repo/blob/main/catalog.Dockerfile, that might be a better fix

flacatus (approved):

chmeliik:
There might be another way, @rcerven informed me there is a `build.appstudio.openshift.io/common-secret` label, so I'll try to set that on the relevant secret as a fix

chmeliik:
Fix attempt: https://github.com/redhat-appstudio/infra-deployments/pull/8430

chmeliik:
It looks like the fix via infra-deployments worked! Closing

Issue #2788, opened 2025-09-15
error handling is ineffective in the `git-clone` task


kdudka:
The script of the `git-clone` task starts with `set -eu`, which makes the shell interpreter exit immediately when a command exits with a non-zero status: https://github.com/konflux-ci/build-definitions/blob/8ddc2b52768ed7b8c0794b59ad5b3d944354b15e/task/git-clone/0.1/git-clone.yaml#L183

Consequently, the bodies of all the ` if [ "${*_EXIT_CODE}" != "0" ]` conditions are effectively dead code.  For example, the `git status` command is never executed when a merge conflict occurs: https://github.com/konflux-ci/build-definitions/blob/8ddc2b52768ed7b8c0794b59ad5b3d944354b15e/task/git-clone/0.1/git-clone.yaml#L281

PR #2848, opened 2025-10-02
Remove set -e option for git-clone task

ralphbean:
There's lots of error handling code in this task that is all skipped because of this flag. Remove it, so we can handle errors more explicitly.

Fixes https://github.com/konflux-ci/build-definitions/issues/2788


gbenhaim:
In think that the features provided by the bash script should be implemented in the git-init binary (https://github.com/tektoncd-catalog/git-clone), which can be better tested an would be useful for other tekton users as well.

PR #2863, opened 2025-10-08
Set CPU limit for prefetch task

hugares:
We have seen strange behavior where prefetch memory consumption is abnormally high when CPU is unlimited and this leads to the pod being OOMKilled.

Setting the cpu limit brought back memory usage under the 3Gi limit during our experimentations so setting it on the task, to avoid setting this in every tenant pipeline.

[KONFLUX-10360](https://issues.redhat.com//browse/KONFLUX-10360)


eskultety (approved):
LGTM

a-ovchinnikov (approved):

PR #2886, opened 2025-10-14
feat: add user and group mention support to slack notification

zhujian7:
## Summary

Add optional `user-ids` and `group-ids` array parameters to the slack-webhook-notification task to allow mentioning multiple Slack users and groups in notifications.

## Changes

- Added `user-ids` parameter (optional array) - accepts list of Slack user IDs (e.g., U024BE7LH)
- Added `group-ids` parameter (optional array) - accepts list of Slack group IDs (e.g., S0614TZR7)
- When set, mentions are prepended to the notification message using Slack's mention syntax
  - User: `<@USER_ID>` for each user in the array
  - Group: `<!subteam^GROUP_ID>` for each group in the array
- Updated README.md with parameter documentation

## Use Case

This allows common pipeline specifications to conditionally mention users or groups without hardcoding empty mention syntax in messages. Teams can pass user/group IDs as parameters based on their needs.

## Test plan

- [x] Test with empty arrays (backward compatibility)
- [x] Test with single user-id
- [x] Test with multiple user-ids
- [x] Test with single group-id
- [x] Test with multiple group-ids
- [x] Test with both user-ids and group-ids set
- [x] Verify mentions appear correctly in Slack using test script

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/task/slack-webhook-notification-oci-ta/0.1/README.md
+++ b/task/slack-webhook-notification-oci-ta/0.1/README.md
@@ -7,10 +7,12 @@ Sends message to slack using incoming webhook
 |---|---|---|---|
 |SOURCE_ARTIFACT|The Trusted Artifact URI pointing to the artifact with the application source code.||true|
 |files|List of file to dump. The content will be added to the message.|[]|false|
+|group-ids|List of Slack group IDs to mention (e.g., S0614TZR7). If set, the groups will be mentioned in the notification.|[]|false|
 |key-name|Key in the key in secret which contains webhook URL for slack.||true|
 |message|Message to be sent||true|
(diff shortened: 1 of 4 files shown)


MartinBasti:
This could be easily implemented by adding it directly into slack message param, do we really need separate params?

MartinBasti:
using tekton array ofc https://tekton.dev/docs/pipelines/tasks/#array-type

MartinBasti:
> Group is also necessary for us because the people in the group may change

I meant group as list as well

zhujian7:
@MartinBasti Thanks for your review. I have addressed the comments, PTAL.

MartinBasti (approved):
Thank you for contribution

PR #2892, opened 2025-10-16
Modify fbc directory structure to only validate that theirs a package directory inside  the conffolder.

ashwgit:
The previous fbc directory structure check was marked as too restrictive cause it was checking that all the package had catalog.json file inside it. After discussion the only requirement finalized was the conffolder (generally named "configs" in case of IIB) have a operator package  sub-directory inside it. 
 
 This merge request modifies the existing check of fbc file structure and simplifies it. 

I will keep the request as Draft , till we decide on how do we actually confirm that the sub-directory is a operator package and not just a simple directory.

### DIFF
--- a/task/validate-fbc/0.1/validate-fbc.yaml
+++ b/task/validate-fbc/0.1/validate-fbc.yaml
@@ -345,13 +345,13 @@ spec:
         echo "Validating FBC directory structure."
         PACKAGE_DIRS=$(find ".${conffolder}" -mindepth 1 -maxdepth 1 -type d)
 
-        # The only requirement is that at least one package subdirectory exists.
+        # First, check if any package subdirectories exist at all.
         if [[ -z "${PACKAGE_DIRS}" ]]; then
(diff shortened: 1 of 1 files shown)


task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -345,43 +345,21 @@ spec:
...
+          echo "    ${conffolder}/"
+          echo "    └── <any-package-name>/"
+
+          # Set the note for the final error report
+          note="FBC directory structure validation failed: No package subdirectories were found inside the '${conffolder}' directory."
+
           failure_num=$((failure_num + 1))
           TESTPASSED=false
         else
-          # Array to hold directories that fail validation
```
yashvardhannanavati: we also need to ensure the package directories are not empty.

ashwgit: @yashvardhannanavati implemented

PR #2918, opened 2025-10-27
ensure correct digests are returned after format conversion

fmudry:
When buildah push converts formats (e.g., OCI to Docker), it creates new remote digests. However, the task was inspecting a stale local cache and a floating tag, causing it to output the old digests. This broke downstream attestations, which targeted the wrong, pre-conversion manifests.

This commit fixes this by:
  - Clearing the local cache (buildah manifest rm) to force a remote lookup.
  - Inspecting the immutable digest-based reference (repo@sha256:...) from the image-digest file to avoid race conditions.

## Testing

I've tested the new task in [this PR](https://github.com/fmudry-konflux/devfile-sample-hello-world/pull/19) (see only `the last 2 commits`).
Commits:
  - First commit (second from last): the `new task` implementation
  - Second commit (last): Failed on ec as expected (see [the comment](https://github.com/fmudry-konflux/devfile-sample-hello-world/pull/19#issuecomment-3451514416))

### DIFF
--- a/task/build-image-index/0.1/build-image-index.yaml
+++ b/task/build-image-index/0.1/build-image-index.yaml
@@ -178,26 +178,39 @@ spec:
           exit 1
       fi
 
+      # Remove local cache to force remote 'inspect'.
+      # When 'buildah push' converts formats (e.g., OCI->Docker), it creates new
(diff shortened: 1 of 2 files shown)


arewm (dismissed):
thanks

PR #2923, opened 2025-10-28
add activation-key support to build-vm-image

Victoremepunto:


ralphbean:
## Issue: Silent failure on certificate copy

At line 256 in the new code, there's a potential issue with silent failure:

```bash
sudo cp /etc/pki/entitlement/*.pem $BUILD_DIR/entitlement/ || true
```

The `|| true` silently ignores failures. If registration succeeded but no certificates were generated, the build would proceed without entitlements and fail later with confusing errors.

**Recommendation**: Verify certificates were copied and provide feedback:

```bash
if \! sudo cp /etc/pki/entitlement/*.pem $BUILD_DIR/entitlement/ 2>/dev/null; then
  echo "Warning: No entitlement certificates found after registration"
  ls -la /etc/pki/entitlement/
fi
```

This way, if something goes wrong, there's visibility into what happened rather than a silent failure that causes confusing errors downstream.

PR #2934, opened 2025-10-30
Isv 6411 + ISV-6412: Stop scanning source dir using Syft and make it configurable

Allda:
This pull request combines 2 separate jiras but since they are interconnected, I am opening one PR for both.

feat([ISV-6412](https://issues.redhat.com//browse/ISV-6412)): Stop scanning source code using Syft
    
Based on the agreement with ProdSec we are removing the Syft scan of the
source directory. Scanning a source code (git repo) leads to SBOMs with
packages that are not in the product/image itself. It produces a lot of
false positives in the CVE remediation and makes SBOMs bigger than
needed.

feat([ISV-6411](https://issues.redhat.com//browse/ISV-6411)): Add task param to customize Syft catalogers

For subset of the product team the default set of Syft catalogers is not
optimal and brings more false positives. This change make the Syft
catalogers cofigurable.

JIRA: [ISV-6411](https://issues.redhat.com//browse/ISV-6411) [ISV-6412](https://issues.redhat.com//browse/ISV-6412)

### DIFF
--- a/task/buildah/0.6/buildah.yaml
+++ b/task/buildah/0.6/buildah.yaml
@@ -140,6 +140,14 @@ spec:
       https://github.com/anchore/syft/wiki/Package-Cataloger-Selection
     type: string
     default: ""
+  - name: SBOM_SOURCE_SCAN_ENABLED
+    description: >-
+      Flag to enable or disable SBOM generation from source code. The scanner
+      of the source code is enabled only for non-hermetic builds and can be disabled if
+      the SBOM_SYFT_SELECT_CATALOGERS can't turn off catalogers that cause false
+      positives on source code scanning.
+    type: string
+    default: "true"
   - name: BUILDAH_FORMAT
     description: The format for the resulting image's mediaType. Valid values are oci (default) or docker.
     type: string
@@ -255,6 +263,8 @@ spec:
       value: $(params.SBOM_TYPE)
     - name: SBOM_SYFT_SELECT_CATALOGERS
       value: $(params.SBOM_SYFT_SELECT_CATALOGERS)
+    - name: SBOM_SOURCE_SCAN_ENABLED
+      value: $(params.SBOM_SOURCE_SCAN_ENABLED)
     - name: ANNOTATIONS_FILE
       value: $(params.ANNOTATIONS_FILE)
     - name: WORKINGDIR_MOUNT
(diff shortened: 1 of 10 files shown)


chmeliik (commented):
> Based on the agreement with ProdSec we are removing the Syft scan of the
source directory. Scanning a source code (git repo) leads to SBOMs with
packages that are not in the product/image itself. It produces a lot of
false positives in the CVE remediation and makes SBOMs bigger than
needed.

Is ProdSec aware how many false *negatives* this can cause for non-hermetic builds?

For example, for non-hermetic Rust builds we will have 0 Rust dependencies in the SBOM. Syft gets the dependency info only from Cargo.lock or from "auditable binaries" (which are not the standard way of building Rust binaries). See https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/rust/cataloger.go

chmeliik:
> Is ProdSec aware how many false _negatives_ this can cause for non-hermetic builds?
> 
> For example, for non-hermetic Rust builds we will have 0 Rust dependencies in the SBOM. Syft gets the dependency info only from Cargo.lock or from "auditable binaries" (which are not the standard way of building Rust binaries). See https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/rust/cataloger.go

- Add task param to customize Syft catalogers

^ I could see this being a reasonable remediation, but IMO it should only apply to the source scan. Scanning the built image rarely results in false positives, it's far more likely to result in false negatives

p-rog:
> > Is ProdSec aware how many false _negatives_ this can cause for non-hermetic builds?
> > For example, for non-hermetic Rust builds we will have 0 Rust dependencies in the SBOM. Syft gets the dependency info only from Cargo.lock or from "auditable binaries" (which are not the standard way of building Rust binaries). See https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/rust/cataloger.go
> 
> * Add task param to customize Syft catalogers
> 
> ^ I could see this being a reasonable remediation, but IMO it should only apply to the source scan. Scanning the built image rarely results in false positives, it's far more likely to result in false negatives

Running syft against the image with default settings can also generate false negatives, depending on what content is detected, like RPM. Being able to disable some of the Syft catalogers allows reducing impact and improve the final SBOM which later is used for vulnerability scanning. This option will be used really only in a corner cases. I do not expect anyone will really use this.

task/buildah/0.6/buildah.yaml
```diff
@@ -132,6 +132,21 @@ spec:
...
+      The value corresponds to Syft's CLI flag --select-catalogers. The details
+      about available catalogers can be found here:
+      https://github.com/anchore/syft/wiki/Package-Cataloger-Selection
+    type: string
+    default: ""
+  - name: SBOM_SOURCE_SCAN_ENABLED
+    description: >-
+      Flag to enable or disable SBOM generation for source code. By default the scanner
+      of the source code is enabled and can be disabled if the SBOM_SYFT_SELECT_CATALOGERS
+      can't turn off catalogers that cause false positives on source code scanning.
```
chmeliik: Let's mention that source scan is only enabled for non-hermetic builds - `SBOM_SOURCE_SCAN_ENABLED=true + HERMETIC=true` skips the source scan

chmeliik (approved):

PR #2947, opened 2025-11-05
task/buildah: add SOURCE_DATE_EPOCH for reproducible builds

jlebon:
This just wires through to buildah's new support for reproducible builds
via the `--source-date-epoch` and `--rewrite-timestamp` switches.

There's a much larger story here of course about enabling reproducible
builds more widely. Konflux is well-placed for this, especially for
hermetic builds where it already today knows to prefetch inputs and
shove them in an image. (A major issue right now though I think is that
too much implementation logic lives in the task entrypoints rather than
being baked in the buildah-task image.)

Anyway, a first step in all this is to be able to use the buildah
features, hence this patch.

An easy way to use this is to wire it up to the `commit-timestamp`
result from the `git-clone` task. (For hermetic builds, there's a
better timestamp available; the commit timestamp of the last git commit
which touched any of the Hermeto input files. But we can tweak that in
a follow-up.)

### DIFF
--- a/task/buildah-min/0.6/buildah-min.yaml
+++ b/task/buildah-min/0.6/buildah-min.yaml
@@ -202,7 +202,7 @@ spec:
     type: string
   - default: ""
     description: Defines the single build time for all buildah builds in seconds since
-      UNIX epoch
+      UNIX epoch. Conflicts with SOURCE_DATE_EPOCH.
     name: BUILD_TIMESTAMP
     type: string
   - default: ""
@@ -217,6 +217,16 @@ spec:
     description: Omit build history information from the resulting image. Improves
       reproducibility by excluding timestamps and layer metadata.
     name: OMIT_HISTORY
+    type: string
+  - default: ""
+    description: Timestamp in seconds since Unix epoch for reproducible builds. Sets
+      image created time and SOURCE_DATE_EPOCH build arg. Conflicts with BUILD_TIMESTAMP.
+    name: SOURCE_DATE_EPOCH
+    type: string
+  - default: "false"
+    description: Clamp mtime of all files to at most SOURCE_DATE_EPOCH. Does nothing
+      if SOURCE_DATE_EPOCH is not defined.
+    name: REWRITE_TIMESTAMP
     type: string
   results:
   - description: Digest of the image just built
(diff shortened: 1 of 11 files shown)


ralphbean:
Relates to:

* https://github.com/konflux-ci/build-definitions/issues/1268
* https://konflux-ci.dev/docs/building/labels-and-annotations/#generating-dynamic-labels-or-annotations

task/buildah-min/0.6/buildah-min.yaml
```diff
@@ -581,6 +588,14 @@ spec:
         BUILDAH_ARGS+=("--inherit-labels=false")
       fi
 
+      if [ -n "${BUILDAH_SOURCE_DATE_EPOCH}" ]; then
+        BUILDAH_ARGS+=("--source-date-epoch=${BUILDAH_SOURCE_DATE_EPOCH}")
+        BUILDAH_ARGS+=("--rewrite-timestamp")
```
cgwalters: Note that this is a separate option for a reason; it's a big hammer, and very crucially doesn't cover things like Python files with embedded timestamps - which are one of the things currently breaking rhel-bootc reproducibility. To solve things like that we need to wire up https://github.com/keszybz/add-determinism or similar.

(really what we want is a combination of add-determinism and layer splitting)

task/buildah/0.6/buildah.yaml
```diff
@@ -568,6 +575,14 @@ spec:
         BUILDAH_ARGS+=("--inherit-labels=false")
       fi
 
+      if [ -n "${BUILDAH_SOURCE_DATE_EPOCH}" ]; then
+        BUILDAH_ARGS+=("--source-date-epoch=${BUILDAH_SOURCE_DATE_EPOCH}")
+        BUILDAH_ARGS+=("--rewrite-timestamp")
+        if [ -z "$BUILD_TIMESTAMP" ]; then
+          BUILD_TIMESTAMP="$BUILDAH_SOURCE_DATE_EPOCH"
```
chmeliik: Since we already have the `BUILD_TIMESTAMP` param, WDYT about using its value instead of adding a new param? Adding a new param would:
- create an opportunity for mismatched timestamps
- make it so that the existing users of BUILD_TIMESTAMP don't benefit from your changes

jlebon: Sigh, sorry, your second comment appeared after I commented and some React websocket on the page decided to reactivate and render it. :)

> That's enough of a difference to warrant a separate parameter. Can we instead make BUILD_TIMESTAMP and SOURCE_DATE_EPOCH mutually exclusive?

Yeah, that's probably fair I guess. Let me look into this.

jlebon: >> That's enough of a difference to warrant a separate parameter. Can we instead make BUILD_TIMESTAMP and SOURCE_DATE_EPOCH mutually exclusive?
>
> Yeah, that's probably fair I guess. Let me look into this.

OK did that! Still need to test it. I did though still made it set the `BUILD_TIMESTAMP` variable so that we get all the non-standard labels like `build-date` to agree.

jlebon: OK yup, that check works as expected! :heavy_check_mark:

jlebon:
Rebased and addressed remaining comments!

chmeliik (approved):
Thanks! :tada:

PR #2953, opened 2025-11-05
feat(ISV-6451): added SBOM_SKIP_VALIDATION to mobster-related tasks

bclindner:
Hey, first PR here. I am still working out the local dev/testing process, so this may take me a bit to get right.

This updates mobster to the latest Quay tag & adds the ability to add the new `--skip-validation` argument to `mobster generate` commands using `SBOM_SKIP_VALIDATION` in relevant tasks where we do that. I think this is primarily to avoid performance bottlenecks with the contextual SBOM.

Some minor refactors to how we were calling mobster in the tasks had to happen here as the new flag has to be after `generate` rather than the subcommand, mind the big changes to the tasks to use arrays for arguments.

The ticket specified skipping validation is only a default on the `buildah` task (& derivatives).

### DIFF
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -43,6 +43,7 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |IMAGE| The target image and tag where the image will be pushed to.| None| '$(params.output-image)'|
 |IMAGES| List of Image Manifests to be referenced by the Image Index| None| '['$(tasks.build-images.results.IMAGE_REF[*])']'|
 |IMAGE_EXPIRES_AFTER| Delete image tag after specified time resulting in garbage collection of the digest. Empty means to keep the image tag. Time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| ""| '$(params.image-expires-after)'|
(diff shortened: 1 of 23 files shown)


chmeliik:
> I think this is primarily to avoid performance bottlenecks with the contextual SBOM.

Hmm, I'm not sure I see the connection. It feels like validation would take the same amount of time regardless of whether the SBOM is contextual or not.

Could you elaborate on the reasons for adding this?

ezopezo:
> > > I think this is primarily to avoid performance bottlenecks with the contextual SBOM.
> > 
> > 
> > Hmm, I'm not sure I see the connection. It feels like validation would take the same amount of time regardless of whether the SBOM is contextual or not.
> > Could you elaborate on the reasons for adding this?
> 
> @ezopezo can answer this best

This performance flaw is happening outside of the contextual SBOM workflow (even though it was exposed during exploration of the performance improvement of the ctx SBOM). 

After any SBOM generation mobster is triggering validation of produced SBOM with validators offered by spdx_tools and cyclonedx libraries. These validation operations are costly (taking long time for large SBOM-s) and non-specific (they are not considering RH specifics of the produced SBOM, they only offer a general overview) and output is only printed out to stdout as warnings. 

Thus we decided to make it optional, freeing computational time.

ralphbean (dismissed):
Approved (approval scoped just to my tasks). +1 to Adam's comment about an explanatory note.

bclindner:
There we go. @ralphbean @chmeliik if you could please re-review that would be great!

chmeliik (approved):

chmeliik (approved):

PR #2964, opened 2025-11-06
task/buildah: add SKIP_INJECTIONS parameter

jlebon:
Today, the buildah task automatically adds `COPY` directives into the Containerfile to inject dynamically-generated files (`content-sets.json` and `labels.json`). This kind of dynamic mutation of the build process makes it very hard to reproduce the build outside of Konflux (especially `labels.json`, which requires parsing the Containerfile).

Really, all the information in those files already lives in the upstream repo, and it's not hard for a project to take over that resposibility.

Add a new SKIP_INJECTIONS parameter which when set to `true` skips all those mutations.

To reiterate more explicitly: all our images should have these files. They are consumed by scanners like Clair. The goal is NOT to ship images without them, but simply to allow projects to take over that responsibility. We will eventually have checks in place that verify the presence and contents of these files. This will block anyone using this parameter without the necessary work upstream to add them.


p-rog:
> > We will eventually have checks in place
> 
> What kind of checks do you have in mind?

@chmeliik There is ECP rule for this labels already:
https://github.com/release-engineering/rhtap-ec-policy/pull/149

PR #2966, opened 2025-11-07
chore(readme): update migration example

MartinBasti:
Cover how to do migration based on the task name, to avoid cases when users just change task name to something else.


tkdchen (dismissed):
LGTM. A suggestion is to mention this explicitly in the changes, whatever in the readme text or the script by comments:

> A migration script should find out tasks from a Pipeline by the referenced real task name.

MartinBasti:
Added

tkdchen (approved):

PR #2988, opened 2025-11-13
feat(KONFLUX-10515): add parallel processing to fips-operator-check

zxiong:
Introduces parallel image scanning with MAX_PARALLEL parameter (default 8) to improve performance when checking multiple relatedImages for FIPS compliance.

Performance tests comparing the original `fips-operator-check-step-action` with the optimized version.
 Metrics | Original | After optimization (8 parallels) | Improved %
-- | -- | -- | --
Duration | 0:14:15 | 0:01:44 | 87.83%

Test env & test data: this evaluation was conducted in the Konflux staging environment, with 51 related images.

BTW, this PR only updates the `fips-operator-check-step-action`. A follow-up PR will update the dependent tasks’ step action references and add the `MAX_PARALLEL` parameter, making it configurable by users.


stepactions/fips-operator-check-step-action/0.2/fips-operator-check-step-action.yaml
```diff
@@ -0,0 +1,253 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: StepAction
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.2"
```
yashvardhannanavati: @zxiong do we need a new stepaction version? IIUC, the new parameter isn't mandatory and wouldn't require a migration. Can we make this change in the existing stepaction?

yashvardhannanavati: I'm fine with a new version too if you feel strongly about it. Also, please keep in mind we might have to bump the resources for the step using this stepAction in the FIPS tasks as we are running parallel scans.

zxiong: > @zxiong do we need a new stepaction version? IIUC, the new parameter isn't mandatory and wouldn't require a migration. Can we make this change in the existing stepaction?

Good point, we don’t need to bump the version.

> I'm fine with a new version too if you feel strongly about it. Also, please keep in mind we might have to bump the resources for the step using this stepAction in the FIPS tasks as we are running parallel scans.

Yes, a follow-up PR will address that once this is merged. Thanks for pointing it out.

zxiong:
> LGTM. Thanks zhiming. Do we need to increase the the memory in task since processing images in parallel?

Yes, I increase resources in the follow-up PR for the tasks that reference this stepaction. Thanks for pointing it out.

yashvardhannanavati (approved):

PR #2998, opened 2025-11-17
Disinherit base image labels

ralphbean:
This changes the default value of INHERIT_BASE_IMAGE_LABELS from true to false.

---

This will cause some disruption and require product teams to set certain labels explicitly, but defaulting to inherit base image labels is not what we want.

For example, inheriting base image labels means that if you happen to omit setting `name` or `cpe` labels, then your image will *appear* to be ubi9 if you build on top of ubi9. The safer choice is to not inherit labels from base images by default. Only inherit labels from base images if the user explicitly sets it with a param.

This is particularly important for scanners that use labels (and /root/buildinfo/labels.json) in order to identify images to look them up in vulnerability databases.

### DIFF
--- a/task/buildah-min/0.7/buildah-min.yaml
+++ b/task/buildah-min/0.7/buildah-min.yaml
@@ -209,7 +209,7 @@ spec:
     description: The image is built from this URL.
     name: SOURCE_URL
     type: string
-  - default: "false"
+  - default: "true"
     description: Determines if SBOM will be contextualized.
     name: CONTEXTUALIZE_SBOM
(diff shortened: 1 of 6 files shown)


chmeliik (commented):
Would it be worth it to bump buildah version to 0.7 for this? Otherwise users will just see the behavior unexpectedly changing between two SHAs of buildah:0.6

ralphbean:
Also, fixed docs in 6b5ac7befadc21a07979a6c302c310eb1a0facb0.

PR #3019, opened 2025-11-25
Always add org.opencontainers.image.created

tnevrlka:
org.opencontainers.image.created is currently added only when a `BUILD_TIMESTAMP` or `SOURCE_DATA_EPOCH` parameter is provided.

Doing so should help reproducible builds because the timestamp is consistent.
This is different from trying to get the current timestamp in multiple parallel buildah task instances used for multi-arch builds where the timestamp will almost certainly differ.

Hovewer, if the timestamps are not specified, the inconsistency already occurs in multi-arch image indexes anyway in other attributes like `build-date` or `created`

Therefore, it seems better to always include the org.opencontainers.image.created even without parameters for a bit more consistency between builds

### DIFF
--- a/task/buildah-min/0.7/buildah-min.yaml
+++ b/task/buildah-min/0.7/buildah-min.yaml
@@ -680,15 +680,17 @@ spec:
         ANNOTATIONS+=("--annotation" "org.opencontainers.image.source=${SOURCE_URL}")
       fi
       [ -n "$IMAGE_EXPIRES_AFTER" ] && DEFAULT_LABELS+=("--label" "quay.expires-after=$IMAGE_EXPIRES_AFTER")
+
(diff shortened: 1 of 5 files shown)


chmeliik (approved):
> However, having at least some annotation even though it might not be consistent throughout builds between multiple architectures seems to be better than having no such annotation at all.

To add to this - the individual images in multi-arch image indexes *already* have inconsistent timestamps when the user doesn't specify `BUILD_TIMESTAMP` or `SOURCE_DATE_EPOCH`. Both in the `build-date` label and in the `created` attribute in the image config (`skopeo inspect --config docker://... | jq .created`)

So we're not introducing *more* inconsistency with this change and we're removing the inconsistency where some builds did have `org.opencontainers.image.created` and others didn't. LGTM 👍

PR #3022, opened 2025-11-25
Update conforma cli

joejstuart:
The ec task checks was running an old version of the cli that has a bug in processing rule_data. Updating to the latest.

https://issues.redhat.com/browse/EC-1554


joejstuart:
Depends on https://github.com/conforma/policy/pull/1579

PR #3023, opened 2025-11-25
Fix ko pipeline to use oci buildah format and push multiple tags.

ggallen:
Fix the following in the `ko` task and pipeline:

- Change the buildah format to be `oci`.
- Add an additional tag to the image pushed by `ko`.

### DIFF
--- a/pipelines/ko-build-oci-ta/ko-build-oci-ta.yaml
+++ b/pipelines/ko-build-oci-ta/ko-build-oci-ta.yaml
@@ -170,7 +170,7 @@ spec:
       value:
       - $(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)
     - name: BUILDAH_FORMAT
-      value: docker
+      value: oci
     runAfter:
     - build-container
     taskRef:
--- a/task/ko-oci-ta/0.1/ko-oci-ta.yaml
+++ b/task/ko-oci-ta/0.1/ko-oci-ta.yaml
@@ -121,6 +121,8 @@ spec:
           value: $(params.KO_DOCKER_REPO)
         - name: TAG
           value: $(params.TAG)
+        - name: TASKRUN_NAME
+          value: $(context.taskRun.name)
       script: |
         #!/bin/bash
         set -euo pipefail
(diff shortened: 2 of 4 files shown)


pipelines/ko-build-oci-ta/ko-build-oci-ta.yaml
```diff
@@ -170,7 +170,7 @@ spec:
       value:
       - $(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)
     - name: BUILDAH_FORMAT
-      value: docker
+      value: oci
```
arewm: For consistency with the other tasks, do we want to comment that this is temporary? Some artrifacts building with this pipeline would still need to produce `docker` by default. 

This could just be a temporary change that you make locally until the index image is fixed.

task/ko-oci-ta/0.1/ko-oci-ta.yaml
```diff
@@ -147,15 +149,15 @@ spec:
         sbom_url=$(grep 'Published SBOM' /tmp/build.log | awk '{print $5}')
         sbom_file=$(grep 'Writing SBOM to' /tmp/build.log | awk '{print $6}')
 
-        cosign attach sbom --type=spdx --input-format=json --sbom="${sbom_file}" "${url}@${digest}"
+        cosign attach sbom --type=spdx --input-format=json --sbom="${sbom_file}" "${ref}"
```
arewm: What is this ref you are using?

pipelines/ko-build-oci-ta/ko-build-oci-ta.yaml
```diff
@@ -170,7 +175,7 @@ spec:
       value:
       - $(tasks.build-container.results.IMAGE_URL)@$(tasks.build-container.results.IMAGE_DIGEST)
     - name: BUILDAH_FORMAT
-      value: docker
+      value: $(params.buildah-format)
```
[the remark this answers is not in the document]
ggallen: I originally hard-coded this as `oci` for `ko`, but you can see an earlier review from @arewm questioned that. So I reverted to the original version that had it as a parameter. For KubeArchive builds I will be setting the parameter to `oci`.

PR #3051, opened 2025-12-03
Use the new location in Quay for Conforma image refs

simonbaird:
The `quay.io/enterprise-contract` quay org was replaced by `quay.io/conforma`. For now there are updates being pushed to both places, but it's better to stop using the deprecated image refs. (See commit messages for extra details.)

Update: Actually the pinned task definition was quite old, as noticed by @joejstuart, and it's being updated to the latest version here. So for the task definition in `pipelines/enterprise-contract.yaml` there is a potentially non-trivial upgrade happening here.

Ref: https://issues.redhat.com/browse/EC-1570

### DIFF
--- a/policies/all-tasks.yaml
+++ b/policies/all-tasks.yaml
@@ -2,7 +2,7 @@
 # These policies are meant to be applied to all of the Tasks in this repo.
 sources:
   - policy:
-      - oci::quay.io/enterprise-contract/ec-task-policy:latest
+      - oci::quay.io/conforma/task-policy:latest
     data:
       - oci::quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles:latest
       - github.com/release-engineering/rhtap-ec-policy//data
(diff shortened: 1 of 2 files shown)


pipelines/enterprise-contract.yaml
```diff
@@ -120,7 +120,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
-            value: quay.io/enterprise-contract/tekton-task:latest@sha256:88755b87abb38842824ed2489e7bb78d3d6fb9179ee046ee6e2542e50a5e24e0
+            value: quay.io/conforma/tekton-task:latest@sha256:88755b87abb38842824ed2489e7bb78d3d6fb9179ee046ee6e2542e50a5e24e0
```
joejstuart: I think this is the digest `2d5d51c5402cae2d521affbb280fcd00d2ea39fe63be80375f2663afe6c30e96`

simonbaird: This digest bump happened in recently merged #3046 , so after the rebase, this is back to a repo change only.

joejstuart (approved):

PR #3052, opened 2025-12-03
Remove obsolete verify-enterprise-contract task

simonbaird:
As explained in the comments at the top of the task definition, the verify-enterprise-contract task used by Konflux is defined in https://github.com/conforma/cli/tree/main/tasks/verify-enterprise-contract/0.1 and published at `quay.io/conforma/tekton-task`.

Previously, (fka) RHTAP synced tasks from the build-definitions repo to its own repo, so this task was kept here so that would work, but RHTAP doesn't do that any more, so there's no longer any reason to keep it.

Ref: https://issues.redhat.com/browse/EC-1571
Ref: https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1764100311926569


lcarva (approved):
Yay!

joejstuart (approved):

dheerajodha (approved):

PR #3056, opened 2025-12-04
fix(KFLUXVNGD-519): bump init task to 0.2.5

avi-biton:
- bump init task to 0.2.5
- Add a migration file `0.2.5.sh` to fix the issue where the migration script failed to detect
variations of buildah task and therefor did
not update them.


avi-biton:
@chmeliik Can you please help me force merge it, I dont have the permissions

PR #3075, opened 2025-12-11
Add convert-image-tasks-to-digests in run-opm-command-oci-ta task

rauhersu:
**Why these changes?**

When building olm fbc catalogs, image references might be generated with tags (e.g., `registry.redhat.io/my-bundle:v1.0`) if the corresponding catalog template was based on tags too. Exclusive tags based catalog templates for released bundle images are a powerful development pattern when combined with the nudge mechanism: should all the bundle images include tags but the unreleased bundle (digest), we can automatically streamline the corresponding catalog build, without leveraging any extra logic, just the bundle nudge for the unreleased bundle image.

Telco operators want to leverage that pattern in our fbc pipelines and it will require a final digest mapping for the tags based images, which has been implemented as an extra step, `convert-image-tags-to-digests`, in `run-opm-command-oci-ta` task. The final catalog generated will be entirely based on digests.

**Implementation Details**

- `task/run-opm-command-oci-ta/0.1/run-opm-command-oci-ta.yaml` : added new `convert-image-tags-to-digests` step with `CONVERT_TAGS_TO_DIGESTS` parameter (default: `true`). Leverages skopeo for the tag to digest mapping.

-  `task/run-opm-command-oci-ta/0.1/README.md` : updated documentation with new parameter and functionality description.

### DIFF
--- a/task/run-opm-command-oci-ta/0.1/run-opm-command-oci-ta.yaml
+++ b/task/run-opm-command-oci-ta/0.1/run-opm-command-oci-ta.yaml
@@ -36,6 +36,18 @@ spec:
       type: string
       description: Optional, path for ImageDigestMirrorSet file. It defaults to '.tekton/images-mirror-set.yaml'
       default: ".tekton/images-mirror-set.yaml"
+    - name: CONVERT_TAGS_TO_DIGESTS
+      type: string
+      description: Optional. Convert all image tags in the catalog output to sha256 digests. Enabled by default. Set to 'false' to disable.
+      default: "true"
+    - name: caTrustConfigMapName
+      type: string
+      description: The name of the ConfigMap containing the CA bundle for TLS verification.
+      default: trusted-ca
+    - name: caTrustConfigMapKey
+      type: string
+      description: The key in the ConfigMap containing the CA bundle.
+      default: ca-bundle.crt
   results:
     - name: SOURCE_ARTIFACT
       description: The Trusted Artifact URI pointing to the artifact with the application source code with generated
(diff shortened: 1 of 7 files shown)


task/run-opm-command-oci-ta/0.1/run-opm-command-oci-ta.yaml
```diff
@@ -108,6 +112,175 @@ spec:
...
+
+          # Process each image and get its digest
+          for image in "${images[@]}"; do
+            # Fail on empty image references - indicates malformed catalog
+            if [[ -z "${image}" ]]; then
+              echo "Error: Found empty image reference in catalog. This indicates a malformed catalog entry." >&2
+              return 1
+            fi
+
+            # Skip images that already use digest format: <url>@sha256:...
```
yashvardhannanavati: This part of the logic could potentially utilize the `parse_image_url` function here: https://github.com/konflux-ci/konflux-test/blob/main/test/utils.sh#L190

provided you decide to use the konflux-test image

task/run-opm-command-oci-ta/0.1/tests/test-convert-tags-to-digests.yaml
```diff
@@ -0,0 +1,270 @@
+---
+apiVersion: tekton.dev/v1
```
yashvardhannanavati: this is amazing!

task/run-opm-command-oci-ta/0.1/run-opm-command-oci-ta.yaml
rauhersu: The new changes use the parse_image_url provided by utils.sh.

yashvardhannanavati (approved):

yashvardhannanavati:
looks good to me. thank you for your contribution! :)

PR #3116, opened 2026-01-05
Filter RPMs in hermetic multi-arch builds

brunoapimentel:
In multi-arch builds, a single prefetch is made so that the content from all target architectures is fetched. This brings the problem that the SBOM produced by each specific arch will have content from all other architectures.

For example, the SBOM for a `x86_64` image could look like this:

```json
{
  "name": "gzip",
  "purl": "pkg:rpm/redhat/gzip@1.12-1.el9?arch=s390x&checksum=sha256:50034ee6281864a218a5f3bc47de5afb434400fb8415907fd31d8351adbdc5a6&repository_id=ubi-9-for-s390x-baseos-rpms",
  "version": "1.12",
  "type": "library"
},
{
  "name": "gzip",
  "purl": "pkg:rpm/redhat/gzip@1.12-1.el9?arch=x86_64&checksum=sha256:c8b3e0414d55b1eedb0185a564ac6cb2368bee2fd5f995447d045f6a714488ac&repository_id=ubi-9-for-x86_64-baseos-rpms",
  "version": "1.12",
  "type": "library"
}
```

In order to solve this, we need to do two things:

1- Allow only the content from a single architecture to be available to the `buildah bud` command. This is being solved by a small script added in this commit that solves the problem in each of the versions of the buildah task.

2- Filter out the mixed content from the Hermeto SBOM. This is being done by the newer [version](https://github.com/konflux-ci/mobster/pull/307) of Mobster which was updated in a previous commit.

### DIFF
--- a/task/buildah/0.8/buildah.yaml
+++ b/task/buildah/0.8/buildah.yaml
@@ -647,6 +647,8 @@ spec:
         PREFETCH_ARCH="$(uname -m)"
         echo "$PREFETCH_ARCH" > /shared/prefetch-arch
 
+        echo "Prefetched content will be made available"
+
         cp -r "$(workspaces.source.path)/cachi2" /tmp/
         chmod -R go+rwX /tmp/cachi2
 
@@ -673,7 +675,6 @@ spec:
             -e 'H;1h;$!d;x' \
             -e 's@^\s*(run((\s|\\\n)+-\S+)*(\s|\\\n)+)@\1. /cachi2/cachi2.env \&\& \\\n    @igM' \
             "$dockerfile_copy"
-        echo "Prefetched content will be made available"
 
         prefetched_repo_for_my_arch="/tmp/cachi2/output/deps/rpm/$(uname -m)/repos.d/cachi2.repo"
         if [ -f "$prefetched_repo_for_my_arch" ]; then
--- a/task/buildah-min/0.8/buildah-min.yaml
+++ b/task/buildah-min/0.8/buildah-min.yaml
@@ -658,6 +658,8 @@ spec:
         PREFETCH_ARCH="$(uname -m)"
         echo "$PREFETCH_ARCH" > /shared/prefetch-arch
 
+        echo "Prefetched content will be made available"
+
         cp -r "$(workspaces.source.path)/cachi2" /tmp/
         chmod -R go+rwX /tmp/cachi2
 
@@ -684,7 +686,6 @@ spec:
             -e 'H;1h;$!d;x' \
             -e 's@^\s*(run((\s|\\\n)+-\S+)*(\s|\\\n)+)@\1. /cachi2/cachi2.env \&\& \\\n    @igM' \
             "$dockerfile_copy"
-        echo "Prefetched content will be made available"
 
         prefetched_repo_for_my_arch="/tmp/cachi2/output/deps/rpm/$(uname -m)/repos.d/cachi2.repo"
         if [ -f "$prefetched_repo_for_my_arch" ]; then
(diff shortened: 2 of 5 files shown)


task/buildah-remote-oci-ta/0.7/buildah-remote-oci-ta.yaml
```diff
@@ -748,6 +748,40 @@ spec:
...
+          # Clean up the components from non-matching architectures from the SBOM
+
+          jq ".components |= map(
+          select(
+              # If a component is not of pkg:rpm type, keep it
+              (.purl | startswith(\"pkg:rpm\") | not) or
+              # Otherwise, only keep it if the purl contains a matching architecture
+              (.purl | test(\"arch=$PREFETCH_ARCH(&|$)\"))
+            )
+          )" "$PREFETCH_DIR/bom.json" > "$PREFETCH_DIR/filtered-bom.json"
```
chmeliik: Minor suggestion to avoid injecting the variable directly into the jq code and having to escape double-quotes

```suggestion
          jq --arg prefetch_arch "$PREFETCH_ARCH" '.components |= map(
          select(
              # If a component is not of pkg:rpm type, keep it
              (.purl | startswith("pkg:rpm") | not) or
              # Otherwise, only keep it if the purl contains a matching architecture
              (.purl | test("arch=\($prefetch_arch)(&|$)"))
            )
          )' "$PREFETCH_DIR/bom.json" > "$PREFETCH_DIR/filtered-bom.json"
```

brunoapimentel:
@chmeliik @mmorhun Do you think this work justifies bumping up the buildah task version?

task/buildah-remote/CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ If that's not something you ever plan to do, consider removing this section.
...
 
+## 0.8.2
+
+### Changed
+
+- The task now makes sure that only RPMs that match the architecture being built are
+  passed to the `buildah bud` command. It also removes the same packages from the
+  Hermeto SBOM to more accurately represent the build.
+  This change should be a noop for this task, but it was added here so that the
+  auto-generated `buildah-remote` task would benefit from it.
```
chmeliik: buildah-remote(-oci-ta) should probably have slightly different text 😄

chmeliik (commented):
LGTM, small nitpicks

task/buildah/0.8/buildah.yaml
```diff
@@ -645,6 +645,24 @@ spec:
...
         chmod -R go+rwX /tmp/cachi2
+
+        # In case RPMs were prefetched and this is a multi-arch build,
+        # clean up the packages that do not match the architecture being built
+        RPM_PREFETCH_DIR="/tmp/cachi2/output/deps/rpm"
+        if [ -d "$RPM_PREFETCH_DIR" ] && [ "$(find $RPM_PREFETCH_DIR | wc -l)" -gt 1 ]; then
+          echo "Removing prefetched RPMs from non-matching architectures"
+          PREFETCH_ARCH="$(uname -m)"
+          for path in "$RPM_PREFETCH_DIR"/*; do
+            [[ -e "$path" ]] || break
```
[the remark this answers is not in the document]
brunoapimentel: This was forced by Checkton. I just abided :dark_sunglasses:

task/buildah/0.8/buildah.yaml
```diff
@@ -1182,6 +1186,7 @@ spec:
         --image-digest "$IMAGE_DIGEST"
         --parsed-dockerfile-path "/shared/parsed_dockerfile.json"
         --base-image-digest-file "/shared/base_images_digests"
+        --arch "$PLATFORM"
```
chmeliik: This should be removed I think

task/buildah-remote/CHANGELOG.md
chmeliik: I still see the same content

PR #3125, opened 2026-01-09
Remove sast-coverity-check from pko pipeline

eqrx:
Check assumes that image is built from Dockerfile, which it isn't.


MartinBasti:
sast-coverity was removed from all pipelines in different PR, no need to resolve conflicts, closing

PR #3259, opened 2026-02-18
feat(buildah-oci-ta): add INCLUDE_PREFETCH_SBOM and INCLUDE_SOURCE_SBOM params

polasudo:
## Summary

Add optional params `INCLUDE_PREFETCH_SBOM` and `INCLUDE_SOURCE_SBOM` to the `buildah-oci-ta` task to control whether prefetch (Cachi2/Hermeto) and source SBOM are included in the final merged SBOM. Both default to `"true"` for full backward compatibility.

## Problem

In **Yarn workspace monorepos**, Hermeto/Cachi2 resolves the root `yarn.lock` when prefetching for any workspace path, producing a prefetch SBOM that covers **all workspaces** — not just the target. This is a known limitation of Yarn workspace resolution in Hermeto (not a general monorepo behavior).

For context: RHDH productizes dozens of Backstage plugins from a shared monorepo. All plugins share a single `yarn.lock` at the workspace root. When Hermeto prefetches for one plugin (e.g. `distgit/containers/rhdh-hub`), it resolves the entire lockfile, producing an SBOM with dependencies from every sibling workspace. Each plugin image then ships an SBOM containing dependencies it does not actually use, causing:

- **~16× CVE duplication** across plugin images (same CVE reported in plugins that don't use the affected dep)
- **SBOM noise** and inflated package counts (6,344 packages vs 5,247 actual)
- **ProdSec ticket churn** — false findings opened for build/dev deps that aren't runtime risks

**Alternative approaches considered:**
- **Separate yarn.locks per plugin** — not viable; Yarn workspaces use a single lockfile by design, splitting requires unsupported manual derivation
- **Hermeto per-workspace resolution** — ideal long-term fix, but not currently supported (potential feature request)
- **Disable prefetch entirely** — would lose hermetic build guarantees

## Solution

Allow pipelines to opt out of prefetch and/or source SBOM inclusion when the prefetch SBOM is known to be inaccurate for the specific artifact:

- `INCLUDE_PREFETCH_SBOM=false` — skip copying and merging `sbom-prefetch.json`
- `INCLUDE_SOURCE_SBOM=false` — skip syft source scan and its merge into the final SBOM

Hermetic builds still run normally — only the SBOM merge step is affected. The build environment remains pristine and isolated.

## Conforma Policy Gating

Per @chmeliik's [requirement](https://github.com/konflux-ci/build-definitions/pull/3259#issuecomment-2661613827), usage of these params is gated by Conforma policy:

**Policy PR:** [conforma/policy#1674](https://github.com/conforma/policy/pull/1674)

The policy denies release if `INCLUDE_PREFETCH_SBOM=false` or `INCLUDE_SOURCE_SBOM=false` is detected in the PipelineRun attestation. Pipelines that legitimately need `false` (e.g. RHDH's Yarn monorepo) must configure an explicit policy exception. This follows the same pattern as existing `PRIVILEGED_NESTED` and `ADD_CAPABILITIES` enforcement.

## Evidence

RHDH hub build comparison on Konflux (`feature/sbom-pruner-buildah-oci-ta-test` branch):

| Configuration | Original SBOM packages | Notes |
|---|---|---|
| `INCLUDE_PREFETCH_SBOM=true`, `INCLUDE_SOURCE_SBOM=true` | 6,344 | Includes all prefetch workspace deps |
| `INCLUDE_PREFETCH_SBOM=false`, `INCLUDE_SOURCE_SBOM=false` | 5,247 | Image-scanned packages only (~17% noise removed) |

Note: `INCLUDE_SOURCE_SBOM` has no effect when `HERMETIC=true` (source scan only runs for non-hermetic builds), so the package count difference is entirely from prefetch SBOM exclusion.

## Backward compatibility

- Both params default to `"true"` — existing pipelines unchanged
- No breaking changes
- Conforma policy blocks `"false"` usage unless explicitly excepted

## Changes (1 file)

1. **Param declarations** — two new string params with `default: "true"`
2. **Env passthrough** — wires params into the build step
3. **Build step** — gates `cp` of `sbom-prefetch.json` on `INCLUDE_PREFETCH_SBOM`
4. **Syft step** — adds `INCLUDE_SOURCE_SBOM` check to source scan condition
5. **Mobster step** — gates both `--from-syft` and `--from-hermeto` merge paths

Related: [RHIDP-11640](https://issues.redhat.com/browse/RHIDP-11640)

[RHIDP-11640]: https://redhat.atlassian.net/browse/RHIDP-11640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chmeliik (commented):
> In monorepo scenarios, prefetch-dependencies produces a workspace-wide SBOM that gets merged into every plugin image.

I don't think this is true. Prefetch produces an SBOM that contains whatever you configured to be prefetched. If it contains every dependency in the repo, that's probably because you have the same global prefetch configuration in every build pipeline.

chmeliik:
> Can we test out these proposed changes in a staged environment to see if the output is what we expect?

You can try out modified tasks like this: https://github.com/konflux-ci/build-definitions?tab=readme-ov-file#pushing-task-bundles-for-inner-loop-development

But to clarify, these parameters would be the very last resort if there's no other option to solve this.

> plugins are part of the same workspace and they would need to have their yarn.locks derived and filtered from the workspace yarn.lock

It sounds like you want separate sets of dependencies per plugin and yarn's workspace system isn't compatible with that. Should you really be using workspaces?

nickboldt:
I've opened https://issues.redhat.com/browse/PSX-1037 to track the workarounds we'll have to do if this isn't implemented. 

Our choices are:
* Fork the buildah-oci.ta task to omit the prefetched content, which we don't want in the SBOM
* start building non-hermetically, leading to the issues with non-deterministic builds

Or, if feasible, we could try creating a SECOND SBOM using a custom task, and then upload/apply that filtered SBOM as a replacement for the default one Konflux creates w/ mobster. I'm not sure if that would be allowed as it could look like "tampering" if something stores the SBOM's digest to ensure it's unchanged.

eskultety:
A tiny comment on the PR description:
Doesn't PREFETCH_SBOM and SOURCE_SBOM in Konflux's context mean exactly the same thing? I think we need to be aligned with the correct terminology: https://www.cisa.gov/resources-tools/resources/types-software-bill-materials-sbom. There's no such thing as a prefetch SBOM, Hermeto produces a source SBOM. My understanding of what AI's notion of SOURCE_SBOM is here is an analyzed SBOM, because IIUC we only run Syft on the produced artifacts to enhance the source SBOM from Hermeto to make it a more complex and more accurate byproduct.

polasudo:
Closing this issue as resolved PR has been Merged https://github.com/hermetoproject/hermeto/pull/1480

PR #3295, opened 2026-02-27
fix(buildah): preserve label values containing equals signs in labels.json

pigeio:
### Summary

Fixes a silent truncation bug during `labels.json` generation where label values containing `=` signs were incorrectly truncated, losing data.

### Root Cause

The `jq` expression used to convert `key=value` label pairs into JSON uses `split("=")` to separate the key from the value. 
However, `split("=")` splits the string on *every* equals sign, producing an array of all segments. The old code then used `.[1]` to select only the second element, which silently discarded anything after the first `=` in the value.
**Example of the bug:**
A label like `description=foo=bar=baz` would produce:
`{"description": "foo"}`  (data lost)

### The Fix

This PR replaces `.[1]` with `.[1:] | join("=")` to re-join all segments after the key, preserving the full original value.
The same example now correctly produces:
`{"description": "foo=bar=baz"}` 
This fix has been applied consistently to all five buildah task variants:
- [buildah.yaml]
- [buildah-min.yaml]
- [buildah-remote.yaml]
- [buildah-oci-ta.yaml]
- [buildah-remote-oci-ta.yaml]

### How this was found

This was discovered during a manual code audit of the Buildah Tekton tasks to ensure edge case robustness in the scripts.

### DIFF
--- a/task/buildah/CHANGELOG.md
+++ b/task/buildah/CHANGELOG.md
@@ -19,6 +19,18 @@ If that's not something you ever plan to do, consider removing this section.
   locally stored OCI image index. This is not a user-facing change. It is for
   optimizing buildah-remote task.
 
+## 0.9.1 - 2026-03-10
+
+### Fixed
+
+- Preserve label values containing equals signs in labels.json
+
+## 0.8.4 - 2026-03-10
+
+### Fixed
+
+- Preserve label values containing equals signs in labels.json
+
 ## 0.9.1
 
 ### Changed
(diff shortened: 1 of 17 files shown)


task/buildah/CHANGELOG.md
```diff
@@ -11,6 +9,25 @@ If that's not something you ever plan to do, consider removing this section.
 
 *Nothing yet.*
 
+## 0.9.1 - 2026-03-10
```
MartinBasti: was this autogenerated?

pigeio (commented):
Thanks for the review @MartinBasti! All three issues are fixed in the latest push. Let me know if anything else needs updating, or if it's ready for an /ok-to-test.

slimreaper35:
Buildah tasks moved to https://github.com/konflux-ci/build-pipeline-tasks

Issue #1264, opened 2024-08-07
Add `zstd:chunked` compression to built images.


ralphbean:
There's a long arc of changes in container tooling to support `zstd:chunked` compression that supports better compression and better download capabilities in clients. However, old versions of docker (pre-2023) don't support it, which has held up adoption.

The request here is to add `zstd:chunked` compression *in addition* to the normal `gzip` compression for images when we push an OCI image index. To illustrate, for an ordinary multi-arch image index, you would have 4 entries, one for each arch. The idea here is to have 8 entries, one for each arch with gzip compression, and one for each arch with zstd:chunked compression.

**Idea number 1**: I looked, and `buildah manifest push` has a `--add-compression zstd:chunked` option which (I think) will automatically add the extra image manifests on push.

However, I think we might have issues with image provenance. In a Konflux multi-arch build pipeline, we build each arch separately in a buildah task. Those produce image manifests, and their digests are recorded and provenance information is associated with them. We then produce the image index, record its digest, and record stuff there too.

When we validate an image at release time, we ensure that we can find provenance for both the image index as well as all of the referenced image manifests.

If we "magically" add an extra four manifests when we create the image index, then our policy check will fail unless their digests and pullspecs are somehow also exposed, so that chains can attest to their origin.

**Idea number 2**: Instead, perhaps our buildah tasks should be modified to first push the image with gzip compression, and then call `buildah push` again with a `--compression-format zstd:chunked` option to generate the second image manifest then, earlier in the process. Our buildah task would need to expose _both_ digests.. hm.

**Idea number 3**: Perhaps we should add a new intermediary task that accepts the digest of the image-manifest producing buildah task as a param, pulls it, and then re-pushes it with `zstd:chunked`. This new task would expose the pullspec and digest of the zstd:chunked manifest as a result, which will induce chains to record provenance for it. The task that builds the image index would then take 8 inputs instead of 4. All 8 would be explicitly added to the image index.

arewm:
**Idea 1**: It should be possible to modify the image index generation to add this functionality (optionally?) if desired. I added functionality last week to expose all image manifests linked to Chains: https://github.com/konflux-ci/build-definitions/pull/1204/files. We will need to be sensitive to the Tekton results limitations. Some users were hitting the limitation for the buildah task when we referenced 3-4 base images in the BASE_IMAGE_DIGESTS. This isn't a direct comparison because the quota for results was split across multiple task steps for buildah. 

[...]

PR #3299, opened 2026-03-01
buildah: support compression-format and force-compression during push

aryanmishracodes:
Adds support for the following parameters to the buildah Task push step:

- COMPRESSION_FORMAT
- FORCE_COMPRESSION

These parameters are now applied to both push operations:
1. The unique TaskRun tag push
2. The final image push with digest

## Details

- Introduces --compression-format and --force-compression flags when invoking `buildah push`
- Applies the flags conditionally to preserve existing behavior
- Regenerates Trusted Artifact (OCI-TA) variants
- Rebuilds kustomize-generated manifests

## Backward Compatibility

- Default values preserve existing behavior
- No changes occur unless parameters are explicitly set

Fixes: #3188

### DIFF
--- a/task/buildah-min/0.8/buildah-min.yaml
+++ b/task/buildah-min/0.8/buildah-min.yaml
@@ -170,6 +170,14 @@ spec:
       oci (default) or docker.
     name: BUILDAH_FORMAT
     type: string
+  - default: ""
+    description: Compression format for image layers (e.g., gzip, zstd, zstd:chunked)
+    name: COMPRESSION_FORMAT
+    type: string
+  - default: "false"
+    description: Force recompression of all layers including base image layers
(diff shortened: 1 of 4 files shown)


chmeliik (commented):
Is this compatible with #1264?

aryanmishracodes:
Thanks for the feedback.

You’re right, I have not yet validated this change against multi-platform builds or the full build pipeline introduced around [#1264](https://github.com/konflux-ci/build-definitions/issues/1264).

So far, validation has only been via CI and single-task execution.

I will:
- Review the interaction with image index generation
- Test compatibility with multi-arch builds
- Validate behavior when parameters are unset to confirm full backward compatibility

If preferred, I can mark this PR as draft until compatibility testing is completed.

PR #3405, opened 2026-04-02
feat(ta): set 4Gi memory request/limit for use-trusted-artifact step

sfowl:
The create-trusted-artifact step already sets a 3Gi limit via the generator code. Apply similar resource limits to the use-trusted-artifact step to ensure consistent memory allocation.

The -min variant patches remove the explicit computeResources so those tasks inherit the smaller stepTemplate defaults instead.

Related: https://github.com/konflux-ci/konflux-sast-tasks/pull/103


MartinBasti (dismissed):

eqrx:
Is this still active? :)

PR #3420, opened 2026-04-10
task(fbc-fips-check): right-size resources from fleet p95 analysis

smodak-rh:
## Why

The task still used **coarse defaults** (for example **12Gi** memory and **1** / **8** CPU on heavy steps) that do not match **measured** usage. Internal **Konflux fleet** analysis (tasks-and-steps-resource-analyzer, **15-day** window) gives per-step **p95**-style signals per cluster; we use that to **right-size** requests while keeping **memory request = memory limit** and **CPU requests only** (no CPU limits).

## What

- **`task/fbc-fips-check/0.1/fbc-fips-check.yaml`:** Set **`computeResources`** from a **fleet p95 envelope** (max across clusters of per-step **`mem_p95_mb`** / **`cpu_p95`** from the analyzer, then rounded up):
  - **`get-unique-related-images`:** **`4Gi`** / **`705m`** (memory req = limit).
  - **`fips-operator-check-step-action`:** **`512Mi`** / **`2059m`** (memory req = limit).
  - **`parse-images-processed-result`:** **`64Mi`** / **`50m`** (memory req = limit).
- **`task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml`:** Same resources for the shared steps; add **`use-trusted-artifact`** with **`64Mi`** / **`50m`** (memory req = limit). Regenerate from base with **`./hack/generate-ta-tasks.sh`** when changing the base task; re-apply **`use-trusted-artifact`** resources if the generator drops them.
- **No behavior change** to scripts, params, or step order — **only** resource requests/limits as above.

## Follow-up

- If **OOM** or **CPU starvation** shows up on rare bundles, bump requests using **max** or higher percentiles from the same analyzer (p95 is not worst-case).

### DIFF
--- a/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
+++ b/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
@@ -47,12 +47,6 @@ spec:
       args:
         - use
         - $(params.SOURCE_ARTIFACT)=/var/workdir/source
-      computeResources:
-        requests:
-          memory: 64Mi
-          cpu: "50m"
-        limits:
-          memory: 64Mi
     - name: get-unique-related-images
       image: quay.io/konflux-ci/konflux-test:v1.4.50@sha256:8b9c7be18bef49f60ac9926830174072cc2010c620c7b3d0a5bb53ba02fcf71d
       env:
--- a/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
+++ b/task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
@@ -306,10 +306,10 @@ spec:
         echo "${images_processed_template/\[%s]/[$digests_processed_string]}" >"/tekton/home/images_processed.txt"
       computeResources:
         limits:
-          memory: 4Gi
+          memory: 6Gi
         requests:
           cpu: 705m
-          memory: 4Gi
+          memory: 6Gi
       securityContext:
         capabilities:
           add:
@@ -320,10 +320,10 @@ spec:
           value: $(params.MAX_PARALLEL)
       computeResources:
         limits:
-          memory: 512Mi
+          memory: 10Gi
         requests:
           cpu: 2059m
-          memory: 512Mi
+          memory: 10Gi
       ref:
         params:
           - name: url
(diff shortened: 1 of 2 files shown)


smodak-rh:
For reviewers, I have posted additional details regarding the following points directly in the Jira (https://redhat.atlassian.net/browse/KONFLUX-13033):

1. Normal (fbc-fips-check) vs oci-ta (fbc-fips-check-oci-ta) — how we used the data
2. Question for reviewers (use-trusted-artifact vs non-oci-ta)
3. Suggestion:
- [ ] Prefer a follow-up PR that extends the generator to attach minimal computeResources to the injected use-trusted-artifact step (small, localized change in ta.go + tests), rather than hand-editing fbc-fips-check-oci-ta.yaml after each ./hack/generate-ta-tasks.sh run.
Cc: @jhutar

jhutar (approved):
/lgtm

task/fbc-fips-check-oci-ta/0.1/fbc-fips-check-oci-ta.yaml
```diff
@@ -306,11 +306,10 @@ spec:
         echo "${images_processed_template/\[%s]/[$digests_processed_string]}" >"/tekton/home/images_processed.txt"
       computeResources:
         limits:
-          cpu: "1"
-          memory: 12Gi
+          memory: 4Gi
```
yashvardhannanavati, line 309: I did some back of the napkin calculation and asked Claude to estimate. The CPU limit change looks good to me. For get-unique-related-images , I would prefer somewhere between 6-8Gi since it renders the RH catalog image (could be using around 4Gi ) and some leeway to process the FBC fragment itself. For fips-operator-check-step-action , I think somewhere around 12Gi because it supports parallel execution and images are pulled locally. Each image could be between 1-2Gi for large

yashvardhannanavati (changes requested):

yashvardhannanavati:
For posterity, this PR is associated to [KONFLUX-13279](https://redhat.atlassian.net/browse/KONFLUX-13279)

[KONFLUX-13279]: https://redhat.atlassian.net/browse/KONFLUX-13279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

yashvardhannanavati (approved):

PR #3421, opened 2026-04-10
Add enable-package-registry-proxy parameter for prefetch

taylormadore:
This parameter controls whether package registry requests are routed through the package registry proxy when prefetching dependencies. Includes migration scripts to automatically wire the parameter into existing pipelines.

Related to: https://github.com/konflux-ci/konflux-build-cli/pull/90

### DIFF
--- a/pipelines/template-build/template-build.yaml
+++ b/pipelines/template-build/template-build.yaml
@@ -61,6 +61,9 @@ spec:
     - name: enable-cache-proxy
       description: Enable cache proxy configuration
       default: "false"
+    - name: enable-package-registry-proxy
+      description: Use the package registry proxy when prefetching dependencies
+      default: "true"
   tasks:
     - name: init
       params:
@@ -90,6 +93,8 @@ spec:
       params:
         - name: input
           value: $(params.prefetch-input)
+        - name: enable-package-registry-proxy
+          value: "$(params.enable-package-registry-proxy)"
       runAfter:
         - clone-repository
       taskRef:
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -134,6 +134,9 @@ spec:
   - default: "false"
     description: Enable cache proxy configuration
     name: enable-cache-proxy
+  - default: "true"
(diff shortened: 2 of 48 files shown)


pipelines/template-build/template-build.yaml
```diff
@@ -61,6 +61,9 @@ spec:
     - name: enable-cache-proxy
       description: Enable cache proxy configuration
       default: "false"
+    - name: enable-package-registry-proxy
```
MartinBasti, line 64: How does this work, when registry proxy service isn't available?
Is it a core service?

taylormadore: The simple answer is that Hermeto will fail the prefetch and yes, Nexus is supposed to be core infra AFAIK.

The extended explanation:

Whether the package registry proxy is used is controlled by two flags:
- **enable-package-registry-proxy**: a pipeline-level parameter that defaults to true. This is the per-user/per-pipeline opt-in
- **allow-package-registry-proxy**: a cluster-level setting in the konflux-info/cluster-config ConfigMap. It indicates whether the package registry proxy can/should be used globally on the cluster.

Both must be true for the proxy to be used. If the proxy service becomes unavailable, Hermeto would fail to fetch packages from it. However, the cluster administrator can respond by toggling allow-package-registry-proxy to false, which disables the proxy for all pipelines on the cluster without requiring any pipeline changes.

We decided it made more sense to fail rather than fallback if the proxy was unavailable because:
1. It's a security control
2. Conforma will eventually enforce that the proxy be used

MartinBasti (dismissed):

taylormadore:
Rebased on main

taylormadore:
This is going to need a rebase to switch this new version of the prefetch-dependencies task to 0.3.2 assuming that https://github.com/konflux-ci/build-definitions/pull/3440 gets merged first.

a-ovchinnikov (approved):

MartinBasti (approved):

slimreaper35 (approved):

amisstea (approved):

PR #3439, opened 2026-04-15
fix: enable helm charts with OCI-only dependencies

solacelost:
For charts that have OCI dependencies at all, attempting to add the repo fails as there is no method to do that for OCI artifact-packaged Helm charts. Additionally, if all dependencies are OCI dependencies, then `helm repo update` fails. Set `|| true` there as `helm dependency build .` should properly fail if helm is unable to resolve dependencies from available, configured repositories.

I don't see any applicable tests at all for this task. If there's anything you'd like added here, I'm happy to do it.

### DIFF
--- a/task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
+++ b/task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
@@ -283,16 +283,25 @@ spec:
           echo "Building Helm dependencies from Chart.lock..."
 
           # Add required repositories from Chart.yaml dependencies
-          yq -r '.dependencies[].repository' Chart.yaml | sort -u | while read -r repo_url; do
+          http_repo_used=false
+          while read -r repo_url; do
             if [ -n "$repo_url" ]; then
-              repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
-              echo "Adding repository: $repo_name ($repo_url)"
-              helm repo add "$repo_name" "$repo_url" || true
+              proto=$(echo "$repo_url" | cut -d: -f1)
+              if [ "$proto" = "oci" ]; then
+                echo "Skipping adding helm repository for OCI dependency: $repo_url"
+              else
+                http_repo_used=true
+                repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
+                echo "Adding repository: $repo_name ($repo_url)"
+                helm repo add "$repo_name" "$repo_url" || true
+              fi
             fi
-          done
+          done < <(yq -r '.dependencies[].repository' Chart.yaml | sort -u)
 
           # Update repository index
-          helm repo update
+          if [ "$http_repo_used" = "true" ]; then
+            helm repo update
+          fi
 
           # Build dependencies
           helm dependency build .
--- a/task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
+++ b/task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
@@ -273,14 +273,17 @@ spec:
           # Add required repositories from Chart.yaml dependencies
           yq -r '.dependencies[].repository' Chart.yaml | sort -u | while read -r repo_url; do
             if [ -n "$repo_url" ]; then
-              repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
-              echo "Adding repository: $repo_name ($repo_url)"
-              helm repo add "$repo_name" "$repo_url" || true
+              proto=$(echo "$repo_url" | cut -d: -f1)
+              if [ "$proto" != "oci" ]; then
+                repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
+                echo "Adding repository: $repo_name ($repo_url)"
+                helm repo add "$repo_name" "$repo_url" || true
+              fi
             fi
           done
 
           # Update repository index
-          helm repo update
+          helm repo update || true
 
           # Build dependencies
           helm dependency build .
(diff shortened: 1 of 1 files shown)


solacelost:
For reference, here's an example Chart.yaml that failed to build:

https://git.jharmison.dev/game-servers/icarus/src/branch/main/charts/icarus/Chart.yaml

And you can see the pipeline logs, if you self-register an account, here:
https://konflux.jharmison.dev/ns/tenant-game-servers/applications/icarus/pipelineruns/icarus-chart-on-push-gs5zq

Here's the relevant text, though:

```
STEP-USE-TRUSTED-ARTIFACT
  Using token for registry.jharmison.com
  Restored artifact registry.jharmison.com/game-servers/helm-icarus@sha256:c658209e0ac945ebd28aed6e8b73884622c0edf047e69b2716504b405b0812f3 to /var/workdir/source
  
  
 
STEP-PACKAGE-AND-PUSH
  + VALUES_FILES_ARGS=values.yaml
  + cd source/charts/icarus
  + REPO=registry.jharmison.com/game-servers/helm-icarus
  + export chart_name=helm-icarus
  + chart_name=helm-icarus
  + yq -i '.name = strenv(chart_name)' Chart.yaml
  + '[' '[]' '!=' '[]' ']'
  + '[' -n 1.0.0 ']'
  + echo 'Using provided chart version: 1.0.0'
  Using provided chart version: 1.0.0
  + chart_version=1.0.0
  + '[' -f Chart.yaml ']'
  + yq -e .dependencies Chart.yaml
  + echo 'Building Helm dependencies from Chart.lock...'
  Building Helm dependencies from Chart.lock...
  + yq -r '.dependencies[].repository' Chart.yaml
  + sort -u
  + read -r repo_url
  + '[' -n oci://registry.jharmison.com/game-servers ']'
  ++ echo oci://registry.jharmison.com/game-servers
  ++ sed 's|https://||'
  ++ sed 's|http://||'
  ++ sed 's|/.*$||'
  ++ sed 's|\.|_|g'
  + repo_name=oci:
  + echo 'Adding repository: oci: (oci://registry.jharmison.com/game-servers)'
  Adding repository: oci: (oci://registry.jharmison.com/game-servers)
  + helm repo add oci: oci://registry.jharmison.com/game-servers
  Error: looks like "oci://registry.jharmison.com/game-servers" is not a valid chart repository or cannot be reached: failed to perform "FetchReference" on source: invalid reference
  + true
  + read -r repo_url
  + helm repo update
  Error: no repositories found. You must add one before updating
  ```

File: task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml [hunk 1]
```diff
@@ -273,14 +273,17 @@ spec:
...
+                repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
+                echo "Adding repository: $repo_name ($repo_url)"
+                helm repo add "$repo_name" "$repo_url" || true
+              fi
             fi
           done
 
           # Update repository index
-          helm repo update
+          helm repo update || true
```
yftacherzog: I think that instead of masking any failure here, we should check if there's nothing to updated (i.e. all `oci`) and then skip this line.

yftacherzog (changes requested):
You're right about the tests. We need to move this task to another repo and have it re-implemented in language that provides proper test framework.

For the specific change, I left a comment.

task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
solacelost: I was just following convention I saw elsewhere. The existing `helm repo add` line also did `|| true` before I got here. I'm not sure of the condition it was guarding against there, but it's very possible it was this `oci://` dependency issue. Do you think I should pull it out of there, or should we leave that dragon alone?

solacelost:
Had to restructure a bit to address the comment as the subshell resulting from the piped `while read` would have prevented communicating that condition back out of the subshell to the main script, so this is a little more disruptive than I like. Hopefully works for you!

task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
```diff
@@ -271,16 +271,23 @@ spec:
...
-              repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
-              echo "Adding repository: $repo_name ($repo_url)"
-              helm repo add "$repo_name" "$repo_url" || true
+              proto=$(echo "$repo_url" | cut -d: -f1)
+              if [ "$proto" != "oci" ]; then
+                http_repo_used=true
+                repo_name=$(echo "$repo_url" | sed 's|https://||' | sed 's|http://||' | sed 's|/.*$||' | sed 's|\.|_|g')
+                echo "Adding repository: $repo_name ($repo_url)"
+                helm repo add "$repo_name" "$repo_url" || true
+              fi
```
yftacherzog, line 285: Can we add an `else` with a similar print when skipped?

task/build-helm-chart-oci-ta/0.3/build-helm-chart-oci-ta.yaml
```diff
@@ -271,16 +271,23 @@ spec:
...
             fi
-          done
+          done < <(yq -r '.dependencies[].repository' Chart.yaml | sort -u)
 
           # Update repository index
-          helm repo update
+          if $http_repo_used; then
```
yftacherzog: Can you address the AI comment about the boolean check? E.g. replace with `[ "$http_repo_used" = "true" ]`.

solacelost:
I don't really agree with the `true`/`false` thing, since it was explicitly defined statically in the only execution path it was checked, and I think it's actually harder to read now. I implemented it anyways 🤷‍♂. The other suggestion was sensible enough.

yftacherzog (dismissed):

yftacherzog:
> I don't really agree with the `true`/`false` thing, since it was explicitly defined statically in the only execution path it was checked, and I think it's actually harder to read now. I implemented it anyways 🤷‍♂. The other suggestion was sensible enough.

Thank you!

This task was initially developed for kind of own use, but lately getting some more users, and the lack of maintainability becomes an issue. I've been playing yesterday with having a [python-based implementation](https://github.com/konflux-ci/tools/pull/687) that could help with that.

Hopefully, we'll be able to get some progress with that and have an easier to maintain task.

solacelost:
I saw in the linter that I'd typo'd a variable name in the commit that moved to the value check. I went ahead and squashed everything down while fixing that.

yftacherzog (approved):

solacelost:
Any update?

yftacherzog:
> Any update?

I did not realize this is still waiting. Thank you for letting me know.

PR #3480, opened 2026-04-27
replace appstudio-utils with task-runner

fmudry:
`appstudio-utils` image will be decommissioned soon and the replacement image is the `task-runner`

### Testing
- Changes in the `.tekton` folder will be tested by the CI.
- Changes in the `task-generator` are safe to do, since the actual tasks already use the `task-runner`.

### DIFF
--- a/.tekton/tasks/e2e-test.yaml
+++ b/.tekton/tasks/e2e-test.yaml
@@ -96,7 +96,10 @@ spec:
       - name: TEKTON_CHAINS_NS
         value: "openshift-pipelines"
     - name: capture-log
-      image: quay.io/konflux-ci/pull-request-builds:appstudio-utils-$(params.revision)
+      image: quay.io/konflux-ci/task-runner:1.6.0@sha256:1abfe4e50d4e961d0fd9790202565f93ee650fe8dfc50932c94989acba10485f
+      securityContext:
+        runAsUser: 0
+        runAsNonRoot: false
       script: |
         #!/bin/bash
         set -euo pipefail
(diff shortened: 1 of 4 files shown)


fmudry:
The `capture-log` in e2e tests failed, due to permission issues. Fixed in the last commit.
```
STEP-CAPTURE-LOG

Storing e2e-tests pod log...
/tekton/scripts/script-1-w9zbz: line 5: /workspace/artifact-dir/e2e-tests.log: Permission denied
```

mkosiarc (approved):

.tekton/tasks/e2e-test.yaml
```diff
@@ -96,7 +96,10 @@ spec:
       - name: TEKTON_CHAINS_NS
         value: "openshift-pipelines"
     - name: capture-log
-      image: quay.io/konflux-ci/pull-request-builds:appstudio-utils-$(params.revision)
+      image: quay.io/konflux-ci/task-runner:1.6.0@sha256:1abfe4e50d4e961d0fd9790202565f93ee650fe8dfc50932c94989acba10485f
+      securityContext:
+        runAsUser: 0
```
mkosiarc, line 101: If you were inclined to do it, you could leave a bit of information on why is this needed. E.g we are not really changing the scope of permissions here, it's just the image is setup differently and we need to explicitly set root here, right?

fmudry: Yes, exactly. I'll add it to the commit message

PR #3512, opened 2026-05-07
task(git-clone): set resource limits for clone and symlink-check steps

smodak-rh:
## Problem

The `git-clone`, `git-clone-oci-ta`, `git-clone-oci-ta-min`, and
`pnc-prebuild-git-clone-oci-ta` tasks currently have no resource
requests or limits set for the `clone` and `symlink-check` steps
(`computeResources: {}`), making them effectively unbound.

## Analysis

Resource usage was analyzed using **72,702 pod executions** of
`git-clone-oci-ta` across 9 production clusters (the OCI-TA dataset
is 42× larger than the base `git-clone` dataset of 1,738 pods and
is therefore used as the authoritative source for both tasks):

| Step | mem Max | mem P95 | cpu Max | cpu P95 |
|---|---|---|---|---|
| `clone` | 1,801 MiB | 894 MiB | 2,125m | 302m |
| `symlink-check` | 10 MiB | 5 MiB | 2m | ~0m |

## Sizing decision — why Max+5% for `clone` instead of P95+5%

The `create-trusted-artifact` step in `git-clone-oci-ta` already
carries **3Gi/1CPU** (Max+5%) to protect against OOM on large
monorepos (max observed: 3,051 MiB; P95/Max ratio: 127×). The same
large repos that drive the high memory usage in `create-trusted-artifact`
also require up to 1,801 MiB to clone. Setting `clone` to P95+5%
(960Mi) while `create-trusted-artifact` stays at Max (3Gi) would be
internally inconsistent and would cause clone-step OOM for those
tenants. For this reason, `clone` is also sized at Max+5% → **2Gi**.

For `symlink-check`, both P95+5% and Max+5% converge at the minimum
floor → **64Mi / 50m**.

## Changes

Only `task/git-clone/0.1/git-clone.yaml` is edited directly.
The remaining three task YAMLs are auto-generated:

| File | Change | Method |
|---|---|---|
| `git-clone/0.1/git-clone.yaml` | Add `computeResources` to `clone` and `symlink-check` | Direct edit |
| `git-clone-oci-ta/0.1/git-clone-oci-ta.yaml` | Inherits `clone` + `symlink-check` resources | `hack/generate-ta-tasks.sh` |
| `git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml` | Inherits `clone` + `symlink-check`; kustomize patch keeps `create-trusted-artifact` at 256Mi/100m | `hack/build-manifests.sh` |
| `pnc-prebuild-git-clone-oci-ta/0.1/pnc-prebuild-git-clone-oci-ta.yaml` | Inherits `clone` + `symlink-check`; `preprocessor` (512Mi) and `create-trusted-artifact` (3Gi) unchanged | `hack/build-manifests.sh` |

## Steps not changed

- `create-trusted-artifact`: already set to 3Gi/1CPU — confirmed correct by analysis (heavy tail ratio 127×)
- `preprocessor` in `pnc-prebuild-git-clone-oci-ta`: already set to 512Mi — no observability data available (zero pods in 15 days)

Jira: https://issues.redhat.com/browse/KONFLUX-13040

Signed-off-by: Subrata Modak <smodak@redhat.com>
Assisted-by: CursorAI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
+++ b/task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
@@ -148,10 +148,10 @@ spec:
   steps:
   - computeResources:
       limits:
-        memory: 2Gi
+        memory: 256Mi
       requests:
-        cpu: 350m
-        memory: 2Gi
+        cpu: 100m
+        memory: 256Mi
     env:
     - name: HOME
       value: $(params.userHome)
--- a/task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
+++ b/task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
@@ -146,7 +146,13 @@ spec:
   - description: The precise URL that was fetched by this Task.
     name: url
   steps:
-  - env:
+  - computeResources:
+      limits:
+        memory: 2Gi
+      requests:
+        cpu: 350m
+        memory: 2Gi
+    env:
     - name: HOME
       value: $(params.userHome)
     - name: PARAM_URL
@@ -346,7 +352,13 @@ spec:
       readOnly: true
     - mountPath: /var/workdir
       name: workdir
-  - env:
+  - computeResources:
+      limits:
+        memory: 64Mi
+      requests:
+        cpu: 50m
+        memory: 64Mi
+    env:
     - name: PARAM_ENABLE_SYMLINK_CHECK
       value: $(params.enableSymlinkCheck)
     - name: CHECKOUT_DIR
--- a/task/git-clone-oci-ta-min/0.1/patch.yaml
+++ b/task/git-clone-oci-ta-min/0.1/patch.yaml
@@ -1,10 +1,23 @@
 - op: replace
   path: /metadata/name
   value: git-clone-oci-ta-min
-# kustomize build task/git-clone-oci-ta/0.1 | yq '.spec.steps.[].name' | nl -v 0 
+# kustomize build task/git-clone-oci-ta/0.1 | yq '.spec.steps.[].name' | nl -v 0 
 #      0	clone
 #      1	symlink-check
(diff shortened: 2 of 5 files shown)


task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
```diff
@@ -146,7 +146,13 @@ spec:
   - description: The precise URL that was fetched by this Task.
     name: url
   steps:
-  - env:
+  - computeResources:
+      limits:
+        memory: 2Gi
+      requests:
+        cpu: 350m
+        memory: 2Gi
```
MartinBasti: min variant of git clone cannot use 2GB of memory

MartinBasti (changes requested):
Min variant must be updated

task/git-clone-oci-ta-min/0.1/git-clone-oci-ta-min.yaml
smodak-rh: Thanks @MartinBasti for pointing that out. Have updated with min variant with new resource definition(s). Please review again.

MartinBasti (dismissed):

MartinBasti (approved):

smodak-rh:
@MartinBasti, any idea when it will be added to the merge queue?

smodak-rh:
Ping for a merge request.

chmeliik:
I see that this PR isn't yet approved in the Red Hat internal approval spreadsheet

smodak-rh:
@MartinBasti, we have approval on the sheet, can we please add this to the MERGE queue. Also there was a suggestion from "Eyal Edri" here: https://redhat-internal.slack.com/archives/C04F4NE15U1/p1779373198237929?thread_ts=1779372966.035709&cid=C04F4NE15U1

PR #3519, opened 2026-05-13
Fix the use of note variable

chandwanitulsi:


task/validate-fbc/0.1/validate-fbc.yaml
```diff
@@ -391,6 +392,7 @@ spec:
 
         if ! opm validate ."${conffolder}"; then
           echo "!FAILURE! - opm validate check failed."
+          note="Step extract-and-validate failed: opm validate check failed. For details, check Tekton task log."
```
yashvardhannanavati, line 395: If another subsequent check fails after this check, wouldn't that overwrite the `note` variable?

chandwanitulsi: Yes, that was the case earlier also and in the previous [PR](https://github.com/konflux-ci/build-definitions/pull/3519) that was closed I had collected all failure messages and showed those together but there was an issue with result limit.
With this I made note is set in every case so it does not remain undefined. The undefined issue was also partially fixed in https://github.com/konflux-ci/build-definitions/pull/3534 as the variable got defined in the start

PR #3523, opened 2026-05-14
feat: Migrate e2e tests from monorepo to build-definitions

flacatus:


flacatus:
cc @tisutisu @mmorhun

.tekton/tasks/e2e-test.yaml
```diff
@@ -36,29 +36,39 @@ spec:
...
-      # a la infra-deployment updates, when PRs merge in e2e-tests, PRs will be opened
-      # against build-definitions to update this tag
-      args: [
-        "--ginkgo.label-filter=build-templates-e2e",
-        "--ginkgo.junit-report=$(params.artifact_dir)/e2e-report.xml",
-        "--ginkgo.no-color",
-        "--ginkgo.v",
-        "--ginkgo.timeout=2h"
-      ]
+      image: quay.io/redhat-user-workloads/konflux-qe-team-tenant/konflux-e2e/konflux-e2e-tests:latest
```
tisutisu: Will it be good to avoid `latest` tag? CI can be broken by some change in e2e-tests repo, instead using a tag which is tied to a specific commit from e2e-tests repo?

.tekton/tasks/e2e-test.yaml
```diff
@@ -36,29 +36,39 @@ spec:
...
+        #!/bin/bash
+        set -euo pipefail
+        ginkgo build ./tests/build/
+        ginkgo \
+          --label-filter=build-templates-e2e \
+          --junit-report="${ARTIFACT_DIR}/e2e-report.xml" \
+          --no-color \
+          -v \
+          --timeout=2h \
+          ./tests/build/ 2>&1 | tee "${ARTIFACT_DIR}/e2e-tests.log"
```
tisutisu, line 65: will it be good to add `--fail-on-empty` flag with this command, helps to fail the command when no spec runs

.tekton/tasks/e2e-test.yaml
tisutisu: we are not using anything from the e2e-tests repo here, can we use a different image like `quay.io/konflux-ci/task-runner:1.6.0@sha256:1abfe4e50d4e961d0fd9790202565f93ee650fe8dfc50932c94989acba10485f` used in the next `capture-log` step?

flacatus: We need something with go and the tools. Can you sugfess something else?

flacatus: Also can continue to use the e2e tests image since will be maintained by build team

tisutisu: looks fine to me now, as long as we are not using floating tag like `latest`, not sure about list of tools needed by the tests (apart from go-toolset), so for now we can use it and later evaluate list of dependent tools needed by the tests and update the image accordingly

tisutisu (approved):

tisutisu:
@flacatus since we merged this PR now, suspecting [this task](https://github.com/konflux-ci/e2e-tests/blob/51052ba6ac77e5c2e1170a577cbba4a7882ec52a/.tekton/konflux-e2e-tests-push.yaml#L473) will start to fail, maybe we need remove that task as well as part of the move

PR #3524, opened 2026-05-15
fix: validate base image architecture before build

fmudry:
When a base image is pinned by a single-arch digest (e.g. amd64), buildah pulls the manifest regardless of the host architecture and builds the image via emulation. Not only is it against the RH policy, but the resulting image-index contains manifests of the same architecture.

We must fail in this scenario. I've implemented the check after pulling the image, but before build. I've considered using skopeo instead of buildah pull+inspect, but the only benefit is that we would fail before pulling an image. Since konflux-build-cli already implements pre-pull logic regardless if it's a non-/hermetic build, this approach is simpler. Moreover, the bug doesn't happen that often, so failing a little later is a trade-off between user experience and simplicity/maintainability.

I've also removed the unshare flags from the hermetic pre-pulling, since it's redundant.

## Validation
Tested on staging — no regressions observed.
Staging PR: https://github.com/fmudry-konflux/devfile-sample-python-basic/pull/20
(you can focus only on the comments, the following commits are just more test)

## Risk Assessment
**Risk Level:** Low
**Description:** fail builds done via emulation
**Rollback:** Revert PR and redeploy the task

### DIFF
--- a/task/buildah-min/0.9/buildah-min.yaml
+++ b/task/buildah-min/0.9/buildah-min.yaml
@@ -6,7 +6,7 @@ metadata:
     tekton.dev/pipelines.minVersion: 0.12.1
     tekton.dev/tags: image-build, konflux
   labels:
-    app.kubernetes.io/version: 0.9.3
+    app.kubernetes.io/version: 0.9.4
     build.appstudio.redhat.com/build_type: docker
   name: buildah-min
 spec:
--- a/task/buildah-min/0.9/buildah-min.yaml
+++ b/task/buildah-min/0.9/buildah-min.yaml
@@ -539,26 +539,47 @@ spec:
             tr -d "'"
       )
 
+      echo "[$(date --utc -Ins)] Validate base image architectures"
+
+      host_arch=$(uname -m)
+      case "$host_arch" in
+        x86_64)  oci_arch="amd64" ;;
+        aarch64) oci_arch="arm64" ;;
+        *)       oci_arch="$host_arch" ;;
+      esac
+
+      set_proxy
+
+      # Pull each base image and verify its architecture matches the host, prevents building via emulation
+      # Pulled images are then reused when calling buildah build
+      buildah_retries=3
+      for image in $BASE_IMAGES; do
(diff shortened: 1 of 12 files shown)


task/buildah-min/0.9/buildah-min.yaml
chmeliik: Could you bump the version (patch number is fine) and update the changelog?

task/buildah-min/0.9/buildah-min.yaml
```diff
@@ -539,26 +539,45 @@ spec:
...
+          exit 1
+        fi
+
+        if ! image_arch=$(buildah inspect --type=image --format '{{.OCIv1.Architecture}}' "$image"); then
+          echo "ERROR: Failed to inspect base image ${image}" >&2
+          exit 1
+        fi
+
+        if [ "$image_arch" != "$oci_arch" ]; then
+          echo "ERROR: Base image ${image} has architecture '${image_arch}', expected '${oci_arch}'. Use a multi-arch image reference instead of a single-architecture digest." >&2
```
chmeliik, line 570: We know it's a single-arch reference, because if it was an index reference without the correct arch, `buildah pull` would have already failed above?

It might be worth mentioning that in a comment, it wasn't immediately clear to me why the error message is correct here

chmeliik: It might also be worth adding integration tests for both cases (index without the right arch, wrong single-arch ref) when adding this check to konflux-build-cli

fmudry: - Yes. Good point, we shouldn't "hide the behavior" :)
- Thanks, I'll make sure to include them

chmeliik (approved):

mmorhun (approved):

jedinym (approved):

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-05-25T11:11:32.252Z |
| **Risk Level** | low |

PR #3526, opened 2026-05-15
Update git-clone to use konflux-build-cli

tnevrlka:
git-clone logic was implemented in Go in the [konflux-build-cli](https://github.com/konflux-ci/konflux-build-cli) repository. 
Use it instead of the original Bash in YAML implementation.

### DIFF
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -191,7 +191,7 @@ spec:
     - init
     taskRef:
       name: git-clone
-      version: "0.1"
+      version: "0.2"
     workspaces:
     - name: output
       workspace: workspace
--- a/task/git-clone-oci-ta-min/0.2/MIGRATION.md
+++ b/task/git-clone-oci-ta-min/0.2/MIGRATION.md
@@ -1,0 +1,14 @@
+# Migration from 0.1 to 0.2
+
+## What changed
+
+- **Removed parameters:** `gitInitImage`, `verbose`, `userHome`.
+- **New parameter:** `logLevel` (default `info`) replaces `verbose`. Valid values: `debug`, `info`, `warn`, `error`.
+- **Image changed:** The task now uses `konflux-build-cli` instead of the previous `git-clone` image.
+- **Step consolidation:** The separate `symlink-check` step has been merged into the `clone` step. Symlink checking is now handled internally by the Go binary.
+
+## Action from users
+
+The migration should be done automatically by renovate.
+For users who don't have automatic updates, remove `gitInitImage`, `verbose`, and `userHome` parameters if present.
+Replace `verbose: "true"` with `logLevel: "debug"` if needed.
--- a/task/git-clone-oci-ta-min/0.2/README.md
+++ b/task/git-clone-oci-ta-min/0.2/README.md
@@ -1,0 +1,51 @@
+# git-clone-oci-ta-min task
+
+The git-clone-oci-ta Task will clone a repo from the provided url and store it as a trusted artifact in the provided OCI repository.
+
+## Parameters
+|name|description|default value|required|
+|---|---|---|---|
+|caTrustConfigMapKey|The name of the key in the ConfigMap that contains the CA bundle data.|ca-bundle.crt|false|
+|caTrustConfigMapName|The name of the ConfigMap to read CA bundle data from.|trusted-ca|false|
+|depth|Perform a shallow clone, fetching only the most recent N commits.|1|false|
+|enableSymlinkCheck|Check symlinks in the repo. If they're pointing outside of the repo, the build will fail. |true|false|
+|fetchTags|Fetch all tags for the repo.|false|false|
+|httpProxy|HTTP proxy server for non-SSL requests.|""|false|
+|httpsProxy|HTTPS proxy server for SSL requests.|""|false|
+|mergeSourceDepth|Perform a shallow fetch of the target branch, fetching only the most recent N commits. If empty, fetches the full history of the target branch. |""|false|
+|mergeSourceRepoUrl|URL of the repository to fetch the target branch from when mergeTargetBranch is true. If empty, uses the same repository (origin). This allows merging a branch from a different repository. |""|false|
+|mergeTargetBranch|Set to "true" to merge the targetBranch into the checked-out revision.|false|false|
+|noProxy|Opt out of proxying HTTP/HTTPS requests.|""|false|
+|ociArtifactExpiresAfter|Expiration date for the trusted artifacts created in the OCI repository. An empty string means the artifacts do not expire.|""|false|
+|ociStorage|The OCI repository where the Trusted Artifacts are stored.||true|
+|refspec|Refspec to fetch before checking out revision.|""|false|
+|revision|Revision to checkout. (branch, tag, sha, ref, etc...)|""|false|
+|shortCommitLength|Length of short commit SHA|7|false|
+|sparseCheckoutDirectories|Define the directory patterns to match or exclude when performing a sparse checkout.|""|false|
+|sslVerify|Set the `http.sslVerify` global git config. Setting this to `false` is not advised unless you are sure that you trust your git remote.|true|false|
+|submodulePaths|Comma-separated list of specific submodule paths to initialize and fetch. Only submodules in the specified directories and their subdirectories will be fetched. Empty string fetches all submodules. Parameter "submodules" must be set to "true" to make this parameter applicable. |""|false|
+|submodules|Initialize and fetch git submodules.|true|false|
+|targetBranch|The target branch to merge into the revision (if mergeTargetBranch is true).|main|false|
+|url|Repository URL to clone from.||true|
+|userHome|Absolute path to the user's home directory. Set this explicitly if you are running the image as a non-root user. |/tekton/home|false|
+|verbose|Log the commands that are executed during `git-clone`'s operation.|false|false|
+
+## Results
+|name|description|
+|---|---|
+|CHAINS-GIT_COMMIT|The precise commit SHA that was fetched by this Task. This result uses Chains type hinting to include in the provenance.|
+|CHAINS-GIT_URL|The precise URL that was fetched by this Task. This result uses Chains type hinting to include in the provenance.|
+|SOURCE_ARTIFACT|The Trusted Artifact URI pointing to the artifact with the application source code.|
+|commit|The precise commit SHA that was fetched by this Task.|
+|commit-timestamp|The commit timestamp of the checkout|
+|merged_sha|The SHA of the commit after merging the target branch (if the param mergeTargetBranch is true).|
+|short-commit|The commit SHA that was fetched by this Task limited to params.shortCommitLength number of characters|
+|url|The precise URL that was fetched by this Task.|
+
+## Workspaces
+|name|description|optional|
+|---|---|---|
+|basic-auth|A Workspace containing a .gitconfig and .git-credentials file or username and password. These will be copied to the user's home before any git commands are run. Any other files in this Workspace are ignored. It is strongly recommended to use ssh-directory over basic-auth whenever possible and to bind a Secret to this Workspace over other volume types. |true|
+|ssh-directory|A .ssh directory with private key, known_hosts, config, etc. Copied to the user's home before git commands are executed. Used to authenticate with the git remote when performing the clone. Binding a Secret to this Workspace is strongly recommended over other volume types. |true|
+
+## Additional info
--- a/task/git-clone-oci-ta-min/0.2/README.md
+++ b/task/git-clone-oci-ta-min/0.2/README.md
@@ -12,6 +12,7 @@ The git-clone-oci-ta Task will clone a repo from the provided url and store it a
 |fetchTags|Fetch all tags for the repo.|false|false|
 |httpProxy|HTTP proxy server for non-SSL requests.|""|false|
 |httpsProxy|HTTPS proxy server for SSL requests.|""|false|
+|logLevel|Log level for the git-clone command.|info|false|
 |mergeSourceDepth|Perform a shallow fetch of the target branch, fetching only the most recent N commits. If empty, fetches the full history of the target branch. |""|false|
 |mergeSourceRepoUrl|URL of the repository to fetch the target branch from when mergeTargetBranch is true. If empty, uses the same repository (origin). This allows merging a branch from a different repository. |""|false|
 |mergeTargetBranch|Set to "true" to merge the targetBranch into the checked-out revision.|false|false|
@@ -27,8 +28,6 @@ The git-clone-oci-ta Task will clone a repo from the provided url and store it a
 |submodules|Initialize and fetch git submodules.|true|false|
 |targetBranch|The target branch to merge into the revision (if mergeTargetBranch is true).|main|false|
 |url|Repository URL to clone from.||true|
-|userHome|Absolute path to the user's home directory. Set this explicitly if you are running the image as a non-root user. |/tekton/home|false|
-|verbose|Log the commands that are executed during `git-clone`'s operation.|false|false|
 
 ## Results
 |name|description|
--- a/task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
+++ b/task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+declare -r pipeline_file=${1:?missing pipeline file}
+
+tasks_names=()
+tasks_selector="(.spec.tasks[], .spec.pipelineSpec.tasks[])"
+
+for task_refname in "git-clone" "git-clone-oci-ta" "git-clone-oci-ta-min" "pnc-prebuild-git-clone-oci-ta"; do
+    task_selector="${tasks_selector} | select(.taskRef.params[] | (.name == \"name\" and .value == \"${task_refname}\"))"
+    if yq -e "$task_selector" "$pipeline_file" >/dev/null 2>&1; then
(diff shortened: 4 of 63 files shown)


task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
```diff
@@ -0,0 +1,34 @@
...
+    task_name_selector="${tasks_selector} | select(.name == \"${task_name}\")"
+    verbose_val=$(yq -e "${task_name_selector} | .params[] | select(.name == \"verbose\") | .value" "$pipeline_file" 2>/dev/null) || true
+    if [ "${verbose_val}" = "true" ]; then
+        log_level="debug"
+    fi
+
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param gitInitImage
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param verbose
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param userHome
+    pmt modify -f "$pipeline_file" task "$task_name" add-param LOG_LEVEL "$log_level"
```
mmorhun: Do we need to explicitly add the parameter for the default value?

task/git-clone-oci-ta-min/0.2/MIGRATION.md
```diff
@@ -0,0 +1,13 @@
...
+## What changed
+
+- **Removed parameters:** `gitInitImage`, `verbose`, `userHome`.
+- **New parameter:** `LOG_LEVEL` (default `info`) replaces `verbose`. Valid values: `debug`, `info`, `warn`, `error`.
+- **Image changed:** The task now uses `konflux-build-cli` instead of the previous `git-clone` image.
+- **Step consolidation:** The separate `symlink-check` step has been merged into the `clone` step. Symlink checking is now handled internally by the Go binary.
+
+## Action from users
+
+Remove `gitInitImage`, `verbose`, and `userHome` parameters if present.
```
mmorhun: Won't it be done by `renovate` and `pmt`?
Probably we should mention that this part should be done automatically.

task/git-clone/0.2/tests/test-git-clone-full-depth.yaml
```diff
@@ -0,0 +1,41 @@
...
+    Test the git-clone task with full commit history (depth=0)
+  workspaces:
+    - name: tests-workspace
+  tasks:
+    - name: run-task
+      taskRef:
+        name: git-clone
+      params:
+        - name: url
+          value: https://github.com/kelseyhightower/nocode
```
mmorhun: Not sure about this.
Way safer would be to use a repository from `konflux-ci` org.

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,239 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.2"
```
mmorhun, line 5: Should we deprecate version `0.1`, with due date in around 3 months?

mmorhun (commented):
Kudos for the tests.
Nevertheless, since the change affect all users, a test in a real Konflux pipeline is highly recommended (if you've done it, a note or screenshot is appreciated).

pipelines/core-services/core-services.yaml
```diff
@@ -191,7 +191,7 @@ spec:
     - init
     taskRef:
       name: git-clone
-      version: "0.1"
+      version: "0.2"
```
chmeliik, line 194: We should also bump the git-clone version in the .tekton/ pipelines

task/git-clone/0.2/git-clone.yaml
```diff
@@ -71,18 +71,6 @@ spec:
...
-    name: verbose
-    type: string
-  - default: ""
-    description: Deprecated. Has no effect. Will be removed in the future.
-    name: gitInitImage
-  - default: /tekton/home
-    description: |
-      Absolute path to the user's home directory. Set this explicitly if you are running the image as a non-root user.
-    name: userHome
-    type: string
```
chmeliik: Could you mention in the commit message why these two were removed?

Or maybe split this removal into a separate commit? As far as I can tell, it's not related to konflux-build-cli

task/git-clone/0.2/tests/test-git-clone-merge-target-branch.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+              # merged_sha must be set and equal to commit (both point to the merge commit)
+              [[ "$MERGED_SHA" =~ ^[0-9a-f]{40}$ ]]
+              [ "$MERGED_SHA" = "$COMMIT" ]
+
+              # The merge commit must differ from the original tag SHA
+              [ "$COMMIT" != "$TAG_SHA" ]
+
+              # The workspace should contain files from both the tag and master
+              test -f "$(workspaces.output.path)/source/README.md"
```
chmeliik: README.md is in both `master` and `v1.0.0`, so I think this would pass even if the target branch merging didn't work.

It also feels fragile to use an external repo for this test

task/pnc-prebuild-git-clone-oci-ta/0.2/migrations/0.2.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ ${#tasks_names[@]} -eq 0 ]; then
+    echo "No git-clone tasks found"
+    exit 0
+fi
+
+for task_name in "${tasks_names[@]}"; do
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param gitInitImage
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param verbose
+    pmt modify -f "$pipeline_file" task "$task_name" remove-param userHome
+    pmt modify -f "$pipeline_file" task "$task_name" add-param LOG_LEVEL "info"
```
chmeliik: Should this also have the logic that sets LOG_LEVEL based on the `verbose` value? It's currently missing

task/git-clone-oci-ta/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## 0.2
+
+- Use git-clone implementation from konflux-build-cli instead of inline Bash.
+- Removed `verbose` and `userHome` parameters.
```
chmeliik: `task/git-clone/CHANGELOG.md` mentions the removed `gitInitImage` param but the other changelogs don't

task/git-clone-oci-ta-min/0.2/MIGRATION.md
chmeliik: +1, let's mention that but also keep the instructions

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,239 @@
...
+    - name: HTTP_PROXY
+      value: $(params.httpProxy)
+    - name: HTTPS_PROXY
+      value: $(params.httpsProxy)
+    - name: NO_PROXY
+      value: $(params.noProxy)
+    - name: KBC_LOG_LEVEL
+      value: $(params.LOG_LEVEL)
+    image: quay.io/konflux-ci/konflux-build-cli:latest@sha256:b296232c9b0d478c0bd1f48911ead97cd786eebdc737b877797564567fda8eae
+    computeResources: {}
```
chmeliik: FYI I just added https://github.com/konflux-ci/build-definitions/pull/3512 to the merge queue. Version 0.2 will need the computeResources added as well

task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
mmorhun: @tnevrlka please add the `LOG_LEVEL` only if it has non default value.

task/git-clone-oci-ta-min/0.2/MIGRATION.md
```diff
@@ -0,0 +1,16 @@
...
+
+- **Removed parameters:** `gitInitImage`, `verbose`, `userHome`.
+- **New parameter:** `LOG_LEVEL` (default `info`) replaces `verbose`. Valid values: `debug`, `info`, `warn`, `error`.
+- **Image changed:** The task now uses `konflux-build-cli` instead of the previous `git-clone` image.
+- **Step consolidation:** The separate `symlink-check` step has been merged into the `clone` step. Symlink checking is now handled internally by the Go binary.
+
+## Action from users
+
+The task version bump in pipelines is handled automatically by renovate and pmt.
+The parameter changes below may still require manual action:
```
mmorhun: That's confusing for end users.
We should say something like:
> The migration should be done automatically by renovate.
For users who don't have automatic updates, remove ...

mmorhun (commented):
@tnevrlka since the change affect all users, a test in a real Konflux pipeline is highly recommended (if you've done it, a note or screenshot is appreciated).

task/git-clone/0.2/git-clone.yaml
chmeliik: Still not seeing the computeResources from #3512 in version 0.2

task/git-clone/0.2/git-clone.yaml
chmeliik: I'm still not seeing the reasoning documented for why these were removed. Mentioning it in the changelog would also work for me, and would be good to do so that users understand as well

eqrx (dismissed):

chmeliik (commented):
@tnevrlka what is the status here? Needs a rebase and a few threads resolved, then should be good to go?

task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
tnevrlka: I don't know how to go around this. Doing so causes CI check regarding migrations to fail

chmeliik: Right, it fails if the migration script doesn't do any changes. We can either ignore that and merge anyway, or always add the param. Or we can kind of hack it, add it unconditionally for git-clone-oci-ta-min but conditionally for the ones that people actually use :smile:

task/git-clone-oci-ta-min/0.2/patch.yaml
```diff
@@ -0,0 +1,19 @@
...
+  value: create-trusted-artifact
+- op: replace
+  path: /spec/steps/1/computeResources/limits/memory
+  value: 256Mi
+- op: replace
+  path: /spec/steps/1/computeResources/requests/memory
+  value: 256Mi
+- op: replace
+  path: /spec/steps/1/computeResources/requests/cpu
+  value: 100m
```
chmeliik, line 32: I think we need the overrides for step 0 as well, currently it's inheriting the 2Gi memory

chmeliik:
LGTM apart from https://github.com/konflux-ci/build-definitions/pull/3526#discussion_r3381666421

task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
tnevrlka: imo always adding it is the cleanest, @mmorhun are you okay with it?

task/git-clone-oci-ta/0.2/git-clone-oci-ta.yaml
```diff
@@ -0,0 +1,283 @@
...
+    tekton.dev/pipelines.minVersion: 0.21.0
+    tekton.dev/platforms: linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
+    tekton.dev/tags: git
+  labels:
+    app.kubernetes.io/version: "0.2"
+spec:
+  description: The git-clone-oci-ta Task will clone a repo from the provided
+    url and store it as a trusted artifact in the provided OCI repository.
+  params:
+    - name: LOG_LEVEL
```
MartinBasti: all other params are lower case with camel case, why we don't keep consistency?

chmeliik:
Also tested this in https://github.com/konflux-ci/task-runner/pull/201. A reasonably tricky repo with git submodules which also needs to fetch the tags in those submodules.

worked :heavy_check_mark:

task/git-clone-oci-ta/0.2/git-clone-oci-ta.yaml
tnevrlka: ok, sounds reasonable

MartinBasti:
LGTM, it needs rebase and resolve conflicts

chmeliik (dismissed):

task/git-clone/0.2/tests/test-git-clone-merge-target-branch.yaml
```diff
@@ -0,0 +1,63 @@
...
+              set -eux
+
+              TAG_SHA="aab8f135d7b2857463be3a2dec98789363e7087c"
+
+              # commit is the current HEAD (post-merge), must be a valid SHA
+              [[ "$COMMIT" =~ ^[0-9a-f]{40}$ ]]
+
+              # merged_sha must be set and equal to commit (both point to the merge commit)
+              [[ "$MERGED_SHA" =~ ^[0-9a-f]{40}$ ]]
+              [ "$MERGED_SHA" = "$COMMIT" ]
```
chmeliik: Hmm, fullsend makes a valid point here - why is the `MERGED_SHA` result the same as the `COMMIT` result? I think it wouldn't have been the case with git-clone 0.1

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,244 @@
...
+    - name: KBC_GIT_CLONE_BASIC_AUTH_DIRECTORY
+      value: $(workspaces.basic-auth.path)
+    - name: KBC_GIT_CLONE_SSH_DIRECTORY
+      value: $(workspaces.ssh-directory.path)
+    - name: HTTP_PROXY
+      value: $(params.httpProxy)
+    - name: HTTPS_PROXY
+      value: $(params.httpsProxy)
+    - name: NO_PROXY
+      value: $(params.noProxy)
```
chmeliik: Another potentially valid point from fullsend - is it okay to set the proxy env vars unconditionally, even if the values are empty strings?

chmeliik (commented):
After reading the fullsend review, I think two points warrant addressing

task/git-clone-oci-ta-min/0.2/README.md
```diff
@@ -0,0 +1,50 @@
...
+|refspec|Refspec to fetch before checking out revision.|""|false|
+|revision|Revision to checkout. (branch, tag, sha, ref, etc...)|""|false|
+|shortCommitLength|Length of short commit SHA|7|false|
+|sparseCheckoutDirectories|Define the directory patterns to match or exclude when performing a sparse checkout.|""|false|
+|sslVerify|Set the `http.sslVerify` global git config. Setting this to `false` is not advised unless you are sure that you trust your git remote.|true|false|
+|submodulePaths|Comma-separated list of specific submodule paths to initialize and fetch. Only submodules in the specified directories and their subdirectories will be fetched. Empty string fetches all submodules. Parameter "submodules" must be set to "true" to make this parameter applicable. |""|false|
+|submodules|Initialize and fetch git submodules.|true|false|
+|targetBranch|The target branch to merge into the revision (if mergeTargetBranch is true).|main|false|
+|url|Repository URL to clone from.||true|
+|logLevel|Log level for the git-clone command.|info|false|
```
webbnh: I don't want to delay this merge, but...in #3565, which is based on the branch for this PR, the [fullsend-ai-review](https://github.com/apps/fullsend-ai-review) Bot [is complaining](https://github.com/konflux-ci/build-definitions/pull/3565#issuecomment-4685447932) that `logLevel` is not in proper alphabetical order, here and in some of the other Git-Clone `README.md` files.

task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
mmorhun: @tnevrlka I don't think it's necessary. Majority of users won't need it, so we add a parameter which hardly ever changes.

mmorhun: Also (and maybe it's really a micro management but still), having less items will reduce ETCD usage.

tnevrlka: So when there's a CI failure regarding that, you're comfortable with force-merging that? I don't have anything against it, just asking

task/git-clone/0.2/tests/test-git-clone-merge-target-branch.yaml
tnevrlka: Oh right, so switch the order of mergeTargetBranch in https://github.com/konflux-ci/konflux-build-cli/blob/0a7d5e97b52d83a2c1914290e3f3a0cbe11d83ce/pkg/commands/gitclone/gitclone.go#L108-L125?

chmeliik: I think so, yeah

tnevrlka: PTAL https://github.com/konflux-ci/konflux-build-cli/pull/152

mmorhun (approved):
LGTM

task/git-clone-oci-ta-min/0.2/migrations/0.2.sh
mmorhun: I think we can force merge it if people ignores the review request.
However, let's get at least a couple of approvals.

MartinBasti (approved):

chmeliik (approved):

ILpinto:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ILpinto |
| **Timestamp** | 2026-06-24T11:55:22.035Z |

PR #3585, opened 2026-06-10
fix(modelcar-oci-ta): raise create-modelcar-base-image memory to 512Mi

ralphbean:
The create-modelcar-base-image step runs oras copy --to-oci-layout to pull a full base image. The 64Mi floor value set in cd975229 is insufficient, causing OOMKilled in CI. Raise to 512Mi to match the copy-model-files step.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>


Victoremepunto (approved):
LGTM

sanialisar (approved):
LGTM

PR #3587, opened 2026-06-10
Remove latest tag for images using :latest@digest pattern

hugares:
Having the latest tag make k8s set the imagepullpolicy to always which cause the pull to not leverage node cache.

 Update the trusted-artifacts generator so it can reuse an existing pinned build-trusted-artifacts image when regenerating tasks. The old logic only matched references containing "/build-trusted-artifacts:",
 which broke after switching to digest-only pullspecs (...@sha256:...). It now parses each step image with go-containerregistry and checks whether the repository is quay.io/konflux-ci/build-trusted-artifacts, so
 both legacy tag+digest and new digest-only formats are recognized and the existing digest is preserved instead of being re-resolved from :latest

[KFLUXINFRA-3096](https://redhat.atlassian.net/browse/KFLUXINFRA-3096)

[KFLUXINFRA-3096]: https://redhat.atlassian.net/browse/KFLUXINFRA-3096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/task-generator/trusted-artifacts/golden/buildah/ta.yaml
+++ b/task-generator/trusted-artifacts/golden/buildah/ta.yaml
@@ -130,7 +130,7 @@ spec:
       - mountPath: /var/workdir
         name: workdir
   steps:
-  - image: quay.io/konflux-ci/build-trusted-artifacts:latest@sha256:resolved
+  - image: quay.io/konflux-ci/build-trusted-artifacts@sha256:resolved
     name: use-trusted-artifact
     args:
       - use
--- a/task-generator/trusted-artifacts/golden/git-clone/ta.yaml
+++ b/task-generator/trusted-artifacts/golden/git-clone/ta.yaml
@@ -251,7 +251,7 @@ spec:
       fi
 
   - name: create-trusted-artifact
-    image: quay.io/konflux-ci/build-trusted-artifacts:latest@sha256:existing
+    image: quay.io/konflux-ci/build-trusted-artifacts@sha256:existing
     env:
     - name: IMAGE_EXPIRES_AFTER
       value: $(params.ociArtifactExpiresAfter)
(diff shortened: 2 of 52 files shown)


yftacherzog (approved):

task-generator/trusted-artifacts/golden/buildah/ta.yaml
```diff
@@ -130,7 +130,7 @@ spec:
       - mountPath: /var/workdir
         name: workdir
   steps:
-  - image: quay.io/konflux-ci/build-trusted-artifacts:latest@sha256:resolved
+  - image: quay.io/konflux-ci/build-trusted-artifacts@sha256:resolved
     name: use-trusted-artifact
```
filariow, line 134: if we want to maintain the same semantic we could add a renovate annotation to explicitly tell renovate to follow the `latest` tag. I'm not an expert, but as per my understanding we'll need something like the following:

```suggestion
  - name: use-trusted-artifact
    # renovate: image=quay.io/konflux-ci/build-trusted-artifacts tag=latest
    image: quay.io/konflux-ci/build-trusted-artifacts@sha256:resolved
```

yftacherzog: I don't think there's something that digests standard comments out of the box.
It needs to be accompanied by changes in renovate.json.

chmeliik (commented):
Let's not change the tags but instead explicitly set `imagePullPolicy: IfNotPresent` everywhere (except the tasks that already set something else)

The Kubernetes behavior is magical and frankly nonsensical. Tags will inevitably creep back in over time. An explicit policy works much better IMO

hugares:
> Let's not change the tags but instead explicitly set `imagePullPolicy: IfNotPresent` everywhere (except the tasks that already set something else)
> 
> The Kubernetes behavior is magical and frankly nonsensical. Tags will inevitably creep back in over time. An explicit policy works much better IMO

Thanks for the feedback. I thought this could be a low hanging fruit but the comment and the fact there are many task to deal with, this is only part 1, I will instead close this PR and create Jira work items for each team to deal with theirs

PR #3622, opened 2026-06-24
ART-21342: expose EFFECTIVE_TIME param in enterprise-contract pipeline

ashwindasr:
## Summary

- Expose the `EFFECTIVE_TIME` parameter in the `enterprise-contract` pipeline, passing it through to the `verify-enterprise-contract` task
- The underlying `verify-enterprise-contract` task already supports this parameter (passed as `--effective-time` to ec-cli), but the pipeline did not expose it

## Motivation

Build systems need to evaluate EC policies as of a future date -- specifically the scheduled release ship date -- so that rules with upcoming `effective_on` dates are caught during build time, before they become enforced at release time.

Without this pass-through, callers have no way to override the evaluation time and policies are always checked against "now".

Relates to https://redhat.atlassian.net/browse/ART-20711 and https://redhat.atlassian.net/browse/EC-1910

## Changes

- Added `EFFECTIVE_TIME` pipeline param (type: string, default: `"now"`)
- Wired it through to the `verify` task params

Default behavior is unchanged -- existing callers that do not set `EFFECTIVE_TIME` will continue to use `"now"`.

## Test plan

- Verify existing EC pipeline runs still pass (default "now" behavior unchanged)
- Verify that passing a future date string causes ec-cli to evaluate time-gated rules accordingly

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -82,7 +82,8 @@ spec:
       default: "true"
     - name: EFFECTIVE_TIME
       type: string
-      description: Run policy checks with the provided time.
+      description: Run policy checks with the provided time. Expects a
+        RFC3339 formatted value, e.g. "2026-06-24T00:00:00Z"
       default: "now"
   results:
     - name: TEST_OUTPUT
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -80,6 +80,10 @@ spec:
         Setting to false is useful on specific conditions but will always mark the integration test as successful and
         humans will tend to ignore the test results if they failed. Use with caution.
       default: "true"
+    - name: EFFECTIVE_TIME
+      type: string
+      description: Run policy checks with the provided time.
+      default: "now"
   results:
     - name: TEST_OUTPUT
       value: "$(tasks.verify.results.TEST_OUTPUT)"
@@ -137,6 +141,8 @@ spec:
           value: "$(params.SINGLE_COMPONENT)"
         - name: SINGLE_COMPONENT_CUSTOM_RESOURCE
           value: "$(params.SINGLE_COMPONENT_CUSTOM_RESOURCE)"
+        - name: EFFECTIVE_TIME
+          value: "$(params.EFFECTIVE_TIME)"
       taskRef:
         resolver: bundles
         params:


pipelines/enterprise-contract.yaml
```diff
@@ -80,6 +80,10 @@ spec:
         Setting to false is useful on specific conditions but will always mark the integration test as successful and
         humans will tend to ignore the test results if they failed. Use with caution.
       default: "true"
+    - name: EFFECTIVE_TIME
+      type: string
+      description: Run policy checks with the provided time.
```
simonbaird: Mentioning the format might be useful, something like this:


```suggestion
      description: Run policy checks with the provided time. Expects a
        RFC3339 formatted value, e.g. "2026-06-24T00:00:00Z"
```

simonbaird (dismissed):
Lgtm.

simonbaird:
FYI this pipeline uses the `verify-enterprise-contract` task, not `verify-conforma-konflux-ta`, which his the one used in the release-service-catalog pipelines.

See https://github.com/konflux-ci/release-service-catalog/tree/development/pipelines/managed .

pipelines/enterprise-contract.yaml
ashwindasr: added in https://github.com/konflux-ci/build-definitions/pull/3622/commits/eae8dc52b5338487a7d7978c8f1d33ee56f56d31

simonbaird:
Pipeline ran successfully [here](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/rhtap-contract-tenant/applications/ec-main/taskruns/ec-main-enterprise-contract-test-kjwlb-verify/), so should be safe to merge.

pipelines/enterprise-contract.yaml
```diff
@@ -80,6 +80,11 @@ spec:
         Setting to false is useful on specific conditions but will always mark the integration test as successful and
         humans will tend to ignore the test results if they failed. Use with caution.
       default: "true"
+    - name: EFFECTIVE_TIME
```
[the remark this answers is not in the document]
simonbaird: Valid point I guess, but I don't think it's blocker.

pipelines/enterprise-contract.yaml
```diff
@@ -80,6 +80,11 @@ spec:
...
+      type: string
+      description: Run policy checks with the provided time. Expects a
```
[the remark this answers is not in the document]
simonbaird: Not a big deal IMO.

pipelines/enterprise-contract.yaml
```diff
@@ -82,7 +82,8 @@ spec:
       default: "true"
     - name: EFFECTIVE_TIME
       type: string
-      description: Run policy checks with the provided time.
+      description: Run policy checks with the provided time. Expects a
+        RFC3339 formatted value, e.g. "2026-06-24T00:00:00Z"
```
simonbaird, line 86: I guess you could take the bot's recommendation, e.g. 

```
RFC3339 formatted value, e.g. "2026-06-24T00:00:00Z" or the string "now".
```

simonbaird (approved):

PR #3630, opened 2026-06-25
buildah: add option to allow crossplatfrom parent images

MartinBasti:
New task option ALLOW_CROSS_PLATFORM_IMAGES allows user to use images which doen't match platform of the build host.

### DIFF
--- a/task/buildah/0.10/buildah.yaml
+++ b/task/buildah/0.10/buildah.yaml
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.10.2"
+    app.kubernetes.io/version: "0.10.3"
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
@@ -232,6 +232,11 @@ spec:
     description: Log level for the build command.
     type: string
     default: info
+  - name: ALLOW_CROSS_PLATFORM_IMAGES
+    description: Allows to use parent images that don't match the build host architecture. This
+      option must be used with caution as it may create incompatible images.
+    type: string
+    default: "false"
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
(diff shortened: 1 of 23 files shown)


task/buildah/0.10/buildah.yaml
```diff
@@ -311,9 +316,11 @@ spec:
       value: $(params.SBOM_SKIP_VALIDATION)
     - name: SKIP_INJECTIONS
       value: $(params.SKIP_INJECTIONS)
+    - name: KBC_BUILD_ALLOW_CROSS_PLATFORM_IMAGES
+      value: $(params.ALLOW_CROSS_PLATFORM_IMAGES)
```
chmeliik, line 320: This would be the only one (except LOG_LEVEL) that we pass as an env var instead of a CLI flag. IMO we should keep it consistent with CLI flags

task/buildah/0.10/buildah.yaml
```diff
@@ -232,6 +232,11 @@ spec:
     description: Log level for the build command.
     type: string
     default: info
+  - name: ALLOW_CROSS_PLATFORM_IMAGES
```
chmeliik, line 235: Could you update the version and CHANGELOG?

task/buildah/0.10/buildah.yaml
MartinBasti: updated to use CLI instead

task/buildah/0.10/buildah.yaml
MartinBasti: added

task/buildah/0.10/buildah.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.10.2"
+    app.kubernetes.io/version: "0.10.3"
```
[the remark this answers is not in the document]
MartinBasti: Why don't you verify it?

chmeliik (approved):

PR #3641, opened 2026-06-26
fix(renovate): group all task-runner updates into a single PR

kelchen123:
Add a matchPackageNames rule for quay.io/konflux-ci/task-runner so that updates to this image are consolidated into one PR rather than split across every file-path group that references it.


chmeliik (changes requested):
This would mean that PRs which update task-runner need approvals by many different codeowners groups. In the current setup, they're separated by codeowners.

PR #3646, opened 2026-06-29
init: use external task

MartinBasti:
Init task has been moved to a separate repo and now it's consumed as an external task.


MartinBasti:
We can waive this check as migrations were moved together with the task
```
 error: It is not allowed to delete or modify existing migration file: archived-tasks/init/0.2/migrations/0.2.1.sh
error: Please bump task version in the label 'app.kubernetes.io/version' and create a new migration file.
```

fmudry (approved):

PR #3650, opened 2026-06-30
fix(task-generator): accept explicit task path for digest reuse

fmudry:
The generator derives the existing task file path from the directory structure, which assumes a versioned structure (task/name/version/).
In flat structures (task/name/) the derivation resolves to `task.yaml` instead of `<name>.yaml`, silently skipping digest reuse and resolving `:latest` from the registry on every run.

Accept an optional second argument to override the derived path.
No second argument preserves the existing behavior.

### DIFF
--- a/task-generator/trusted-artifacts/main.go
+++ b/task-generator/trusted-artifacts/main.go
@@ -21,7 +21,15 @@ func main() {
 
 	taDir := path.Dir(recipePath)
 
-	taTaskPath := path.Join(taDir, path.Base(path.Dir(taDir))+".yaml")
+	segments := strings.Split(recipePath, "/")
+	var name string
+	for i, s := range segments {
+		if s == "task" && i+1 < len(segments) {
+			name = segments[i+1]
+			break
+		}
+	}
+	taTaskPath := path.Join(taDir, name+".yaml")
 
 	if _, err := os.Stat(taTaskPath); err == nil {
 		existing := expectValue(readTask(taTaskPath))


task-generator/trusted-artifacts/main.go
```diff
@@ -21,7 +21,14 @@ func main() {
 
 	taDir := path.Dir(recipePath)
 
+	// Default: derive the task path from the directory structure (works with
+	// versioned layouts like task/<name>/<version>/recipe.yaml). An explicit
+	// second argument overrides this for flat layouts where the derivation
+	// produces the wrong path.
 	taTaskPath := path.Join(taDir, path.Base(path.Dir(taDir))+".yaml")
```
chmeliik: Now that I'm looking at it, I think we can fix it for both layouts without needing the extra argument

Currently, the script makes the assumption that you're calling it with `.../<name>/<version>/recipe.yaml`. The `...` part could be arbitrarily many path segments, but for all practical uses, it's either going to be nothing or `task/`

So we can just take the name from the first path element, or the second if the first one is `task`

chmeliik: The output path argument would still make sense, because currently the situation is that the generator has no idea where you're going to write its output but also assumes it can guess it from the input path.

But we can add it later as a more thorough rework:
- make it an `-o/--output` flag, not a positional arg, so that it's clear it's the output path
- make it affect the output as well, so that the generator doesn't write to stdout but to the output path

fmudry: Updated. I've also updated commit description and added examples.

fmudry: I'll also close shared-ci PR, as it is obsolete now.

task-generator/trusted-artifacts/main.go
```diff
@@ -21,7 +21,11 @@ func main() {
 
 	taDir := path.Dir(recipePath)
 
-	taTaskPath := path.Join(taDir, path.Base(path.Dir(taDir))+".yaml")
+	name := path.Base(path.Dir(taDir))
+	if name == "task" {
+		name = path.Base(taDir)
+	}
+	taTaskPath := path.Join(taDir, name+".yaml")
```
chmeliik: This is still "counting from the back" though. We can't really predict where in the path the task name will be relative to the recipe.yaml file, but what we do know is that the task name will always be at `task/{name}` relative to the repo root

So IMO we should count from the front - second element if first one is `task`, else first element (because the generator script `cd`s to task/ before calling the generator)

fmudry: :man_facepalming: I misunderstood, sorry for that. Updated and rebased.

chmeliik (approved):

task-generator/trusted-artifacts/main.go
chmeliik: > else first element (because the generator script cds to task/ before calling the generator)

Upon checking again, this is unnecessary because the script calls the generator with the absolute recipe path, so the current approach should work

PR #3654, opened 2026-07-01
Chunked model download

ushaket:
Replace bulk oras pull with per-blob oras blob fetch + olot loop. Each model file is downloaded, layered with olot, pushed to the destination registry, and deleted from disk before the next file. Peak disk drops from 2x model size to ~2x single file size.

Merges download-model-files, copy-model-files, apply-image-labels, and push-image into a single build-and-push step. Uses oras blob push for incremental layer uploads and oras manifest push at the end.


sanialisar (approved):
LGTM

PR #3665, opened 2026-07-06
Add retries to init and clone in template-build

ehavryli-wq:
## Summary

PoC for [SPRE-5454](https://redhat.atlassian.net/browse/SPRE-5454): add Tekton
pipeline-level `retries: 2` to the first two setup steps in
`template-build` — `init` and `clone-repository`.

Build pipelines often fail when a TaskRun pod is evicted or the node is
drained mid-run. Retrying from a fresh pod avoids a full pipeline restart
for transient infrastructure failures, following the same approach used
elsewhere in Konflux (e.g. `retries: 2` on `wait-for-results` in
[TFT-4496](https://redhat.atlassian.net/browse/TFT-4496)).

## Changes

- `pipelines/template-build/template-build.yaml`
  - `retries: 2` on pipeline task `init` (task ref: `init` 0.4)
  - `retries: 2` on pipeline task `clone-repository` (task ref: `git-clone` /
    `git-clone-oci-ta` depending on pipeline patch)

No task bundle or version bumps — only pipeline step configuration.

## Scope

`template-build` is the shared base for the docker-build family and other
pipelines that include it via kustomize, including:

- `docker-build` → `docker-build-oci-ta` → `docker-build-oci-ta-min`,
  `docker-build-multi-platform-oci-ta`
- `ko-build-oci-ta`, `maven-zip-build`, `package-operator-package`,
  `tekton-bundle-builder`

This MR intentionally covers **wave 1 only** (setup tasks assessed as
retry-safe). Follow-up MRs will add retries to other idempotent steps
(e.g. `deprecated-base-image-check`, `ecosystem-cert-preflight-checks`,
`rpms-signature-scan`) and handle non-idempotent tasks separately.

## Why these steps are safe to retry

| Step | Task | Rationale |
|------|------|-----------|
| `init` | `init` | Reads cluster ConfigMap and exposes proxy URLs; no registry writes. |
| `clone-repository` | `git-clone` / `git-clone-oci-ta` | Cleans workspace before clone (`deleteExisting: true`); a retry starts from a fresh pod/workdir. |

## Testing

- [ ] Repo CI on this PR (yaml-lint, kustomize build, Tekton checks)
- [ ] Manual review of generated pipeline bundles (kustomize output)
- Not run: live Konflux component build with forced pod eviction (planned as
  follow-up validation after merge)

## Related

- Jira epic: [SPRE-5454](https://redhat.atlassian.net/browse/SPRE-5454) — TaskRun retry resilience for Konflux build pipelines

[SPRE-5454]: https://redhat.atlassian.net/browse/SPRE-5454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TFT-4496]: https://redhat.atlassian.net/browse/TFT-4496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -178,6 +178,7 @@ spec:
     params:
     - name: enable-cache-proxy
       value: $(params.enable-cache-proxy)
+    retries: 2
     taskRef:
       name: init
       version: "0.4"
@@ -187,6 +188,7 @@ spec:
       value: $(params.git-url)
     - name: revision
       value: $(params.revision)
+    retries: 2
     runAfter:
     - init
     taskRef:
--- a/pipelines/docker-build-multi-platform-oci-ta/docker-build-multi-platform-oci-ta.yaml
+++ b/pipelines/docker-build-multi-platform-oci-ta/docker-build-multi-platform-oci-ta.yaml
@@ -109,6 +109,7 @@ spec:
     params:
     - name: enable-cache-proxy
       value: $(params.enable-cache-proxy)
+    retries: 2
     taskRef:
       name: init
       version: "0.4"
@@ -122,6 +123,7 @@ spec:
       value: $(params.output-image).git
     - name: ociArtifactExpiresAfter
       value: $(params.image-expires-after)
+    retries: 2
     runAfter:
     - init
     taskRef:
--- a/pipelines/docker-build-oci-ta-min/docker-build-oci-ta-min.yaml
+++ b/pipelines/docker-build-oci-ta-min/docker-build-oci-ta-min.yaml
@@ -101,6 +101,7 @@ spec:
     params:
     - name: enable-cache-proxy
       value: $(params.enable-cache-proxy)
+    retries: 2
     taskRef:
       name: init
       version: "0.4"
@@ -114,6 +115,7 @@ spec:
       value: $(params.output-image).git
     - name: ociArtifactExpiresAfter
       value: $(params.image-expires-after)
+    retries: 2
     runAfter:
     - init
     taskRef:
--- a/pipelines/docker-build-oci-ta/docker-build-oci-ta.yaml
+++ b/pipelines/docker-build-oci-ta/docker-build-oci-ta.yaml
@@ -103,6 +103,7 @@ spec:
     params:
     - name: enable-cache-proxy
       value: $(params.enable-cache-proxy)
+    retries: 2
     taskRef:
       name: init
       version: "0.4"
@@ -116,6 +117,7 @@ spec:
       value: $(params.output-image).git
     - name: ociArtifactExpiresAfter
       value: $(params.image-expires-after)
+    retries: 2
     runAfter:
     - init
     taskRef:
--- a/pipelines/docker-build/docker-build.yaml
+++ b/pipelines/docker-build/docker-build.yaml
@@ -103,6 +103,7 @@ spec:
     params:
     - name: enable-cache-proxy
       value: $(params.enable-cache-proxy)
+    retries: 2
     taskRef:
       name: init
       version: "0.4"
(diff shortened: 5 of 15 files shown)


jdcasey (approved):
LGTM, these both look like the types of things where retries are a no-brainer

chmeliik:
The pipelines that build from template-build.yaml need regenerating:

```
Did you forget to build the manifests locally?
Please run ./hack/build-manifests.sh and update your PR
Error: Process completed with exit code 1.
```

chmeliik (approved):

eqrx (dismissed):

chmeliik:
@ehavryli-wq it's not necessary to merge the main branch into your branch every time main gets updated. It also resets approvals and CI results every time

chmeliik (dismissed):

gbenhaim (changes requested):
I would prefer that `konflux-build-cli` would handle the retry since it has more context about the exact failure. Also, it can have smarter retry with backoff, jitter etc.

erdii (dismissed):

ehavryli-wq:
@gbenhaim Thanks again for the review  and sorry for the delay getting back on this.

I agree that **application-level** retries belong in `konflux-build-cli`, where we already have failure context and can do smarter retry (exit-code / stderr aware, backoff, jitter), rather than relying on Tekton `PipelineTask.retries` for git network flakes.

To address that, I opened a follow-up on the CLI side:  
https://github.com/konflux-ci/konflux-build-cli/pull/236  

It adds jitter to the shared `Retryer` (used by `git fetch` / git-clone) and includes property-based unit tests for the interval guarantees Adam asked for.

That said, Tekton `retries` and CLI retries cover **different** failure modes:

| Layer | Covers |
|-------|--------|
| **konflux-build-cli** | Process still running — transient network / git fetch errors |
| **Tekton `retries` (this PR)** | Process/pod already gone — eviction, node drain, OOM |

Josh has reviewed this PR and is fine with the Tekton-side changes for the infra/pod case. My proposal is to treat them as complementary:

1. Land the CLI improvements (#236) for the smarter app-level path you asked for  
2. Keep Tekton `retries` on `init` / `clone-repository` for pod-level resilience (this PR)

If you’d still prefer **no** Tekton retries on clone even for eviction recovery, I’m happy to adjust the scope (e.g. drop `clone-repository` and keep only `init`, or close this PR and leave pod-level resilience to platform work).

gbenhaim:
| Build pipelines often fail when a TaskRun pod is evicted or the node is
drained mid-run

@ehavryli-wq I think that the tekton controller should handle such retries, can you please ensure that it's isn't the case? if it isn't the case, the fix should be implemented in https://github.com/tektoncd/pipeline so it will apply to any task.

ehavryli-wq:
@gbenhaim Thanks, good question. I checked how Tekton behaves in this case.

Tekton does not automatically retry a TaskRun when the pod is evicted or drained. If the pod is evicted, the TaskRun fails. It will retry only if retries is explicitly configured. If retries is not set, Tekton does not retry the task.

There is also no built-in option today to retry only for infrastructure failures like eviction or node drain, but not for application failures.

That is why this PR is intentionally limited. I added retries: 2 only to the clearly idempotent setup steps like init and clone-repository. I did not want to enable retries for all build tasks, because some of them may not be safe to run twice and could also make builds more expensive.

I agree that the better long-term solution would be at the Tekton or platform level, ideally with retry logic based on the failure reason.

For now, I see this PR as a small mitigation for the current incidents. I’m happy to open a follow-up issue for the platform/upstream solution as well.

If you prefer, we can also pause this PR and focus only on the platform-level approach.

PR #3670, opened 2026-07-07
Add opt-in reproducibility params to docker-build

JamesOlaitan:
## Summary

This wires the three buildah reproducibility flags through the `docker-build*` pipeline variants so users can opt into byte-wise reproducible image digests. The build tasks have accepted `SOURCE_DATE_EPOCH`, `REWRITE_TIMESTAMP`, and `OMIT_HISTORY` since https://github.com/konflux-ci/build-definitions/pull/2947; this change adds the matching pipeline parameters and passes them through to `build-container` (or `build-images` in the multi-platform variant). All three ship default-off, so existing pipelines are unaffected. This implements the pipeline-level wiring described in [ADR-0069](https://github.com/konflux-ci/architecture/blob/4d68b9cbcfc6e4d96f693713033bed086eca51fb/ADR/0069-reproducible-container-builds.md).

The three parameters:

- `source-date-epoch`: a timestamp in seconds since the Unix epoch. buildah sets the image config's `created` field to this value and exports it as the `SOURCE_DATE_EPOCH` build argument. On its own, it does not touch file timestamps inside layers.
- `rewrite-timestamp`: when `"true"`, buildah clamps file modification times in the layer tarballs to at most the `source-date-epoch` value. This is the flag that makes layer content byte-stable across rebuilds. It does nothing unless `source-date-epoch` is set.
- `omit-history`: when `"true"`, buildah omits the build history block from the image config. Independent of any timestamp setting.

## Implementation notes

1. **The auto-source rule is deferred.** [ADR-0069](https://redhat.atlassian.net/browse/ADR-0069) wants the pipeline to default `SOURCE_DATE_EPOCH` to the clone task's `commit-timestamp` result when the user turns on `rewrite-timestamp` without picking an epoch. Tekton parameter values cannot express that fallback: there is no conditional in the expression language. A helper task cannot express it either. If the helper is gated by a `when` clause and gets skipped, every task consuming its result gets skipped too, which would take `build-container` down for every default build. If the helper always runs, every build pays for an extra pod to serve an opt-in feature. So the pipeline passes `source-date-epoch` through as given, and the fallback moves into konflux-build-cli as a follow-up. The CLI already warns when `rewrite-timestamp` shows up without an epoch. Until the follow-up, `rewrite-timestamp` only takes effect when `source-date-epoch` is also set, and the parameter descriptions say exactly that.

2. **The new parameters land at the end of the parameter lists.** The generated pipelines come from kustomize patches that address existing entries by JSON-patch index. Inserting at the top would shift every index and break the sibling patches, so the parameters are appended instead. The same kustomize inheritance explains why `core-services` and `fbc-builder` also pick up the parameters: both build on the docker-build bases. Their build tasks accept the same parameters, and the defaults keep behavior unchanged there too.

## Testing done

- Every changed YAML parses, and all six generated pipelines declare the three parameters with their documented defaults, wired into the build task.
- Verified the index-addressed downstream patches still target the right entries after the append (for example `fbc-builder` still drops `privileged-nested` and `buildah-format`).
- Re-running `./hack/build-manifests.sh` and `./hack/generate-pipelines-readme.py` changes nothing, which is what the manifest and README CI checks assert.
- Ran the PR's `docker-build-oci-ta` pipeline on a local Konflux kind cluster. The pipeline's `taskRef.version` fields were stripped because that field only makes sense to Konflux's bundle resolver, and plain Tekton rejects it. The conditional check tasks (`clair-scan` and the `sast-*` family) were removed, since they live outside this repo. The build path the PR wires (init, clone, prefetch, buildah, image index) ran unmodified.
- The verification itself: build the same pinned commit twice with `rewrite-timestamp: "true"`, `omit-history: "true"`, and `source-date-epoch` set to that commit's timestamp, the same value the deferred auto-source rule would choose. If the parameters were not reaching buildah, each run would stamp its own `created` time and its own file mtimes, and the two digests would differ. Instead, both scenarios reproduced exactly. The Alpine COPY-only image built to `sha256:0ba0cd63c1b48b68ed7a044e9b5869a243a90dd628bbfa2d8802baee62ce64ed` in both runs, and the Go multi-stage image built to `sha256:92921aef1af6008e59cd2e18c2b786853fbd9f935dcb293fa8fafe79a3b02dff` in both runs.

## Refs

- ADR: https://github.com/konflux-ci/architecture/blob/4d68b9cbcfc6e4d96f693713033bed086eca51fb/ADR/0069-reproducible-container-builds.md
- Task-side parameters: https://github.com/konflux-ci/build-definitions/pull/2947

### DIFF
--- a/pipelines/docker-build/docker-build.yaml
+++ b/pipelines/docker-build/docker-build.yaml
@@ -89,6 +89,22 @@ spec:
       VMs
     name: privileged-nested
     type: string
+  - default: ""
+    description: Sets the image created time and the SOURCE_DATE_EPOCH build argument.
+      On its own, it does not change file timestamps inside the layers (set rewrite-timestamp
+      to "true" for that). Leave empty to keep the actual build time.
+    name: source-date-epoch
+    type: string
+  - default: "false"
+    description: When "true", clamp file modification times in the image layers to
+      at most source-date-epoch. Does nothing unless source-date-epoch is set.
+    name: rewrite-timestamp
+    type: string
+  - default: "false"
+    description: When "true", omit the build history (history timestamps, layer metadata,
+      etc.) from the resulting image.
+    name: omit-history
+    type: string
   results:
   - name: IMAGE_URL
     value: $(tasks.build-image-index.results.IMAGE_URL)
@@ -171,6 +187,12 @@ spec:
       value: $(tasks.init.results.http-proxy)
     - name: NO_PROXY
       value: $(tasks.init.results.no-proxy)
+    - name: SOURCE_DATE_EPOCH
+      value: $(params.source-date-epoch)
+    - name: REWRITE_TIMESTAMP
+      value: $(params.rewrite-timestamp)
+    - name: OMIT_HISTORY
+      value: $(params.omit-history)
     runAfter:
     - prefetch-dependencies
     taskRef:
--- a/pipelines/core-services/core-services.yaml
+++ b/pipelines/core-services/core-services.yaml
@@ -151,6 +151,22 @@ spec:
     name: build-args-file
     type: string
   - default: ""
+    description: Sets the image created time and the SOURCE_DATE_EPOCH build argument.
+      On its own, it does not change file timestamps inside the layers (set rewrite-timestamp
+      to "true" for that). Leave empty to keep the actual build time.
+    name: source-date-epoch
+    type: string
+  - default: "false"
+    description: When "true", clamp file modification times in the image layers to
+      at most source-date-epoch. Does nothing unless source-date-epoch is set.
+    name: rewrite-timestamp
+    type: string
+  - default: "false"
+    description: When "true", omit the build history (history timestamps, layer metadata,
+      etc.) from the resulting image.
+    name: omit-history
+    type: string
+  - default: ""
     name: infra-deployment-update-script
   - default: ""
     name: update-repo-script
@@ -249,6 +265,12 @@ spec:
       value: $(tasks.init.results.http-proxy)
     - name: NO_PROXY
       value: $(tasks.init.results.no-proxy)
+    - name: SOURCE_DATE_EPOCH
+      value: $(params.source-date-epoch)
+    - name: REWRITE_TIMESTAMP
+      value: $(params.rewrite-timestamp)
+    - name: OMIT_HISTORY
+      value: $(params.omit-history)
     - name: IMAGE_APPEND_PLATFORM
       value: "true"
     runAfter:
--- a/pipelines/docker-build-multi-platform-oci-ta/README.md
+++ b/pipelines/docker-build-multi-platform-oci-ta/README.md
@@ -19,13 +19,16 @@ This pipeline is pushed as a Tekton bundle to [quay.io](https://quay.io/reposito
 |git-url| Source Repository URL| None| clone-repository:0.2:url|
 |hermetic| Execute the build with network isolation| false| build-images:0.10:HERMETIC|
 |image-expires-after| Image tag expiration time, time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively.| | clone-repository:0.2:ociArtifactExpiresAfter ; prefetch-dependencies:0.3:ociArtifactExpiresAfter ; build-images:0.10:IMAGE_EXPIRES_AFTER|
+|omit-history| When "true", omit the build history (history timestamps, layer metadata, etc.) from the resulting image.| false| build-images:0.10:OMIT_HISTORY|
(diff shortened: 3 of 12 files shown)


pipelines/docker-build/docker-build.yaml
```diff
@@ -171,6 +187,12 @@ spec:
       value: $(tasks.init.results.http-proxy)
     - name: NO_PROXY
       value: $(tasks.init.results.no-proxy)
+    - name: SOURCE_DATE_EPOCH
+      value: $(params.source-date-epoch)
+    - name: REWRITE_TIMESTAMP
+      value: $(params.rewrite-timestamp)
+    - name: OMIT_HISTORY
+      value: $(params.omit-history)
```
chmeliik, line 195: Note that none of the Pipeline-level changes will automatically propagate to the users who have already onboarded to Konflux and already have the pipelines in their repos.

The way to resolve this is to add a migration script for some Task that all the affected pipelines share, and make the necessary edits in that script. We typically use the init task (which has moved to the build-pipeline-tasks repo) for this purpose. See for example https://github.com/konflux-ci/build-pipeline-tasks/blob/main/task/init/migrations/0.4.1.sh.
(But note it had a bug: in [this section](https://github.com/konflux-ci/build-pipeline-tasks/blob/5cd1647a5a68a04b4b0505192220a3beb6624462/task/init/migrations/0.4.1.sh#L52), it doesn't make a distinction between Pipeline params and PipelineRun params, so it could incorrectly add Pipeline param definitions to a PipelineRun file. This had to be fixed in [migrations/0.4.2.sh](https://github.com/konflux-ci/build-pipeline-tasks/blob/main/task/init/migrations/0.4.2.sh).)

chmeliik (commented):
This seems reasonable. While I don't expect any users to set source-date-epoch manually, having the param shouldn't hurt.

I'd like some more info on the future plans though. Setting the epoch to the commit date makes much more sense to me, so I'd like to make sure that what we're doing here doesn't make that harder.

pipelines/docker-build/docker-build.yaml
JamesOlaitan: On future plans: I agree that the commit date is the right epoch, and that's where this is headed. It's what I also proposed in ADR-0069: when rewrite-timestamp is on, and the user didn't pick an epoch, default SOURCE_DATE_EPOCH to the commit timestamp. I couldn't wire that at the pipeline level because Tekton param values have no conditionals, and a helper task also fails for reasons stated in the PR description. So, the fallback moves into konflux-build-cli instead. This PR doesn't make any of that harder. An explicit source-date-epoch always takes precedence, and empty means "off" by default. When derivation lands, empty plus rewrite-timestamp starts meaning "derive". Nobody who set nothing sees a change, and nobody who set a value gets overridden.

On the migration: thanks for the pointers. I'll do a follow-up PR in build-pipeline-tasks bumping init with a migration script for this. I'll follow the 0.4.1.sh example and avoid the bug you pointed out.

JamesOlaitan: The PR for the migration script: https://github.com/konflux-ci/build-pipeline-tasks/pull/24

chmeliik: > the fallback moves into konflux-build-cli instead

How will konflux-build-cli get the commit timestamp? If the clone task is used with `mergeTargetBranch: true`, then the commit timestamp returned by the clone task will not match the commit timestamp you would get from the HEAD of the source repo. And I think we want the pre-merge timestamp, since the post-merge timestamp will be different every time.

Will we not still need to pass the commit timestamp from the git-clone task to the SOURCE_DATE_EPOCH param of the buildah task?

JamesOlaitan: We do need the pre-merge timestamp, and since konflux-build-cli uses commit hash instead of merge commit,  that's exactly what the clone task's commit-timestamp result holds.  So, the pipeline has to carry it to the buildah task.

Wiring it directly into SOURCE_DATE_EPOCH, though, would set the image created time on every build, which is an always-on behavior we don't want, since it would stamp builds that aren't actually reproducible. So, the plan is a separate param on the buildah task (say COMMIT_TIMESTAMP), wired from the clone result, and the CLI uses it as the epoch only when rewrite-timestamp is on and source-date-epoch is empty.

chmeliik: I'm not on board with making this behavior implicit, where setting rewrite-timestamp would also implicitly set source-date-epoch to the commit timestamp (and the only way to set source-date-epoch=\<commit timestamp\> would be to also enable rewrite-timestamp). But it might be reasonable to accept a special value for source-date-epoch, e.g. `:from-commit-timestamp:`.

We can think about the interface later. Either way, the two-param approach does look like the only reasonable way to do this :+1:

chmeliik (approved):

PR #3677, opened 2026-07-08
Replace deprecated buildah-task image with task-runner image

mmorhun:
Replace deprecated `buildah-task` image with `task-runner` image.
The change includes the following tasks:
- build-paketo-builder-oci-ta
- build-vm-image
- run-script-oci-ta

### DIFF
--- a/task/build-vm-image/0.2/build-vm-image.yaml
+++ b/task/build-vm-image/0.2/build-vm-image.yaml
@@ -5,7 +5,7 @@ metadata:
     tekton.dev/pipelines.minVersion: 0.12.1
     tekton.dev/tags: image-build, konflux
   labels:
-    app.kubernetes.io/version: "0.2"
+    app.kubernetes.io/version: "0.2.1"
     build.appstudio.redhat.com/build_type: disk-image
   name: build-vm-image
 spec:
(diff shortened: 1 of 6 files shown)


chmeliik (commented):
The tasks that don't already set `runAsUser: 0` will need that now I think.

build-vm-image may also need some additional adjustments to avoid breaking SSH: 8cb219a37e053771aa58370ea638c0c34ee50166
> - The new image sets HOME=/home/taskuser, which breaks the remote
  tasks. They create files in ~/.ssh, which now resolves to
  /home/taskuser/.ssh. But the ssh tool itself prefers /etc/passwd over
  $HOME and reads from /root/.ssh instead. Fix this by setting
  HOME=/root for the steps that run as root.

chmeliik (dismissed):

Allda (dismissed):

Victoremepunto (commented):
please could you add build-vm-image/0/2 as well ? we just merged this recently, I can also follow-up

chmeliik:
Fullsend is probably right about this being a problem:

> The `run-script` step embeds the `retry` function into a remote script via:
```bash
cat /usr/bin/retry | sed 1d >> scripts/script-build.sh
```

It never ceases to amaze me what people come up with in Bash scripts :exploding_head:

mmorhun:
Rebased, fixed the retry script path and also updated build-vm-image v0.2

task/build-vm-image/0.2/build-vm-image.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     tekton.dev/pipelines.minVersion: 0.12.1
     tekton.dev/tags: image-build, konflux
   labels:
-    app.kubernetes.io/version: "0.2"
+    app.kubernetes.io/version: "0.2.1"
```
chmeliik, line 8: For the tasks that have changelogs (this and run-script-oci-ta), it would be nice to update them so that there isn't a gap for this version.

chmeliik (commented):
Changelog update would be nice, otherwise LGTM

chmeliik (approved):

Victoremepunto (approved):

Allda (approved):

PR #3681, opened 2026-07-10
buildah*/0.10: add RHSM_MOUNT_CA_CERTS param

edcsnt:
Expose the `--rhsm-mount-ca-certs` option to Task consumers. This is a prerequisite to fix KFLUXSPRT-8389.

The MintMaker image [installs](https://github.com/konflux-ci/mintmaker-renovate-image/blob/ed85567f7a10f71828c0927a070a8f9d02ef9591/Dockerfile#L144-L145) the `subscription-manager-rhsm-certificates` package so that its CA certificates are made available to builds at runtime under `/etc/rhsm/ca`. [konflux-build-cli](https://github.com/konflux-ci/konflux-build-cli/blob/main/README.md#konflux-build-tasks-cli) introduced a new behavior of [mounting](https://github.com/konflux-ci/konflux-build-cli/blob/8bfe33041d7d7213c31bb28d583cedb32ed6f931/pkg/commands/build.go#L1576) at that directory at every `RUN` instruction, which causes the MintMaker Dockerfile to write files to the mount instead of the container filesystem. This parameter allows consumers to opt out of this automatic mount of `/etc/rhsm/ca` during `RUN` instructions so that CA certificates can persist in the resulting image.

### DIFF
--- a/task/buildah/0.10/buildah.yaml
+++ b/task/buildah/0.10/buildah.yaml
@@ -2,7 +2,7 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   labels:
-    app.kubernetes.io/version: "0.10.4"
+    app.kubernetes.io/version: "0.10.5"
     build.appstudio.redhat.com/build_type: "docker"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
@@ -68,7 +68,8 @@ spec:
     type: string
   - default: auto
     description: Mount /etc/rhsm/ca from the host machine into the build. Valid values
-      are 'always', 'auto' (default), 'never'.
+      are 'always', 'auto' (default), 'never'. Only effective if HERMETIC=false and
+      ACTIVATION_KEY or ENTITLEMENT_SECRET is set.
     name: RHSM_MOUNT_CA_CERTS
     type: string
   - name: ADDITIONAL_SECRET
--- a/task/buildah/0.10/buildah.yaml
+++ b/task/buildah/0.10/buildah.yaml
@@ -66,6 +66,11 @@ spec:
     default: activation-key
     description: Name of secret which contains subscription activation key
     type: string
+  - default: auto
+    description: Mount /etc/rhsm/ca from the host machine into the build. Valid values
+      are 'always', 'auto' (default), 'never'.
+    name: RHSM_MOUNT_CA_CERTS
+    type: string
   - name: ADDITIONAL_SECRET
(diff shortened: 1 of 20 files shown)


chmeliik (commented):
From the PR description I don't understand what issue this solves and how

chmeliik:
> From the PR description I don't understand what issue this solves and how

I think I understand now from https://github.com/konflux-ci/mintmaker-renovate-image/pull/494 and the internal user support thread. The MintMaker image wants to install the subscription-manager certificates package, which has the `/etc/rhsm/ca/*` files. But the MintMaker image build also enables the type of subscription-manager integration that causes the buildah task to mount `/etc/rhsm/ca:/etc/rhsm/ca`. This means that installing the certs package makes the certs go into a volume mount instead of going into the container filesystem, and the final container doesn't have them. And this option allows explicitly disabling the mount.

Do I understand that correctly? If yes, would you be able to update the commit message with more details?

task/buildah/0.10/buildah.yaml
```diff
@@ -66,6 +66,11 @@ spec:
     default: activation-key
     description: Name of secret which contains subscription activation key
     type: string
+  - default: auto
+    description: Mount /etc/rhsm/ca from the host machine into the build. Valid values
+      are 'always', 'auto' (default), 'never'.
```
chmeliik: It may be worth mentioning that this has no effect if neither the `ACTIVATION_KEY` secret nor the `ENTITLEMENT_SECRET` exists, or if `HERMETIC=true`. The build doesn't mount the RHSM certs in that case, regardless of the `RHSM_MOUNT_CA_CERTS` value

task/buildah/0.10/buildah.yaml
chmeliik: Please bump the task version to 0.10.5 and add an entry in CHANGELOG.md (for all the task variants)

chmeliik:
The PR has merge conflicts now

chmeliik (approved):

chmeliik:
Now Conforma is failing for unrelated reasons :disappointed: 

```yaml
- filepath: all_tasks-ec/0.1_modelcar-oci-ta.yaml
  success: false
  success-count: 9
  successes: null
  violations:
  - metadata:
      code: step_images.step_images_accessible
      term: quay.io/konflux-ci/release-service-utils:2f93b7ed6a2099e7187bb110a6b95caac3b8bdbc
    msg: Step 1 uses inaccessible image ref 'quay.io/konflux-ci/release-service-utils:2f93b7ed6a2099e7187bb110a6b95caac3b8bdbc'
```

chmeliik:
/retest

restored the deleted tag

PR #3694, opened 2026-07-13
fix: use per-command RETRY_MAX_TRIES override in build-vm-image

Victoremepunto:
## Summary
Follow the established retry pattern from `buildah-remote`: let the imported retry function's own defaults apply, and use a per-command env prefix (`RETRY_MAX_TRIES=6 retry ...`) for the subscription-manager call instead of `export`/`unset` which breaks under `set -u`.

## Root cause
The `export RETRY_MAX_TRIES=6` + `unset RETRY_MAX_TRIES` pair around subscription-manager registration removes the variable entirely. Under `set -euo pipefail`, subsequent `retry` calls (`podman pull`) fail with `RETRY_MAX_TRIES: unbound variable`.

The imported `retry` function already sets its own default (`: "${RETRY_MAX_TRIES=10}"`), so no manual export/unset is needed — just override per-command when a different value is desired, as `buildah-remote` does with `RETRY_STOP_IF_STDERR_MATCHES`.

## Test plan
- [ ] Verify disk image builds with activation-key pass end-to-end

### DIFF
--- a/task/build-vm-image/0.2/build-vm-image.yaml
+++ b/task/build-vm-image/0.2/build-vm-image.yaml
@@ -315,7 +315,6 @@ spec:
           mkdir -p $BUILD_DIR/rhsm-tmp/etc/pki/entitlement
           mkdir -p $BUILD_DIR/rhsm-tmp/etc/pki/consumer
 
-          export RETRY_MAX_TRIES=6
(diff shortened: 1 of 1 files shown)


Victoremepunto:
superseded by https://github.com/konflux-ci/build-definitions/pull/3698

PR #3700, opened 2026-07-14
CI: drop the MIGRATION.md check

chmeliik:
The MIGRATION.md files are no longer the primary mechanism to communicate changes in task versions to users. That role now belongs to CHANGELOG.md.

Decentralized task repos do not have the checks for MIGRATION.md existence anymore and, as explained in ADR 57 [1], they are not expected to be used anymore.

Remove the CI check that requires MIGRATION.md files and the related text in the README. We already have CI checks for CHANGELOG.md presence (though they are warning-level checks at the moment), as well as AI agent rules that AI review should enforce.

Also note that MintMaker will not be trying to link MIGRATION.md files anymore after https://github.com/konflux-ci/mintmaker/pull/569.

[1]: https://github.com/konflux-ci/architecture/blob/main/ADR/0054-task-versioning.md#mintmaker-dont-assume-each-xy-version-has-a-migrationmd

PR #3710, opened 2026-07-15
build-vm-image: add v0.3 with corrected qcow2 artifact type

thozza:
The v0.2 task advertised application/vnd.diskimage.qcow2.gzip as the OCI artifact type for qcow2 images, but never actually gzip-compressed the payload. The qcow2 format uses its own internal compression, so consumers trusting the artifact type and attempting to gunzip would get an error.

This is a breaking change for anything matching on the old artifact type, so it goes into a new minor version (0.3) rather than a patch on 0.2. The MIGRATION.md documents what downstream consumers need to update.

### DIFF
--- a/task/build-vm-image/0.3/MIGRATION.md
+++ b/task/build-vm-image/0.3/MIGRATION.md
@@ -1,0 +1,20 @@
+# Migration from 0.2 to 0.3
+
+## What changed
+
+The OCI artifact type for qcow2 disk images has been corrected from
+`application/vnd.diskimage.qcow2.gzip` to `application/vnd.diskimage.qcow2`.
+
+The previous artifact type was wrong - the task never gzip-compressed the
+qcow2 payload. The qcow2 format uses its own internal compression, so
+the `.gzip` suffix was misleading. Consumers that trusted the artifact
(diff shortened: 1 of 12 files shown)


Victoremepunto (approved):

task/build-vm-image/0.3/MIGRATION.md
```diff
@@ -0,0 +1,20 @@
+# Migration from 0.2 to 0.3
```
chmeliik, line 1: FYI MIGRATION.md files are deprecated, the CI will no longer require them after https://github.com/konflux-ci/build-definitions/pull/3700 and MintMaker will no longer link them on PRs

If any of this needs to be surfaced to users, please document it in CHANGELOG.md instead

thozza: Thanks for the note. I think that the migration requirement is sufficiently mentioned in the CHANGELOG.md. Do you insist on the removal of the MIGRATION.md, or can it be left there as is?

chmeliik: It can be left here

Victoremepunto (approved):

PR #3743, opened 2026-07-29
fix(ci): replace removed deploy-konflux.sh with deploy-local.sh

sanialisar:
## Summary
- `run-task-tests` called `./deploy-konflux.sh`, which no longer exists in the pinned `konflux-ci` checkout
- Switch to `./scripts/deploy-local.sh` (operator-based install) with `DEPLOY_LOCAL_SKIP_KIND=1`, `OPERATOR_INSTALL_METHOD=release`, and `SKIP_SECRETS=true`

### DIFF
--- a/.github/workflows/run-task-tests.yaml
+++ b/.github/workflows/run-task-tests.yaml
@@ -102,9 +102,16 @@ jobs:
 
       - name: Deploying Konflux
         if: steps.tasks-to-be-tested.outputs.tasklist != ''
+        env:
+          # Kind cluster already created by kind-action above.
+          DEPLOY_LOCAL_SKIP_KIND: "1"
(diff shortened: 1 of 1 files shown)


sanialisar:
Addressed the reproducibility review comment:

- Bumped the `konflux-ci` checkout pin to `d510887fc3f38a8f8f4f4d5540f5b7199d4c7f67` (needed for `OPERATOR_RELEASE` support in `deploy-local.sh`)
- Set `OPERATOR_RELEASE: v0.1.13` so operator + CR install is pinned instead of floating `latest`

Chose approach (1) over checkout/`local` install because `make deploy` pulls in `controller-gen` and is heavier than this workflow needs.

PR #3751, opened 2026-07-29
fix(modelcar-oci-ta): fix Kind registry TLS and task test harness

sanialisar:
## Summary
After https://github.com/konflux-ci/build-definitions/pull/3748 (`deploy-local.sh`), `run-task-tests` actually reaches `modelcar-oci-ta` tests. They were failing against `registry-service.kind-registry` and blocking bundle builds.

### Production task (`modelcar-oci-ta.yaml`)
- Declare missing `caTrustConfigMapName` / `caTrustConfigMapKey` params (defaults match other tasks: `trusted-ca` / `ca-bundle.crt`)
- Set `CA_FILE` and export `SSL_CERT_FILE` only when the mounted CA bundle is non-empty (oras / cosign / mobster)
- Optional `INSECURE_REGISTRY` support for oras/cosign TLS opts — **off by default**; production behavior unchanged unless explicitly set

### Test harness only
- Kind registry CA is not reliably trusted under deploy-local CI even with `SSL_CERT_FILE`, so tests use `--insecure` / `--tls-verify=false` / `--allow-insecure-registry`
- `pre-apply-task-hook.sh` sets `INSECURE_REGISTRY=true` and `ORAS_OPTIONS=--insecure` on the **test copy** of the task only
- Mock model push: push individual files (directory push collapsed to one layer)
- Verify expectations: oras titles layers with basenames, so files land at `models/model.bin` (not `models/source/models/model.bin`)

## Test plan
- [x] `run-task-tests` gets past `mock-model` TLS errors

### DIFF
--- a/task/modelcar-oci-ta/0.1/modelcar-oci-ta.yaml
+++ b/task/modelcar-oci-ta/0.1/modelcar-oci-ta.yaml
@@ -80,6 +80,15 @@ spec:
       description: CPE identifier label required for on-prem product releases
       type: string
       default: ""
+    - name: caTrustConfigMapKey
+      description: The name of the key in the ConfigMap that contains the
+        CA bundle data.
+      type: string
+      default: ca-bundle.crt
+    - name: caTrustConfigMapName
+      description: The name of the ConfigMap to read CA bundle data from.
+      type: string
+      default: trusted-ca
   results:
     - name: IMAGE_DIGEST
       description: Digest of the artifact just pushed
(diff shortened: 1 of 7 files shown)


sanialisar:
Follow-up: `SSL_CERT_FILE` alone was not enough — same TLS failure. Under deploy-local CI the Kind `trusted-ca` bundle does not appear to make `oras` trust `registry-service.kind-registry`.

Pushed a second commit that:
- uses `oras --insecure` / `skopeo --tls-verify=false` / `cosign --allow-insecure-registry` in the test pipelines
- sets `ORAS_OPTIONS=--insecure` for build-trusted-artifacts
- gates the task itself on `INSECURE_REGISTRY=true` via `pre-apply-task-hook.sh` (tests only; production unchanged)

sanialisar:
Closing in favor of https://github.com/konflux-ci/build-definitions/pull/3752 (`markVaykhansky:fix-modelcar-tests-ssl-cert`), which carries the same modelcar TLS/test-harness fixes and is merged up to date with `main`.

PR #3756, opened 2026-07-30
Implement crypto-scanner tasks

sguarin:
# Implement crypto scanner tasks

This implements a POC of the long term solution for running the crypto scanner in konflux. See https://docs.google.com/document/d/1zvAc8hakVwI4h3xRwS2ac9bJPd3mnWIcrzAyynLYOK8/edit?tab=t.0#heading=h.66y4kqbj468a

A private image with proprietary rules is being pulled using oidc federation.
It requires to maintain an acl per konflux build service account until [RFE-9645](https://redhat.atlassian.net/browse/RFE-9645) gets implemented.

[RFE-9645]: https://redhat.atlassian.net/browse/RFE-9645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


task/sast-crypto-scan-oci-ta/0.1/sast-crypto-scan-oci-ta.yaml
```diff
@@ -0,0 +1,284 @@
...
+        mkdir /workspace/rules
+        retry oc image extract "${ADDITIONAL_RULES_IMAGE}" --path="/rules/:/workspace/rules"
+        if [ $? -ne 0 ]; then
+          echo "Unable to extract image"
+          exit 0
+        fi
+
+    - name: rh-crypto-scanner
+      # TEST IMAGE, it should be stored under konflux-ci and pinned
+      image: quay.io/exd-guild-security/crypto-finder-image:latest
```
sguarin, line 148: This need to be pinned or possible adjusted to konflux-ci space in case https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/20696 gets accepted, and the image released in that space.

gbenhaim:
We are trying to move away from inline scripts since those pile up in etcd and impact the cluster performance. 
Can you please place the scripts inside of a container image?

avi-biton:
As part of the decentralization of build-definitions we introduced a flow to build tekton tasks as bundles in other repositories and use them as external tasks in build-definitions.

[Building the tasks](https://konflux.pages.redhat.com/docs/users/end-to-end/building-tekton-tasks.html)
[Using external-task](https://github.com/konflux-ci/build-definitions#external-tasks)

I don't think we want to add new tasks to build-definitions.

sguarin:
> As part of the decentralization of build-definitions we introduced a flow to build tekton tasks as bundles in other repositories and use them as external tasks in build-definitions.
> 
> [Building the tasks](https://konflux.pages.redhat.com/docs/users/end-to-end/building-tekton-tasks.html) [Using external-task](https://github.com/konflux-ci/build-definitions#external-tasks)
> 
> I don't think we want to add new tasks to build-definitions.

Do you recommend to create a new repo for this, or does it fit in another already descentralized repository with tasks?

cc @gbenhaim

sguarin:
> We are trying to move away from inline scripts since those pile up in etcd and impact the cluster performance. Can you please place the scripts inside of a container image?

every scripting? even the step related to oidc authentication?

avi-biton:
> > As part of the decentralization of build-definitions we introduced a flow to build tekton tasks as bundles in other repositories and use them as external tasks in build-definitions.
> > [Building the tasks](https://konflux.pages.redhat.com/docs/users/end-to-end/building-tekton-tasks.html) [Using external-task](https://github.com/konflux-ci/build-definitions#external-tasks)
> > I don't think we want to add new tasks to build-definitions.
> 
> Do you recommend to create a new repo for this, or does it fit in another already descentralized repository with tasks?
> 
> cc @gbenhaim

If you already have a repository containing tasks you can use it for this task as well

gbenhaim:
> > We are trying to move away from inline scripts since those pile up in etcd and impact the cluster performance. Can you please place the scripts inside of a container image?
> 
> every scripting? even the step related to oidc authentication?

The preference will be to place any script inside of a container image.

PR #3758, opened 2026-07-30
Mark enterprise-contract pipeline as deprecated

jsmid1:
The source of truth for this pipeline has moved to conforma/cli and is published to conforma/tekton-catalog. This copy is retained for backward compatibility.

Ref: https://issues.redhat.com/browse/EC-1942

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -1,3 +1,9 @@
+# DEPRECATED: The source of truth for this pipeline has moved to:
+#   https://github.com/conforma/cli/blob/main/pipelines/enterprise-contract/0.1/enterprise-contract.yaml
+# It is also published to:
+#   https://github.com/conforma/tekton-catalog/blob/main/pipelines/enterprise-contract/0.1/enterprise-contract.yaml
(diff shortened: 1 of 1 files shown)


simonbaird (approved):
Lgtm.

pipelines/enterprise-contract.yaml
```diff
@@ -1,3 +1,9 @@
+# DEPRECATED: The source of truth for this pipeline has moved to:
+#   https://github.com/conforma/cli/blob/main/pipelines/enterprise-contract/0.1/enterprise-contract.yaml
+# It is also published to:
+#   https://github.com/conforma/tekton-catalog/blob/main/pipelines/enterprise-contract/0.1/enterprise-contract.yaml
+# This copy is retained for backward compatibility but will not receive updates.
```
simonbaird, line 5: Just for the record, it's possible we might want to update this to keep it synced with the new source of truth, even though it will be deprecated.

(No need to change the comment IMO, but I figured it's worth mentioning this possibility.)

simonbaird (approved):
Lgtm.

PR #3779, opened 2026-08-11
fix(modelcar-oci-ta): set name label for container-first security checks

sanialisar:
## Summary
- ModelCar images currently inherit `name=ubi9/ubi-micro` from the base image, which will fail `check-labels` in `rh-advisories` when `enforceContainerFirstSecurityLabels` defaults to true (Aug 23).
- Set the `name` label from `IMAGE_NAME`, or when unset to `IMAGE_NAME_PREFIX/<IMAGE repository basename>` (default prefix `rhai`).
- `cpe` was already set via the existing `CPE` param; this change addresses the mismatched `name` label.

## Test plan
- [ ] Task tests pass
- [ ] Build a ModelCar and confirm `name` is e.g. `rhai/modelcar-…` (not `ubi9/ubi-micro`)
- [ ] For components released to `rhelai1/…`, pass `IMAGE_NAME_PREFIX=rhelai1` or full `IMAGE_NAME=rhelai1/…`
- [ ] Confirm a release pipeline `check-labels` no longer warns/fails on `name`

### DIFF
--- a/task/modelcar-oci-ta/0.1/modelcar-oci-ta.yaml
+++ b/task/modelcar-oci-ta/0.1/modelcar-oci-ta.yaml
@@ -80,6 +80,20 @@ spec:
       description: CPE identifier label required for on-prem product releases
       type: string
       default: ""
+    - name: IMAGE_NAME
+      description: >-
+        Value for the image "name" label (must match the release destination
+        repository path, e.g. rhai/modelcar-foo). Required for rh-advisories
+        check-labels / container-first security label enforcement. When empty,
+        defaults to IMAGE_NAME_PREFIX/<repository-name> derived from IMAGE.
+      type: string
+      default: ""
+    - name: IMAGE_NAME_PREFIX
+      description: >-
+        Registry namespace prefix used when IMAGE_NAME is empty
+        (e.g. rhai or rhelai1). Combined with the IMAGE repository basename.
+      type: string
+      default: "rhai"
     - name: caTrustConfigMapKey
       description: The name of the key in the ConfigMap that contains the
         CA bundle data.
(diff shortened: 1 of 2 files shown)


task/modelcar-oci-ta/0.1/modelcar-oci-ta.yaml
```diff
@@ -416,17 +434,31 @@ spec:
...
         config_path="$BLOBS_DIR/$(jq -r '.config.digest | sub("^sha256:"; "")' "$manifest_path")"
 
+        # Override inherited base "name" (e.g. ubi9/ubi-micro) so check-labels
+        # matches the catalog destination path (rhai/... or rhelai1/...).
+        if [ -n "${IMAGE_NAME}" ]; then
+          NAME_LABEL="${IMAGE_NAME}"
+        else
+          IMAGE_REPO="${IMAGE%%@*}"
+          IMAGE_REPO="${IMAGE_REPO%%:*}"
+          IMAGE_BASENAME="${IMAGE_REPO##*/}"
```
markVaykhansky, line 444: By this point the `org name` of the image will be `org-IMAGE` and not `org/IMAGE`, right? Otherwise there's a bug there

markVaykhansky: Also, if Gemini doesn't lie than:
`registry.example.com/myproject/ubuntu` --> `ubuntu` after this line.
But in the email, the example is:
`registry.redhat.io/openshift-gitops-1/gitops-rhel8-operator` --> `openshift-gitops-1/gitops-rhel8-operator` i.e we would need to keep `rhai/` which is currently dropped. Or am I missing something?

sanialisar: Yes,  catalog paths are flat `rhai/<image>` (e.g. `rhai/redhatai-qwen3-8b`), not nested `rhai/org/image`

So `IMAGE_BASENAME` is the `redhatai-qwen3-8b` and `IMAGE_NAME_PREFIX` is default to `rhai`
then we combain it like this: `NAME_LABEL="${IMAGE_NAME_PREFIX}/${IMAGE_BASENAME}"`
to get `rhai/redhatai-qwen3-8b`.

PR #3782, opened 2026-08-11
Pin conforma tekton task to digest

joejstuart:
Pin this to a digest so we can control when changes are released. This is to sync the changes with the release-service. Without this change, the integration service can end up with changes a couple of weeks before the release service.

### DIFF
--- a/pipelines/enterprise-contract.yaml
+++ b/pipelines/enterprise-contract.yaml
@@ -100,7 +100,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
-            value: quay.io/conforma/tekton-task:konflux
+            value: quay.io/conforma/tekton-task:konflux@sha256:13375ed6d012614f030d36fa6485c16db1ab1e82d6077d5055c0eb1bfb90b644
           - name: name
             value: collect-keyless-params
           - name: kind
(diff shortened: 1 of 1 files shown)


pipelines/enterprise-contract.yaml
```diff
@@ -100,7 +100,7 @@ spec:
         resolver: bundles
         params:
           - name: bundle
-            value: quay.io/conforma/tekton-task:konflux
+            value: quay.io/conforma/tekton-task:konflux@sha256:13375ed6d012614f030d36fa6485c16db1ab1e82d6077d5055c0eb1bfb90b644
```
simonbaird, line 103: Seems to pin a 7 weeks old cli. That okay?

simonbaird (approved):

pipelines/enterprise-contract.yaml
simonbaird: I think yes. See also [changelog (for what is coming)](https://github.com/joejstuart/infra-deployments-ci/blob/1d6782bd4f04901cd85346c124f5077a157da24a/releases/2026-08-11T15%3A14%3A53/changelog.md).

PR #3783, opened 2026-08-12
fix(modelcar-oci-ta): set name label for container-first security checks

markVaykhansky:
## Summary
- ModelCar images currently inherit `name=ubi9/ubi-micro` from the base image, which will fail `check-labels` in `rh-advisories` when `enforceContainerFirstSecurityLabels` defaults to true (Aug 23).
- Set the `name` label from `IMAGE_NAME`, or when unset to `IMAGE_NAME_PREFIX/<IMAGE repository basename>` (default prefix `rhai`).
- `cpe` was already set via the existing `CPE` param; this change addresses the mismatched `name` label.

## Test plan
- [ ] Task tests pass
- [ ] Build a ModelCar and confirm `name` is e.g. `rhai/modelcar-…` (not `ubi9/ubi-micro`)
- [ ] For components released to `rhelai1/…`, pass `IMAGE_NAME_PREFIX=rhelai1` or full `IMAGE_NAME=rhelai1/…`
- [ ] Confirm a release pipeline `check-labels` no longer warns/fails on `name`


sanialisar (approved):
LGTM

PR #3785, opened 2026-08-12
fix(modelcar-oci-ta): merge base-image Syft packages into SBOM

markVaykhansky:
## Summary
- Modelcar SBOMs currently only contain the composition graph (modelcar + base + model), so release has no RPM inventory to load into the package DB.
- Scan the **base image** with Syft during `sbom-generate` and merge those packages into the Mobster modelcar SBOM (keep existing descendant relationships).
- Extend the happy-path test to assert `pkg:rpm/...` purls are present.

## Test plan
- [x] `run-task-tests` / modelcar-oci-ta happy-path passes (composition purls + RPM purls)
- [x] Rebuild a Modelcar with the new task bundle and confirm `cosign download sbom … | jq '.packages[].name'` includes base RPMs (e.g. bash/glibc) in addition to the three composition nodes

Made with [Cursor](https://cursor.com)


sanialisar (approved):
LGTM

PR #3800, opened 2026-08-20
Conditional releases for tekton bundle builder pipeline

slimreaper35:
Tested in https://github.com/slimreaper35/konflux-test-bundle

### DIFF
--- a/task/tkn-bundle/0.2/tkn-bundle.yaml
+++ b/task/tkn-bundle/0.2/tkn-bundle.yaml
@@ -41,6 +41,10 @@ spec:
   - name: REVISION
     type: string
     description: Revision
+  - name: RELEASE_ONLY_IF_VERSION_BUMPED
+    type: string
+    description: Release the task bundle only when the app.kubernetes.io/version is different
+    default: "true"
   results:
   - description: Digest of the image just built
     name: IMAGE_DIGEST
@@ -197,6 +201,8 @@ spec:
       value: $(params.URL)
     - name: REVISION
       value: $(params.REVISION)
+    - name: RELEASE_ONLY_IF_VERSION_BUMPED
+      value: $(params.RELEASE_ONLY_IF_VERSION_BUMPED)
     computeResources:
       limits:
         memory: 64Mi
--- a/task/tkn-bundle/0.2/tkn-bundle.yaml
+++ b/task/tkn-bundle/0.2/tkn-bundle.yaml
@@ -52,6 +52,8 @@ spec:
     name: IMAGE_URL
   - description: Image reference of the built image
     name: IMAGE_REF
+  - description: Whether the task bundle should be released
+    name: SHOULD_RELEASE
   stepTemplate:
     env:
     - name: HOME
@@ -386,6 +388,7 @@ spec:
       echo -n "$IMAGE" >"$(results.IMAGE_URL.path)"
       echo -n "${digest}" >"$(results.IMAGE_DIGEST.path)"
       echo -n "${IMAGE_REF}" >"$(results.IMAGE_REF.path)"
+      echo -n "true" >"$(results.SHOULD_RELEASE.path)"
 
       # cleanup task file
       # shellcheck disable=SC2153
--- a/task/tkn-bundle/0.2/tkn-bundle.yaml
+++ b/task/tkn-bundle/0.2/tkn-bundle.yaml
@@ -265,9 +265,30 @@ spec:
         yq ".metadata.labels.\"${LABEL_ACTUAL_TASK_VERSION}\"" "$task_file" | sed '/null/d' | tr -d '[:space:]'
       }
 
+      get_previous_task_version() {
+        local -r task_file=$1
+        local -r task_relpath="${task_file#*/"${SOURCE_CODE_DIR}"/}"
+
+        local previous_yaml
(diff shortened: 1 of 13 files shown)


task/tkn-bundle/0.2/tkn-bundle.yaml
```diff
@@ -265,10 +265,31 @@ spec:
         yq ".metadata.labels.\"${LABEL_ACTUAL_TASK_VERSION}\"" "$task_file" | sed '/null/d' | tr -d '[:space:]'
       }
 
+      get_previous_task_version() {
+        local -r task_file=$1
+        local -r task_relpath="${task_file#*/"${SOURCE_CODE_DIR}"/}"
+
+        local previous_yaml
+        if ! previous_yaml=$(git -C "${SOURCE_CODE_DIR}" show "HEAD^:${task_relpath}"); then
```
chmeliik, line 273: So this will only work properly in repos that use merge commits or squash merging, right? Not for rebase merging, because it wouldn't pick up versions that changed in commits prior to the last one.

Is there some reliable way to figure out the base sha and get the version number from that?

slimreaper35: > So this will only work properly in repos that use merge commits or squash merging, right? Not for rebase merging, because it wouldn't pick up versions that changed in commits prior to the last one.

Yes, the only problem is when a commit that bumps the version of a task isn't the last one in a series with rebase merging. (It should be probably documented as a limitation)

> Is there some reliable way to figure out the base sha and get the version number from that?

I haven't found one.

chmeliik: > when a commit that bumps the version of a task isn't the last one in a series with rebase merging

I think this will be a common case. Unlike the task-runner repo, where a release PR is expected to only bump the VERSION file, PRs in task repos will often end up with the version bump not being the last commit

chmeliik: > Is there some reliable way to figure out the base sha and get the version number from that

Is there a way to use the PaC `{target_branch}` variable for that? Use git to find the latest sha on the target branch?

This probably wouldn't work in PR pipelines from forks, but that should be ok, this only needs to work in push pipelines

Edit: oh, it's probably too late in the push pipeline, since the target branch will already have the changes :facepalm:

avi-biton:
### Question: Renovate/MintMaker image digest update without a version bump

Consider this scenario:

Renovate (MintMaker) opens a PR that bumps a step image digest in the task YAML - e.g. updating the `task-runner` image to pick up a newer base with security patches. Per the repo conventions, automated image digest bumps [do not require a version bump](https://github.com/konflux-ci/build-definitions/blob/main/.cursor/rules/tekton.mdc):

> Automated image digest bumps do not necessarily require a version bump or changelog entry.

**What happens with this PR's logic:**

1. `actual_task_version` = `"0.2.4"` (current commit — digest changed, version label unchanged)
2. `previous_task_version` = `"0.2.4"` (from `HEAD^`)
3. `"0.2.4" != "0.2.4"` → **false**
4. `SHOULD_RELEASE` = `"false"` → release is skipped

The task content changed (new image digest), but because the version label didn't change, the integration-service would not trigger a release. The updated image would not be rolled out.

**Example timeline:**

```text
commit A: task/foo/0.2/foo.yaml  →  version: "0.2.4", image: quay.io/task-runner@sha256:aaa...
commit B: task/foo/0.2/foo.yaml  →  version: "0.2.4", image: quay.io/task-runner@sha256:bbb...
                                     ↑ MintMaker digest bump
                                     ↑ SHOULD_RELEASE = "false" — new image not released
                                     
Is this an expected trade-off (MintMaker PRs must also bump the version for the bundle to be released)? Or should the feature also account for content changes without a version bump - e.g. by comparing the full task YAML hash instead of just the version label?

**Comment by slimreaper35** [comment]
IIUC, that's not a trade-off, but expected behavior based on [ADR 54](https://github.com/konflux-ci/architecture/blob/main/ADR/0054-task-versioning.md) to reduce unclear image digest updates and always provide some information in the CHANGELOG.md for a particular task version.

For security patches, that means the same process. Conforma can than reject a certain version range of a task in https://github.com/release-engineering/rhtap-ec-policy as it happened recently for prefetch-dependencies.

cc @chmeliik @cursor

File: task/tkn-bundle/0.2/tkn-bundle.yaml [hunk 2]
```diff
@@ -271,12 +300,26 @@ spec:
...
         printf "Label %s is not set in task file %s\n" "$LABEL_ACTUAL_TASK_VERSION" "${FILES[0]}" >&2
         exit 1
       fi
 
+      SHOULD_RELEASE="false"
+      if [[ "$RELEASE_ONLY_IF_VERSION_BUMPED" == "true" ]]; then
+        previous_task_version=$(get_previous_task_version "${original_task_file}")
+        echo "Found previous task version: $previous_task_version"
+
+        if [[ "$actual_task_version" != "$previous_task_version" ]]; then
```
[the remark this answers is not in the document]
slimreaper35: fixed

slimreaper35:
Closing. ~Un~fortunately the Tekton task and builder pipeline will be moved to https://github.com/konflux-ci/container-build-catalog and https://github.com/konflux-ci/konflux-build-cli soon so it makes more sense to implement this feature in Go.

/fs-review Sorry for the burnt tokens.

PR #3803, opened 2026-08-23
Add verify-reproducibility pipeline and task

JamesOlaitan:
## Summary

Implements the standalone verification pipeline from ADR-0069. It's a pipeline rather than a task since a task can't invoke another task to trigger the rebuild.

The rebuild pushes to a caller-supplied scratch repository, not the repo of the image being verified. buildah-oci-ta signs every push unconditionally whenever the cluster has keyless signing configured, with no param to skip it, and cosign's signature tag is repo-scoped. A same-repo rebuild would overwrite the verified image's own signature and SBOM/attestation tags as a side effect of inspecting it.

A matching digest proves the same source, params, and task version produce the same bytes on this platform, not that an independent platform would agree. Verifying on the same build platform that produced the original image can't catch that platform itself being compromised in a way that's still internally deterministic. The ADR names this limitation and points to independent-platform rebuilding from provenance as the stronger property to grow into later.

Single-arch images only for now. This pipeline rebuilds through one buildah-oci-ta run, so it can't produce a matching digest for a multi-platform image index. Multi-arch manifest ordering is still open work per the ADR.

diffoci integration for layer-by-layer diffs on mismatch is left for a follow-up. It isn't vendored anywhere in this repo today, and bringing in the first non-konflux-ci step image is its own conversation.

## Testing

Task-level tests (matching digest, mismatched digest, malformed ref) run automatically in CI. YAML structure, kustomize builds, CODEOWNERS, and shellcheck were all validated locally.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -138,6 +138,10 @@
 # renovate groupName=verify-source
 /task/verify-source @arewm @ralphbean
 
+# renovate groupName=verify-reproducibility
+/task/verify-reproducibility @chmeliik @Allda @jedinym
+/pipelines/verify-reproducibility @chmeliik @Allda @jedinym
+
 # renovate groupName=ko-builder
 /task/ko-oci-ta @ggallen @vdemeester @arewm @skoved @rh-hemartin @maruiz93
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -140,6 +140,7 @@
 
 # renovate groupName=verify-reproducibility
 /task/verify-reproducibility @chmeliik @Allda @jedinym
+/task/verify-distinct-repositories @chmeliik @Allda @jedinym
 /pipelines/verify-reproducibility @chmeliik @Allda @jedinym
 
 # renovate groupName=ko-builder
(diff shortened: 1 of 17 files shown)


chmeliik (commented):
The pipeline is a nice demo of one possible way to go about verifying reproducibility, but I don't think we want to merge this.

As I see it, the pipeline doesn't have much value. It's not seamless enough to be worth using. A user would have to dig into the original build to find the artifact references, set up a scratch repo to push to, and find correct values for every single param this pipeline exposes.

It also has poor maintainability, since it relies on being tightly in sync with the pipeline it's trying to reproduce.

With nobody having a real incentive to use this and no tests exercising the pipeline, it would become abandoned the moment it gets merged.

As an alternative, I would suggest adding a simplified version of this pipeline as an "attachment" to the ADR (e.g. hidden in an html-markdown `<details>` element). I think that would sufficiently fulfill the purpose of showing one example of how to go about reproducing a build.

tnevrlka:
How about storing the pipeline in James's repository? If a user wanted to use it, I guess they could through `kubectl apply -f`, right?

JamesOlaitan:
I'll close this PR and continue working on it in my repo for anyone who wants to use it (maybe as a subcommand that pulls from SLSA provenance attestation – so users won't have to manually enter parameters – to clone, rebuild locally, and compare without pushing to a registry). I'll also update the ADR to include details on the build-twice-and-compare pipeline approach.

PR #3810, opened 2026-08-25
fix(build-helm-chart-oci-ta): skip file:// deps in repo add

rcerven:
The dependency-handling loop only skipped oci:// dependencies from the helm repo add path. For file:// dependencies (valid Helm local path deps), the code fell into the else branch, set http_repo_used=true, and then helm repo update failed because no real HTTP repos were added — aborting before helm dependency build was ever called.

Add file:// to the protocol skip condition alongside oci://. Both protocol types are natively handled by helm dependency build and should not go through the helm repo add path.

Assisted-by: Claude Code


AjayJagan:
I have addressed the suggestions and opened a new PR here: https://github.com/konflux-ci/build-definitions/pull/3812

amisstea:
Closing in favor of #3812.

PR #3814, opened 2026-08-26
feat!: respect Chart.yaml version with timestamp (0.4.0)

rhartman93:
BREAKING CHANGE: The task now respects the version field in Chart.yaml when present, appending a timestamp as a pre-release identifier to ensure uniqueness.

Version priority order:
1. CHART_VERSION parameter (full override)
2. Chart.yaml version field with timestamp (NEW)
3. Git-based calculation (fallback)

This prevents race conditions from concurrent builds with the same Chart.yaml version and enables dynamic tagging via {{ oci_version }} in ReleasePlanAdmissions.

Migration required: Users with placeholder versions in Chart.yaml (e.g., version: 0.1.0) that rely on git-based calculation must remove the version field from Chart.yaml or update it to their actual product version.

Example: Chart.yaml version 1.0.0 becomes 1.0.0-1724684400, which Helm stores in org.opencontainers.image.version annotation.

Related: [KONFLUX-15549](https://redhat.atlassian.net/browse/KONFLUX-15549)

Assisted-by: Claude

[KONFLUX-15549]: https://redhat.atlassian.net/browse/KONFLUX-15549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
+++ b/task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
@@ -1,0 +1,383 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: build-helm-chart-oci-ta
+  labels:
+    app.kubernetes.io/version: "0.4.0"
+spec:
+  description: |-
+    The task packages and pushes a Helm chart to an OCI repository.
+    As Helm charts require to have a semver-compatible version to be packaged, the
+    task relies on git tags in order to determine the chart version during runtime.
+
+    The task computes the version based on the git commit SHA distance from the latest
+    tag prefixed with the value of TAG_PREFIX. The value of that tag will be used as
+    the version's X.Y values, and the Z value will be computed by the commit's distance
+    from the tag, followed by an abbreviated SHA as build metadata.
+    If the CHART_VERSION parameter is provided, that value is used as-is and version calculation is skipped.
+
+    The task also supports image substitution in the chart templates. Use the IMAGE_MAPPINGS
+    parameter to specify source images to be replaced with target images and tags.
+
+    Version 0.2 includes improved digest extraction from skopeo copy operations.
+  params:
+    - name: CA_TRUST_CONFIG_MAP_KEY
+      description: The name of the key in the ConfigMap that contains the
+        CA bundle data.
+      type: string
+      default: ca-bundle.crt
+    - name: CA_TRUST_CONFIG_MAP_NAME
+      description: The name of the ConfigMap to read CA bundle data from.
+      type: string
+      default: trusted-ca
+    - name: CHART_CONTEXT
+      description: Path relative to SOURCE_CODE_DIR where the chart is located
+      default: dist/chart/
+    - name: COMMIT_SHA
+      description: Git commit sha to build chart for
+    - name: IMAGE_MAPPINGS
+      description: |-
+        JSON array of image mappings to substitute in chart templates.
+        Format: [{"source": "localhost/my/repo", "target": "quay.io/myorg/myapp"}]
+        Source images will be replaced with target images in all YAML files in templates/.
+        The task automatically appends the tag format: VERSION_SUFFIX-COMMIT_SHA (or just COMMIT_SHA if VERSION_SUFFIX is empty).
+      default: "[]"
+    - name: IMAGE
+      description: Full image reference with tag (e.g., quay.io/redhat-user-workloads/konflux-vanguard-tenant/caching/squid:on-pr-{{revision}})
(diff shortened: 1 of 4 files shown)


task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
```diff
@@ -0,0 +1,383 @@
...
+
+        # Create a chart version
+        # Priority order:
+        # 1. CHART_VERSION parameter if provided (full override)
+        # 2. Chart.yaml version field + timestamp (respects user input, ensures uniqueness)
+        # 3. Git-based calculation (fallback for repos without Chart.yaml version)
+        if [ -n "$CHART_VERSION" ]; then
+          echo "Using provided chart version: $CHART_VERSION"
+          chart_version="$CHART_VERSION"
+        elif [ -f "Chart.yaml" ] && yq -e '.version' Chart.yaml >/dev/null 2>&1; then
```
amisstea, line 252: The `version` field is required in `Chart.yaml` so the else case after this won't ever be reachable.

I think a better approach would be to use a Konflux specific annotation in the chart to drive the behavior around versioning.

task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
```diff
@@ -0,0 +1,383 @@
...
+          base_version=$(yq -r '.version' Chart.yaml)
+          timestamp=$(date +%s)
```
amisstea, line 254: There's still a minor risk of collision with this. We should use UTC and higher precision. E.g. `date -u +%s%N`.

task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
```diff
@@ -0,0 +1,383 @@
...
+      default: helm-
+    - name: VALUES_FILES
+      description: Array of values file names to process for image substitution
+        (e.g., ["values.yaml", "values-prod.yaml", "values-dev.yaml"])
+      type: array
+      default:
+        - values.yaml
+    - name: VERSION_SUFFIX
+      description: A suffix to be added to the version string
+      default: ""
```
amisstea, line 68: The `VERSION_SUFFIX` in unaccounted for with the new versioning path. We either need to document it more clearly or incorporate it.

task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
```diff
@@ -0,0 +1,383 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: build-helm-chart-oci-ta
+  labels:
+    app.kubernetes.io/version: "0.4.0"
+spec:
+  description: |-
```
amisstea, line 9: This description and similar documentation in the README need to be updated.

task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
amisstea: I also don't think we'll need a version bump for this task if we use the annotation approach.

task/build-helm-chart-oci-ta/0.4/build-helm-chart-oci-ta.yaml
```diff
@@ -0,0 +1,383 @@
...
+    - name: workdir
+      emptyDir: {}
+  stepTemplate:
+    imagePullPolicy: IfNotPresent
+    volumeMounts:
+      - mountPath: /var/workdir
+        name: workdir
+  steps:
+    - name: use-trusted-artifact
+      image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:9b180776a41d9a22a1c51539f1647c60defbbd55b44bbebdd4130e33512d8b0d
```
slimreaper35, line 104:
```suggestion
      image: quay.io/konflux-ci/build-trusted-artifacts:latest@sha256:7b4f5bbf8b8aa663beaf18bfef4fd16ba1cb5ce846d1fc776a64b312c52ee28c
```

yftacherzog (changes requested):
There's a go-based implementation of the task that should address this issue. Example reference: quay.io/konflux-ci/tekton-catalog/task-build-helm-chart-oci-ta:0.4@sha256:bcc69ce9ecf786df7876eb4a47ff7835d4ed85121d8f55dc4531ff955c1095b9

I think this one can be closed. Any remaining issues can be raised in tools or tekton-tools repos.

PR #3825, opened 2026-09-07
Delete another set of deprecated/expired build tasks

slimreaper35:


renovate.json
```diff
@@ -53,10 +53,8 @@
         "external-task/source-build-oci-ta/**",
         "external-task/source-build/**",
         "task/generate-labels/**",
-        "task/show-sbom/**",
         "task/slack-webhook-notification-oci-ta/**",
         "task/slack-webhook-notification/**",
-        "task/summary/**",
         "task/update-infra-deployments/**"
```
chmeliik, line 58: Unrelated: weren't `slack-webhook-notification(-oci-ta)` and `update-infra-deployments` also supposed to be deprecated?

chmeliik (approved):

renovate.json
slimreaper35: The last usage of both tasks was in the `core-services` pipeline so they weren't formally deprecated. But now after deleting the pipeline they should be.

The `slack-webhook-notification` is in https://github.com/konflux-ci/tekton-integration-catalog so maybe a README.md with a redirect there ?

chmeliik: > The last usage of both tasks was in the core-services pipeline so they weren't formally deprecated. But now after deleting the pipeline they should be.

Ah, ok. Should we (re-)open some Jira issues for that?

> The slack-webhook-notification is in https://github.com/konflux-ci/tekton-integration-catalog so maybe a README.md with a redirect there ?

Ideally, tekton-integration-catalog would release the task bundles to `quay.io/konflux-ci/tekton-catalog/task-slack-webhook-notification(-oci-ta)` so that users don't have to migrate manually. But a redirect in the README would be nice as well.

slimreaper35: Okay, so I will create a Jira issue and probably assign it to myself

PR #3830, opened 2026-09-09
build-vm-image: populate artifact manifest config with architecture/os

Victoremepunto:
## Problem

Disk-image artifact manifests produced by `build-vm-image` are built with buildah's **empty config** (`application/vnd.oci.empty.v1+json`, i.e. `{}`). Because of this, the child artifact manifest carries no platform data:

```
child manifest config:
  mediaType: application/vnd.oci.empty.v1+json
  digest:    sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
  size:      2
  data:      e30=   # base64 for "{}"
```

`build-image-index` derives the index entry's `platform` from this config, so the resulting image index ends up with `platform: null`. Downstream release-service tasks (`get-image-architectures` → `apply_mapping.py`) then fail with:

```
arch = arch_infos[0]["platform"]["architecture"]
KeyError: 'platform'
```

The `--arch`/`--os` flags on `buildah manifest add` only stamp platform onto the length-1 index entry, which the task subsequently discards when it overwrites the tag with the bare child manifest — so no platform metadata survives on the published artifact.

## Fix

Write a real OCI image config (`{architecture, os}`) to a temp file after `buildah manifest create`, and pass it via `--artifact-config` / `--artifact-config-type application/vnd.oci.image.config.v1+json` on every `buildah manifest add` call. The artifact manifest now carries platform metadata, and the final image index gets a populated `platform`, matching normal container images.

## Notes

- Applied to `0.3`. `0.2` has the same latent issue; happy to follow up there if desired.
- The existing tests mock the manifest and run the SBOM path locally; the `buildah manifest add` config path executes on the remote build VM, so this is best validated via the task's e2e.

Refs: [AIPCC-1307](https://redhat.atlassian.net/browse/AIPCC-1307)

### DIFF
--- a/task/build-vm-image/0.3/build-vm-image.yaml
+++ b/task/build-vm-image/0.3/build-vm-image.yaml
@@ -387,6 +387,11 @@ spec:
 
         export OUTPUT_IMAGE="$OUTPUT_IMAGE"
         export IMAGE_TYPE="$IMAGE_TYPE"
+        # ARCH is derived from the PLATFORM param (see above) and is already in
+        # OCI naming (amd64/arm64). Pass it into the remote push script so the
+        # artifact config uses OCI arch names, NOT the remote host's kernel arch
+        # ($(arch)/uname -m returns x86_64/aarch64). See AIPCC-1307.
+        export OCI_ARCH="$ARCH"
 
         REMOTESSHEOF
 
@@ -409,8 +414,20 @@ spec:
         # entry's platform from this config, and downstream release tasks
         # (get-image-architectures / apply_mapping) require platform.architecture.
         # See AIPCC-1307.
+        # Determine the OCI architecture name. Prefer OCI_ARCH (derived from the
+        # PLATFORM param on the pod, already in OCI naming). Fall back to the
+        # remote host's kernel arch only if OCI_ARCH is unset, and normalize it:
+        # the archs returned by uname (x86_64/aarch64) are not the names used by
+        # OCI manifests (amd64/arm64), and the --artifact-config blob is stored
+        # verbatim with no normalization (unlike buildah's --arch flag). See AIPCC-1307.
+        OCI_ARCH="${OCI_ARCH:-$(arch)}"
+        case "$OCI_ARCH" in
+          x86_64)  OCI_ARCH=amd64 ;;
+          aarch64) OCI_ARCH=arm64 ;;
+        esac
+
         ARTIFACT_CONFIG_FILE=$(mktemp)
-        jq -cn --arg arch "$(arch)" --arg os "linux" \
+        jq -cn --arg arch "$OCI_ARCH" --arg os "linux" \
           '{architecture: $arch, os: $os}' > "$ARTIFACT_CONFIG_FILE"
         ARTIFACT_CONFIG_ARGS=(--artifact-config "$ARTIFACT_CONFIG_FILE" \
           --artifact-config-type application/vnd.oci.image.config.v1+json)
--- a/task/build-vm-image/0.3/build-vm-image.yaml
+++ b/task/build-vm-image/0.3/build-vm-image.yaml
@@ -402,37 +402,50 @@ spec:
         # Build an image index of length 1 referring to an image manifest with the content
         buildah manifest create "$OUTPUT_IMAGE"
 
+        # Write a real OCI image config for the artifact manifest so it carries
+        # architecture/os. Without this, buildah uses the empty config
+        # (application/vnd.oci.empty.v1+json, i.e. "{}"), which leaves the child
+        # manifest with no platform data. build-image-index derives the index
+        # entry's platform from this config, and downstream release tasks
+        # (get-image-architectures / apply_mapping) require platform.architecture.
+        # See AIPCC-1307.
+        ARTIFACT_CONFIG_FILE=$(mktemp)
+        jq -cn --arg arch "$(arch)" --arg os "linux" \
+          '{architecture: $arch, os: $os}' > "$ARTIFACT_CONFIG_FILE"
+        ARTIFACT_CONFIG_ARGS=(--artifact-config "$ARTIFACT_CONFIG_FILE" \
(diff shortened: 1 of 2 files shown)


thozza:
Thanks for chasing this down, the `KeyError: 'platform'` is real. I'd push back on the mechanism though: I think it makes the artifact non-conformant and bakes a `build-image-index` defect into the on-registry format.

**1. The config blob is not a valid `application/vnd.oci.image.config.v1+json`.** [config.md](https://github.com/opencontainers/image-spec/blob/main/config.md#properties) marks `rootfs` REQUIRED (with `rootfs.type` / `rootfs.diff_ids`), and [schema/config-schema.json](https://github.com/opencontainers/image-spec/blob/main/schema/config-schema.json) has `"required": ["architecture","os","rootfs"]`. `{"architecture":"amd64","os":"linux"}` declares a media type whose schema it violates. Compare go-containerregistry's [empty.Image](https://github.com/google/go-containerregistry/blob/main/pkg/v1/empty/image.go) (basis for cosign signature manifests, the usual "artifact with an image config" prior art): it still emits `RootFS.Type: "layers"`, with the comment "Some clients check this."

To be clear, a *non-empty* config is fine. [manifest.md](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) says `config.mediaType` "MUST be set to a value specific to the artifact type or the empty value." The problem is borrowing the *container image* type specifically. Worth noting that when ORAS explored this same idea in [image-spec#1216](https://github.com/opencontainers/image-spec/issues/1216), even their config-data variant used a custom type (`application/vnd.unknown.config.v1+json`), not the image config type.

**2. That type is a load-bearing discriminator.** [containers/image manifest/oci.go](https://github.com/containers/image/blob/main/manifest/oci.go) gates `Inspect()`, `ImageID()` and `CanChangeLayerCompression()` on `Config.MediaType == MediaTypeImageConfig`. This flips them from "fail loudly with `NonImageArtifactError`" to "succeed with wrong answers." Worst case is `ImageID()`: it would return the config digest, and that blob is byte-identical for every amd64 disk image, forever. Its consumer is `storageImageDestination.computeID`, so all amd64 disk images collide on one containers-storage ID. The guard's comment says exactly why: "for non-image artifacts, we can't expect the config to change any time the other layers (semantically) change."

**3. The manifest becomes self-contradictory.** `artifactType: application/vnd.diskimage.*` says disk image; `config.mediaType` says container image. Consumers following [descriptor.md](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties) ("This is the value of the config descriptor `mediaType` when the descriptor references an image manifest") get the wrong one. buildah's own [--artifact-config docs](https://github.com/containers/buildah/blob/main/docs/buildah-manifest-add.1.md) say the empty default "is the preferred option."

**4. It's the wrong layer.** `platform` is an index descriptor property that "SHOULD be present if its target is platform-specific" ([image-index.md](https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions)). Nothing requires the child manifest to carry a config for that. This exact question was raised upstream in [image-spec#1216](https://github.com/opencontainers/image-spec/issues/1216), and the maintainer answer was consistently "use the index" ([jonjohnsonjr](https://github.com/opencontainers/image-spec/issues/1216#issuecomment-2450438068), [sudo-bmitch](https://github.com/opencontainers/image-spec/issues/1216#issuecomment-2450559662): "Every use case I can think of comes back to this").

Note this task already passes `--arch $(arch) --os linux` and produces a correct 1-entry index, which the subsequent `skopeo copy` of `.manifests[0].digest` then discards. The data isn't missing at the source, it's lost in transit and never reconstructed, because [BuildahManifestAddArgs](https://github.com/konflux-ci/konflux-build-cli/blob/main/pkg/cliwrappers/buildah.go) has no `Arch`/`OS` fields, so `build-image-index` emits a bare `buildah manifest add --all` (and `--all` is a no-op on a bare manifest).

### Suggested alternative

Pass the mapping explicitly instead of inferring it from a config:

1. `build-vm-image` emits an architecture-to-image-reference mapping as a task result, derived from the existing `PLATFORM` param. No remote-builder changes, and backward compatible, since Tekton ignores unused results.
2. `build-image-index` gains an **optional** param accepting that mapping, and `konflux-build-cli` adds `Arch`/`OS` to `BuildahManifestAddArgs` so it can pass `--arch`/`--os` per entry and reconstruct `platform` correctly. When the param is absent, behaviour is unchanged.

This matches what image-spec maintainers describe as the producer's job ([sudo-bmitch](https://github.com/opencontainers/image-spec/issues/1216#issuecomment-2813107699)):

> it's the responsibility of the tooling that is creating the multi-platform index to track each of the entries and the referenced descriptor values, including the platform and any annotations. Those values could be pulled up from the nested image manifest and its config blob, but that's not a requirement, particularly for custom artifacts.

It also puts `platform` where the spec puts it, keeps the artifact manifest conformant.

Defining a genuinely artifact-specific config type (e.g. `application/vnd.diskimage.config.v1+json`, per decision-tree case 3 in manifest.md) would be spec-legal, but it would *not* fix `platform: null`, since buildah only infers platform from a config it recognises as an image config. To me that's the indication that the index layer is the right place to fix this.

One thing worth checking before merge: does anything in the Conforma/EC path branch on `config.mediaType` to decide whether a manifest is a container image? We have an open issue ([KONFLUX-15247](https://redhat.atlassian.net/browse/KONFLUX-15247)) where `labels.required_labels` is structurally incompatible with OCI artifacts. If EC currently skips these because the config is empty, this change may start applying container-image rules to disk images.

Victoremepunto:
Superseded by #3838. Per the discussion here and @thozza's analysis, platform metadata belongs on the image index descriptor, not faked into the artifact config (reusing `application/vnd.oci.image.config.v1+json` is non-conformant and causes ImageID collisions / EC risk). #3838 exposes a `PLATFORM` result from build-vm-image so build-image-index can set platform on each descriptor natively. Closing in favour of that approach.

PR #3833, opened 2026-09-10
build-vm-image: resolve pushed manifest digest without --digestfile

### LINKED ISSUE #3832: build-vm-image/0.3: push step crashes reading empty --digestfile (invalid reference format)
## Problem Description

The `build-vm-image/0.3` task deterministically fails at the push/finalize
step. The disk image builds successfully and `buildah manifest push` succeeds,
but the task then reads an **empty** digestfile and crashes when constructing an
image reference, exiting 125.

The relevant code runs inside a `podman run "$BUILDAH_IMAGE" /script-push.sh`
container on the remote builder VM
(`task/build-vm-image/0.3/build-vm-image.yaml`, ~lines 495-505):

```bash
retry buildah manifest push --digestfile image-digest --authfile /.docker/config.json --retry "$buildah_retries" --all $OUTPUT_IMAGE
...
IMAGE_INDEX_DIGEST=$(cat image-digest)
REPO=${OUTPUT_IMAGE%:*}
MANIFEST_DIGEST=$(buildah manifest inspect --authfile /.docker/config.json $REPO@$IMAGE_INDEX_DIGEST | jq -r '.manifests[0].digest')
```

`--digestfile image-digest` and `cat image-digest` use a **bare relative path**,
and the `podman run` invocation defines no `WORKDIR`, so the file written by
`buildah manifest push` is not where `cat` looks for it. `IMAGE_INDEX_DIGEST`
ends up empty, producing a bare `$REPO@` reference.

## Steps to Reproduce

1. Run a pipeline that uses `build-vm-image/0.3` to produce a disk image
   (e.g. a qcow2 disk-image build).
2. Let the build complete and reach the push step.
3. Observe the task exit with code 125.

## Current Behavior

```
++ cat image-digest
+ IMAGE_INDEX_DIGEST=
+ REPO=quay.io/.../bootc-cuda-qcow2-disk-image
++ buildah manifest inspect --authfile /.docker/config.json quay.io/.../bootc-cuda-qcow2-disk-image@
Error: parsing reference "quay.io/.../bootc-cuda-qcow2-disk-image@": invalid reference format
+ MANIFEST_DIGEST=
...
debug1: Exit status 125
```

The manifest push itself succeeds ("Writing manifest to image destination",
`return 0`) immediately before this. The failure is purely in reading the
digestfile back. Because the task dies before the final `skopeo copy`, the
arch-suffixed image tag is never published.

This reproduces identically on both amd64 and arm64 builder VMs, so it is
deterministic and not a race or infra flake.

## Expected Behavior

The push step reads the pushed digest reliably, resolves the child manifest
digest, overwrites the tag, and records results — the task completes
successfully.

## Other information

The sibling remote-SSH build task `build-paketo-builder-oci-ta/0.2` avoids this
by writing the digestfile to an absolute path on a mounted/shared volume and
reading it back from that absolute path, rather than a bare relative path.

Fix intent: use an absolute digestfile path inside the push container and add a
guard that fails loudly with diagnostics if the digest is empty, instead of
constructing a bare `@` reference. PR to follow.

Victoremepunto:
## Problem

The push step in `build-vm-image/0.3` read the pushed index digest from
`buildah manifest push --digestfile`. Against **quay.io**, a
`manifest push --all` of a length-1 manifest list writes only the **child**
manifest and never the list itself, so the digestfile is left **empty** while
the push still exits `0`. The empty digest yielded a bare `$REPO@` reference and
crashed the task with `invalid reference format` (exit 125) even though the
manifest push itself succeeded.

This was observed on **both** arch legs of a real RHEL AI disk-image build
([AIPCC-1307](https://redhat.atlassian.net/browse/AIPCC-1307)). Note: the earlier absolute-path / `WORKDIR` theory did **not**
hold — the digestfile was empty even with an absolute `mktemp` path, and the
push output confirmed the list was never written (`Writing manifest list to
image destination` / `Storing list signatures` lines absent).

## Fix

- Resolve the child (artifact) manifest digest from the **local** manifest list
  via `buildah manifest inspect` **before** pushing. The list only ever contains
  a single artifact manifest and the index is discarded (the tag is re-pointed
  at the child), so the locally-read child digest is authoritative and
  **registry-independent**.
- Drop `--digestfile` from the push entirely.
- Add a post-push `skopeo inspect` guard that fails loudly if the child manifest
  is not present on the registry, instead of constructing a bare reference.

## Verification

Replayed the exact `manifest create` → `manifest add --artifact` →
`manifest push --all` → `skopeo copy` sequence with buildah 1.43.2 against a
local registry: `MANIFEST_DIGEST` resolves, the child is confirmed present on
the registry, the final tag is the single artifact manifest carrying a real
config (`architecture`/`os`), and `IMAGE_URL` / `IMAGE_DIGEST` /
`IMAGE_REFERENCE` all populate — no bare `$REPO@`.

Fixes #3832

### DIFF
--- a/task/build-vm-image/0.3/build-vm-image.yaml
+++ b/task/build-vm-image/0.3/build-vm-image.yaml
@@ -461,18 +461,41 @@ spec:
         fi
 
         buildah_retries=3
+        REPO=${OUTPUT_IMAGE%:*}
+
+        # Resolve the child (artifact) manifest digest from the LOCAL manifest
+        # list BEFORE pushing. We deliberately do NOT rely on
+        # `buildah manifest push --digestfile` to report the pushed index
+        # digest: against some registries (e.g. quay.io) a `manifest push --all`
+        # of a length-1 list writes only the child manifest and never the list
+        # itself, so the digestfile is left empty and the push still exits 0.
+        # That produced a bare "$REPO@" reference and crashed the task with
+        # "invalid reference format". See issue #3832.
+        #
+        # The list only ever contains a single artifact manifest, and we throw
+        # the index away anyway (we re-tag with the child below), so the child
+        # digest read locally is authoritative and registry-independent.
+        MANIFEST_DIGEST=$(buildah manifest inspect $OUTPUT_IMAGE | jq -r '.manifests[0].digest')
+        if [ -z "$MANIFEST_DIGEST" ] || [ "$MANIFEST_DIGEST" = "null" ]; then
+            echo "Failed to resolve child manifest digest from local manifest list ${OUTPUT_IMAGE}; refusing to build a bare reference" >&2
+            buildah manifest inspect $OUTPUT_IMAGE >&2 || true
+            exit 1
+        fi
+
         echo "Pushing image $OUTPUT_IMAGE to registry"
-        if ! retry buildah manifest push --digestfile image-digest --authfile /.docker/config.json --retry "$buildah_retries" --all $OUTPUT_IMAGE
+        if ! retry buildah manifest push --authfile /.docker/config.json --retry "$buildah_retries" --all $OUTPUT_IMAGE
         then
             echo "Failed to push image ${OUTPUT_IMAGE} to registry"
             exit 1
         fi
 
-        # At this point, we have pushed an image index of length 1 to the registry.
-        # Next, extract a reference to the image manifest and expose that, throwing away the image index.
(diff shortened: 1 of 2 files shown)


Victoremepunto:
Obsolete. This was a workaround from the earlier direct-push exploration, which has been abandoned in favour of the index-layer fix (#3838 + konflux-build-cli#249 + container-build-catalog#166). Closing.

PR #3835, opened 2026-09-11
AIPCC-1307: build-vm-image: push disk image as a single OCI artifact manifest (direct oras push)

Victoremepunto:
> **DRAFT — for discussion, not yet requesting review.** Supersedes #3830 and #3833.

## Summary
This is the full [AIPCC-1307](https://redhat.atlassian.net/browse/AIPCC-1307) work-in-progress branch, opened as a **draft** so the
actual change is visible (the two existing PRs #3830 and #3833 are stale and
predate the redesign).

RHEL AI disk-image OCI artifacts carried buildah's empty config `{}`, so the
image index had `platform: null` and the release `apply-mapping` task failed with
`KeyError: 'platform'`.

## What this branch currently does (⚠️ under active review — see below)
1. Populate the artifact manifest config with a real `{architecture, os}` (from #3830).
2. Use OCI arch names amd64/arm64 derived from the PLATFORM param (from #3830).
3. **Replace `buildah manifest create/add --all` + `skopeo copy .manifests[0]`
   with a direct single OCI artifact manifest push via `oras`** (blob push config
   + layer, hand-build manifest, `oras manifest push`, `oras resolve`). This
   avoids quay returning HTTP 500 on an OCI *index* whose only child is an
   artifact carrying a real config, and removes the `--digestfile` workaround
   (#3832 / #3833).
4. Provide `oras` on the remote builder VM by extracting it from the pinned
   Konflux `task-runner` image (`podman run … cat`), since oras is not in the UBI9
   repos and `podman cp` is not permitted by the VM sudoers.
5. Give oras a registry-scoped auth file (Konflux config.json is repo-scoped;
   oras needs registry-scoped), mirroring `select-oci-auth`.

## ⚠️ Open design concern (thozza, #3830)
@thozza raised that using `application/vnd.oci.image.config.v1+json` as the
artifact config makes a disk-image artifact look like an OCI container image,
which is spec-nonconformant (image config schema requires `rootfs`) and flips
load-bearing discriminators in containers/image, and may trip Conforma/EC
container-image rules. His suggested alternative is to keep the artifact config
empty/artifact-specific and instead have `build-vm-image` emit a
ref→platform mapping result that `build-image-index` optionally consumes to
reconstruct `platform`.

This draft is kept open to compare approaches; the config-media-type question is
unresolved and may change the direction of the fix.

### DIFF
--- a/task/build-vm-image/0.3/build-vm-image.yaml
+++ b/task/build-vm-image/0.3/build-vm-image.yaml
@@ -402,67 +402,68 @@ spec:
 
         # this quoted heredoc prevents expansions and command substitutions. the env vars are evaluated on the remote vm
         cat >>scripts/script-push.sh <<'REMOTESSHEOF'
-        dnf -y install buildah skopeo pigz jq tar xz
+        dnf -y install buildah skopeo oras pigz jq tar xz
 
-        # Build an image index of length 1 referring to an image manifest with the content
-        buildah manifest create "$OUTPUT_IMAGE"
-
-        # Write a real OCI image config for the artifact manifest so it carries
-        # architecture/os. Without this, buildah uses the empty config
-        # (application/vnd.oci.empty.v1+json, i.e. "{}"), which leaves the child
-        # manifest with no platform data. build-image-index derives the index
-        # entry's platform from this config, and downstream release tasks
-        # (get-image-architectures / apply_mapping) require platform.architecture.
-        # See AIPCC-1307.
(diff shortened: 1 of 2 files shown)


Victoremepunto:
Abandoned. The oras direct-push approach existed only to work around placing a real config in the artifact child; the correct fix sets platform on the image index instead (#3838 + konflux-build-cli#249 + container-build-catalog#166), which makes the whole oras detour unnecessary. Closing.

PR #249, opened 2022-12-07
PLNSRVCE-884: Add FBC build pipeline

Michkov:
FBC build pipeline placeholder - copy of docker-builder since it should have only updated tests.

PR #3838, opened 2026-09-11
build-vm-image: emit IMAGE_PLATFORM_MAPPING result for image index platform metadata

Victoremepunto:
## Problem

Multi-arch disk-image OCI artifacts end up with `platform: null` on their image-index entries. Disk-image artifacts carry an empty config (`application/vnd.oci.empty.v1+json`), so `build-image-index` cannot infer each child's platform when assembling the index and leaves the descriptors without `platform.architecture`/`os`. This breaks consumers that select by architecture — e.g. the release `apply-mapping` task, which reads `platform.architecture` and fails with `KeyError: 'platform'`.

The correct place to fix this is the **image index** (per opencontainers/image-spec#1216), not the artifact config. But `build-image-index` needs to know each child's platform, and the empty artifact config carries none — so `build-vm-image` must surface it.

## Change

Add a new `IMAGE_PLATFORM_MAPPING` task result: a single per-matrix-leg string of the form `<IMAGE_REFERENCE>=linux/<arch>` (e.g. `quay.io/org/repo@sha256:aaa=linux/amd64`).

Why the combined string (and not a bare `PLATFORM` result): `build-vm-image` is matrixed, so `IMAGE_REFERENCE[*]` and a bare `PLATFORM[*]` would be two independently-aggregated arrays. Tekton guarantees neither element-wise alignment between them nor any zip/join syntax (the matrix docs explicitly disallow consuming a specific combination of results from a fanned-out task). Binding the image reference to its platform **inside** the matrix leg avoids that entirely: the pipeline just fans the combined value in with `IMAGE_PLATFORM_MAPPING[*]` and passes it to `build-image-index`'s `IMAGES_PLATFORMS` param. This mirrors how the task already builds `IMAGE_REFERENCE` (IMAGE_URL + IMAGE_DIGEST) in-task, for the same isolation reason.

- The value is written after the leg's results are synced back, reading `IMAGE_REFERENCE` from the result the remote build just wrote, so the key matches the `IMAGES` entry exactly.
- `arch` comes from the `PLATFORM` param (OCI names), never `uname`/`$(arch)` on the remote VM; `os` is always `linux` for VM disk images.
- Backward compatible: Tekton ignores unused results.
- README regenerated via `hack/generate-readme.sh`.

## Related (coordinated change set)

This is part 1 of a three-part fix (design by @thozza):
1. **this PR** — `build-vm-image` emits `IMAGE_PLATFORM_MAPPING`
2. konflux-ci/konflux-build-cli#249 — `--images-platforms` → `buildah manifest add --arch/--os` (ref-keyed)
3. konflux-ci/container-build-catalog#166 — `build-image-index` task `IMAGES_PLATFORMS` array param

The disk-image build pipeline then fans in `IMAGE_PLATFORM_MAPPING[*]` → `IMAGES_PLATFORMS[*]`.

## Validation

Proven end-to-end on a real Konflux cluster (bootc RHEL AI disk images), with all three PRs wired via git resolver + a CLI image built from #249:

- **Single-arch (Gaudi qcow2):** built index child carries `platform: {architecture: amd64, os: linux}`; `get-image-architectures` and the release `apply_mapping.py` succeed.
- **Multi-arch (CUDA qcow2, amd64 + arm64):** index has two children, each with correct `platform`; `apply_mapping.py` returns both arches with no `KeyError: 'platform'`.

Supersedes the config-in-artifact approach (#3830), which is non-conformant (fakes an image config on an artifact). See @thozza's analysis for why platform belongs on the index descriptor.

Co-designed with Tomáš Hozza.

### DIFF
--- a/task/build-vm-image/0.3/README.md
+++ b/task/build-vm-image/0.3/README.md
@@ -25,6 +25,7 @@ Build disk images using bootc-image-builder. https://github.com/osbuild/bootc-im
 |IMAGE_URL|Image repository where the built manifest list was pushed|
 |IMAGE_REFERENCE|Image reference (IMAGE_URL + IMAGE_DIGEST)|
 |SBOM_BLOB_URL|Reference of SBOM blob digest to enable digest-based verification from provenance|
+|PLATFORM|OCI platform ("linux/<arch>") of the disk image built by this task run, derived from the PLATFORM param. Lets build-image-index set the platform on each child descriptor of the multi-arch index, since the artifact manifest itself carries an empty config with no platform information.|
 
 
 ## Additional info
--- a/task/build-vm-image/0.3/build-vm-image.yaml
+++ b/task/build-vm-image/0.3/build-vm-image.yaml
@@ -64,6 +64,13 @@ spec:
     - description: Reference of SBOM blob digest to enable digest-based verification from provenance
       name: SBOM_BLOB_URL
       type: string
+    - description: >-
+        OCI platform ("linux/<arch>") of the disk image built by this task run,
+        derived from the PLATFORM param. Lets build-image-index set the platform
+        on each child descriptor of the multi-arch index, since the artifact
+        manifest itself carries an empty config with no platform information.
+      name: PLATFORM
+      type: string
   stepTemplate:
     env:
       - name: OUTPUT_IMAGE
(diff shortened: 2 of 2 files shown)


thozza (commented):
In think that it would be sensible to update the `version` to `0.3.1` and add an entry to the `CHANGELOG.md`.

But otherwise, the diff LGTM. Thanks!

Victoremepunto:
> In think that it would be sensible to update the `version` to `0.3.1` and add an entry to the `CHANGELOG.md`.
> 
> But otherwise, the diff LGTM. Thanks!

did just that, I agree with the API change bump - can you please review again ? thanks!

ralphbean (dismissed):

ralphbean (approved):

PR #3842, opened 2026-09-14
Bump slack-webhook-notification task version

slimreaper35:
The task has been moved to the tekton-integration-catalog repo. Add a migration script for self-replacing the task reference to git resolver.

--
https://github.com/konflux-ci/tekton-integration-catalog/tree/main/tasks/slack-webhook-notification/0.1 https://tekton.dev/docs/pipelines/git-resolver/#task-resolution

Assisted-by: Cursor


policies/step-actions.yaml
```diff
@@ -2,8 +2,10 @@
 # These policies are meant to be applied to all of the Tasks in this repo.
 sources:
   - policy:
-      - github.com/conforma/policy//policy/lib
-      - github.com/conforma/policy//policy/stepaction
+      # Pin the policy repo to a commit that doesn't break EC checks.
+      # https://github.com/conforma/policy/commit/81c889ba791f81ebe5a4251364f7e0f045100447
+      - github.com/conforma/policy?ref=81c889ba791f81ebe5a4251364f7e0f045100447//policy/lib
+      - github.com/conforma/policy?ref=81c889ba791f81ebe5a4251364f7e0f045100447//policy/stepaction
```
chmeliik: Could you share some more info about this? What was the change that broke the checks? What's the plan for fixing it?

task/slack-webhook-notification-oci-ta/0.1/migrations/0.1.1.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+declare -r GIT_URL="https://github.com/konflux-ci/tekton-integration-catalog.git"
+declare -r GIT_REVISION="main"
```
chmeliik: I think we should use a specific revision rather than `main`, as a supply chain security practice

task/slack-webhook-notification-oci-ta/0.1/migrations/0.1.1.sh
```diff
@@ -0,0 +1,39 @@
...
+declare -r GIT_PATH="tasks/slack-webhook-notification/0.1/slack-webhook-notification.yaml"
+
+TASKS_SELECTOR='(.spec.tasks[]?, .spec.finally[]?, .spec.pipelineSpec.tasks[]?, .spec.pipelineSpec.finally[]?)'
+TASK_SELECTOR="${TASKS_SELECTOR} | select(
+  .taskRef.name == \"slack-webhook-notification\" or
+  (.taskRef.params[] | (.name == \"name\" and .value == \"slack-webhook-notification\")) or
+  (.taskRef.params[] | (.name == \"pathInRepo\" and (.value | test(\"task/slack-webhook-notification/\"))))
```
chmeliik: IMO trying to handle git resolver references in this migration script is not necessary and may be a bad idea.

The pipeline-migration-tool only supports migrations for bundle references, not git references. Which means that if this migration script is being executed on a Pipeline, it's because that pipeline references slack-webhook-notification as a bundle. So in the unlikely case that the "pathInRepo" selector actually matches something, it may be a false positive.

task/slack-webhook-notification-oci-ta/0.1/migrations/0.1.1.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+declare -r pipeline_file=${1:?missing pipeline file}
+
+declare -r GIT_URL="https://github.com/konflux-ci/tekton-integration-catalog.git"
+declare -r GIT_REVISION="main"
+declare -r GIT_PATH="tasks/slack-webhook-notification/0.1/slack-webhook-notification.yaml"
```
chmeliik, line 9: We might have a slight problem. This migration is for the oci-ta variant of the task, but references the non-oci-ta variant of the task in tekton-integration-catalog.

But the problem is that tekton-integration-catalog doesn't even have an oci-ta variant for this task. I think we need to add one first.

slimreaper35: I don't see any oci-ta task variants in the repo at all.

It has the optional [workspace](https://github.com/konflux-ci/tekton-integration-catalog/tree/main/tasks/slack-webhook-notification/0.1#workspaces) from "our" slack-notification-oci-ta task.

chmeliik: > It has the optional [workspace](https://github.com/konflux-ci/tekton-integration-catalog/tree/main/tasks/slack-webhook-notification/0.1#workspaces) from "our" slack-notification-oci-ta task.

That's from our slack-webhook-notification (non- oci-ta) task. The oci-ta variant doesn't have it and instead has SOURCE_ARTIFACT

slimreaper35: https://github.com/konflux-ci/tekton-integration-catalog/pull/325

PR #3847, opened 2026-09-18
build-vm-image: fix IMAGE_DIGEST result trailing newline breaking SBOM discovery

Victoremepunto:
## Problem

`build-vm-image`'s build step writes the `IMAGE_DIGEST` Tekton result with `echo $MANIFEST_DIGEST` (no `-n`), which appends a trailing newline to the digest value. That newline is preserved verbatim in the SLSA provenance's task result.

Conforma's `sbom.found` check (via `_fetch_pipelinerun_sbom` in `lib/sbom/sbom.rego`) locates a build task's `SBOM_BLOB_URL` by doing an **exact string match** between the evaluated artifact's digest and the build task's `IMAGE_DIGEST` result:

```rego
expected_image_digest := image.parse(input.image.ref).digest   # "sha256:...ab"
image_digest := tekton.task_result(task, "IMAGE_DIGEST")       # "sha256:...ab\n"
expected_image_digest == image_digest                           # always false
```

Because of the trailing newline, this comparison never matched, so the SBOM produced by `build-vm-image` was never discoverable via the trusted provenance-based path. This was masked by the legacy unsigned `.sbom` cosign tag attachment (which Conforma also discovers, as a fallback) — until a recent Conforma policy change stopped trusting unsigned `.sbom` tags for signature verification, which exposed this pre-existing bug as a `sbom.found` violation.

## Root cause verification

Reproduced against the real deployed Conforma policy bundle using `ec validate image --output policy-input` to dump the exact OPA input, then `opa eval` to confirm:
- `IMAGE_DIGEST` task result literally contains a trailing `\n`.
- `data.lib.sbom._fetch_pipelinerun_sbom` returns 0 results against the real input.
- Stripping just the trailing newline from the captured input's `IMAGE_DIGEST` value makes `data.sbom.deny` evaluate to `[]` (passes).

## Fix

- `echo -n "$MANIFEST_DIGEST" | tee /tekton-results/IMAGE_DIGEST` (added `-n`, matching the existing pattern already used for `IMAGE_URL` and `IMAGE_REFERENCE` two lines below/above).
- Bumped task to `0.3.1`, added CHANGELOG entry.

## Note on test coverage

The task's test suite mocks out the entire "build" step (the SSH-based script where this bug lived) with a lightweight stub that already used `echo -n` correctly, so the real bug was never exercised by CI. Flagging this as a known gap; happy to discuss follow-up test coverage if desired.


thozza (approved):
FWIW, LGTM

task/build-vm-image/0.3/tests/test-build-vm-image-sbom-cyclonedx-fallback.yaml
```diff
@@ -36,7 +36,7 @@ spec:
 
               echo "Pushing dummy source container image (no SBOM attached)"
               cd /tmp && echo "source-content-cyclonedx-fallback" > source.txt
-              oras push --no-tty "${SOURCE_IMAGE}" source.txt
+              oras push --insecure --no-tty "${SOURCE_IMAGE}" source.txt
```
ralphbean, line 39: Wait.

Why `--insecure`?

(there's a lot of this in the PR.)

ralphbean (changes requested):

task/build-vm-image/0.3/tests/test-build-vm-image-sbom-cyclonedx-fallback.yaml
ralphbean: Ah, I see. This is a test and `INSECURE_REGISTRY` is an option in the task. Will look more closely.

ralphbean (approved):

